PR #7, opened 2022-11-20
Add basic unit tests

picklelo:
Added unit tests for all the base classes and made sure they pass.

### DIFF
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -371,13 +371,12 @@ class State(Base, ABC):
         delta = {}
 
         # Return the dirty vars, as well as all computed vars.
-        if len(self.dirty_vars) > 0:
-            delta = {
-                self.get_full_name(): {
-                    prop: getattr(self, prop)
-                    for prop in self.dirty_vars | set(self.computed_vars.keys())
-                }
-            }
+        subdelta = {
+            prop: getattr(self, prop)
+            for prop in self.dirty_vars | set(self.computed_vars.keys())
+        }
+        if len(subdelta) > 0:
+            delta[self.get_full_name()] = subdelta
 
         # Recursively find the substate deltas.
         for substate in self.dirty_substates:
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -368,13 +368,17 @@ class State(Base, ABC):
         Returns:
             The delta for the state.
         """
+        delta = {}
+
         # Return the dirty vars, as well as all computed vars.
-        delta = {
-            self.get_full_name(): {
-                prop: getattr(self, prop)
-                for prop in self.dirty_vars | set(self.computed_vars.keys())
+        if len(self.dirty_vars) > 0:
+            delta = {
+                self.get_full_name(): {
+                    prop: getattr(self, prop)
+                    for prop in self.dirty_vars | set(self.computed_vars.keys())
+                }
             }
-        }
+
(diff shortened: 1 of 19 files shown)


pynecone/state.py
```diff
@@ -368,13 +368,17 @@ def get_delta(self) -> Delta:
...
             The delta for the state.
         """
+        delta = {}
+
         # Return the dirty vars, as well as all computed vars.
-        delta = {
-            self.get_full_name(): {
-                prop: getattr(self, prop)
-                for prop in self.dirty_vars | set(self.computed_vars.keys())
+        if len(self.dirty_vars) > 0:
```
picklelo, line 374: I just realized this may break the clock app - maybe I'll get rid of this `if` statement.

tests/test_app.py
```diff
@@ -36,25 +39,32 @@ def about():
...
     Args:
         app: The app to test.
     """
+    assert app.state() == DefaultState()
     assert app.middleware == [HydrateMiddleware()]
+    assert app.style == Style()
 
 
-def test_add_page_default_route(app: App, index_page, about_page) -> None:
+def test_add_page_default_route(app: App, index_page, about_page):
```
Alek99, line 67: Maybe for later but a nested route test here.

tests/test_utils.py
```diff
@@ -169,3 +169,13 @@ def test_format_cond(condition: str, true_value: str, false_value: str, expected
         expected: The expected output string.
     """
     assert utils.format_cond(condition, true_value, false_value) == expected
+
```
Alek99, line 172: I can't seem to put this comment where it belongs on line 9. I think we should add a test for snake case with a _ coming first for example  (_hover)

pynecone/var.py
```diff
@@ -153,14 +153,15 @@ def __getitem__(self, i: Any) -> Var:
...
-        elif utils.is_dataframe(self.type_):
+        elif utils._issubclass(self.type_, Dict) or utils.is_dataframe(self.type_):
             if isinstance(i, str):
                 i = utils.wrap(i, '"')
             if isinstance(self.type_, _GenericAlias):
                 type_ = utils.get_args(self.type_)[1]
             else:
                 type_ = Any
         else:
+            breakpoint()
```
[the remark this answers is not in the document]
picklelo: Good catch, removed.

pynecone/state.py
picklelo: I removed this to always include computed vars.

tests/components/test_component.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+    style = {
+        component1: Style({"color": "white"}),
+        component2: Style({"color": "black"}),
+    }
+    c1 = component1().add_style(style)  # type: ignore
+    c2 = component2().add_style(style)  # type: ignore
+    assert c1.style["color"] == "white"
+    assert c2.style["color"] == "black"
+
```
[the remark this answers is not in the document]
picklelo: Added a test for this.

tests/test_app.py
picklelo: Added a test for nested routes below.

tests/test_utils.py
picklelo: Good idea, added a test for this

Alek99 (approved):

PR #8, opened 2022-11-21
Create config class

picklelo:
Created a new class: `pc.Config`

Now the new default `pcconfig.py` file looks something like:
```
import pynecone as pc


config = pc.Config(
    app_name="asdf",
)
```

and extra fields can just be passed as keyword arguments inside there.
This way they don't have to see the defaults also.

### DIFF
--- a/pynecone/config.py
+++ b/pynecone/config.py
@@ -19,6 +19,9 @@ class Config(Base):
     # The redis url.
     redis_url: Optional[str] = None
 
+    # The deploy url.
+    deploy_url: Optional[str] = None
+
     # The environment mode.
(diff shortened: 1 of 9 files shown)


pynecone/config.py
```diff
@@ -0,0 +1,26 @@
+from typing import Optional
+
+from pynecone import constants
+from pynecone.base import Base
+
+
+class Config(Base):
+    """A Pynecone config."""
+
```
Alek99, line 9: Deploy URI needs to be added and optional.

Alek99 (changes requested):

Alek99 (approved):

PR #13, opened 2022-11-24
Add cache for GitHub actions

picklelo:
* Add caching for poetry and python packages
* Add black lint check for python check


Alek99 (approved):
Ok so copy to clipboard is just a pynetree thing for now.

PR #25, opened 2022-12-05
Updated README and Contributing

Alek99:


README.md
```diff
@@ -50,160 +48,174 @@ You can run this app in development mode:
...
-        font_size="2em",
-        text_align="center",
-    )
-```
-This function defines the frontend of the app.
-We use different components such as pc.box, pc.button, and pc.heading to build the frontend. Components can be nested to create complex layouts, and can be styled using the full power of CSS.
-    
-Pynecone comes with [50+ built-in components](https://pynecone.io/docs/library) to help you get started. 
-We are actively adding more components, plus it's easy to create your own components.
+Our counter app has two event handlers, process_image to indicate that the image is being generated and get_image, which calls the OpenAI API.
```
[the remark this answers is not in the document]
picklelo: Also, same comment of styling code

README.md
```diff
@@ -50,160 +48,174 @@ You can run this app in development mode:
...
+ ### **UI In Pynecone**
+Lets start by talking about the UI this Pynecone App.
+
+```python 
+def index():
+    return pc.center(
+        ...
+    )
+```
+This index function defines the frontend of the app. We use different components such as center, vstack, input, and button to build the front end. Components can be nested to create complex layouts and styled using CSS's full power. Just pass them in as keyword args.
```
picklelo, line 141: Let's style the functions as inline code, like `index`,  `center`, etc.

README.md
```diff
@@ -50,160 +48,174 @@ You can run this app in development mode:
...
-* ### Routing 
+### **Routing** 
     
-Next we define our app and tell it what state to use.
+Finally we define our app and tell it what state to use.
 ```python
 app = pc.App(state=State)
 ```
-We add a route from the root of the app to the counter component. By default the route
+We add a route from the root of the app to the counter component. We also add a title that will show up in the page preview.
```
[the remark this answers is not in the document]
picklelo: We can also mention the title shows up in the browser tab

README.md
```diff
@@ -138,7 +140,7 @@ def index():
         ...
     )
 ```
-This index function defines the frontend of the app. We use different components such as center, vstack, input, and button to build the front end. Components can be nested to create complex layouts and styled using CSS's full power. Just pass them in as keyword args.
+This index function defines the frontend of the app. We use different components such as ```center```, ```vstack```, ```input```, and ```button``` to build the front end. Components can be nested to create complex layouts and styled using CSS's full power. Just pass them in as keyword args.
```
picklelo, line 143: Thanks for the updates! (I think you can use just one backtick in the future for inline, and 3 for a big code block)

README.md
```diff
@@ -171,15 +173,15 @@ Here the state is comprised of a prompt and image_url. There are also the boolea
...
+Our dalle app has two event handlers, ```process_image``` to indicate that the image is being generated and ```get_image```, which calls the OpenAI API.
 
 ### **Routing** 
     
 Finally we define our app and tell it what state to use.
 ```python
 app = pc.App(state=State)
 ```
-We add a route from the root of the app to the counter component. We also add a title that will show up in the page preview.
+We add a route from the root of the app to the index component. We also add a title that will show up in the page preview/ broweser tab.
```
picklelo, line 184: typo: browser

picklelo (approved):
LGTM! We can fix the typo now or after

PR #30, opened 2022-12-05
Added Visually Hidden + Box Video/Iframe

Alek99:
Added visually hidden component for screen readers and made box able to become an iframe using `element='iframe' ` and `src='some_url'`.


picklelo (approved):
LGTM

PR #32, opened 2022-12-06
Add style for code blocks.

picklelo:
* Rename GitHub actions build name
* Update custom code for components
* Add style for code blocks

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: Python application
+name: build
 
 on:
   push:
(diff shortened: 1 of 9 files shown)


.github/workflows/build.yml
```diff
@@ -1,4 +1,4 @@
-name: Python application
```
Alek99, line 1: You'll just need to update the tests badge in the read me because I use the python-checks.yml file.

picklelo: Good catch, updated this

Alek99 (approved):

Alek99:
Sweet this code component styling is so cool!

PR #84, opened 2022-12-13
Improve prop error messages

picklelo:
Fix a range of errors we found over the launch:
* make sure the app is not named "pynecone"
* check that the user ran `pc init` before `pc run`
* add better error messages when setting invalid prop types
* add some unit tests for components

### DIFF
--- a/pynecone/components/component.py
+++ b/pynecone/components/component.py
@@ -71,6 +71,9 @@ class Component(Base, ABC):
         Args:
             *args: Args to initialize the component.
             **kwargs: Kwargs to initialize the component.
+
+        Raises:
+            TypeError: If an invalid prop is passed.
         """
         # Get the component fields, triggers, and props.
         fields = self.get_fields()
--- a/pynecone/components/component.py
+++ b/pynecone/components/component.py
@@ -97,13 +97,25 @@ class Component(Base, ABC):
 
             # Check whether the key is a component prop.
             if utils._issubclass(field_type, Var):
-                # Convert any constants into vars and make sure the types match.
-                kwargs[key] = Var.create(value)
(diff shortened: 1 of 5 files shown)


pynecone/components/component.py
```diff
@@ -97,13 +100,25 @@ def __init__(self, *args, **kwargs):
...
+                    # Get the passed type and the var type.
+                    passed_type = kwargs[key].type_
+                    expected_type = fields[key].outer_type_.__args__[0]
+                except TypeError:
+                    # If it is not a valid var, check the base types.
+                    passed_type = type(value)
+                    expected_type = fields[key].outer_type_
+                if not utils._issubclass(passed_type, expected_type):
+                    raise TypeError(
+                        f"Invalid var passed for prop {key}, expected type {expected_type}, got value {value} of type {passed_type}."
```
Alek99, line 120: Nice defiantly needed this!

pynecone/pc.py
```diff
@@ -63,7 +75,17 @@ def run(
...
         frontend: Whether to run the frontend.
         backend: Whether to run the backend.
+
+    Raises:
+        Exit: If the app is not initialized.
     """
+    # Check that the app is initialized.
+    if not utils.is_initialized():
+        utils.console.print(
+            f"[red]The app is not initialized. Run [bold]pc init[/bold] first."
```
Alek99, line 85: Awesome I know a couple people brought this up aswell

Alek99 (approved):

pynecone/pc.py
picklelo: Yup, we can just catch any of these small errors over time - if one person had them, I'm sure many will...

PR #87, opened 2022-12-13
Update issue templates

Alek99:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,28 @@
...
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+** Specifics (please complete the following information):**
+ - OS: 
+ - Browser:
```
picklelo, line 24: Let's also add the Python version here. Maybe we can also make the browser optional, since most issues here may be with the framework itself

picklelo (changes requested):
Thanks for making this! Let's just add the Python version also

picklelo (approved):
Nice!

PR #106, opened 2022-12-15
Fix event handler lambda args

picklelo:
* Format constant lambda arguments as JSON
* Update pc run command for empty packages list


picklelo:
TODO: I'm working on a follow-up adding unit tests for this. For now I just ran an example app with the code.

Alek99 (approved):

PR #108, opened 2022-12-15
Fix model json conversion

picklelo:
* Update bug where unset model fields weren't being sent in JSON.
* Fix type in pc var getattr method.


pynecone/app.py
```diff
@@ -228,7 +228,7 @@ def add_page(
         component = component if isinstance(component, Component) else component(*args)
 
         # Add the title to the component.
-        compiler_utils.add_meta(component, title, description, image)
```
Alek99, line 231: Good catch here

PR #118, opened 2022-12-16
Fix pc run backend only

picklelo:
* fix bug where running backend only still checked if the web directory existed

### DIFF
--- a/pynecone/pc.py
+++ b/pynecone/pc.py
@@ -64,7 +64,7 @@ def run(
         Exit: If the app is not initialized.
     """
     # Check that the app is initialized.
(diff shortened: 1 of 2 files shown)


pynecone/pc.py
```diff
@@ -64,7 +64,7 @@ def run(
         Exit: If the app is not initialized.
     """
     # Check that the app is initialized.
-    if not utils.is_initialized():
+    if frontend and not utils.is_initialized():
```
Alek99, line 67: Ok nice this will fix deploy

Alek99 (approved):

PR #152, opened 2022-12-21
Added initial integration test template.

Alek99:


.github/workflows/integration.yml
```diff
@@ -0,0 +1,74 @@
+name: integrations
```
picklelo, line 1: nit: shall we call it `integration` to match the file name?

.github/workflows/integration.yml
```diff
@@ -0,0 +1,74 @@
...
+      uses: actions/cache@v2
+      with:
+        path: .venv
+        key: python-${{ matrix.python-version }}-pydeps-${{ hashFiles('**/poetry.lock') }}
+
+    # Begin the integration test
+    - run: poetry install --no-interaction --no-root
+      if: steps.cache-deps.outputs.cache-hit != 'true'
+    - run: ls -R
+    - name: Poetry Install
```
picklelo, line 63: I think this `poetry install` is repeated twice. We may only need the first one, followed by the `pip install` below.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,74 @@
...
+      id: cache-deps
+      uses: actions/cache@v2
+      with:
+        path: .venv
+        key: python-${{ matrix.python-version }}-pydeps-${{ hashFiles('**/poetry.lock') }}
+
+    # Begin the integration test
+    - run: poetry install --no-interaction --no-root
+      if: steps.cache-deps.outputs.cache-hit != 'true'
+    - run: ls -R
```
picklelo, line 62: Do we need this?

.github/workflows/integration.yml
```diff
@@ -0,0 +1,74 @@
...
+      run: poetry install --no-interaction
+    - name: Install Requirements
+      working-directory: ./pynecone-website/pcweb
+      run: poetry run pip install -r requirements.txt
+    - name: Init Website
+      working-directory: ./pynecone-website/pcweb
+      run: poetry run pc init
+    - name: Run Website
+      working-directory: ./pynecone-website/pcweb
+      run: timeout 1m poetry run pc run || exit 0
```
picklelo, line 74: So will this return a success if the server starts or how does it work?

Alek99: If it doesn't error after a minute running it shuts this step dow

.github/workflows/integration.yml
Alek99: ah yes ill get rid of it duplicated

.github/workflows/integration.yml
Alek99: Ill name it to integration-test

picklelo (approved):
Looks good to me

PR #153, opened 2022-12-21
Improve event processing performance

picklelo:
* Reorder if statements in format_state to handle most common vars first (this provided a big speedup)
* Use `super().__getattribute__` in functions in the critical path of event processing. It's ugly but it's faster.
* Add `onopen` for web socket to hydrate the page on initial load.

### DIFF
--- a/pynecone/.templates/web/utils/state.js
+++ b/pynecone/.templates/web/utils/state.js
@@ -124,10 +124,15 @@ export const updateState = async (state, result, setResult, router, socket) => {
  * @param endpoint The endpoint to connect to.
  */
 export const connect = async (socket, state, result, setResult, router, endpoint) => {
+  // Create the socket.
   socket.current = new WebSocket(endpoint);
+
+  // Once the socket is open, hydrate the page.
   socket.current.onopen = () => {
     updateState(state, result, setResult, router, socket.current)
   }
+
+  // On each received message, apply the delta and set the result.
   socket.current.onmessage = function (update) {
     update = JSON.parse(update.data);
     applyDelta(state, update.delta);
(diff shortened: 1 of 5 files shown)


picklelo:
Test code: 
```
# Perf testing...
import asyncio

from pynetree.pynetree import app
from pynecone.event import Event
from pynecone.app import process
token = '0'

async def do_loop():
    event = Event(token=token, name="counter_example_state.increment")
    update = await process(app, event)
    print(update)

async def main():
    for i in range(10000):
        print("loop", i)
        await do_loop()
    
asyncio.run(main())
```
Before: 
Screen Shot 2022-12-20 at 4 08 02 PM

After:
Screen Shot 2022-12-20 at 10 01 12 PM

FHU-yezi:
Remove the `print` line in performance benchmark, it will slow this code down.

picklelo:
> Remove the `print` line in performance benchmark, it will slow this code down.

It does slow the runtime, but I'm only visualizing inside the `process` function so the results should still be valid.

pynecone/.templates/web/utils/state.js
```diff
@@ -123,8 +123,16 @@ export const updateState = async (state, result, setResult, router, socket) => {
  * @param setResult The function to set the result.
  * @param endpoint The endpoint to connect to.
  */
-export const connect = async (socket, state, setResult, endpoint) => {
+export const connect = async (socket, state, result, setResult, router, endpoint) => {
+  // Create the socket.
   socket.current = new WebSocket(endpoint);
```
Alek99, line 128: Will we eventually change this so its on socket per app

pynecone/utils.py
```diff
@@ -851,8 +851,16 @@ def format_state(value: Any) -> Dict:
     Raises:
         TypeError: If the given value is not a valid state.
     """
+    # Handle dicts.
```
Alek99, line 854: Nice the reordering speed up

Alek99 (approved):

pynecone/.templates/web/utils/state.js
picklelo: I'm not sure tbh, we will need to investigate this.

PR #159, opened 2022-12-22
Add Node Version Error Handling

Alek99:


pynecone/utils.py
```diff
@@ -298,6 +299,44 @@ def get_config() -> Config:
...
+            ["node", "-v"], stdout=subprocess.PIPE, stderr=subprocess.PIPE
+        )
+        # The output will be in the form "vX.Y.Z", so we can split it on the "v" character and take the second part
+        version = result.stdout.decode().strip().split("v")[1]
+        # Split the version string on the "." character and convert each part to an integer
+        version_parts = [int(x) for x in version.split(".")]
+        # Split the minimum version string on the "." character and convert each part to an integer
+        min_version_parts = [int(x) for x in min_version.split(".")]
+        # Compare the version parts to the minimum version parts
+        if version_parts[0] < min_version_parts[0] or (
```
picklelo, line 323: I *think* you can simplify this to just
```
return version_parts > min_version_parts
```
Since Python should do an element-wise comparison on the lists.

pynecone/utils.py
```diff
@@ -306,7 +345,16 @@ def get_package_manager() -> str:
 
     Raises:
         FileNotFoundError: If bun or npm is not installed.
+        Exit: If the app directory is invalid.
+
     """
+    # Check that the node version is valid.
+    if not (check_node_version(constants.MIN_NODE_VERSION)):
```
picklelo, line 352: Don't need the parentheses around the function call

pynecone/utils.py
```diff
@@ -298,6 +299,44 @@ def get_config() -> Config:
...
+        Whether the version of Node.js is high enough.
+    """
+    try:
+        # Run the node -v command and capture the output
+        result = subprocess.run(
+            ["node", "-v"], stdout=subprocess.PIPE, stderr=subprocess.PIPE
+        )
+        # The output will be in the form "vX.Y.Z", so we can split it on the "v" character and take the second part
+        version = result.stdout.decode().strip().split("v")[1]
+        # Split the version string on the "." character and convert each part to an integer
```
picklelo, line 318: Right here I think you can just
```
return version.split(".") > min_version.split(".")
```

pynecone/utils.py
```diff
@@ -298,6 +299,44 @@ def get_config() -> Config:
...
+    Returns:
+        Whether the version of Node.js is high enough.
+    """
+    try:
+        # Run the node -v command and capture the output
+        result = subprocess.run(
+            ["node", "-v"], stdout=subprocess.PIPE, stderr=subprocess.PIPE
+        )
+        # The output will be in the form "vX.Y.Z", so we can split it on the "v" character and take the second part
+        version = result.stdout.decode().strip().split("v")[1]
```
picklelo, line 317: I don't think we need this line either if we don't cast to int

pynecone/utils.py
Alek99: Yeah good catch, gpt complicated it

pynecone/utils.py
Alek99: I actually need this line because the output has v

picklelo (approved):
Sweet

PR #163, opened 2022-12-22
Improve event processing performance

picklelo:
* Remove all special `getattr` and `setattr` logic previously included to handle some edge cases of substates accessing parent state vars. From now on, that use case will have to be more explicit.
* Change gunicorn log level to `critical` - we had left it as `debug` which may also have impacted performance

Performance after this change:
Screen Shot 2022-12-21 at 6 14 44 PM

Compared to before: https://github.com/pynecone-io/pynecone/pull/153

### DIFF
--- a/tests/test_state.py
+++ b/tests/test_state.py
@@ -411,12 +411,15 @@ def test_get_child_attribute(TestState, ChildState, ChildState2, GrandchildState
         GrandchildState: The grandchild state class.
     """
     test_state = TestState()
+    child_state = test_state.get_substate(["child_state"])
+    child_state2 = test_state.get_substate(["child_state2"])
+    grandchild_state = child_state.get_substate(["grandchild_state"])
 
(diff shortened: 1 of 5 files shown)


FHU-yezi:
Maybe use error level for production and debug level for debug mode?

picklelo:
> Maybe use error level for production and debug level for debug mode?

Made this update

tests/test_state.py
```diff
@@ -695,18 +637,3 @@ async def test_process_event_substate(TestState, ChildState, GrandchildState):
         "test_state.child_state.grandchild_state": {"value2": "new"},
         "test_state": {"sum": 3.14, "upper": ""},
     }
-
```
Alek99, line 698: Nice I think it won't be a big deal getting rid of this

Alek99 (approved):
Nice looks good glad

PR #164, opened 2022-12-22
Add template version check

picklelo:
* Add `pcversion.txt` file in the template - we should update this whenever we change the template.
* If the app was initialized with an old version, ask the user to run `pc init` again.
* Update the poetry project to v0.1.11


pynecone/.templates/web/pcversion.txt
```diff
@@ -0,0 +1 @@
+0.1.11
```
Alek99, line 1: What the reason for using a txt file vs having a variable in the constants.py

Alek99 (changes requested):

pynecone/.templates/web/pcversion.txt
Alek99: Oh wait nvm I get it because if they update pc the web folder could be made from the older one

Alek99 (approved):
This makes sense to me nicely done clever with the txt file

pynecone/.templates/web/pcversion.txt
picklelo: Yeah I tried the config way but this seemed easier

Issue #157, opened 2022-12-21
Add option to specify port


picklelo:
Either in the `pcconfig` or as a flag to `pc run` we should be able to set the port on which we run the app.

Alek99:
Both would be nice I see the config being useful for repeat runs but flag would be nice for sh scripts

FHU-yezi:
And throw an error if this port is already in use, our default behavior now is to find another port avaliable.

pysqz:
please see this pr [#192](https://github.com/pynecone-io/pynecone/pull/192)

PR #192, opened 2022-12-30
Add option to specify port

pysqz:
fix issue [#157](https://github.com/pynecone-io/pynecone/issues/157)

### DIFF
--- a/pynecone/utils.py
+++ b/pynecone/utils.py
@@ -14,6 +14,7 @@ import string
 import subprocess
 import sys
 import uvicorn
+from urllib.parse import urlparse
 from collections import defaultdict
 from subprocess import PIPE
 from types import ModuleType
@@ -500,7 +501,10 @@ def run_frontend(app: App):
 
     # Run the frontend in development mode.
     console.rule("[bold green]App Running")
-    subprocess.Popen([get_package_manager(), "run", "dev"], cwd=constants.WEB_DIR)
+    os.environ["PORT"] = get_config().port
+    subprocess.Popen(
+        [get_package_manager(), "run", "dev"], cwd=constants.WEB_DIR, env=os.environ
+    )
 
 
 def run_frontend_prod(app: App):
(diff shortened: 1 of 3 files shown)


pynecone/utils.py
```diff
@@ -533,6 +537,13 @@ def get_num_workers() -> int:
     return (os.cpu_count() or 1) * 2 + 1
 
 
+def get_api_port():
```
picklelo, line 540: This needs a return type annotation `-> str` as well as a doctoring to make the build happy

pynecone/utils.py
```diff
@@ -533,6 +537,13 @@ def get_num_workers() -> int:
...
+    result = urlparse(get_config().api_url)
+    if result.port is None:
+        return urlparse(constants.API_URL).port
```
picklelo, line 543: You may need to add an assert here for Pyright
```
port = urlparse(...)
assert port is not None
return port
```

pynecone/utils.py
```diff
@@ -559,6 +571,8 @@ def run_backend_prod(
     """
     num_workers = get_num_workers()
     command = constants.RUN_BACKEND_PROD + [
+        "--bind",
+        "0.0.0.0:%s"%get_api_port(),
```
picklelo, line 575: Can we use an f-string here for consistency with the rest of the code?

picklelo (changes requested):
Thanks for making this change!

Alek99:
Yes thanks for implementing this feature defiantly needed! 🙂

pysqz:
update the commit, pls review again~ @picklelo @Alek99

picklelo (approved):
Looks good to me - it just needs to be linted. You can run
```
poetry run black pynecone
```
and it should format the code for you, then we can merge this

pysqz:
> Looks good to me - it just needs to be linted. You can run
> 
> ```
> poetry run black pynecone
> ```
> 
> and it should format the code for you, then we can merge this

done

PR #207, opened 2023-01-04
Access parent state vars from substate

picklelo:


Alek99 (approved):
Nice re adding of this feature

PR #214, opened 2023-01-05
Added victory graphs.

Alek99:


pynecone/components/graphing/victory.py
```diff
@@ -0,0 +1,465 @@
...
+from pynecone.var import Var
+
+
+def format_line(x: List, y: List) -> List:
+    """Format line data."""
+    data = []
+    if len(x) != len(y):
+        raise ValueError("x and y must be the same length")
+
+    for i in range(len(x)):
```
picklelo, line 27: This can be simplified to
```
return [{"x": x[i], "y": y[i]} for i in range(len(x))]
```
I think it may be worth it to move that line out to a helper function since it's used so much in this file, such that you can just call something like.
```
return create_data({"x": x, "y": y})
```
So you can just give it the dict of lists and it can convert it to a list of dicts.
It can even the the length assertions within the helper functions instead of repeating it.

pynecone/components/graphing/victory.py
```diff
@@ -0,0 +1,465 @@
+"""Component for displaying a plotly graph."""
```
picklelo, line 1: Docstring needs update

Alek99:
Not ready yet just testing checks.

pynecone/components/graphing/victory.py
```diff
@@ -0,0 +1,465 @@
...
+    elif graph == "candlestick":
+        return format_candlestick(x, **kwargs)
+    elif graph == "error_bar":
+        return format_error_bar(x, y, **kwargs)  # type: ignore
+    else:
+        raise ValueError("Invalid graph type")
+
+
+class Victory(Component):
+    """A component that wraps a plotly lib."""
```
picklelo, line 393: Update docstring

tests/components/test_graphing.py
```diff
@@ -0,0 +1,161 @@
+from typing import List, Set, Type
```
picklelo, line 1: Great job adding these tests

pynecone/components/graphing/victory.py
```diff
@@ -0,0 +1,567 @@
...
+
+    Raises:
+        ValueError: If x and y are not the same length.
+    """
+    if x is None or y is None:
+        raise ValueError("x and y must be provided")
+
+    if amount is None:
+        return format_xy(x, y)
+    else:
```
picklelo, line 60: Don't need the `else` here since the `if` case returns. (Similarly in the other functions below)

pynecone/components/graphing/victory.py
```diff
@@ -0,0 +1,567 @@
...
+
+    if y is not None:
+        return format_xy(x, y)
+
+    else:
+        if min_ is None or max_ is None or median is None or q1 is None or q3 is None:
+            raise ValueError(
+                "min, max, median, q1, and q3 must be specified if y is not provided"
+            )
+        elif (
```
picklelo, line 149: The `elif` can just be `if` since the previous `if` raises an exception

pynecone/components/graphing/victory.py
```diff
@@ -0,0 +1,567 @@
...
+            or len(x) != len(median)
+            or len(x) != len(q1)
+            or len(x) != len(q3)
+        ):
+            raise ValueError(
+                "x, min, max, median, q1, and q3 must be the same length and specified if y is not provided"
+            )
+        for i in range(len(x)):
+            row = {}
+            if x is not None:
```
picklelo, line 161: Don't you already check that these are not `None` above?

pynecone/components/graphing/victory.py
```diff
@@ -0,0 +1,567 @@
...
+        y: The y values.
+
+    Returns:
+        The formatted data.
+
+    Raises:
+        ValueError: If x or y is not provided.
+        ValueError: If x and y are not the same length.
+    """
+    data = []
```
picklelo, line 236: Don't need this variable. Also, I think the `None` check can be moved into the `format_xy` function

pynecone/components/graphing/victory.py
```diff
@@ -0,0 +1,567 @@
...
+        The formatted data.
+
+    Raises:
+        ValueError: If graph is not provided.
+        ValueError: If graph is not supported.
+    """
+    if graph == "box_plot":
+        return format_box_plot(x, y, **kwargs)
+    elif graph == "bar":
+        return format_bar(x, y)  # type: ignore
```
picklelo, line 324: Why do these need the type ignores?

pynecone/components/graphing/victory.py
```diff
@@ -0,0 +1,567 @@
...
+    elif graph == "candlestick":
+        return format_candlestick(x, **kwargs)
+    elif graph == "error_bar":
+        return format_error_bar(x, y, **kwargs)  # type: ignore
+    else:
+        raise ValueError("Invalid graph type")
+
+
+class Victory(Component):
+    """A component that wraps a plotly lib."""
```
picklelo, line 346: plotly -> victory

pynecone/components/graphing/victory.py
```diff
@@ -0,0 +1,567 @@
...
+    # Sets the border width of the error bars.
+    borderWidth: Var[float]
+
+
+class Group(Victory):
+    """Display a victory group."""
+
+    tag = "VictoryGroup"
+
+    # Optional prop that defines a color scale to be applied to the children of the group. Takes in an array of colors. Default color scemes are: "grayscale", "qualitative", "heatmap", "warm", "cool", "red", "green", "blue".
```
picklelo, line 533: typo: color schemes

pynecone/components/graphing/victory.py
```diff
@@ -0,0 +1,567 @@
...
+    color_scale: Var[List]
```
picklelo, line 534: Should we make it `Var[List[str]]`

picklelo (commented):
Great stuff

pynecone/components/graphing/victory.py
Alek99: Its color scale for group

Alek99: Ill update the comment

pynecone/components/graphing/victory.py
Alek99: The reason is sometimes y is not provide and was getting a None type error

pynecone/components/graphing/victory.py
Alek99: Ah yeah that was from before when I was making it so you could pass in some of them but changed it to need all if y inset specified

picklelo (approved):
LGTM

PR #228, opened 2023-01-08
Add path event

### LINKED ISSUE #138: Access current page within state
Currently within the state event handlers (and computed var functions), we don't have access to the current page. We can add a method to the `pc.State` class called `get_current_page` that returns this info (potentially along with route params, query params, etc.)

Lendemor:
Closes #138

I needed this feature so I figured I would try to add it, turn out it didn't require too much effort.

Notes:
- There is a modification in .templates/web/utils/state.js , if I'm getting it right, this file is copied during `pc init`, so I'm not sure how to update it for existing project? I haven't seen a `pc update` command to refresh `.web`?
- I tried to pick an intuitive name for the variable, if it's no good I can change it of course.

EDIT: I need to read slower, I noticed original issue mention a `get_current_page()` method, I'll have to change some things then.

### DIFF
--- a/pynecone/.templates/web/utils/state.js
+++ b/pynecone/.templates/web/utils/state.js
@@ -71,7 +71,8 @@ export const applyDelta = (state, delta) => {
 export const applyEvent = async (event, router, socket) => {
   // Handle special events
 
-  event.path = router.pathname;
+  event.router_data = router;
+
   if (event.name == "_redirect") {
     router.push(event.payload.path);
     return;
--- a/pynecone/.templates/web/utils/state.js
+++ b/pynecone/.templates/web/utils/state.js
@@ -70,6 +70,8 @@ export const applyDelta = (state, delta) => {
  */
 export const applyEvent = async (event, router, socket) => {
   // Handle special events
+
+  event.path = router.pathname;
   if (event.name == "_redirect") {
     router.push(event.payload.path);
     return;
--- a/pynecone/.templates/web/utils/state.js
+++ b/pynecone/.templates/web/utils/state.js
@@ -71,7 +71,7 @@ export const applyDelta = (state, delta) => {
 export const applyEvent = async (event, router, socket) => {
   // Handle special events
 
-  event.router_data = router;
+  event.router_data = (({ pathname, query }) => ({ pathname, query }))(router);
 
   if (event.name == "_redirect") {
     router.push(event.payload.path);
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -262,7 +262,7 @@ class State(Base, ABC):
             field.default = default_value
 
     def get_current_page(cls) -> str:
-        """Obtain the path of current page from the router data"""
+        """Obtain the path of current page from the router data."""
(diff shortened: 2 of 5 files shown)


pynecone/.templates/web/utils/state.js
```diff
@@ -70,6 +70,9 @@ export const applyDelta = (state, delta) => {
  */
 export const applyEvent = async (event, router, socket) => {
   // Handle special events
+
+  event.router_data = router;
```
picklelo, line 74: There seems to be many extra fields in `router` - we can subselect to just the ones we need:
```
event.router_data = (({ pathname, query }) => ({ pathname, query }))(router);
```

pynecone/state.py
```diff
@@ -257,6 +260,11 @@ def _set_default_value(cls, prop: BaseVar):
             field.required = False
             field.default = default_value
 
+    @classmethod
+    @functools.lru_cache()
```
picklelo, line 264: This shouldn't be a classmethod since it applies to the instance of the state, can just be:
```
def get_current_page(self) -> str:
    return self.router_data["pathname"]
```

This also needs a docstring.

pynecone/state.py
```diff
@@ -45,6 +45,9 @@ class State(Base, ABC):
     # The set of dirty substates.
     dirty_substates: Set[str] = set()
 
+    # The routing path that triggered the state
+    router_data: Dict[str, Any]
```
picklelo, line 49: We can default this to an empty dictionary:
```
router_data: Dict[str, Any] = {}
```

Also, we need to update:
* this file on line 90, add `router_data` to the set of `skip_vars`
* `tests/test_state.py`, on line 151, add `router_data` to the tuple of fields to get the unit test passing

picklelo (changes requested):
Thanks for this change! Just a few comments

picklelo:
> There is a modification in .templates/web/utils/state.js , if I'm getting it right, this file is copied during pc init, so I'm not sure how to update it for existing project? I haven't seen a pc update command to refresh .web?

You can just run `pc init` on your existing project and it will update the template files. We will make this more clear in the docs.

picklelo (approved):
Awesome

picklelo:
Seems to be a couple lint errors. Can you run these and make sure they pass locally:

```
poetry run pydocstyle pynecone tests
find pynecone tests -name "*.py" -not -path pynecone/pc.py | xargs poetry run darglint
```



```

File: pynecone/state.py [hunk 4]
```diff
@@ -257,6 +261,10 @@ def _set_default_value(cls, prop: BaseVar):
             field.required = False
             field.default = default_value
 
+    def get_current_page(cls) -> str:
+        """Obtain the path of current page from the router data."""
```
picklelo, line 265: Needs a return docstring:
```
"""Obtain the path of current page from the router data.

Returns:
    The current page.
"""
```

Lendemor:
Note:  `darglint` was not installed by poetry to run the tests, had to manually install it.

Issue #237, opened 2023-01-10
Add Datepicker Component (pc.datepicker)


Alek99:
This would be a good first issue for someone as seen in our docs someone can wrap  a datepicker.

For an intro on how to add a component see our docs: https://pynecone.io/docs/advanced-guide/wrapping-react.

The specs for this component would depend on the data picker library wrapped, or we could event use the base html datepicker.

qu3vipon:
I did some research and couldn't complete this work, but here is a summary of my research.
Let me first reveal that I am not familiar with ReactJS.

1. There are two different libraries: `react-date-picker` and `react-datepicker`
    - I can't tell which one is better. 
    - Both of them have a little problem when importing to pynecone.
        - `react-date-picker` has some conflicts when used with next.js.
        - So I had to replace import statements like below.
            ```javascript
            // index.js
            import DatePicker from "react-date-picker/dist/entry.nostyle"
            import "react-date-picker/dist/DatePicker.css"
            import "react-calendar/dist/Calendar.css"
            ```
       - Ref: https://github.com/wojtekmaj/react-date-picker/issues/340)
       
2. I choose `react-datepicker` because its documentation is more detailed. Here is my code.
    ```python
    class DatePicker(Component):
        library = "react-datepicker"
        tag = "DatePicker"  
      
        value: Var[str]  
        placeholder_text: Var[str]
    
	    # date_format: Var[str]
	    # time_format: Var[str]
	    # locale: Var[str]
      
        @classmethod  
        def get_controlled_triggers(cls) -> Set[str]:  
            return {"on_change", "on_select"}  
      
        def _get_imports(self) -> ImportDict:  
            return utils.merge_imports(  
                super()._get_imports(), {"": {"react-datepicker/dist/react-datepicker.css"}},  
            )
    
    class State(pc.State): 
        date: str = ""  
      
    def index():  
        return pc.center(  
            pc.date_picker(  
                value=State.date,  
                on_change=State.set_date,  
                placeholder_text="Select a date..",  
            )  
        )
    ```
    
3. After I compiled the app, it was rendered nicely. But when I selected a date, the value returned by the callback function was a string in iso format (i.e. 2023-01-01T00:00:00Z). I tried to format this string with the props the library provides, but every prop didn't work as I want.

Hope that it helps.

FHU-yezi:
Can we use `datetime.fromisoformat` to deal with this in Python side?

qu3vipon:
[...]

PR #240, opened 2023-01-10
FEATURE: added a datepicker module(#237)

muddi900:
Adds #237 

Right now it only had the `value` prop, but I will add more as per comments.

### DIFF
--- a/pynecone/components/__init__.py
+++ b/pynecone/components/__init__.py
@@ -15,6 +15,7 @@ from .media import *
 from .navigation import *
 from .overlay import *
 from .typography import *
+from .datepicker import *
 
 # Add the convenience methods for all the components.
 locals().update(
@@ -26,6 +27,8 @@ locals().update(
 )
 
 # Add responsive styles shortcuts.
+
+
 def mobile_only(*children, **props):
     """Create a component that is only visible on mobile.
 
--- a/pynecone/components/datepicker/__init__.py
+++ b/pynecone/components/datepicker/__init__.py
@@ -1,0 +1,1 @@
+from .datepicker import DatePicker
--- a/pynecone/components/datepicker/datepicker.py
+++ b/pynecone/components/datepicker/datepicker.py
@@ -1,8 +1,9 @@
-import pynecone as pc
+from pynecone.components.component import Component
 from pynecone.var import Var
 
 
-class DatePicker(pc.Component):
+class DatePicker(Component):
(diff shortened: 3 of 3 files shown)


Alek99:
Thanks for taking this on a couple things. The library seems to be `react-datepicker` does your code work locally. You will also need to add this library to our package.json

As for props and events 
-In addition to onChange we should support onSelect

More props can be these but snake case: closeOnScroll, dateFormat, startDate, endDate, minDate, maxDate, isClearable, placeholderText, showWeekNumbers, shouldCloseOnSelect

muddi900:
I have added the new props. However, should the date values be of `datetime` type? Or the framework handle it on its end.

I ahve not tested locally. I am working in a Codespace. The tests are running fine.

Alek99:
Looking good thanks for the changes, can you add comment and doc strings like our other components also can you text to see if it runs locally. https://github.com/pynecone-io/pynecone/blob/main/pynecone/components/datadisplay/datatable.py

muddi900:
I have added the docstring.

I am trying to test locally but I getting a `ReferenceError: date_picker_state is not defined`. I am following the [wrapping docs](https://pynecone.io/docs/advanced-guide/wrapping-react) quite literally. Do I need to define the state separately in the components?

here is my code.

```python
class DatePickerState(pc.State):
    date = "1/1/2001"


def index():
    return pc.center(
        pc.vstack(
            pc.heading(DatePickerState.date),
            date_picker(on_change=DatePickerState.set_date),
            pc.box("Get started by editing ",
                   pc.code(filename, font_size="1em")),
            pc.link(
                "Check out our docs!",
                href=docs_url,
                border="0.1em solid",
                padding="0.5em",
                border_radius="0.5em",
                _hover={
                    "color": "rgb(107,99,246)",
                },
            ),
            spacing="1.5em",
            font_size="2em",
        ),
        padding_top="10%",
    )


app = pc.App(state=State)
app.add_page(index)
app.compile()
```

picklelo:
It looks like you're passing the wrong State class to your app.

You probably want to do
```
app = pc.App(state=DatePickerState)
```

Or if you meant to use substates, you need to make sure your new class subclasses your base state rather than `pc.State`:

```
class State(pc.State):
    """"Base State"""

class DatePickerState(State):
    date = "1/1/2001"
```

Alek99:
Going to close this pr seems inactive.

PR #241, opened 2023-01-10
fix to wait for router isReady on page load

Lendemor:
fix needed to receive the content of `query` when the page initially load


picklelo (approved):
amazing

PR #245, opened 2023-01-11
Add var length operation

picklelo:
* Add var length operation
* Update template to v0.1.12 in preparation for next release
* Fix some router handling
* Add var length operation


pynecone/components/component.py
```diff
@@ -480,7 +488,7 @@ def get_custom_components(self) -> Set[CustomComponent]:
         return (
             {self}
             | super().get_custom_components()
-            | self.get_component().get_custom_components()
+            # | self.get_component().get_custom_components()
```
picklelo, line 491: I will add this back in a follow-up, it's currently causes an infinite recursion in some cases.

Alek99 (commented):
looks good

Alek99 (approved):

PR #252, opened 2023-01-13
Feature/113 follow up

ElijahAhianyo:
113-Pynecone-support-for-windows-symlink-follow-up

- [x] Removed symlink logic for non-windows
- [x] Added Test case for test setting up frontend logic


picklelo (approved):
Great job!

PR #260, opened 2023-01-14
Add optional catch all handling

Lendemor:
Added some methods to handle the optional catchall from nextjs dynamic routes.

Include some refactoring to make it work all together.

### DIFF
--- a/pynecone/app.py
+++ b/pynecone/app.py
@@ -214,8 +214,8 @@ class App(Base):
         Args:
             new_route: the route being newly added.
         """
-        catchall_pattern = utils.catchall_in_route(new_route)
-        if not catchall_pattern:
+        newroute_catchall = utils.catchall_in_route(new_route)
+        if not newroute_catchall:
             return
 
         for route in self.pages:
@@ -226,16 +226,13 @@ class App(Base):
                     f"You cannot define a route with the same specificity as a optional catch-all route ('{route}' and '{new_route}')"
                 )
 
-            if utils.catchall_in_route(route) and catchall_pattern:
+            route_catchall = utils.catchall_in_route(route)
+            if route_catchall and newroute_catchall:
                 # both route have a catchall, check if preceding path is the same
-                # ic("double slug", route, new_route)
                 if utils.catchall_prefix(route) == utils.catchall_prefix(new_route):
                     raise ValueError(
-                        f"You cannot use different slug names for the same dynamic path ({route} !== {new_route})"
+                        f"You cannot use multiple catchall for the same dynamic path ({route} !== {new_route})"
                     )
-
-            if route == "[...slug]" and new_route == "[[...slug]]":
-                raise ValueError("")
 
     def compile(self, force_compile: bool = False):
         """Compile the app and output it to the pages folder.
--- a/pynecone/app.py
+++ b/pynecone/app.py
@@ -236,6 +236,7 @@ class App(Base):
 
             if route == "[...slug]" and new_route == "[[...slug]]":
                 raise ValueError("")
+
     def compile(self, force_compile: bool = False):
         """Compile the app and output it to the pages folder.
 
--- a/pynecone/app.py
+++ b/pynecone/app.py
@@ -199,10 +199,43 @@ class App(Base):
 
         # Format the route.
         route = utils.format_route(path)
-
         # Add the page.
+        self._check_routes_conflict(route)
         self.pages[route] = component
 
+    def _check_routes_conflict(self, new_route: str):
+        """Verify if there is any conflict between the new route and any existing route.
+
+        Based on conflicts that NextJS would throw if not intercepted.
+
+        Raises:
+            ValueError: exception showing which conflict exist with the path to be added
+
+        Args:
+            new_route: the route being newly added.
+        """
+        catchall_pattern = utils.catchall_in_route(new_route)
+        if not catchall_pattern:
+            return
+
+        for route in self.pages:
+            route = "" if route == "index" else route
+
+            if new_route.startswith(route + "/[[..."):
+                raise ValueError(
+                    f"You cannot define a route with the same specificity as a optional catch-all route ('{route}' and '{new_route}')"
+                )
+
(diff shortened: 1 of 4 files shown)


pynecone/app.py
```diff
@@ -199,10 +199,44 @@ def add_page(
...
+            route = "" if route == "index" else route
+
+            if new_route.startswith(route + "/[[..."):
+                raise ValueError(
+                    f"You cannot define a route with the same specificity as a optional catch-all route ('{route}' and '{new_route}')"
+                )
+
+            if utils.catchall_in_route(route) and catchall_pattern:
+                # both route have a catchall, check if preceding path is the same
+                # ic("double slug", route, new_route)
```
picklelo, line 231: leftover debug statement

pynecone/app.py
```diff
@@ -199,10 +199,44 @@ def add_page(
...
+                if utils.catchall_prefix(route) == utils.catchall_prefix(new_route):
+                    raise ValueError(
+                        f"You cannot use different slug names for the same dynamic path ({route} !== {new_route})"
+                    )
+
+            if route == "[...slug]" and new_route == "[[...slug]]":
+                raise ValueError("")
```
picklelo, line 238: Should we have a message in this error?

pynecone/constants.py
```diff
@@ -188,3 +188,17 @@ def get_url(self) -> str:
 
         # Return the url.
         return url
+
+
+# ROUTE REGEXs
+class Regex(str, Enum):
```
picklelo, line 194: Maybe call this class `RouteRegex` for clarity?

pynecone/constants.py
```diff
@@ -188,3 +188,17 @@ def get_url(self) -> str:
...
+    """Regex used for extracting path args in path."""
+
+    ARG = r"\[(?!\.)([^\[\]]+)\]"
+    CATCHALL = r"(\[?\[\.{3}(?![0-9]).*\]?\])"  # group return the catchall pattern (i.e "[[..slug]]")
```
picklelo, line 198: nitpick: Can we put the comments directly above the field like:
```
# Group return the catchall pattern (i.e "[[..slug]]")
CATCHALL = r"(\[?\[\.{3}(?![0-9]).*\]?\])"
```

pynecone/utils.py
```diff
@@ -799,41 +794,79 @@ def get_path_args(path: str) -> Dict[str, str]:
...
-    Raises:
-        ValueError: explains what is wrong with the path.
-
     Returns:
         The path arguments.
     """
-    # Import here to avoid circular imports.
-    from pynecone.var import BaseVar
+
+    def add_path_arg(match, type_):
```
picklelo, line 801: Missing type annotations for args

pynecone/utils.py
```diff
@@ -799,41 +794,79 @@ def get_path_args(path: str) -> Dict[str, str]:
...
+        if arg_name in args:
+            raise ValueError(
+                f"arg name [{arg_name}] is used more than once in this URL"
+            )
+        args[arg_name] = type_
 
     # Regex to check for path args.
-    check = re.compile(r"^\[(.+)\]$")
-    check_catchall = re.compile(r"^\[\.\.\.(.+)\]$")
+    check = re.compile(constants.Regex.ARG)
```
picklelo, line 819: Maybe we can just compile these within the `constants` rather than doing it every time in this function. So the Enum contains the regexes instead of the str value. Not sure if that would complicate things though

pynecone/utils.py
```diff
@@ -799,41 +794,79 @@ def get_path_args(path: str) -> Dict[str, str]:
...
     args = {}
     for part in path.split("/"):
-        match = check_catchall.match(part)
-        if match:
-            arg_name = match.groups()[0]
-            if arg_name in args:
-                raise ValueError(f"arg [{arg_name}] is used more than once in this URL")
+        match_opt = check_opt_catchall.match(part)
+        if match_opt:
+            add_path_arg(match_opt, "catchall")
```
picklelo, line 828: Let's define `"catchall"` and `"patharg"` in the constants rather than passing string values (also in `state.py` where it's used)

pynecone/utils.py
```diff
@@ -799,41 +794,79 @@ def get_path_args(path: str) -> Dict[str, str]:
...
             # Add the path arg to the list.
-            arg_name = match.groups()[0]
-            if arg_name in args:
-                raise ValueError(f"arg [{arg_name}] is used more than once in this URL")
-            args[arg_name] = "patharg"
+            add_path_arg(match, "patharg")
     return args
 
 
+def catchall_in_route(route) -> str:
```
picklelo, line 843: Missing arg type annotation (I'll try to make the linter stricter on this to catch earlier)

pynecone/utils.py
```diff
@@ -799,41 +794,79 @@ def get_path_args(path: str) -> Dict[str, str]:
...
+    """Extract the catchall part from a route.
+
+    Args:
+        route: the route from which to extract
+
+    Returns:
+        str: the catchall part of the URI
+    """
+    pattern = re.compile(constants.Regex.CATCHALL)
```
picklelo, line 852: Same comment about maybe having it compiled in the constants already

picklelo (changes requested):
Just a couple style comments :)

pynecone/app.py
Lendemor: it's leftover code from the testing, I'll remove it altogether, it's never used

pynecone/utils.py
Lendemor:
```
class ArgKind(Enum):
    SINGLE = "arg_simple"
    LIST = "arg_list"
```

pynecone/utils.py
FHU-yezi: Yes it will make regexs faster and make startup time a little bit longer, but I have tested it and found compile regex for ten thousand times only cost less than 0.01s, so it is worth to do this.

picklelo (approved):
Thanks for the updates!

PR #265, opened 2023-01-15
add 404 custom page handling

### LINKED ISSUE #234: Create custom 404 page

Lendemor:
Closes #234 

Allow to add a default 404 page:

Will add a catchall route that match all undefined route, and display the given page.

### DIFF
--- a/pynecone/app.py
+++ b/pynecone/app.py
@@ -234,6 +234,38 @@ class App(Base):
                         f"You cannot use multiple catchall for the same dynamic path ({route} !== {new_route})"
                     )
 
+    def add_custom_404_page(self, component, title=None, image=None, description=None):
+        """Define a custom 404 page for any url having no match.
+
+        If there is no page defined on 'index' route, add the 404 page to it.
+        If there is no global catchall defined, add the 404 page with a catchall
+
+        Args:
+            component: The component to display at the page.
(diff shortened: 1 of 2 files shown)


pynecone/app.py
```diff
@@ -234,6 +234,38 @@ def _check_routes_conflict(self, new_route: str):
                         f"You cannot use multiple catchall for the same dynamic path ({route} !== {new_route})"
                     )
 
+    def add_custom_404_page(self, component, title=None, image=None, description=None):
```
picklelo, line 237: What if we make it so the user never has to call this method, instead when they do
```
app.add_page(my_404_component, path="/404")
```
we can call this if we detect the path is `/404`? Just to reduce the number of methods we expose to the end user.

Then also a lot of the common logic for the `add_meta`, etc can stay in one function and this one only handles modifying the route.

picklelo (commented):
Code looks good, just an idea to possibly simplify the API we expose

pynecone/app.py
picklelo: Talked offline - let's keep this since it provides clarity/flexibility. We can possibly combine it into `add_page` with a flag later.

picklelo (approved):
Awesome

PR #269, opened 2023-01-16
Empty .web/pages directory on recompile

andy-verstraeten:
Empty the .web/pages directory on recompile so that old pages that don't exist anymore in the python code get removed.

### DIFF
--- a/pynecone/compiler/compiler.py
+++ b/pynecone/compiler/compiler.py
@@ -211,7 +211,6 @@ def compile_components(components: Set[CustomComponent]):
 
 
 def purge_web_pages_dir():
-    """Empty out .web directory.
-    """
-    template_files = ["_app.js","404.js"]
+    """Empty out .web directory."""
+    template_files = ["_app.js", "404.js"]
     utils.empty_dir(constants.WEB_PAGES_DIR, ignored=template_files)
--- a/pynecone/compiler/compiler.py
+++ b/pynecone/compiler/compiler.py
@@ -213,4 +213,4 @@ def compile_components(components: Set[CustomComponent]):
 def purge_web_pages_dir():
     """Empty out .web directory."""
     template_files = ["_app.js", "404.js"]
-    utils.empty_dir(constants.WEB_PAGES_DIR, ignored=template_files)
+    utils.empty_dir(constants.WEB_PAGES_DIR, keep_files=template_files)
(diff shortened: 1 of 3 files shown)


pynecone/compiler/compiler.py
```diff
@@ -208,3 +208,9 @@ def compile_components(components: Set[CustomComponent]):
     # Compile the components.
     code = _compile_components(components)
     return output_path, code
+
+
+def purge_web_pages_dir():
+    """Empty out .web directory.
+    """
+    utils.empty_dir(constants.WEB_PAGES_DIR)
```
picklelo, line 216: There's two special files in the pages directory - `._app.js` and `404.js` that are part of the template and not recompiled. I think we need to exclude those two files when purging the dir.

picklelo (changes requested):
Thanks for digging into this!

andy-verstraeten:
Both files were added to an ignore list.

pynecone/compiler/compiler.py
```diff
@@ -208,3 +208,10 @@ def compile_components(components: Set[CustomComponent]):
     # Compile the components.
     code = _compile_components(components)
     return output_path, code
+
+
+def purge_web_pages_dir():
+    """Empty out .web directory.
+    """
+    template_files = ["_app.js","404.js"]
```
picklelo, line 216: I think you forgot to pass this in as an argument

pynecone/compiler/utils.py
```diff
@@ -304,3 +304,17 @@ def write_page(path: str, code: str):
...
+
+def empty_dir(path, ignored=[]):
+    """Remove all files and folders in a directory except for the ignored file- or foldernames.
+
+    Args:
+        path (str): The path to the directory that will be emptied
+        excluded (list, optional): List of filenames or foldernames that will not be deleted. Defaults to [].
+    """
+    directory_contents = os.listdir(path)
+    print(directory_contents)
```
picklelo, line 316: We can remove this print

andy-verstraeten:
My bad.

picklelo (approved):
Awesome!

picklelo:
Sorry, needs one more change to make the linter happy (I think it wants the docstring on a single line). Can you make sure the following checks pass:
```
poetry run pyright pynecone tests
poetry run pydocstyle pynecone tests
find pynecone tests -name "*.py" -not -path pynecone/pc.py | xargs poetry run darglint
```

picklelo:
Thank you for the persistence 😄

PR #315, opened 2023-01-22
Polars df compatibility

suarzac:
Hi, not sure if this should be an issue or not but hopefully this will work to start a discussion.
This is really just a quick workaround for making the compiler work with polars dataframes.

### DIFF
--- a/pynecone/utils.py
+++ b/pynecone/utils.py
@@ -980,6 +980,7 @@ def get_default_app_name() -> str:
     """
     return os.getcwd().split(os.path.sep)[-1].replace("-", "_")
 
+
 def is_dataframe(value: Type) -> bool:
     """Check if the given value is a dataframe.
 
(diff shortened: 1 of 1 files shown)


picklelo:
Thanks for making this! I'll test it locally then we can merge it in

pynecone/utils.py
```diff
@@ -1019,10 +1019,18 @@ def format_state(value: Any) -> Dict:
 
     # Convert pandas dataframes to JSON.
     if is_dataframe(type(value)):
-        return {
-            "columns": value.columns.tolist(),
-            "data": value.values.tolist(),
-        }
+        try:
```
picklelo, line 1022: Maybe instead of doing a try/except, we can add a similar function called `is_polars_dataframe`. The only catch is we don't want `polars` as a dependency to pynecone, so we need to do something similar where we check the `__name__` of the value.

picklelo (changes requested):

pynecone/utils.py
suarzac: Hey, sorry for the late reply.
I've added an is_polars_dataframe function that checks the str of the type object for 'polars'.

picklelo (approved):
Sweet!

picklelo:
I think this needs to be linted, see the bottom of [this](https://github.com/pynecone-io/pynecone/blob/main/CONTRIBUTING.md)

picklelo (changes requested):
Run black on this

PR #316, opened 2023-01-22
Add event token to router_data

### LINKED ISSUE #296: Expose client token in state
Sometimes the event handlers may need to know the specific client that is calling it, to do logic such as rate-limiting.

Each client has a token already, so we can just make this accessible in the state class as `self.get_token()`

Lendemor:
* also clean up other part of the code related to router_data, by adding some literal in `constants.py`
* added some tests that were missing for new State methods

closes #296


picklelo (approved):
Awesome!

Issue #282, opened 2023-01-18
Stream did not contain valid UTF-8


Iced0368:
**Describe the bug**
When characters such as Chinese characters and Hangul are initially rendered, they are converted into invalid characters and entered in index.js. Therefore, a stream did not contain valid UTF-8 error occurs.

For example:
```
def index():
    return pc.container(
        pc.text("한글")
    )

app = pc.App()
app.add_page(index)
app.compile()
```
this code makes an error

```
class State(pc.State):
    text = "한글"

def index():
    return pc.container(
        pc.text(State.text)
    )

app = pc.App(state=State)
app.add_page(index)
app.compile()
```

but this works.
This obviously looks like a bug.

**Screenshots**
![invalid](https://user-images.githubusercontent.com/103271471/213213765-b5fc0353-9138-4b2d-aabd-945d87047282.png)

index.js when the first code was compiled

** Specifics (please complete the following information):**
 - Python Version: 3.10.xx
 - Pynecone Version: 0.1.12
 - OS: Windows

martiniifun:
I think my problem is similar.
I wrote Pynecone's "To-do List" demo in your gallery.
The placeholder on <Input> component, almost hangul characters go like
![image](https://user-images.githubusercontent.com/62679294/213522276-55de5daf-7651-4170-97a2-3bb73ca45b45.png)

nice2cu1:
same quetision.I use Chinese to display the text and I hope to be able to solve this problem as soon as possible. @Alek99

LumiaGG:
I have a similar problem.After hard coding Chinese in the .py, then the compiled Corresponding Chinese of .js is garbled.

Alek99:
Fix will be push soon this is the pr https://github.com/pynecone-io/pynecone/pull/332

DeathShort:
The bug is caused by pynecone.components.tags.tag ,when format the props, the json.dumps function user ascii encode as default. The solution is add 'ensure_ascii=False' in every case of json.dumps uses.
For example, change the code
 ` return json.dumps(prop.full_name)`
 to 
`return json.dumps(prop.full_name, ensure_ascii=False)`

PR #332, opened 2023-01-24
fix  Stream did not contain valid UTF-8 (#282)

LumiaGG:


Alek99:
I think you need to run black

Alek99 (approved):
Nice, thanks

Issue #310, opened 2023-01-22
Uncaught JS Error : Passing a var instead of event handler to an event_trigger 


Lendemor:
**Describe the bug**
When you pass the wrong kind of argument to an event trigger (var instead of event handler), it result in a runtime JS error instead of compilation python error.

**To Reproduce**
```
class State(pc.State):
    some_var: bool

def index():
    return pc.box(pc.icon(tag="AddIcon", on_click=State.some_var))
```

**Expected behavior**
Compilation should catch this error and throw a Python traceback to identify the problem

**Screenshots**
![image](https://user-images.githubusercontent.com/4015177/213922749-8ec72fa4-8d0a-46d2-bbe8-cce9e496f963.png)

** Specifics (please complete the following information):**
 - Python Version: 3.10
 - Pynecone Version: 0.1.13
 - OS: Ubuntu 22.04
 - Browser (Optional): Chrome

PR #338, opened 2023-01-25
fix: validate event

qu3vipon:
Resolve: #310 

Validate EventChain explicitly, so other types will be caught as a ValueError

> ValueError: Invalid event chain: {state. some_var}

### DIFF
--- a/pynecone/components/component.py
+++ b/pynecone/components/component.py
@@ -157,9 +157,10 @@ class Component(Base, ABC):
         Raises:
             ValueError: If the value is not a valid event chain.
         """
-        # If it's already an event chain, return it.
-        if isinstance(value, EventChain):
-            return value
+        # If it's a custom component and var, return it.
+        if isinstance(self, CustomComponent):
+            if isinstance(value, Var):
+                return value
 
         arg = self.get_controlled_value()
 
--- a/pynecone/components/component.py
+++ b/pynecone/components/component.py
@@ -158,7 +158,7 @@ class Component(Base, ABC):
             ValueError: If the value is not a valid event chain.
         """
         # If it's already an event chain, return it.
-        if isinstance(value, Var):
+        if isinstance(value, EventChain):
             return value
 
         arg = self.get_controlled_value()


pynecone/components/component.py
```diff
@@ -158,7 +158,7 @@ def _create_event_chain(
             ValueError: If the value is not a valid event chain.
         """
         # If it's already an event chain, return it.
-        if isinstance(value, Var):
+        if isinstance(value, EventChain):
```
picklelo, line 161: I think here the code is correct but the comment is incorrect.

Can we revert the code, but fix the comment to say:
```python
# If it's a var, return it.
```

picklelo: Maybe in addition, we need the `EventChain` check? But we definitely need this current one. What code did you run that raised the `ValueError`?

picklelo (changes requested):

pynecone/components/component.py
qu3vipon: EventChain check is for CustomComponent. 

I run the same code from #310.
```python
class State(pc.State):
    some_var: bool

def index():
    return pc.box(pc.icon(tag="AddIcon", on_click=State.some_var))
```

some_var is passed as a BaseVar, it is a child class of Var, so it is caught at first if statement.

qu3vipon: Can you explain more about why we definitely need the current one when we create EventChain? BaseVar shouldn't be returned in my opinion.

picklelo: Ahh I see - you're right this is causing the other bug it looks like. We need this line for the [CustomComponent](https://github.com/pynecone-io/pynecone/blob/a4ac2979c815d417e84699c2a764dbefa873776e/pynecone/components/component.py#L415) feature that I'm working on (it's not released yet, but it will allow you to create [memoized React components ](https://beta.reactjs.org/reference/react/memo)for better performance).

But I think in adding that line it caused this other bug. Let me think a bit more on how to resolve this.

picklelo: Maybe we can add a line like
```python
if isinstance(self, CustomComponent):
    if isinstance(value, Var):
        return value
```
Then it should fix this bug, but still work for the new feature.

qu3vipon: Ok. Should I change the code as you suggest? or just close this pr and wait for your new feature?

qu3vipon: I just did a quick fix as you suggest.

picklelo (approved):
thanks for the fix!

Issue #334, opened 2023-01-24
Create pc.dict and pc.list


picklelo:
Currently Pynecone doesn't detect mutations in list/dict state vars which is annoying and can lead to bugs if one is not careful.

Maybe we could create classes `pc.dict` and `pc.list` that wrap any dict/list State vars, and modify them such that they detect mutations.

TommyDew42:
I can work on this. Can you provide an example how `pc.dict` & `pc.list` would look like? My guess would be sth like this:
```py
def index():
  colors = pc.list(['red', 'oragen'])

  return pc.vstack(
    pc.text('Number of colors:' + len(colors)),
    pc.button('+', on_click=lambda : colors.append('another color'))
  )
```

Lendemor:
There was some discussion on Discord about that, but the idea would be to use `pc.list` and `pc.dict` behind the hood.

If you declare:
```
class State(pc.State):
    somevar: list[str] = []
```
then using State.somevar should behave like a list, even if we actually replace it by a `pc.list`

(same thing with dict)

But maybe it can be done in two step: 
1. Making the actual classes `pc.list` and `pc.dict` (so they can be used explicitly)
2. Use them implicitly in `pc.State`

FHU-yezi:
Yeh, we can use a real list / dict to store data, and rewrite __setitem__ to detect any change to it.

picklelo:
@TommyDew42 implemented `pc.list`, now we just need `pc.dict`

picklelo:
Had to [comment this out temporaril](https://github.com/pynecone-io/pynecone/pull/407)y due to an issue with pickling the pc.list, which we need for Redis. Will prioritize getting this back in.

ElijahAhianyo:
> Had to [comment this out temporarily](https://github.com/pynecone-io/pynecone/pull/407) due to an issue with pickling the pc.list, which we need for Redis. Will prioritize getting this back in.

Is there an issue created for this?

Lendemor:
We need to use the actual name `PCList` instead of `pc.list`, because the latter refer to the List component in `pynecone/components/datadisply/list.py`

Since `PCList` is used completely internally ( pynecone users should not have to declare a PCList ) then it doesn't need a shortcut in this style :  `pc.*`

And for consistency, refer to the `dict` wrapper as `PCDict`.

PR #339, opened 2023-01-25
Add and use `pc.list` to have list mutation detection

TommyDew42:
To add and use `pc.list`/`PcList` to have list mutation detection.

It's not as straightforward as I thought. Or is my approach too complicated?

For #334. `pc.dict` would be in another PR.

### DIFF
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -13,7 +13,7 @@ from redis import Redis
 from pynecone import constants, utils
 from pynecone.base import Base
 from pynecone.event import Event, EventHandler, window_alert
-from pynecone.var import BaseVar, ComputedVar, Var
+from pynecone.var import BaseVar, ComputedVar, PcList, Var
 
 Delta = Dict[str, Any]
 
@@ -60,6 +60,38 @@ class State(Base, ABC):
         # Setup the substates.
         for substate in self.get_substates():
             self.substates[substate.get_name()] = substate().set(parent_state=self)
+
+        self._init_mutable_fields()
+
+    def _init_mutable_fields(self):
+        """Initialize mutable fields.
+
+        So that mutation to them can be detected by the app:
+        * list
+        """
+        for field in self.base_vars.values():
+            if field.type_ is list:
+                setattr(
+                    self,
+                    field.name,
+                    PcList(
+                        getattr(self, field.name),
+                        reassign_field=self._reassign_field,
+                        field_name=field.name,
+                    ),
+                )
+
+    def _reassign_field(self, field_name: str):
+        """Reassign the given field.
+
+        Args:
+            field_name (str): The name of the field we want to reassign
+        """
+        setattr(
+            self,
+            field_name,
+            getattr(self, field_name),
+        )
 
     def __repr__(self) -> str:
         """Get the string representation of the state.
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -70,15 +70,16 @@ class State(Base, ABC):
         * list
         """
         for field in self.base_vars.values():
-            if field.type_ is list:
-                setattr(
-                    self,
-                    field.name,
-                    PcList(
-                        getattr(self, field.name),
-                        reassign_field=lambda: self._reassign_field(field.name),
-                    ),
-                )
+            value = getattr(self, field.name)
+
+            value_in_pc_data = _convert_mutable_datatypes(
+                value, self._reassign_field, field.name
+            )
+
+            if utils._issubclass(field.type_, List):
+                setattr(self, field.name, value_in_pc_data)
+
+        self.clean()
 
     def _reassign_field(self, field_name: str):
         """Reassign the given field.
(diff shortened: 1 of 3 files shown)


pynecone/var.py
```diff
@@ -748,3 +748,54 @@ def type_(self):
         if "return" in self.fget.__annotations__:
             return self.fget.__annotations__["return"]
         return Any
+
+
+class PcList(list):
```
TommyDew42, line 753: `PcList` sounds more reasonable to me than just `List` cuz it's a list object specifically for pcnecone. IMO, `List` is appropriate when it is just a list but with more generalised capabilities.

pynecone/var.py
```diff
@@ -748,3 +748,54 @@ def type_(self):
...
+
+    def __delitem__(self, *args, **kargs):
+        """Delete item.
+
+        Args:
+            args: The args passed.
+            kargs: The kwargs passed.
+        """
+        super().__delitem__(*args, **kargs)
+        self._reassign_field()
```
TommyDew42, line 801: Did I miss any methods? We have `append`, `__setitem__`, and `__delitem__` now

pynecone/state.py
```diff
@@ -61,6 +61,39 @@ def __init__(self, *args, **kwargs):
...
 
+        self._init_mutable_fields()
+
+    def _init_mutable_fields(self):
+        """Initialize mutable fields.
+
+        So that mutation to them can be detected by the app:
+        * list
+        """
+        for field in self.base_vars.values():
```
picklelo, line 72: This will work for the common use case - one thing we need to think deeper about is how to treat nested types.

For example, a dict with a list value, a subclass of `pc.Base` with a list field, etc. We may later need to add some sort of recursive inspection and convert everything to `PcList`. Or state clearly that we only support this use case and the mutable notice still applies for the other case.

pynecone/state.py
```diff
@@ -61,6 +61,39 @@ def __init__(self, *args, **kwargs):
...
+            if field.type_ is list:
```
picklelo, line 73: We can use `utils._issubclass(field.type_, List)` here which will work for things like `typing.List[int]` also.

Also let's use `List` everywhere because we need to support Python 3.7+

pynecone/state.py
```diff
@@ -61,6 +61,39 @@ def __init__(self, *args, **kwargs):
...
+                setattr(
+                    self,
+                    field.name,
+                    PcList(
+                        getattr(self, field.name),
+                        reassign_field=lambda: self._reassign_field(field.name),
+                    ),
+                )
+
+    def _reassign_field(self, field_name: str):
```
picklelo, line 83: Nice this is a clean way to do this

pynecone/var.py
picklelo: Yup sounds good!

pynecone/var.py
```diff
@@ -748,3 +748,54 @@ def type_(self):
...
+
+        Args:
+            original_list (list): The original list
+            reassign_field (Callable):
+                The method in the parent state to reassign the field.
+                Default to be a no-op function
+        """
+        self._reassign_field = reassign_field
+
+        super().__init__(original_list)
```
picklelo, line 771: Put the super call first?

pynecone/var.py
picklelo: We could add these too: https://www.w3schools.com/python/python_ref_list.asp (can be in a follow up)

pynecone/var.py
```diff
@@ -748,3 +748,54 @@ def type_(self):
...
             return self.fget.__annotations__["return"]
         return Any
+
+
+class PcList(list):
+    """A custom list that pynecone can detect its mutation."""
+
+    def __init__(
+        self,
+        original_list: list,
```
picklelo, line 758: Use `List[Any]` here

pynecone/state.py
TommyDew42: Created `_convert_mutable_datatypes` in e2dbc362265d0f434180bbb81fc2fd6439965e6e to recursively convert `list` to `PcList`. But now only `list` & `dict` would be iterated to recursively do the conversion.

pynecone/state.py
```diff
@@ -61,6 +61,40 @@ def __init__(self, *args, **kwargs):
...
+        """Initialize mutable fields.
+
+        So that mutation to them can be detected by the app:
+        * list
+        """
+        for field in self.base_vars.values():
+            value = getattr(self, field.name)
+
+            value_in_pc_data = _convert_mutable_datatypes(
+                value, self._reassign_field, field.name
```
TommyDew42, line 76: I found sth interesting. The previous way where we use `lambda` to bind the `field.name` to `self._reassign_field` wouldn't work. The `field.name` would be dynamically evaluated to be the last `field.name` instead of the `field.name` in the current iteration.

So let's say you have 3 fields in a state and the last `state.name` is `friends_in_nested_list` (i.e. the case in the unit test), the `field_name` would always be `friends_in_nested_list`. 😓 I have no choice but to pass both into the class, instead of use `lambda` to piece them together. 

Any ideas for this problem?

pynecone/var.py
TommyDew42: More methods are added in e2dbc362265d0f434180bbb81fc2fd6439965e6e

pynecone/var.py
kbrgl: Small nit, but `PcList` doesn't follow PEP-8 naming convention, which we should defer to when possible. Prefer `PCList`:

> *Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.*

Source: [PEP-8](https://peps.python.org/pep-0008/)

picklelo:
Sorry for the delay in reviewing this, will get to it soon!

pynecone/var.py
TommyDew42: Thanks for catching that! [fd7c722](https://github.com/pynecone-io/pynecone/pull/339/commits/fd7c722759bc0e45631f10030e5446935b2f5df3)

pynecone/state.py
picklelo: Ahh I think @Lendemor was seeing a similar issue in handling the dynamic routing - I think this code may be relevant where he used a factory? https://github.com/pynecone-io/pynecone/blob/5aae6a122d8935aea599b5a9f8bf84622fda9cdb/pynecone/state.py#L291

I'm not entirely sure though may have to give this a deeper look.

pynecone/state.py
```diff
@@ -61,6 +61,40 @@ def __init__(self, *args, **kwargs):
...
+            )
+
+            if utils._issubclass(field.type_, List):
```
picklelo, line 79: Should we create the `value_in_pc_data` inside this `if` statement? Since it's not used outside of it

pynecone/var.py
picklelo: Nice thanks

picklelo (approved):
Awesome job! Tested this out on an app locally and it works nicely. May be some edge cases we need to fix in the future, but will go ahead and merge this in now :)

TommyDew42:
@picklelo @kbrgl thanks for the reviews.

@picklelo What are the edge cases we haven't covered in this PR? Will have a follow up PR for them.

PR #340, opened 2023-01-25
Added anonymous telemetry helpers.

Alek99:


pynecone/utils.py
```diff
@@ -1228,3 +1230,46 @@ def get_redis() -> Optional[Redis]:
...
+    return multiprocessing.cpu_count()
+
+
+def add_telemetry() -> Dict:
+    """Add telemetry to an event.
+
+    Returns:
+        The payload with telemetry added.
+    """
+    payload = {}
```
picklelo, line 1268: I think we can create a class `Telemetry(pc.Base)` with these fields to make it strongly typed - then we can call the `.json()` method to get this dict.

We can make a new file `telemetry.py` and define the class and this method there

tests/test_telemetry.py
```diff
@@ -0,0 +1,29 @@
+import pytest
```
picklelo, line 1: Thanks for adding tests

pynecone/telemetry.py
```diff
@@ -0,0 +1,51 @@
...
+    """Add telemetry to an event.
+
+    Returns:
+        The payload with telemetry added.
+    """
+    payload = {}
+    payload["os"] = get_os()
+    payload["cpu_count"] = get_cpu_count()
+    # Display memory in MB.
+    payload["memory"] = psutil.virtual_memory().total >> 20
```
picklelo, line 48: We can put this in a helper also in a follow-up

picklelo (approved):
Nice

pynecone/telemetry.py
Alek99: Yeah not done yet still need to make the class aswell for telemetry

PR #348, opened 2023-01-26
Use ReconnectingWebSocket instead of vanilla WebSocket

advo-kat:
Currently Pynecone doesn't handle websocket onclose events. Using reconnecting-websocket handles such events automatically. 

Note that this does not 100% solve the issues which arise on mobile devices, especially iPhones which aggressively control network connection to preserve battery.


picklelo (approved):
Thanks for adding

Issue #349, opened 2023-01-26
Throw better error message for invalid event handler signature


picklelo:
Some event handlers are expected an argument.

For example when setting an input, you need

```python

class State(pc.State):
    text: str
    
    def change_text(self, text):
        self.text = text.upper()
```
Then call it with
```python

pc.input(on_change=State.change_text)
```

But if by mistake you define your event handler as

```python

def change_text(self):
    self.text = text.upper()
```

it throws a gross list index error. This should instead be a nice clean error message.

PR #363, opened 2023-01-27
error message for invalid event handler signature

qu3vipon:
Resolve: #349 

Couldn't get enough context inside this util function, so just let a developer know which function used is invalid.

It prints out,
> TypeError: Event handler has an invalid signature, needed a method with a parameter, got fn=<function State.change_text at 0x113131670>.


r0b2g1t (approved):

picklelo (approved):
Nice thanks for the explicit message

PR #366, opened 2023-01-28
Reconnect websocket on window focus as a solution to disconnects on mobile devices

advo-kat:
There appears to be some issue with ReconnectingWebSocket (and likely the base websocket) that causes onclose onerror events not to tigger when the device hibernates the window, especially on iOS Safari, preventing syncing of the front-end unless the page is refreshed. 

Instead of refreshing the page the best and simplest solution I've found is to simply reconnect the current websocket on window focus. This happens quickly and is seamless and ensures there's always a live websocket available. 

It isn't the perfect approach but it solves the problem until perhaps a different websocket library (such as Websocket.IO) is used.

You may be wondering what `{SOCKET}.current !== 'undefined'` is doing. 

Diagnosing the problem and trying to reproduce it is sometimes hard because there's different ways in which for e.g. sleeps tabs/pages. It might be voodoo, but I've found performing this check seems to make the fix work more consistently. I have no idea why that would be the case so it might be unnecessary.


pynecone/compiler/templates.py
```diff
@@ -164,8 +164,14 @@ def format_state(
...
         "    return;",
         "  }}",
-        f"  if (!{SOCKET}.current) {{{{",
-        f"    connect({SOCKET}, {{state}}, {{set_state}}, {RESULT}, {SET_RESULT}, {ROUTER}, {EVENT_ENDPOINT})",
+        "  const reconnectSocket = () => {{",
+        f"    {SOCKET}.current.reconnect()",
+        "  }}",
+        f"  if (typeof {SOCKET}.current !== 'undefined') {{{{",
+        f"    if (!{SOCKET}.current) {{{{",
+        f"      window.addEventListener('focus', reconnectSocket)",
```
picklelo, line 172: Good find

picklelo (approved):
Sweet!

Issue #353, opened 2023-01-27
Support multiple style dicts for components


picklelo:
Currently you can only pass a single style to a component:

```python
pc.box(style=style_dict)
```

We should allow passing a list of styles (such that the styles are applied sequentially, with the later ones overriding keys in the earlier ones):

```python
pc.box(style=[style1, style2])
```

Which is basically the equivalent of the current:

```python
pc.box(style={**style1, **style2})
```

PR #370, opened 2023-01-28
feat: multiple styles for components

qu3vipon:
Resolve: #353 

looping through a list and overwriting earlier ones.

This renders a yellow box:
```python
def index():
    return pc.box("box", style=[{"background_color": "orange"}, {"background_color": "yellow"}])

app = pc.App(state=State, style={pc.Box: {"background_color": "red"}})
app.add_page(index)
app.compile()
```


picklelo (approved):
Awesome!

Issue #346, opened 2023-01-26
Investigate dynamic var default value


picklelo:
In @Lendemor's work in dynamic vars, we encountered an issue that the default value for the dynamic vars is always `None`. 

```python
@classmethod
def _set_default_value(cls, prop: BaseVar, default_value: Any = None):
    """Set the default value for the var.

    Args:
       prop: The var to set the default value for.
    """
    # Get the pydantic field for the var.
    field = cls.get_fields()[prop.name]
    default_value = default_value if default_value else prop.get_default_value()
    if field.required and default_value is not None:
        field.required = False
        field.default = default_value
```

Setting the default value here doesn't work.

Lendemor:
I think this issue can be closed, default value problems was solved in the PR above.

picklelo:
Ah okay sweet, closing this.

PR #381, opened 2023-01-29
Dynamically add vars to a State

### LINKED ISSUE #303: Dynamically add state vars
Currently all state vars need to be defined as fields within the State class. Sometimes it is useful to dynamically add vars after the state class is already defined.

Minimally this requires:
* logic to set all the relevant state var info: https://github.com/pynecone-io/pynecone/blob/main/pynecone/state.py#L114:
```python
        for prop in cls.base_vars.values():
            if not utils._issubclass(prop.type_, utils.StateVar):
                raise TypeError(
                    "State vars must be primitive Python types, "
                    "Plotly figures, Pandas dataframes, "
                    "or subclasses of pc.Base. "
                    f'Found var "{prop.name}" with type {prop.type_}.'
                )
            cls._set_var(prop)
            cls._create_setter(prop)
            cls._set_default_value(prop)
```
* adding it as a Pydantic field

Then we should be able to do something like
```python
State.add_var(pc.BaseVar(name="myvar", type_="str"))
```

Lendemor:
## Description:
Add a new classmethod `add_var` to `pc.State` allowing to add dynamically a var to a State, that can then be used in the same way as a var defined statically.

To add a variable called `"dynamic_var"` of type `int` with default value `42`: 
```
State.add_var("dynamic_var", int, 42)
```
This variable can then be accessed in two way : 
```
State.dynamic_var # access it the same way you would a static var
getattr(State, "dynamic_var") # access it like this if the name of the variable to access change at runtime
```

Closes #303 #346 

> Note: I added some basic tests for the new method but they could probably be improved.
 
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

### DIFF
--- a/pynecone/base.py
+++ b/pynecone/base.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 from typing import Any, Dict, TypeVar
 
 import pydantic
+from pydantic.fields import ModelField
 
 # Typevar to represent any class subclassing Base.
 PcType = TypeVar("PcType")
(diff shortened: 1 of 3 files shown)


picklelo (approved):
Great job finding this

PR #386, opened 2023-01-29
Check if curl and unzip are installed, fix #383

### LINKED ISSUE #383: pc init does not fail when curl and unzip are not installed.
**Description**
When curl and unzip is not installed the `pc init` command shows Finished Initializing even though it failed.  When you go on to do a `pc run` pynecone fails to find bun.  I hesitate to call this a bug; however, this might improve usability for first time users.

**To Reproduce**
 - On a linux pc without curl or unzip installed create a conda environment using mamba
 - mamba create -n pc python
 - mamba activate pc
 - mamba install nodejs
 - pip install pynecone-ip
 - pc init

**Expected behavior**
Expect Failed Initializing <app name >: curl was not installed.
Or add curl, unzip to the dependencies.

**Screenshots**

 - Python Version: 3.10
 - Pynecone Version: 0.1.8
 - OS:  WSL Debian
 - Browser (Optional): NA
 
**Additional context**

HellAmbro:
As @brianmcconnel noted, Pynecone initialization may fail if curl and unzip are not installed.
Checks whether they are installed, if not the user is notified with a proper message.


picklelo (approved):
Thanks for adding

pynecone/utils.py
```diff
@@ -422,6 +422,17 @@ def install_bun():
     # Only install if bun is not already installed.
     if not os.path.exists(get_package_manager()):
         console.log("Installing bun...")
+
+        # Check if curl is installed
+        curl_path = which("curl")
+        if curl_path is None:
+            raise FileNotFoundError("Pynecone requires curl to be installed.")
```
picklelo, line 429: We need to add this to the docstring:
```
"""Install bun onto the user's system.

Raises:
    FileNotFoundError: If the required packages are not installed.
"""
```

picklelo (changes requested):
Just one change to make the linter happy

pynecone/utils.py
HellAmbro: Done.

picklelo (approved):
Awesome

PR #390, opened 2023-01-30
allow backend variables starting with _

### LINKED ISSUE #307: Allow backend-only state vars
Currently all state vars must be JSON serializable to send the the frontend.

However sometimes it could be useful to have state fields that are only on the backend (i.e. generator objects) that never are called within render functions. These fields would be accessible within the event handlers, but not from the frontend. And they wouldn't have to adhere to the serialization requirement.

Lendemor:
## Description
Allow the use of variables starting with _ within a State.
Those variable will be considered as backend variables, and will never be sent to the frontend.
You can read/write a backend variable from within an event handler.

Backend variables can also be used within a computed var to generate a value that will be sent to the frontend.  

(This is a non-breaking change because pydantic already ignored any variable starting with _)

Closes #307 

> Note: I added test for the utils function `is_backend_variable()` but I'm not too sure how to test the backend variable, I'm open to suggestions if I need to add them.

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

### DIFF
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -33,6 +33,9 @@ class State(Base, ABC):
     # Vars inherited by the parent state.
     inherited_vars: ClassVar[Dict[str, Var]] = {}
 
+    # Backend vars that are never sent to the client.
+    backend_vars: ClassVar[Dict[str, Any]] = {}
+
     # The parent state.
     parent_state: Optional[State] = None
 
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -117,6 +117,12 @@ class State(Base, ABC):
         if parent_state is not None:
             cls.inherited_vars = parent_state.vars
 
+        cls.backend_vars = {
+            name: value
+            for name, value in cls.__dict__.items()
+            if utils.is_backend_variable(name)
+        }
+
         # Set the base and computed vars.
         skip_vars = set(cls.inherited_vars) | {
             "parent_state",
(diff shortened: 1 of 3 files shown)


pynecone/state.py
```diff
@@ -117,6 +117,12 @@ def __init_subclass__(cls, **kwargs):
         if parent_state is not None:
             cls.inherited_vars = parent_state.vars
 
+        cls.backend_vars = {
```
picklelo, line 120: Don't we need to add this as a Pydantic field to the `pc.State` class?

```python
class State(Base, ABC):
    """The state of the app."""
    ...
    # Vars inherited by the parent state.
    backend_vars: ClassVar[Dict[str, Var]] = {}
```

picklelo (commented):
Sweet - just one question otherwise looks good :)

pynecone/state.py
Lendemor: I didn't notice any problem without the declaration, but I agree we should add it.

However, `Dict[str, Var]` => `Dict[str, Any]` (because backend variables are never transformed into Var, they are simply pure Python variables.
I'm not sure what the `ClassVar` does, but if it's needed I can add it.

picklelo: ClassVar just makes it so it's not tied to the instance but the class instead. So it won't appear in the `.dict` and `.json` methods I believe.

pynecone/utils.py
```diff
@@ -1231,5 +1231,17 @@ def get_redis() -> Optional[Redis]:
     return Redis(host=redis_url, port=int(redis_port), db=0)
 
 
+def is_backend_variable(name: str) -> bool:
+    """Check if this variable name correspond to a backend variable.
+
+    Args:
+        name (str): The name of the variable to check
```
picklelo, line 1238: Nit: don't need the type in the docstring since we use type annotations

picklelo (approved):

PR #402, opened 2023-01-30
Make db_url optional

Wazarr94:
This pull request allows to set db_url to None to prevent the creation of the sqlite file and the use of SQLAlchemy within the applicaton in general.
I have let the default of db_url as it is, I think it's best to let it this way.


picklelo (approved):
Nice

PR #407, opened 2023-01-31
Comment out pclist

picklelo:
We have to comment this out for now until it can be pickled to support Redis.


Alek99 (approved):

TommyDew42:
Thanks for reverting the change for me. We might want to have an unit test for the Redis behaviour here.

PR #417, opened 2023-02-01
Allow the user to change ports

HellAmbro:
With this update an user can change the port the app is running on at runtime, without killing any process.
Maybe the prompts can be changed.
I attach a screenshot to show how it works. 
I've also specified a custom port.
Schermata del 2023-02-01 01-05-04

### DIFF
--- a/pynecone/pc.py
+++ b/pynecone/pc.py
@@ -51,26 +51,26 @@ def init():
 
 @cli.command()
 def run(
-    env: constants.Env = typer.Option(
-        constants.Env.DEV, help="The environment to run the app in."
-    ),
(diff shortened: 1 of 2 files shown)


Alek99:
Nice I'll review this tonight, I like this feature though thanks for adding

HellAmbro:
Nice, I think that sometimes the user don't want to kill a process, and changing every time pcconfig and run again the app can be time consuming. In this way he can change the ports directly from CLI. It also checks if the new port is also in use and prompts the user again. 
Maybe prompts can be changed to be more readable.

Alek99 (approved):
Looks good to me we can improve prompts/cli message in following prs. Nice!

PR #436, opened 2023-02-04
Added a password component.

Imnaseli:
Feature addition: Password component

I added a password component.
I found out chakra dont have a separate password component, the type is just changed between "text" and "password".
so i simply imported the input component in a new class and changed the "type_" to password.
so that instead of pc.input(type_="password") users can just use pc.password().

### DIFF
--- a/pynecone/utils.py
+++ b/pynecone/utils.py
@@ -503,7 +503,7 @@ def export_app(app: App, zip: bool = False):
         if os.name == "posix":
             cmd = r"cd .web/_static && zip -r ../../frontend.zip ./* && cd ../.. && zip -r backend.zip ./* -x .web/\* ./assets\* ./frontend.zip\* ./backend.zip\*"
         if os.name == "nt":
-            cmd = r'powershell -Command "Set-Location .web/_static; Compress-Archive -Path .\* -DestinationPath ..\..\frontend.zip; Set-Location ..\..; Compress-Archive -Path .\* -DestinationPath backend.zip -Exclude .web\*, assets\*, frontend.zip, backend.zip"'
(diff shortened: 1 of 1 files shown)


picklelo (approved):
Awesome

pynecone/components/forms/password.py
```diff
@@ -0,0 +1,12 @@
...
+from typing import Set
+from .input import Input
+from pynecone.components.component import EVENT_ARG
+from pynecone.var import Var
+
+
+class Password(Input):
+    """The Password component is a component that is used to get password input from users."""
+
+    type_: Var[str] = "password"
```
picklelo, line 12: I think you need a `#type: ignore` at the end of this line to make Pyright happy

picklelo (changes requested):

pynecone/components/forms/password.py
Imnaseli: Okay will do that.

Imnaseli: I fixed it and pyright is now happy :D

picklelo (changes requested):
I think there's still lint errors: make sure all the checks in the [Contributing guide](https://github.com/pynecone-io/pynecone/blob/main/CONTRIBUTING.md) pass

Imnaseli:
im sorry i keep getting an error concerning the 'TestState' everytime i test it with pytest.
`
cannot collect test class 'TestState' because it has a __init__ constructor (from: tests/test_state.py)
 `
i believe it is the same error that the pull request of  #437 is trying to solve.
should i just close the pull request.

picklelo:
> im sorry i keep getting an error concerning the 'TestState' everytime i test it with pytest.
> `cannot collect test class 'TestState' because it has a __init__ constructor (from: tests/test_state.py)`
> i believe it is the same error that the pull request of #437 is trying to solve.
> should i just close the pull request.

The tests look like they're all passing in the Github Actions?

I think you just have to run
```
```

to lint the files, and it should be good to go!

PR #437, opened 2023-02-04
Fix test fails in windows PC

PeterYusuke:
### All Submissions:

### Description
Test output have fails releted windows routing.
I change route string and set TestState as not test class.

### DIFF
--- a/tests/test_state.py
+++ b/tests/test_state.py
@@ -21,6 +21,9 @@ class Object(Base):
 class TestState(State):
     """A test state."""
 
+    # Set this class as not test one
+    __test__ = False
+
     num1: int
     num2: float = 3.14
     key: str
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -97,7 +97,7 @@ def test_add_page_set_route(app: App, index_page, windows_platform: bool):
         index_page: The index page.
         windows_platform: Whether the system is windows.
     """
-    route = "\\test" if windows_platform else "/test"
+    route = "test" if windows_platform else "/test"
     assert app.pages == {}
     app.add_page(index_page, route=route)
     assert set(app.pages.keys()) == {"test"}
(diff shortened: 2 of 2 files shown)


tests/test_state.py
```diff
@@ -1,657 +1,660 @@
...
-    assert getattr(test_state, "dynamic_list") == [5, 10]
-
-    # how to test that one?
-    # test_state.dynamic_list.append(15)
-    # assert test_state.dynamic_list == [5, 10, 15]
-
-    test_state.add_var("dynamic_dict", Dict[str, int], {"k1": 5, "k2": 10})
-    assert test_state.dynamic_dict == {"k1": 5, "k2": 10}
-    assert getattr(test_state, "dynamic_dict") == {"k1": 5, "k2": 10}
+from typing import Dict, List
```
picklelo, line 1: Looks good, but why is the diff so large on this file? (I think there may be some extra carriage returns or some weird thing like that) - can we resolve this

picklelo (changes requested):

tests/test_state.py
FHU-yezi: Line breaks in Windows and Unix-like OS? When I switch from Windows to Linux a few months ago, similar thing happend.

PeterYusuke: I think this was caused line breaks too.
The diff is only few line at TestState class.

```py
class TestState(State):
    """A test state."""

    # Set this class as not test one   # <- these 2 lines
    __test__ = False                   # <-
```

I think I should revert them before the commit in order to delete line breaks, then commit again.
What do you think?

Alek99: I think that should work try that, or rebase on the current main branch and commit again

PeterYusuke:
I rebase on the current main branch and change line breaks.
Please check this again.

picklelo (approved):
Awesome!

Issue #287, opened 2023-01-18
Make websockets optional


picklelo:
We can have a config option to disable websockets and use normal http requests instead.

Advantages:
* Allow support for certain browsers where websockets aren't supported well
* May allow for easier deployment in some cases
* Can support rate-limiting

Disadvantages:
* We will have to support/test two code paths and make sure we always work with both websockets and http. But since the networking section of the code is pretty contained, this shouldn't be too bad.

FHU-yezi:
Maybe we can use HTTP as a fallback plan if:

- Error occured while create Websocket connection ( The browser does not support Websocket or there are something wrong in our backend )
- Websocket connection broken for a few times ( The network route can't handle it correctly, such as some firewalls )

And we should give our user a way to define which plan to use, maybe a `-c` / `-communication` arg with `ws`, `perfer-ws`, `http`?

FHU-yezi:
Maybe we can close it as #449 already merged.

picklelo:
Yup we now have fallback to HTTP

PR #449, opened 2023-02-05
Use Socket.IO for message transport

### LINKED ISSUE #258: Websockets connection drops on Safari/iOS
**Describe the bug**
The websocket connection is inconsistent on mobile sometimes for Safari/iOS.

**To Reproduce**
See related issue: https://github.com/HeinrichApfelmus/threepenny-gui/issues/130

**Expected behavior**
The connection should be resilient and reconnect when dropped.

### LINKED ISSUE #423: Brave browser websocket compatibility issue with Pynecone
**Describe the bug**
The way the Brave browser handles websockets often breaks many apps, not just Pynecone (like WhatsApp web). See: https://github.com/brave/brave-browser/issues/15410

**To Reproduce**
Use Brave browser and test https://counter.pynecone.app/

This has started occurring since the switch to ReconnectingWebSocket. The reconnection attempts likely trigger some sort of blocking mechanism in the Brave browser.

**Expected behavior**
Pynecone should work in Brave. But currently most of the onus to fix this seems to lie on Brave’s end as it affects multiple other platforms. Potentially using Socket.IO instead would work, since it has an automatic http fallback mechanism which reverts to long polling.

 - Python Version:
 - Pynecone Version:
 - OS: 
 - Browser (Optional):
 
**Additional context**

### LINKED ISSUE #424: Websocket was closed meanwhile asyncio tasks is running
**Describe the bug**
When you minimize your web page or go to another application, if you had any asyncio tasks running then your app will crash in the background.
Meaning that on the UI side, you have no feedback, everything seems to be ok but your app is now frozen and you can't do anything on it.
And you have to force refresh the page for it to work again. You can see the specific logs :
```
ERROR:    Exception in ASGI application
...
...
...
websockets.exceptions.ConnectionClosedOK: received 1000 (OK); then sent 1000 (OK)
```

**To Reproduce**
 - [Clock project example](https://github.com/pynecone-io/pynecone-examples/tree/main/clock)
 - ```pc init```
 - ```pc run```
 - Enable the switch ON to start the clock
 - Minimize your web browser or go to another app while the clock is running
 - Come back to the web browser clock page
 - Check your ```pc run``` logs and you will see the error and the clock will be stuck

**Expected behavior**
I dunno if this is expected behavior but I think background asyncio running tasks shouldn't make the UI crash if you leave the page.

**Screenshots**

 - Python Version: 3.10
 - Pynecone Version: 0.1.14
 - OS: WSL2
 - Browser (Optional): Edge

advo-kat:
### All Submissions:

### Type of change

### New Feature Submission:

### Description

A rework of the websocket implementation in Pynecone.

Benefits
1. [Socket.IO](https://socket.io/) is a modern socket library used by major companies
2. Starts connections using long polling (http/https) and upgrades to websocket as a fallback if websockets are blocked for some reason
3. It fixes the issues with Safari iOS and mobile devices (more testing needed for edge cases), potentially closes #258
4. Allows for seamless connections even when WiFi connection is lost and device switches to data
5. Removes need for reconnections on focus fix, which closes #424
6. Works on Brave browser which closes #423 
7. It is now theoretically possible to make websockets optional as a config option, as in #287, without separate code paths
8. Because this PR uses `AsyncServer`, it will be easier to implement implement server side / async events in the future
9. It is easy to create events which are emitted to all users, or groups of users ("rooms"), making it possible to build chat apps or even a Discord clone in Pynecone in the future 
10. Requires no change in server setup to deploy, the default `/socket.io` endpoint has been replaced with the usual `/event` endpoint

I've tested this PR locally and on a production server.


pynecone/app.py
```diff
@@ -85,24 +100,6 @@ def __call__(self) -> FastAPI:
         """
         return self.api
 
-    def add_default_endpoints(self):
```
picklelo, line 88: We should keep the `ping` endpoint

pynecone/app.py
```diff
@@ -85,24 +100,6 @@ def __call__(self) -> FastAPI:
...
-        """Add the default endpoints."""
-        # To test the server.
-        self.api.get(str(constants.Endpoint.PING))(ping)
-
-        # To make state changes.
-        self.api.websocket(str(constants.Endpoint.EVENT))(event(app=self))
-
-    def add_cors(self):
```
picklelo, line 96: Do we not need this anymore?

pynecone/app.py
```diff
@@ -327,52 +324,6 @@ def compile(self, force_compile: bool = False):
         compiler.compile_components(custom_components)
 
 
-async def ping() -> str:
```
picklelo, line 330: This can be useful for health checks, etc.

picklelo (commented):
Awesome job!! Tested locally and it worked nicely - just a couple comments

pynecone/app.py
```diff
@@ -64,10 +66,23 @@ def __init__(self, *args, **kwargs):
...
-
         self.api = FastAPI()
-        self.add_cors()
-        self.add_default_endpoints()
+
+        # Set up the Socket.IO AsyncServer.
+        self.sio = AsyncServer(async_mode="asgi", cors_allowed_origins="*")
+
+        # Create the socket app. Note 'event' replaces the default 'socket.io' path.
+        socket_app = ASGIApp(self.sio, socketio_path="event")
```
picklelo, line 75: Let's use constants.Endpoint.EVENT instead of a string here (and in the EventNameSpace below)

picklelo:
Also, can you update the `pynecone/.templates/web/pcversion.txt` file to `v0.1.15` - then it will let people know they have to `pc init` again

pynecone/app.py
advo-kat: I'll do that. Note that `str(constants.Endpoint.EVENT)` returns `"/event"` not `"event"` so I'll also update state.js accordingly, since those have to match.

advo-kat: Actually that might cause a bunch of confusion in the future, for example you call ping with `'ping'` but it returns on `'\ping'`. So I'll modify constants.py `Endpoint' to allow it to return the literal value without the leading '\'. 

Now that I think about it it's still confusing. There's also potential confusion between the URL endpoint `/event` and the socket namespace `event`. I'll separate those in constants.py.

pynecone/app.py
advo-kat: I've added it back in. For future reference it can be tested like so:

```
    socket.current.emit('ping')
    console.log('ping')
    socket.current.on('ping', function(data) {
      console.log(data);
    });
```

In the console you should see:

```
ping
pong
```

pynecone/app.py
advo-kat: Now that I think about it it's still confusing. There's also potential confusion between the URL endpoint `/event` and the socket `event` (they're separate concepts in socket.io). I'll separate those in constants.py.

pynecone/app.py
advo-kat: It's taken care of when the `AsyncServer` is created on the `App` `__init__` :

        self.sio = AsyncServer(async_mode="asgi", cors_allowed_origins="*")

There's also `cors_credentials=` which defaults to `True` already. 

python-scoketio does not have CORS options like `allow_methods` and `allow_headers` but I believe it's configured to avoid cross-origin problems with browsers by allowing all headers and methods anyway.

advo-kat:
Thanks for the feedback. Please see the latest commits and let me know if there's anything else!

JeremyLG:
Magnifique, you're the boss @advo-kat 
I can confirm it completely fixes the linked issue

You rock !

pynecone/app.py
picklelo: Okay got it nice

picklelo (approved):
Sweet great job!

pynecone/app.py
picklelo: Sounds good!

Issue #416, opened 2023-02-01
Add anchor tag component


picklelo:
We need a way to have anchor tags so you can navigate to a particular section of a page.

For example on our doc pages, you could then link a specific subsection.

PeterYusuke:
Hi!
Do you mind if I work on this issue?
I’m new to React thou.

picklelo:
Yes for sure! I don't have much experience in how to do this, but feel free to start and you can always hop on our Discord to brainstorm ideas or get help.

dhawalpatil200:
But there is already Link component for navigation ? Are there any other use case by adding anchor tag component ?

PR #469, opened 2023-02-07
Add anchor tag for hash link

PeterYusuke:
PR for issue #416.

### All Submissions:

### Type of change

### New Feature Submission:

Added anchor tag that created with key property at any component.
This allows us to jump to to a specific place using hash in URL like "destination/url/#target".

Usage:

```py
    pc.box(
        pc.text("This is target"),
        key="target",
    )
```

Above code creates this.
Anchor tag is placed at the first child tag.
```js
    <Box key="target">
        <a name="target" id="target"></a>
        <Text>{`This is target`}</Text>
    </Box>
```


Lendemor:
Is it not possible to add the `id` directly to the <Box> tag in this case?

PeterYusuke:
> Is it not possible to add the `id` directly to the `<Box>` tag in this case?

I think it's possible. Should I change this?

Alek99:
> > Is it not possible to add the `id` directly to the `<Box>` tag in this case?
> 
> I think it's possible. Should I change this?

Yeah it would simplify the compiled code a bit, can you try this and see if it works. Thanks for the contribution

picklelo (commented):
Good calling adding this `id` field . I'm a bit confused though - looks like the Anchor class got reverted? Don't we need that?

picklelo (approved):
Nvm, tried this locally and it works great - awesome work!

Lendemor:
> Good calling adding this `id` field . I'm a bit confused though - looks like the Anchor class got reverted? Don't we need that?

I think it was just there as a support for the `id` prop, but the Anchor itself didn't have a purpose.
Adding the `id` directly on the component work the same.

mister-rao:
was looking for anchor hash navigation. Adding id to components work, but is there any way to make the scrolling smooth ?

Issue #472, opened 2023-02-07
[Bug] Cannot export on Windows due to zip command


Cheesy-Brik:
**Describe the bug**
```
'zip' is not recognized as an internal or external command,
operable program or batch file.
```
Getting this error when trying to export, it says the files have been made but they aren't at the path it says they are so I assume it doesn't actually make them.

Windows 10, python 3.10.6

r0b2g1t:
Which command parameters do you use to export?

picklelo:
This is from running `pc export` - I think this issue is that `zip` doesn't exist on Windows - we need some alternative.

r0b2g1t:
Yes, it need zip as a command to export. I have also recognised that you can't disable zip for export.

r0b2g1t:
Maybe a PR which allows to read a parameter for enable of disable zip. Like the frontend and backend params for run.

I will try to create the PR tomorrow.

Alek99:
> Maybe a PR which allows to read a parameter for enable of disable zip. Like the frontend and backend params for run.
> 
> I will try to create the PR tomorrow.

Thanks let me know if this fixes this issue

r0b2g1t:
@Alek99, @Cheesy-Brik and @picklelo. I have create the PR #479 which includes the optional parameter to disable zipping.

Cheesy-Brik:
> @Alek99, @Cheesy-Brik and @picklelo. I have create the PR #479 which includes the optional parameter to disable zipping.

@r0b2g1t I raise you one better, exact same command but refactored for windows 
```
powershell -Command "Set-Location .web/_static; Compress-Archive -Path .\* -DestinationPath ..\..\frontend.zip; Set-Location ..\..; Compress-Archive -Path .\* -DestinationPath backend.zip -Exclude .web\*, assets\*, frontend.zip, backend.zip"
```
Should be added in https://github.com/pynecone-io/pynecone/blob/main/pynecone/utils.py#L502 using os.name == 'nt' to check for windows.

r0b2g1t:
That's awesome, I will take a look for integration it by identify the the os. thanks. :-)

r0b2g1t:
@Cheesy-Brik I have add the powershell command and also the os check.

PR #479, opened 2023-02-08
Add --no-zip pc CLI Parameter

r0b2g1t:
### All Submissions:

### Type of change

### New Feature Submission:

The PR provides the option to disable zipping for the backend and frontend exports via the `pc export --no-zip` command. That was requested by @Cheesy-Brik in the issue #472.


pynecone/pc.py
```diff
@@ -159,10 +163,17 @@ def export():
     # Compile the app in production mode and export it.
     utils.console.rule("[bold]Compiling production app and preparing for export.")
     app = utils.get_app().app
-    utils.export_app(app, zip=True)
-    utils.console.rule(
-        "Backend & Frontend compiled. See [green bold]backend.zip[/green bold] and [green bold]frontend.zip[/green bold]."
-    )
+    utils.export_app(app, zip=zipping)
+    if zipping:
```
picklelo, line 167: Now that you've added a way to do the zipping in Windows, should we remove this flag and just always do it?

pynecone/utils.py
```diff
@@ -500,7 +500,10 @@ def export_app(app: App, zip: bool = False):
 
     # Zip up the app.
     if zip:
-        cmd = r"cd .web/_static && zip -r ../../frontend.zip ./* && cd ../.. && zip -r backend.zip ./* -x .web/\* ./assets\* ./frontend.zip\* ./backend.zip\*"
+        if os.name == "posix":
+            cmd = r"cd .web/_static && zip -r ../../frontend.zip ./* && cd ../.. && zip -r backend.zip ./* -x .web/\* ./assets\* ./frontend.zip\* ./backend.zip\*"
+        if os.name == "nt":
```
picklelo, line 505: Would the check `platform.system() == "Windows"` work? This is what we use in other parts of the code, so would be good to stay consistent. We should probably just move this to `utils.py` as a function `def is_windows` or something later.

pynecone/pc.py
r0b2g1t: Maybe someone want to export it without zipping, so we can leave the parameter in. What do you think about it?

pynecone/utils.py
r0b2g1t: I also thought about that. The differences between os.name and platform.system() is that the os provides for Linux and OSX the value posix and for Windows the value nt. On the other side, platform.system provides for Linux  the value Linux, for OSX the value Darwin and for Windows Windows. That produce an additional check in the if statement. We can harmonise it to use platform.system if you think it is better. But I don't recommend to introduce a new is_windows function because the feature is already in os.name or platform.system.

picklelo: That makes sense - I think in other places we just put the Linux/OSX within the `else` case so we don't do an explicit check. But this looks good to me, let's ship it!

picklelo (approved):

Issue #420, opened 2023-02-01
Allowing any case style on all string built ins 


Al3xlol:
pc.button(
    pc.icon(tag="MoonIcon"),
    on_click=pc.toggle_color_mode,
)
has same behavior as pc.button(
    pc.icon(tag="moonicon"),
    on_click=pc.toggle_color_mode,
)
so make it non case sensitive by using .lower() thanks and make sure to have this for any strings related to Pynecone

elyanah-aco:
I'd like to work on this. Just need to know what specific components need to be changed under this

Alek99:
Hi thanks for taking a look at this, I think a good place to start is the icon component. Feel free to join the discord if you need any help or guidance, or post follow ups here

Al3xlol:
> Hi thanks for taking a look at this, I think a good place to start is the icon component. Feel free to join the discord if you need any help or guidance, or post follow ups here

^ yea icon components and other places in code base where strings are used I feel non case sensitive should be used

elyanah-aco:
This was a little more complicated to solve than I actually thought. The below code addition for `icon.py` works in testing but doesn't feel rigorous enough. What do you think?
```
 if "tag" not in props.keys():
            raise AttributeError("Missing 'tag' keyword-argument for Icon")
        else:
            tag_keywords = [
                "add", "allowed", "arrow", "back", "circle", "clock", "close", "down", "forward",
                "handle", "icon", "left", "link", "off", "outline", "right", "sign", "two", "up"
            ]
            props['tag'] = props['tag'].title()
            for keyword in tag_keywords:
                props['tag'] = re.sub(keyword, keyword.title(), props['tag'])
```

masenf:
Seems fixed via #553 unless there are other areas that are unaddressed.

```python
rx.icon(tag="RePeaT"),
```

is rendering correctly

PR #486, opened 2023-02-09
Remove case sensitivity for string built-in in icon component

elyanah-aco:
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

PR removes case sensitivity for the `tag` argument in `pc.Icon` as requested in #420. Admittedly, this fix doesn't feel 'rigorous' enough for me and may need to be updated if future icons are added, so feel free to make suggestions.


pynecone/components/media/icon.py
```diff
@@ -36,5 +38,30 @@ def create(cls, *children, **props):
             )
         if "tag" not in props.keys():
             raise AttributeError("Missing 'tag' keyword-argument for Icon")
+        else:
+            tag_keywords = [
```
FHU-yezi, line 42: May use tuple instead of list to avoid change this by accident?

picklelo: I'm a bit confused - what are the `tag_keywords` used for? Shouldn't the `.title()` call be enough?

elyanah-aco: No, that isn't enough for "multiword" tags (eg. `ExternalLinkIcon`, `RepeatClockIcon`). From what I found in testing, `tag` should still match actual Chakra icon name case to work

picklelo: Ah I see - maybe in those cases the user can specify the tag in `snake_case` like `pc.icon(tag="repeat_clock_icon")` and we can use the `utils.to_title_case` function in `utils.py`. What do you think?

elyanah-aco: Yes, that sounds better and more futureproof. We just need to make it clear that the user has to use `snake_case` in documentation

picklelo:
I think something got messed up on the rebase there's a lot of extra changes - could you try again to rebase on `main`?

elyanah-aco:
Will create new PR for this

PR #503, opened 2023-02-11
334: create PCDict

### LINKED ISSUE #334: Create pc.dict and pc.list

ElijahAhianyo:
This PR adds a PCDict feature to handle mutable dicts

### All Submissions:

### Type of change

fixes #334

### DIFF
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -5,7 +5,18 @@ import asyncio
 import functools
 import traceback
 from abc import ABC
-from typing import Any, Callable, ClassVar, Dict, List, Optional, Sequence, Set, Type
+from typing import (
+    Any,
+    Callable,
+    ClassVar,
+    Dict,
+    List,
+    Optional,
+    Sequence,
+    Set,
+    Type,
+    Union,
+)
 
 import cloudpickle
 from redis import Redis
@@ -79,9 +90,7 @@ class State(Base, ABC):
                 value, self._reassign_field, field.name
             )
 
-            if utils._issubclass(field.type_, List) or utils._issubclass(
-                field.type_, Dict
-            ):
+            if utils._issubclass(field.type_, Union[List, Dict]):
                 setattr(self, field.name, value_in_pc_data)
 
         self.clean()
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -13,7 +13,7 @@ from redis import Redis
 from pynecone import constants, utils
 from pynecone.base import Base
 from pynecone.event import Event, EventHandler, window_alert
-from pynecone.var import BaseVar, ComputedVar, PCList, Var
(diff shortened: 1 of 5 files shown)


pynecone/state.py
```diff
@@ -79,7 +79,9 @@ def _init_mutable_fields(self):
                 value, self._reassign_field, field.name
             )
 
-            if utils._issubclass(field.type_, List):
+            if utils._issubclass(field.type_, List) or utils._issubclass(
```
picklelo, line 82: I think this can be simplified to 
```python
utils_.issubclass(field.type_, Union[List, Dict])
```

pynecone/var.py
```diff
@@ -773,72 +773,155 @@ def __init__(
 
         super().__init__(original_list)
 
-    def append(self, *args, **kargs):
+    def append(self, *args, **kwargs):
```
picklelo, line 776: Thanks for these fixes!

pynecone/var.py
```diff
@@ -773,72 +773,155 @@ def __init__(
...
+        super().remove(*args, **kwargs)
+        self._reassign_field()
+
+
+class PCDict(dict):
+    """A custom dict that pynecone can detect its mutation."""
+
+    def __init__(
+        self,
+        original_dict: dict,
```
picklelo, line 852: We need to support Python 3.7 so we should make this uppercase `Dict`

pynecone/var.py
```diff
@@ -773,72 +773,155 @@ def __init__(
...
+        reassign_field: Callable = lambda _field_name: None,
+        field_name: str = "",
+    ):
+        """Initialize PCList.
```
picklelo, line 856: Comment needs tone updated to PCDict

pynecone/var.py
```diff
@@ -773,72 +773,155 @@ def __init__(
...
+
+        Args:
+            original_dict (Dict): The original dict
```
picklelo, line 859: We can remove the types from the docstrings since we use type annotations

picklelo (commented):
Looks good! I think there's some Pyright issues - you can add a `#type: ignore` if they are just some weird cases.

sid77x (approved):

pynecone/var.py
ElijahAhianyo: ah right, didn't have time to finish this up, I'll do that as soon as I can :)

picklelo (approved):
Awesome - tested it locally and it's working nicely!

PR #510, opened 2023-02-11
Resolve #495: foreach issue in cond

### LINKED ISSUE #495: Compile error when use `pc.foreach()` in `pc.cond()`
**Describe the bug**

When I try to use `pc.foreach()` inside of `pc.cond()`, a compile error occurred.

Full log:

```
❯ pc run
Starting Pynecone App
Installing frontend packages
bun install v0.5.5 (242dcea2)
App Running
$ next dev
ready - started server on 0.0.0.0:3000, url: http://localhost:3000
event - compiled client and server successfully in 2.4s (757 modules)
wait  - compiling /404 (client and server)...
event - compiled client and server successfully in 270 ms (761 modules)
warn  - Fast Refresh had to perform a full reload. Read more: https://nextjs.org/docs/basic-features/fast-refresh#how-it-works
wait  - compiling / (client and server)...
error - ./pages/index.js
Error: 
  x An object member cannot be declared optional
    ,----
 50 | {true ? {["a", "b", "c"].map((ktqunmvq, i) => <Text key={i}>{ktqunmvq}</Text>)} : <Text>{`anything else`}</Text>}
    :       ^
    `----

  x Unexpected token `{`. Expected ... , *,  (, [, :, , ?, = or an identifier
    ,----
 50 | {true ? {["a", "b", "c"].map((ktqunmvq, i) => <Text key={i}>{ktqunmvq}</Text>)} : <Text>{`anything else`}</Text>}
    :         ^
    `----

Caused by:
    0: failed to process input file
    1: Syntax Error
wait  - compiling /_error (client and server)...
error - ./pages/index.js
Error: 
  x An object member cannot be declared optional
    ,----
 50 | {true ? {["a", "b", "c"].map((ktqunmvq, i) => <Text key={i}>{ktqunmvq}</Text>)} : <Text>{`anything else`}</Text>}
    :       ^
    `----

  x Unexpected token `{`. Expected ... , *,  (, [, :, , ?, = or an identifier
    ,----
 50 | {true ? {["a", "b", "c"].map((ktqunmvq, i) => <Text key={i}>{ktqunmvq}</Text>)} : <Text>{`anything else`}</Text>}
    :         ^
    `----

Caused by:
    0: failed to process input file
    1: Syntax Error
```

Compile result ( index.js ):

```jsx
import {useEffect, useRef, useState} from "react"
import {useRouter} from "next/router"
import {E, connect, updateState} from "/utils/state"
import "focus-visible/dist/focus-visible"
import {Text, useColorMode} from "@chakra-ui/react"
import NextHead from "next/head"

const EVENT = "ws://localhost:8000/event"
const [state, setState] = useState({"add_modal_desc": "", "add_modal_priority": "", "add_modal_title": "", "is_add_modal_open": false, "todo_items": [], "events": [{"name": "state.hydrate"}]})
const [result, setResult] = useState({"state": null, "events": [], "processing": false})
const router = useRouter()
const socket = useRef(null)
const { isReady } = router;
const { colorMode, toggleColorMode } = useColorMode()
const Event = events => setState({
  ...state,
  events: [...state.events, ...events],
})
useEffect(() => {
  if(!isReady) {
    return;
  }
  const reconnectSocket = () => {
    socket.current.reconnect()
  }
  if (typeof socket.current !== 'undefined') {
    if (!socket.current) {
      window.addEventListener('focus', reconnectSocket)
      connect(socket, state, setState, result, setResult, router, EVENT)
    }
  }
  const update = async () => {
    if (result.state != null) {
      setState({
        ...result.state,
        events: [...state.events, ...result.events],
      })
      setResult({
        state: null,
        events: [],
        processing: false,
      })
    }
    await updateState(state, setState, result, setResult, router, socket.current)
  }
  update()
})
return (
{true ? {["a", "b", "c"].map((ktqunmvq, i) => <Text key={i}>{ktqunmvq}</Text>)} : <Text>{`anything else`}</Text>}
)
}
```

**To Reproduce**

That is the shortest code snippet to reproduce this bug:

```python
def index() -> None:
    return pc.cond(
        True,
        pc.foreach(
            ["a", "b", "c"],
            lambda x: pc.text(x),
        ),
        pc.text("anything else"),
    )
```

**Expected behavior**

`pc.foreach()` works will with `pc.cond()`. if the condition is True, show "a b c", otherwise show "anything else".

 - Python Version: 3.10.9
 - Pynecone Version: 0.0.14 ( I'm sure this problem won't be fixed in v0.0.15 )
 - OS: Manjaro Linux
 - Browser (Optional): Firefox 109.0.1
 
**Additional context**

This issue may be related to #435 , but I didn't see any error info in this issue's description.

TommyDew42:
Closes #495.

### DIFF
--- a/tests/components/layout/test_cond.py
+++ b/tests/components/layout/test_cond.py
@@ -1,10 +1,12 @@
 from typing import Any
 
 import pytest
+import pytest_mock
 
 import pynecone as pc
 from pynecone.components import cond
 from pynecone.components.layout.cond import Cond
+from pynecone.components.layout.foreach import Foreach
 from pynecone.components.layout.fragment import Fragment
 from pynecone.components.tags.tag import PropCond
 from pynecone.components.typography.text import Text
@@ -86,3 +88,21 @@ def test_cond_no_else():
     # Props do not support the use of cond without else
     with pytest.raises(ValueError):
         cond(True, "hello")
+
+
+def test_cond_foreach(mocker: pytest_mock.MockFixture):
+    var_name = "abc"
+    mocker.patch("pynecone.utils.get_unique_variable_name", return_value=var_name)
+
+    cond_component = Cond.create(
+        True,  # pyright: reportGeneralTypeIssues=false
+        Foreach.create(
+            ["Tommy", "is", "cool"],  # pyright: reportGeneralTypeIssues=false
+            lambda x: Text.create(x),
+        ),
+        Text.create("anything else"),
+    )
+
+    assert str(cond_component) == (
+        f'{{true ? ["Tommy", "is", "cool"].map(({var_name}, i) => <Text key={{i}}>{{{var_name}}}</Text>) : <Text>{{`anything else`}}</Text>}}'
+    )
--- a/pynecone/components/component.py
+++ b/pynecone/components/component.py
@@ -369,13 +369,17 @@ class Component(Base, ABC):
             child.add_style(style)
         return self
 
-    def render(self) -> str:
+    def render(self, *args, **kwargs) -> str:
         """Render the component.
+
+        Args:
+            args: ordered arguments
+            kwargs: key word arguments
 
         Returns:
             The code to render the component.
         """
-        tag = self._render()
+        tag = self._render(*args, **kwargs)
         return str(
             tag.add_props(
                 **self.event_triggers, key=self.key, sx=self.style, id=self.id
--- a/pynecone/components/layout/cond.py
+++ b/pynecone/components/layout/cond.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 from typing import Any, Optional
 
 from pynecone.components.component import Component
+from pynecone.components.layout.foreach import Foreach
 from pynecone.components.layout.fragment import Fragment
 from pynecone.components.tags import CondTag, Tag
 from pynecone.var import Var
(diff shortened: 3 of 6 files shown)


tests/components/layout/test_cond.py
```diff
@@ -86,3 +88,21 @@ def test_cond_no_else():
...
+
+
+def test_cond_foreach(mocker: pytest_mock.MockFixture):
+    var_name = "abc"
+    mocker.patch("pynecone.utils.get_unique_variable_name", return_value=var_name)
+
+    cond_component = Cond.create(
+        True,  # pyright: reportGeneralTypeIssues=false
+        Foreach.create(
+            ["Tommy", "is", "cool"],  # pyright: reportGeneralTypeIssues=false
```
TommyDew42, line 100: Suggested we only see pyright error as a warning until it's fixed.

The pyright error seems off with our type hints:
```
Argument of type "list[atr]" cannot be assigned to parameter "iterable" of type "Var" in function "create"
```

Lendemor:
I was doing some testing on this and found that the following diff also solve the problem : 

image

picklelo:
Thanks for this change! Can we try the solution that @Lendemor posted since I think the diff will be smaller then?

TommyDew42:
Please go ahead! Feel free to close this one if we find a simpler solution.

picklelo:
Thanks! Will close this as we just merged in #521 🙂

PR #545, opened 2023-02-15
Add class name propety to component

### LINKED ISSUE #539: Add class support for objects
It would be great if objects would support classes, like pc.box("Text", class_name="example")

A use case for this is using google material symbols, which requires that objects have a specific class. According to the documentation ( https://developers.google.com/fonts/docs/material_symbols#using_material_symbols ) material symbols are used by importing the font as stylesheet and then giving the element the class "material-symbols-outlined". 

See the following example:
```
<link href="https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:opsz,wght,FILL,GRAD@20,100,1,200" rel="stylesheet" />

<span class="material-symbols-outlined">search</span>
```

(Disclaimer: yes, I know that there is also pc.icons, but this is extremly limited - and there are use cases beyond google material icons). 

**Additional context**
I think this functionality is a per-requisit for another feature request ("Tailwind support", https://github.com/pynecone-io/pynecone/issues/141 ). 

In my (honestly very limited understanding) of css this feature would also enable pynecone users to import complex stylesheets that depend on classes of the objects that should be formatted.

PeterYusuke:
### All Submissions:

### Type of change

### New Feature Submission:

### Description
Add class_name property to component.
We can add class_name to at any component.

This change add class name to the component.
Will not break default class name, like chackra ones.

### Usage
class_name property covers type of str, List and tuple.

#### str
python code.
```py
pc.text("new class", class_name="class-name-1")
```

js code.
```js
<Text ClassName="class-name-1">{`new class`}</Text>
```
---
#### List
python code.
```py
pc.text("new class", class_name=["class-name-1", "class-name-list"])
```

js code.
```js
<Text ClassName="class-name-1 class-name-list">{`new class`}</Text>
```

---
#### tuple
python code.
```py
pc.text("new class", class_name=("class-name-1", "class-name-tuple"))
```

js code.
```js
<Text ClassName="class-name-1 class-name-tuple">{`new class`}</Text>
```

close #539

### DIFF
--- a/pynecone/components/component.py
+++ b/pynecone/components/component.py
@@ -47,6 +47,9 @@ class Component(Base, ABC):
     # The id for the component.
     id: Any = None
 
+    # The class name for the component.
+    class_name: Any = None
+
     @classmethod
     def __init_subclass__(cls, **kwargs):
         """Set default properties.
@@ -144,6 +147,11 @@ class Component(Base, ABC):
                 **{attr: value for attr, value in kwargs.items() if attr not in fields},
             }
         )
+
+        # Convert class_name to str if it's list
+        class_name = kwargs.get("class_name", "")
+        if isinstance(class_name, (List, tuple)):
+            kwargs["class_name"] = " ".join(class_name)
 
         # Construct the component.
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


pynecone/components/component.py
```diff
@@ -370,7 +378,11 @@ def render(self) -> str:
         tag = self._render()
         return str(
             tag.add_props(
-                **self.event_triggers, key=self.key, sx=self.style, id=self.id
+                **self.event_triggers,
+                key=self.key,
+                sx=self.style,
+                id=self.id,
+                className=self.class_name,
```
picklelo, line 385: I think you can have the key here as `class_name` - it should get converted to `className` during compilation

picklelo (commented):
Nice addition! Just one small comment

pynecone/components/component.py
PeterYusuke: Right! I changed this and worked.

picklelo (approved):
Awesome!

PR #552, opened 2023-02-16
Generate app sitemap

picklelo:
Added a new field in the pcconfig.py called `deploy_url` - which will be used to generate the urls for the sitemap.


Alek99 (approved):
lgtm

Issue #515, opened 2023-02-12
Production mode not working on Windows OS


LiteEagle262:
**Describe the bug**
```
import fcntl
ModuleNotFoundError: No module named 'fcntl'
```
Sollution: https://stackoverflow.com/questions/45228395/error-no-module-named-fcntl

**To Reproduce**
Try production mode on windows

**Expected behavior**
Production Mode to work

picklelo:
PR #570 merged and should resolve this

PR #570, opened 2023-02-20
Windows production mode command

PeterYusuke:
### All Submissions:

### Type of change

### Description
Add production mode command for Windows PC.
We can not use Gunicorn on Windows now, so I add Uvicorn command.
I kept the code for the Linux-based machine because it's not better to use only Uvicorn than Gunicorn.

### DIFF
--- a/pynecone/utils.py
+++ b/pynecone/utils.py
@@ -799,31 +799,34 @@ def run_backend_prod(
     setup_backend()
 
     num_workers = get_num_workers()
-    command = []
-    if platform.system() == "Windows":
-        command += constants.RUN_BACKEND_PROD_WINDOWS + [
+    command = (
+        constants.RUN_BACKEND_PROD_WINDOWS
+        + [
             "--host",
             "0.0.0.0",
             "--port",
             str(port),
-            "--workers",
-            str(num_workers),
             "--log-level",
             loglevel,
             f"{app_name}:{constants.APP_VAR}",
         ]
-    else:
-        command += constants.RUN_BACKEND_PROD + [
+        if platform.system() == "Windows"
+        else constants.RUN_BACKEND_PROD
+        + [
             "--bind",
             f"0.0.0.0:{port}",
-            "--workers",
-            str(num_workers),
             "--threads",
             str(num_workers),
             "--log-level",
             str(loglevel),
             f"{app_name}:{constants.APP_VAR}()",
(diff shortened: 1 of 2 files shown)


PeterYusuke:
Close #515.

pynecone/utils.py
```diff
@@ -799,17 +799,31 @@ def run_backend_prod(
...
-        f"0.0.0.0:{port}",
-        "--workers",
-        str(num_workers),
-        "--threads",
-        str(num_workers),
-        "--log-level",
-        str(loglevel),
-        f"{app_name}:{constants.APP_VAR}()",
-    ]
+    command = []
```
picklelo, line 802: Don't need to define the variable here, we can just do it within the if/else statements

pynecone/utils.py
```diff
@@ -799,17 +799,31 @@ def run_backend_prod(
...
+    if platform.system() == "Windows":
+        command += constants.RUN_BACKEND_PROD_WINDOWS + [
```
picklelo, line 804: These can just be `command =` instead of `+=`

pynecone/utils.py
```diff
@@ -799,17 +799,31 @@ def run_backend_prod(
...
+            "--port",
+            str(port),
+            "--workers",
+            str(num_workers),
+            "--log-level",
+            loglevel,
+            f"{app_name}:{constants.APP_VAR}",
+        ]
+    else:
+        command += constants.RUN_BACKEND_PROD + [
```
picklelo, line 816: Here also

picklelo (commented):
Thanks for the update! Just a small comment

pynecone/utils.py
picklelo, line 816: It seems like all the args we pass in are the same. Can we then change this to just
```python
command = [constants..RUN_BACKEND_PROD_WINDOWS] if platform.system() == "Windows" else [constants.RUN_BACKEND_PROD]
command += [
    # Common flags here
]

```
If there are some special flags, we can put them in the if/else but move the common ones out.

PeterYusuke: OK. Only workers arg is the same, others are little bit different.
I will make common arg list.

PeterYusuke:
Changed command list. Please check.

picklelo (approved):
nice

Issue #585, opened 2023-02-21
Event chain missing events


lucashofer:
**Describe the bug**
Using a list of events for ``on_click`` misses every other event

**To Reproduce**
```
from pcconfig import config
import pynecone as pc


class State(pc.State):
    
    var1: str = "Hello"
    var2: int = 10


    def print_var1(self):
        return pc.console_log(self.var1)


    def print_var2(self):
        return pc.console_log(self.var2)


def index():
    return pc.box(
        pc.button("Chain", 
                  on_click=[State.print_var1,
                            State.print_var2,
                            State.print_var1,
                           ]
        )
    )


# Add state and page to the app.
app = pc.App(state=State)
app.add_page(index)
app.compile()
```

**Expected behavior**
The console should have the following printed:

```
> Hello
> 10
> Hello
```

instead it gives:

```
> Hello
> Hello
```

Furthermore if I add more events than three it misses every other one.

**Specifics**
 - Python Version: Python 3.10
 - Pynecone Version: 0.1.18
 - OS: Ubuntu 20.04

Alek99:
Ok thanks will get this fixed asap

lucashofer:
Amazing! Thanks!

PR #596, opened 2023-02-23
Multiple Page Load Events Added

lucashofer:
Previously only a single State event handler could be called when loading a page. 

Now either a single event handler or a list of event handlers can be passed. This allows multiple actions to be performed when loading a page which significantly enhances code re-usability in the State classes.

### All Submissions:

### Type of change

### New Feature Submission:

### DIFF
--- a/pynecone/app.py
+++ b/pynecone/app.py
@@ -53,7 +53,7 @@ class App(Base):
     middleware: List[Middleware] = []
 
     # Event handlers to trigger when a page loads.
-    load_events: Dict[str, EventHandler] = {}
+    load_events: Dict[str, Union[EventHandler, List[EventHandler]]] = {}
 
     def __init__(self, *args, **kwargs):
         """Initialize the app.
@@ -197,7 +197,7 @@ class App(Base):
         title: str = constants.DEFAULT_TITLE,
         description: str = constants.DEFAULT_DESCRIPTION,
         image=constants.DEFAULT_IMAGE,
-        on_load: Optional[EventHandler] = None,
+        on_load: Optional[Union[EventHandler, List[EventHandler]]] = None,
         path: Optional[str] = None,
         meta: List[Dict] = constants.DEFAULT_META_LIST,
     ):
@@ -213,7 +213,7 @@ class App(Base):
             title: The title of the page.
             description: The description of the page.
             image: The image to display on the page.
-            on_load: The event handler that will be called each time the page load.
+            on_load: The event handler(s) that will be called each time the page load.
             meta: The metadata of the page.
         """
         if path is not None:
--- a/pynecone/middleware/hydrate_middleware.py
+++ b/pynecone/middleware/hydrate_middleware.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Optional
 
 from pynecone import constants, utils
-from pynecone.event import Event
+from pynecone.event import Event, EventHandler
 from pynecone.middleware.middleware import Middleware
 from pynecone.state import Delta, State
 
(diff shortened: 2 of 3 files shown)


Lendemor:
To match event handlers being used in regular triggers, page load should be able to handle EventChain.

(List of EventHandler is a good step toward that, but it doesn't handle the case where an event handler return another event handler to make chain of events, like it does in other events).

The following should work when you pass `load_method1` to `on_load`:
```
class State(pc.State):
    def load_method1(self):
        # method 1 body
        return self.load_method2()

    def load_method2(self):
        # method 2 body
```
and I don't think the current solutions will handle this.

lucashofer:
A few notes on this. 

Currently ``on_load``  does not support it's event handler calling another event handler. This is due to the way the event handler is called in the middleware (i.e. ``load_event.fn(ex_state)``). My proposed change maintains this structure. 

I really like Pynecone, but I've been struggling a lot with code re-usability. This is due to two main issues:

1. Event chains only work for events that can be executed asynchronously.
2. Calling one event handler from another allows very limited code re-usability 

I give a small example below with two integer state variables. Now say I have three buttons, one which increments count1, another which adds the counts together and a third which increments count1 and then adds them together. 

```
class State(pc.State):
    
    count1: int = 0
    count2: int = 1
    sum_count: int = 0

    def increment_count1(self):
        time.sleep(1)
        self.count1 += 1

    def add_counts(self):
        self.sum_count = self.count1 + self.count2

    def increment_add(self):
        self.increment_count1()
        self.add_counts()
```

Currently the given ``increment_add`` function won't work in Pynecone. We also can't use an event chain like ``on_click=[State.increment_count1, State.add_counts]`` because event two relies on event one and it executes asynchronously. This only leaves us with one option 

```
def increment_add(self):
    self.count1 += 1
    return self.add_counts()
```
which means we have to rewrite the ``increment_count1`` function. This isn't bad for this toy example, but it makes using Pynecone difficult when the functions are more complex. 

This is a broader issue not just related to the ``on_load`` event. I'm not opposed to event chains, but is there a flag we could set to make them run synchronously? This would allow them to be used more broadly and avoid having to rewrite code in the state event functions.

Lendemor:
> Currently on_load does not support it's event handler calling another event handler.

Yes I know, this is one of the reason I didn't push a similar PR to what you proposed (one that I had started working on), because it wasn't satisfying to me.

As the one who did the first PR for on_load event, it was a very naive approach to the problem, one that doesn't really cover all use case, but simply provided a "basic solution".

> My proposed change maintains this structure.

Based on what I just explained above, current structure is very naive, and need improvement.

I think your PR could be accepted as it is, but with knowing that we will have to rework it again anyway to improve on it.

> We also can't use an event chain like on_click=[State.increment_count1, State.add_counts] because event two relies on event one and it executes asynchronously.

To me this feel weird, the basic principle of an event chain is that they should execute one after another.
If we want them to execute at the same time, we would need to define something like an `EventGroup` which would start all of  its event asynchronously like you said .
 
As for your example, it would be possible to write : 
```
def increment_add(self):
    return [self.increment_count1(), self.add_counts()]
```

But I do agree on this way of doing thing not being very intuitive.

picklelo:
Thanks @lucashofer and @Lendemor for adding this feature and the discussion.

> Event chains only work for events that can be executed asynchronously.

Event chains should work synchronously. Currently the architecture is that the frontend sends one event at a time to the backend to process. It sets a `processing` flag while this is happening to avoid race conditions. If you see any async behavior with event chains, it's a bug.

> Calling one event handler from another allows very limited code re-usability

I think I may have a better solution for how we handle the `on_load` events. 

This is the line where we add the initial events for a page. Currently it is the same for every page and it's just the hydrate event: https://github.com/pynecone-io/pynecone/blob/9a9a7317661ac2a39c7b5bb8a4e7379d237865c9/pynecone/compiler/utils.py#L106

We could compile in `on_load` event for the page directly into the beginning event queue, and its should then handle it like any other event (allowing events that return other events, etc.).

We can merge in this PR for now since it's a good start, but then we can work on rearchitecting the `on_load` to this more flexible design if it works well.

What do you guys think?

picklelo:
> Currently the given increment_add function won't work in Pynecone. We also can't use an event chain like on_click=[State.increment_count1, State.add_counts] because event two relies on event one and it executes asynchronously. This only leaves us with one option

I agree that this is a flaw - I think it's something we can actually fix pretty easily. Created this issue: https://github.com/pynecone-io/pynecone/issues/602

picklelo (approved):
This looks good - we can rethink the `on_load` design for a more flexible approach in a follow-up

lucashofer:
@picklelo I agree incorporating the `on_load` events into the beginning event queue makes more sense long term.

Sequential execution of the event chains currently is not working (see [#585](https://github.com/pynecone-io/pynecone/issues/585)) and it'd be good to have that fixed before trying to incorporate them into `on_load`

I'd really love to see [#602](https://github.com/pynecone-io/pynecone/issues/602), but will move to that discussion thread. Thanks!

PR #650, opened 2023-03-09
635-integration tests

### LINKED ISSUE #635: Fix integration tests
We should fix the integration tests to run pcweb in prod mode.

ElijahAhianyo:
This PR adds an integration test. It clones a pynecone repo, runs it, and checks for errors.

fixes #635

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -4,7 +4,6 @@ on:
   pull_request_review:
     types: [submitted]
 
-
 permissions:
   contents: read
 
@@ -35,8 +34,8 @@ jobs:
     - name: Clone Pynecone Website Repo
       uses: actions/checkout@v3
       with:
-        repository: pynecone-io/pcweb
-        path: pcweb
+        repository: pynecone-io/pynecone-examples
+        path: pynecone-examples
     
     # Install poetry
     - name: cache poetry install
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yml
```diff
@@ -58,17 +57,34 @@ jobs:
...
+      working-directory: ./pynecone-examples/counter
+      run: poetry run pip install -r requirements.txt
+    - name: Init Website
+      working-directory: ./pynecone-examples/counter
+      run: poetry run pc init
+    - name: Run Website and check for errors
+      working-directory: ./pynecone-examples/counter
+      run: |
+        curl -fsSL https://bun.sh/install | bash -s -- bun-v0.5.5
+        timeout 2m poetry run pc run &
```
picklelo, line 73: When we add `pcweb` let's run `pc run --env prod` since it will force compile all the pages to JS. For this app it doesn't matter since it's only a single app.

.github/workflows/integration.yml
```diff
@@ -58,17 +57,34 @@ jobs:
...
+        sleep 10
+        URL="127.0.0.1:3000"
+
+        # make the curl request and save the response and HTTP status code
+        RESPONSE=$(curl -s -w "\n%{http_code}" $URL)
```
picklelo, line 78: Nice check! We can also ping there `8000` port to make sure the backend is running.

picklelo (approved):
Nice! Just some minor comments for a follow up

PR #666, opened 2023-03-12
Replace smelly utils module

iron3oxide:
### All Submissions:

Having a `utils` module is considered a code smell in most cases, including this one. It did way too much different stuff and even required circular imports in some places. This PR distributes the code formerly contained in the `utils` module in a logically coherent manner into new and existing modules. It also includes minor fixups that were necessary to pass all tests; the most notable one being the clear separation of web directory creation, frontend package installation and frontend setup. This fixup should increase testability in general, as the frontend setup called the frontend package installation function in the previous code, which made it way harder to mock said installation.

### Type of change

Refactoring

### Changes To Core Features:


picklelo (changes requested):
Thanks for breaking this up - that file was getting big and had many circular imports.

Could we make the following changes:
* Create a `pynecone/utils` directory where all these new files will go under
* Move the new files into the new folder e.g. `pynecone/utils/format.py` and `pynecone/utils/imports.py`
* In the files that import these, import the module rather than the individual functions - for example `from pynecone.utils import format` and then call the functions like `format.format_state`

iron3oxide:
Done, that probably won't fix the workflow error from the first run though. I'll dig into that.

iron3oxide:
Seems like `from __future__ import annotations` was never needed in `route.py` until now. I think this should fix the workflow errors, but I haven't been able to set `rtx` up to test with 3.7 or 3.8 yet, so only a rerun will tell for sure.

iron3oxide:
Forgot to lint the docstrings 🤦‍♂️. Sorry for the delay, should work now.

picklelo:
Thanks for fixing the build! We're cutting a release tonight, I'll merge this in afterwards.

pynecone/compiler/utils.py
```diff
@@ -20,12 +20,14 @@
     Script,
     Title,
 )
-from pynecone.components.component import Component, CustomComponent, ImportDict
+from pynecone.components.component import Component, CustomComponent
+from pynecone.event import get_hydrate_event
 from pynecone.state import State
 from pynecone.style import Style
+from pynecone.utils import imports, path_ops
```
picklelo, line 27: We need to import `format` here also it seems

picklelo (changes requested):
Just one issue, otherwise seems to work well!

pynecone/compiler/utils.py
iron3oxide: Done

picklelo (approved):
thanks for the huge refactor

picklelo:
Talked with @ElijahAhianyo offline - looks like this is because we need to update pcweb and pynecone at the same time. I ran the integration test manually and it passed, so I'll merge this in.

PR #683, opened 2023-03-15
[Fix: issue 652] Fix escaping quotes in dict values

d29107d:
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

### Description
Fix escaping quotes in dict values
Close https://github.com/pynecone-io/pynecone/issues/652

### DIFF
--- a/pynecone/components/tags/tag.py
+++ b/pynecone/components/tags/tag.py
@@ -127,6 +127,8 @@ class Tag(Base):
         # Format all the props.
         return os.linesep.join(
             f"{name}={self.format_prop(prop)}"
+            if name != "sx"
+            else f"{name}={self.format_style(prop)}"
             for name, prop in sorted(self.props.items())
             if prop is not None
         )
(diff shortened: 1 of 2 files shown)


Alek99:
Ok nice testing this locally now.

pynecone/components/tags/tag.py
```diff
@@ -209,3 +211,23 @@ def is_valid_prop(prop: Optional[Var]) -> bool:
...
+        Returns:
+            The formatted style.
+        """
+        ret = utils.json_dumps(prop)
+
+        # This substitution is necessary to unwrap var values.
+        if ret.find('"{') > 0:
+            ret = re.sub('"{', "", ret)
+            ret = re.sub('}"', "", ret)
+            ret = re.sub('\\\\"', '"', ret)
```
Alek99, line 231: I remember this specific sub was causing errors for something a while back, do you remember the specific case causing this @picklelo, I'll test it

picklelo: I think this sub was necessary for fixing another issue with indexing state vars i.e. `prop=State.dict["key"]` was getting formatted wrong without this. I added unit tests for that though, so if they are passing we should be good.

picklelo: Oh but now they may not be valid within style props. We need to make sure we can pass in the state indexing a string like above^ for style props also.

d29107d: @picklelo 

> Oh but now they may not be valid within style props. We need to make sure we can pass in the state indexing a string like above^ for style props also.

How to make sure? Is there a case to check it?

danik292 (approved):

Chixonya22 (approved):

d29107d:
@Alek99 @picklelo Could you have a review?

PR #691, opened 2023-03-15
Bugfix: Call event handlers from other event handlers

### LINKED ISSUE #602: Call event handlers from other event handlers
Currently the only way to call another event handler is to return it and create a chain. However this leads to poor code reusability in many cases.

Example in #596 

```python
class State(pc.State):
    
    count1: int = 0
    count2: int = 1
    sum_count: int = 0

    def increment_count1(self):
        time.sleep(1)
        self.count1 += 1

    def add_counts(self):
        self.sum_count = self.count1 + self.count2

    def increment_add(self):
        self.increment_count1()
        self.add_counts()
```

Currently the given increment_add function won't work in Pynecone. We also can't use an event chain like on_click=[State.increment_count1, State.add_counts] because event two relies on event one and it executes asynchronously. This only leaves us with one option

This should be very possible to implement.

ElijahAhianyo:
This PR allows an event handler to be called from another event handler

fixes #602

### DIFF
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -50,6 +50,9 @@ class State(Base, ABC):
     # Backend vars inherited
     inherited_backend_vars: ClassVar[Dict[str, Any]] = {}
 
+    # The event handlers.
+    event_handlers: ClassVar[Dict[str, EventHandler]] = {}
+
     # The parent state.
     parent_state: Optional[State] = None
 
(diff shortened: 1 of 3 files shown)


pynecone/state.py
```diff
@@ -180,8 +183,18 @@ def __init_subclass__(cls, **kwargs):
...
         for name, fn in events.items():
             event_handler = EventHandler(fn=fn)
+            cls.event_handlers[name] = event_handler
             setattr(cls, name, event_handler)
 
+    @classmethod
+    def convert_handlers_to_fns(cls):
+        """Convert the event handlers to functions.
+
+        This is done so that the event handlers can be pickled.
```
picklelo, line 193: I think we should say this is done so that we can call the state functions as normal functions during runtime

pynecone/state.py
```diff
@@ -544,7 +557,8 @@ async def process(self, event: Event) -> StateUpdate:
         path = event.name.split(".")
         path, name = path[:-1], path[-1]
         substate = self.get_substate(path)
-        handler = getattr(substate, name)
+        # handler = getattr(substate, name)
```
picklelo, line 560: remove

picklelo (commented):
Nice! I think this needs to be rebased on the latest `main`. Also does this handle the case for returning event handlers?

picklelo (approved):
Awesome, seems to work well

PR #743, opened 2023-03-29
Use Strict (In)Equality in Generated Javascript

jonaengs:
Use strict equality and inequality in generated javascript equality checks.  

**Reason**: Javascript's strict equality semantically more similar to Python's notion of equality than the previously used equality. This difference mainly stems from Javascript's heavy reliance on type coercion when evaluating the equality operator. The strict equality operator avoids this type coercion.

Some examples showing why this is a good change:
```javascript
// Javascript equality:
2 == "2"; // true
"1,2" == [1, 2]; // true
```
```python
# Python equality:
2 == "2" # False
"1,2" == [1, 2]  # False
```

```javascript
// Javascript strict equality:
2 === "2"; // false
"1,2" === [1, 2]; // false
```

The one drawback to this change is that the following correct (read: same as Python) behavior will disappear:
```javascript
1 == true; // true
0 == false // true
``` 
But I believe that this is a rather uncommon expression. This change will not affect the behavior of `pc.cond(State.some_integer, ...)`. The change only affects this exact equality operation, with these exact operands.

Sources:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Equality https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Inequality https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_inequality https://tc39.es/ecma262/#sec-isstrictlyequal
https://docs.python.org/3/reference/expressions.html

### Type of change

### DIFF
--- a/pynecone/var.py
+++ b/pynecone/var.py
@@ -373,7 +373,7 @@ class Var(ABC):
         Returns:
             A var representing the equality comparison.
         """
-        return self.compare("==", other)
+        return self.compare("===", other)
 
     def __ne__(self, other: Var) -> Var:
         """Perform an inequality comparison.
@@ -384,7 +384,7 @@ class Var(ABC):
         Returns:
             A var representing the inequality comparison.
         """
-        return self.compare("!=", other)
+        return self.compare("!==", other)
 
     def __gt__(self, other: Var) -> Var:
         """Perform a greater than comparison.


picklelo (changes requested):
Good catch! Looks like we need to update the unit tests - then we can merge this. (Also make sure to lint, etc. https://github.com/pynecone-io/pynecone/blob/main/CONTRIBUTING.md)

jonaengs:
TIL that renaming a branch with an open PR attached to it will get your PR closed. The new PR is #750.

Issue #603, opened 2023-02-25
Improve `on_load` events


picklelo:
As a follow up to #596 we should rethink the design and potentially compile the on_load events directly into the page's initial event queue. This should allow them to follow the same flow as all other events.

picklelo:
We should also make sure to handle async events:

```python
class State(...):
    async def onload(self): # was never awaited exception
        async with websockets.connect(...) as ws:
            ...
            recv = await ws.recv()
            self.data = recv
@pc.route("/", on_load=State.onload)
def index():
    return ...
```

picklelo:
Merged in #773

johanfforsberg:
I know this issue is closed, and perhaps I should open a new one, but I would like to comment that this kind of thing mentioned here still doesn't work AFAICT:

```python
class State(...):
    async def onload(self): # was never awaited exception
        async with websockets.connect(...) as ws:
            ...
            recv = await ws.recv()
            self.data = recv

@pc.route("/", on_load=State.onload)
def index():
    return ...
```
When I do something like this e.g. just
```python
    async def onload(self):
        i = 0
        while True:
            await asyncio.sleep(1)
            self.data = i
            i +=1
```
I get no updates. It seems nothing will happen until the `onload` function returns.

I can work around it by returning another event handler from `onload` and then let it return itself, to set up an infinite loop. But it seems more obscure and probably less efficient than just looping over an iterator.

pynecone 0.1.28
Python 3.10.6

PR #773, opened 2023-04-04
Fix Event chaining in the on_load event handler return not working

### LINKED ISSUE #610: Event chaining in the on_load event handler return is not working.
**Describe the bug**

Event chaining in the on_load event handler return is not working.

**To Reproduce**
 - Code/Link to Repo:

I implemented it like this

```python
class BaseState(pc.State):
    ...


class QueryState(BaseState):
    label: str

    def init(self):
        print("init!")
        return [
            self.reset_label,
        ]

    def reset_label(self):
        print("reset label!")
        self.label = ""

app = pc.App(state=state.BaseState)
app.add_page(query.index, route="/", on_load=QueryState.init)
```

When I ran the code, the result was as follows.

```
init!
```

The `reset_label()` method doesn't seem to have actually been executed.

**Expected behavior**

The `reset_label()` method should be executed.
This means that the result of its execution should be

```
init!
reset label!
```

**Screenshots**

 - Python Version: 3.11.0
 - Pynecone Version: 0.1.18
 - OS: Mac OS Monterey 12.6
 - Browser (Optional): .
 
**Additional context**

ElijahAhianyo:
This PR fixes a couple of issues with the way on-load events are handled. In this PR, the following should work now:

- event-chaining( handlers that return other handlers)
- list of on-load events (eg. `app.add_page('index', on_load = [State.first_handler, State.second_handler]`)
- Async handlers

fixes #610 , #603

### DIFF
--- a/pynecone/app.py
+++ b/pynecone/app.py
@@ -148,7 +148,7 @@ class App(Base):
             allow_origins=["*"],
         )
 
-    def preprocess(self, state: State, event: Event) -> Optional[Delta]:
+    async def preprocess(self, state: State, event: Event) -> Optional[Delta]:
         """Preprocess the event.
 
         This is where middleware can modify the event before it is processed.
@@ -165,7 +165,7 @@ class App(Base):
             An optional state to return.
         """
         for middleware in self.middleware:
-            out = middleware.preprocess(app=self, state=state, event=event)
+            out = await middleware.preprocess(app=self, state=state, event=event)
             if out is not None:
                 return out
 
@@ -400,7 +400,7 @@ class App(Base):
 
 async def process(
     app: App, event: Event, sid: str, headers: Dict, client_ip: str
-) -> StateUpdate:
+) -> Union[StateUpdate, List[StateUpdate]]:
     """Process an event.
 
     Args:
@@ -411,7 +411,7 @@ async def process(
         client_ip: The client_ip.
 
     Returns:
-        The state update after processing the event.
+        The state update(s) after processing the event.
     """
     # Get the state for the session.
     state = app.state_manager.get_state(event.token)
@@ -430,21 +430,27 @@ async def process(
     state.router_data[constants.RouteVar.CLIENT_IP] = client_ip
 
     # Preprocess the event.
-    pre = app.preprocess(state, event)
-    if pre is not None:
+    pre = await app.preprocess(state, event)
+    if pre is not None and not isinstance(pre, List):
         return StateUpdate(delta=pre)
 
     # Apply the event to the state.
-    update = await state.process(event)
+    updates = pre if pre else await state.process(event)
     app.state_manager.set_state(event.token, state)
 
+    updates = updates if isinstance(updates, List) else [updates]
+
     # Postprocess the event.
-    post = app.postprocess(state, event, update.delta)
-    if post is not None:
-        return StateUpdate(delta=post)
+    post_list = []
+    for update in updates:
+        post = app.postprocess(state, event, update.delta)  # type: ignore
+        post_list.append(post) if post else None
+
+    if post_list:
+        return [StateUpdate(delta=post) for post in post_list]
 
     # Return the update.
-    return update
+    return updates
 
 
 async def ping() -> str:
@@ -582,7 +588,13 @@ class EventNamespace(AsyncNamespace):
         update = await process(self.app, event, sid, headers, client_ip)
 
         # Emit the event.
-        await self.emit(str(constants.SocketEvent.EVENT), update.json(), to=sid)
+        if isinstance(update, List):
+            for single_update in update:
+                await self.emit(
+                    str(constants.SocketEvent.EVENT), single_update.json(), to=sid
+                )
+        else:
+            await self.emit(str(constants.SocketEvent.EVENT), update.json(), to=sid)
 
     async def on_ping(self, sid):
         """Event for testing the API endpoint.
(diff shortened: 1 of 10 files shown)


pynecone/app.py
```diff
@@ -148,7 +148,7 @@ def add_cors(self):
             allow_origins=["*"],
         )
 
-    def preprocess(self, state: State, event: Event) -> Optional[Delta]:
+    async def preprocess(self, state: State, event: Event) -> Optional[Delta]:
```
picklelo, line 151: Should we make `postprocess` async also for consistency?

pynecone/app.py
```diff
@@ -400,7 +400,7 @@ def compile(self, force_compile: bool = False):
 
 async def process(
     app: App, event: Event, sid: str, headers: Dict, client_ip: str
-) -> StateUpdate:
+) -> Union[StateUpdate, List[StateUpdate]]:
```
picklelo, line 403: We can just always return `List[StateUpdate]` to simplify things

pynecone/app.py
```diff
@@ -430,21 +430,27 @@ async def process(
     state.router_data[constants.RouteVar.CLIENT_IP] = client_ip
 
     # Preprocess the event.
-    pre = app.preprocess(state, event)
-    if pre is not None:
+    pre = await app.preprocess(state, event)
+    if pre is not None and not isinstance(pre, List):
         return StateUpdate(delta=pre)
```
picklelo, line 435: We can listify this here

pynecone/app.py
```diff
@@ -582,7 +588,13 @@ async def on_event(self, sid, data):
         update = await process(self.app, event, sid, headers, client_ip)
 
         # Emit the event.
-        await self.emit(str(constants.SocketEvent.EVENT), update.json(), to=sid)
+        if isinstance(update, List):
```
picklelo, line 591: We can make sure update is always a List so we don't need this check

pynecone/middleware/hydrate_middleware.py
```diff
@@ -37,20 +39,42 @@ def preprocess(self, app: App, state: State, event: Event) -> Optional[Delta]:
...
+                    load_event = [load_event]
+                updates = []
+                for single_event in load_event:
+                    updates.append(
+                        await self.execute_load_event(
+                            state, single_event, event.token, event.payload
+                        )
+                    )
+                return updates
             return format.format_state({state.get_name(): state.dict()})
```
picklelo, line 52: We can maybe just make this a `StateUpdate` object to simplify the API

pynecone/middleware/hydrate_middleware.py
```diff
@@ -37,20 +39,42 @@ def preprocess(self, app: App, state: State, event: Event) -> Optional[Delta]:
...
+            ValueError: If the state value is None.
         """
         substate_path = format.format_event_handler(load_event).split(".")
         ex_state = state.get_substate(substate_path[:-1])
-        load_event.fn(ex_state)
+        if ex_state:
+            return await state.process_event(
+                handler=load_event, state=ex_state, payload=payload, token=token
+            )
+        else:
```
picklelo, line 77: Don't need the `else` since the first case returns

pynecone/state.py
```diff
@@ -546,37 +546,65 @@ def get_substate(self, path: Sequence[str]) -> Optional[State]:
...
         """
         # Get the event handler.
         path = event.name.split(".")
         path, name = path[:-1], path[-1]
         substate = self.get_substate(path)
         handler = substate.event_handlers[name]  # type: ignore
 
-        # Process the event.
-        fn = functools.partial(handler.fn, substate)
+        if substate:
```
picklelo, line 566: We can change this to

```python
if not substate:
    raise ValueError(...)

return ...

```

pynecone/middleware/hydrate_middleware.py
```diff
@@ -37,20 +39,42 @@ def preprocess(self, app: App, state: State, event: Event) -> Optional[Delta]:
...
-    def execute_load_event(self, state: State, load_event: EventHandler) -> None:
+    async def execute_load_event(
+        self, state: State, load_event: EventHandler, token: str, payload: Dict
+    ) -> StateUpdate:
         """Execute single load event.
 
         Args:
             state: The client state.
             load_event: A single load event to execute.
+            token: client token
```
picklelo, line 62: nit: can we format the comments with uppercase and periods like "A client token." and "The event payload."

pynecone/state.py
```diff
@@ -546,37 +546,65 @@ def get_substate(self, path: Sequence[str]) -> Optional[State]:
...
+            )
+
+    async def process_event(
+        self, handler: EventHandler, state: State, payload: Dict, token: str
+    ) -> StateUpdate:
+        """Process event.
+
+        Args:
+            handler: Eventhandler to process.
+            state: state to process the handler.
```
picklelo, line 585: Same nit about formatting the arg comments

pynecone/state.py
```diff
@@ -546,37 +546,65 @@ def get_substate(self, path: Sequence[str]) -> Optional[State]:
...
+                state=substate,
+                payload=event.payload,
+                token=event.token,
+            )
+        else:
+            raise ValueError(
+                "The value of state cannot be None when processing an event."
+            )
+
+    async def process_event(
```
picklelo, line 578: Nice way to abstract this!

picklelo (commented):
Nice work! Mostly minor comments

pynecone/app.py
ElijahAhianyo: yep, I agree

picklelo (approved):
thanks for the updates!

PR #788, opened 2023-04-08
Fix event handler returns

picklelo:
This fixes the current broken clock example. I created a follow-on ticket to solve the more general cases: https://github.com/pynecone-io/pynecone/issues/787


Alek99 (approved):
nice looks good to me

PR #796, opened 2023-04-10
Fix bun install

picklelo:


Alek99 (approved):
Approved but you will need to change version to 0.1.23

PR #808, opened 2023-04-11
Initialize an EventHandler for dynamic vars

barnjamin:
### Summary

When constructing a class dynamically using `add_var`, the auto generated `set_XXX` method is not an `EventHandler` so pc complains:
```
ValueError: Lambda <function app_state.add.set_a at 0x7fb248e5f520> must have 0 or 1 arguments.
```

This changes the `_create_setter` method to wrap the prop setter in an `EventHandler` and also adds a check in the `__init__subclass__` to prevent double wrapping of the `EventHandler`

### Type of change

### New Feature Submission:

### Changes To Core Features:

### DIFF
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -180,7 +180,9 @@ class State(Base, ABC):
         events = {
             name: fn
             for name, fn in cls.__dict__.items()
-            if not name.startswith("_") and isinstance(fn, Callable)
+            if not name.startswith("_")
+            and isinstance(fn, Callable)
+            and not isinstance(fn, EventHandler)
         }
         for name, fn in events.items():
             event_handler = EventHandler(fn=fn)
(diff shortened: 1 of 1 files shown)


Alek99:
I remember adding 'add_var' when we first released the library, this was when list and dictionary did auto update when modified. I'm thinking now there's not really a situation when you would want a bunch of individual vars vs using a data structure such as dictionary that can contain all the things you are trying to keep track of.

barnjamin:
I'm using `add_var` as I'm constructing a State subclass dynamically here

https://github.com/barnjamin/blah/blob/master/blah/pc_app.py#L119

Maybe there is a better way?

Chixonya22 (approved):

danik292 (approved):

ElijahAhianyo:
> I'm using `add_var` as I'm constructing a State subclass dynamically here
> 
> https://github.com/barnjamin/blah/blob/master/blah/pc_app.py#L119
> 
> Maybe there is a better way?



> I'm using `add_var` as I'm constructing a State subclass dynamically here
> 
> https://github.com/barnjamin/blah/blob/master/blah/pc_app.py#L119
> 
> Maybe there is a better way?

Went through your example and it makes sense to use it that way

picklelo (approved):
This looks good to me!

ElijahAhianyo (approved):

PR #819, opened 2023-04-14
Update Dockerfile

cheuora:
Docker Container using Dockerfile in pynecone doesn't contain auto install of nextjs, bunjs and other mandatory libs. So I added install scripts to this file.

### DIFF
--- a/docker-example/Dockerfile
+++ b/docker-example/Dockerfile
@@ -2,7 +2,6 @@ FROM python:3.11-slim as base
 
 RUN adduser --disabled-password pynecone
 
-
 FROM base as build
 
 WORKDIR /app
@@ -25,6 +24,9 @@ RUN apt-get update && apt-get install -y \
     nodejs \
     unzip \
     && rm -rf /var/lib/apt/lists/*
+RUN curl -fsSL https://bun.sh/install | bash 
+
+
 
 ENV PATH="/app/venv/bin:$PATH"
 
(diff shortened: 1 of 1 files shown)


docker-example/Dockerfile
```diff
@@ -25,6 +24,9 @@ RUN apt-get update && apt-get install -y \
     nodejs \
     unzip \
     && rm -rf /var/lib/apt/lists/*
+RUN curl -fsSL https://bun.sh/install | bash
```
ElijahAhianyo, line 27: we should definitely specify the version of bun to install for compatibility

Alek99: You probably can get rid of Bun install because we install it automatically install it for the user if they don't have it

cheuora: > You probably can get rid of Bun install because we install it automatically install it for the user if they don't have it

if you delete `RUN curl -fsSL https://bun.sh/install | bash` , you will meet `Cannot'find /pynecone/.bun` error !

mouadessalim:
actually working locally but not on web servers, i've tryed usingt you're dockerfile but apparently there is some port problems

r0b2g1t:
> actually working locally but not on web servers, i've tryed using you're dockerfile in [fly.io](https://fly.io) but apparently there is some port problems

Maybe you can explain the which error message you get when you try to run it on fly.io?

r0b2g1t:
I had reproduced the error why the packages are not installed during the start.
This a [bug](https://github.com/oven-sh/bun/issues/1590) of bun in non interactive environments.
It is not necessary to edit the Dockerfile to install anything additionally.
So, the PR can closed.

Alek99:
Closing as https://github.com/pynecone-io/pynecone/pull/860 fixes this.

PR #840, opened 2023-04-19
Only update ComputedVar when dependent vars change

### LINKED ISSUE #830: Computed vars are recomputed and retransmitted too often
**Describe the bug**

Whenever any value in a state changes, all computed values are recomputed and retransmitted over the websocket. If any computed value takes > 300ms or so to execute, it will create problems for app responsiveness, particularly with fully-controlled inputs, but also simple events will get blocked waiting for vars to recompute.

**To Reproduce**

Here's my contrived example

```python
import time

import pynecone as pc


class State(pc.State):

    should_take_a_long_time: bool = True
    input_value: str = ""
    filter_value: str = ""

    def toggle_should_take_a_long_time(self) -> None:
        self.should_take_a_long_time = not self.should_take_a_long_time

    def set_filter_value(self) -> None:
        self.filter_value = self.input_value

    @pc.var
    def filtered_items(self) -> list[str]:
        if self.filter_value:
            print(f"FILTERING on {self.filter_value}")
            if self.should_take_a_long_time:
                # simulate especially slow operation
                time.sleep(0.1 * len(self.filter_value))
            return [self.filter_value] * 3
        return []


def index() -> pc.Component:
    return pc.center(
        pc.vstack(
            pc.input(placeholder="filter by", value=State.input_value, on_change=State.set_input_value),
            pc.button("Speed: ", pc.cond(State.should_take_a_long_time, "Slow", "Fast"), on_click=State.toggle_should_take_a_long_time),
            pc.button("Filter", on_click=State.set_filter_value),
            pc.text("Filtered items: ", pc.foreach(State.filtered_items, pc.box)),
            spacing="1.5em",
            font_size="2em",
        ),
        padding_top="10%",
    )


app = pc.App(state=State)
app.add_page(index)
app.compile()
```

In this example, the app presents a fully controlled text input, a button that toggles whether the computed operation is slow or fast, a button to set a state value upon which a computed variable directly depends.

So the text input sends every change event over websocket to set `State.input_value` Var.

The Speed button toggles the boolean `State.should_take_a_long_time` Var.

The Filter button sets `State.filter_value` to the current `State.input_value`.

The ComputedVar `filtered_items` uses the `filter_value` and `should_take_a_long_time` vars to simulate some data crunching operation. Note that `filtered_items` explicitly does NOT depend on `input_value`, which changes a lot and would be expensive to "re-filter" the data as the user is typing.

**Expected behavior**
1. `filtered_items` is only recomputed when the `Var` that it uses change ("reactive state").
2. After initial filtering, the text box is basically unusable, because the server is blocked processing the slow recompute operation. Would expect text input to rely less on server side state (can we use async computed vars?).

 - Python Version: 3.11
 - Pynecone Version: 1.0.24
 - OS: macOS 13.2.1
 - Browser (Optional): Firefox, Safari
 
**Additional context**

It's obvious from [reading the code](https://github.com/pynecone-io/pynecone/blob/c3148d348cc1fb29eb281c9d9a065da81568c3a4/pynecone/state.py#L631-L635) that it's intentionally recalculating and retransmitting all `ComputedVar` in the state when retrieving a delta. This makes it really hard to have controlled inputs mixed with expensive computed vars in the same state, so the natural solution is to move to substates...

```python
import time

import pynecone as pc


class State(pc.State):

    should_take_a_long_time: bool = True
    filter_value: str = ""

    def toggle_should_take_a_long_time(self) -> None:
        self.should_take_a_long_time = not self.should_take_a_long_time

    @pc.var
    def filtered_items(self) -> list[str]:
        if self.filter_value:
            print(f"FILTERING on {self.filter_value}")
            if self.should_take_a_long_time:
                # simulate especially slow operation
                time.sleep(0.1 * len(self.filter_value))
            return [self.filter_value] * 3
        return []


class ControlledInputState(State):
    input_value: str = ""

    def set_filter_value(self) -> None:
        self.filter_value = self.input_value


def index() -> pc.Component:
    return pc.center(
        pc.vstack(
            pc.input(placeholder="filter by", value=ControlledInputState.input_value, on_change=ControlledInputState.set_input_value),
            pc.button("Speed: ", pc.cond(State.should_take_a_long_time, "Slow", "Fast"), on_click=State.toggle_should_take_a_long_time),
            pc.button("Filter", on_click=ControlledInputState.set_filter_value),
            pc.text("Filtered items: ", pc.foreach(State.filtered_items, pc.box)),
            spacing="1.5em",
            font_size="2em",
        ),
        padding_top="10%",
    )


app = pc.App(state=State)
app.add_page(index)
app.compile()
```

This actually hits the exact same problem as the original example, even though it _looks_ pretty good. It seems like the `input_value` would be relegated to a child state `ControlledInputState` and not affect the parent... but because `filtered_values` is computed, and computed values are _always_ recalculated, the fact that the child state triggered a change, means that the app asks for a delta from the parent state (and recalculates the expensive computed operations).

Am I holding it wrong?

```python
import time

import pynecone as pc


class State(pc.State):
    should_take_a_long_time: bool = True
    filter_value: str = ""

    def toggle_should_take_a_long_time(self) -> None:
        self.should_take_a_long_time = not self.should_take_a_long_time


class FilterState(State):
    @pc.var
    def filtered_items(self) -> list[str]:
        if self.filter_value:
            print(f"FILTERING on {self.filter_value}")
            if self.should_take_a_long_time:
                # simulate especially slow operation
                time.sleep(0.1 * len(self.filter_value))
            return [self.filter_value] * 3
        return []


class ControlledInputState(State):
    input_value: str = ""

    def set_filter_value(self) -> None:
        self.filter_value = self.input_value


def index() -> pc.Component:
    return pc.center(
        pc.vstack(
            pc.input(placeholder="filter by", value=ControlledInputState.input_value, on_change=ControlledInputState.set_input_value),
            pc.button("Speed: ", pc.cond(State.should_take_a_long_time, "Slow", "Fast"), on_click=State.toggle_should_take_a_long_time),
            pc.button("Filter", on_click=ControlledInputState.set_filter_value),
            pc.text("Filtered items: ", pc.foreach(FilterState.filtered_items, pc.box)),
            spacing="1.5em",
            font_size="2em",
        ),
        padding_top="10%",
    )


app = pc.App(state=State)
app.add_page(index)
app.compile()
```

☝️ this one _really_ seems like it should address the `input_value` and `filtered_items` coupling by putting the ComputedVar into a separate child state, a sibling to the `input_value`, forcing all communication between these states to go through the common parent Var, `filter_value`.

**This one actually doesn't work at all** 😲.

Even though the ComputedVar `FilterState.filtered_items` directly depends on its parent Var `filter_value`, we actually have the opposite problem as above: **the child state does NOT recalculate or retransmit computed vars when the dependent vars in the parent state change!**

# Solutions

I prototyped a solution to address both of these issues, as well as #432 while actually improving performance.

You can see the work in my [per-var-dirty](https://github.com/masenf/pynecone/tree/per-var-dirty) branch.

In essence, `State` will now keep track of when it sees `Var` access from a `ComputedVar` (`computed_var_dependencies`) via the existing `__getattribute__` implementation. Additionally, it will also keep track when a substate accesses an inherited var (`substate_var_dependencies`).

Using these new mappings, the `mark_dirty` logic can now iterate through known substate dependencies and mark them dirty when the variables they depend on have changed.

When getting the state delta, the `dirty_vars` set is used to reach into `computed_vars_dependencies` to identify only the `ComputedVars` that depend on the variables that have changed.

masenf:
When a `ComputedVar` function accesses a `Var`, `ComputedVar`, or backend var via a state instance (`State.__getattribute__`), the new `computed_var_dependencies` dict will keep track of `requested_variable -> Set[dependent ComputedVar]`. This is implemented via introspecting the stack (`_get_previous_recursive_frame_info`) to find the previous frame (`__getattribute__` called from the `ComputedVar`) and accessing the `name` of the requesting var.

During `get_delta` calculation, instead of recalculating all `ComputedVar`, only recalculate the `ComputedVar` that depend on the set of changed vars. This determination is made via recursive function (`State._dirty_computed_vars`) to accommodate `ComputedVar` that
depend on other `ComputedVar`.

### All Submissions:

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue) #830 
  - If existing `ComputedVar` is not accessing all of its dependencies on all paths (i.e. some Var access is conditional), this change could result in the `ComputedVar` NOT updating in response to state changes when it would have before.
  - New semantics (and especially a warning about above) should be added to the "State" page of the docs. 

### New Feature Submission:

### Changes To Core Features:

Closes #830

### DIFF
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import asyncio
 import functools
+import inspect
 import traceback
 from abc import ABC
 from typing import (
@@ -14,6 +15,7 @@ from typing import (
     Optional,
     Sequence,
     Set,
+    Tuple,
     Type,
     Union,
 )
@@ -50,6 +52,9 @@ class State(Base, ABC):
 
     # Backend vars inherited
     inherited_backend_vars: ClassVar[Dict[str, Any]] = {}
+
+    # Mapping of var name to set of computed variables that depend on it
+    computed_var_dependencies: ClassVar[Dict[str, Set[str]]] = {}
 
     # The event handlers.
     event_handlers: ClassVar[Dict[str, EventHandler]] = {}
@@ -171,6 +176,7 @@ class State(Base, ABC):
             **cls.base_vars,
             **cls.computed_vars,
         }
+        cls.computed_var_dependencies = {}
 
         # Setup the base vars at the class level.
         for prop in cls.base_vars.values():
@@ -470,12 +476,28 @@ class State(Base, ABC):
 
         If the var is inherited, get the var from the parent state.
 
+        If the Var is a dependent of a ComputedVar, track this status in computed_var_dependencies.
+
         Args:
             name: The name of the var.
 
         Returns:
             The value of the var.
         """
+        vars = {
+            **super().__getattribute__("vars"),
+            **super().__getattribute__("backend_vars"),
+        }
+        if name in vars:
+            parent_frame, parent_frame_locals = _get_previous_recursive_frame_info()
+            if parent_frame is not None:
+                computed_vars = super().__getattribute__("computed_vars")
+                requesting_attribute_name = parent_frame_locals.get("name")
+                if requesting_attribute_name in computed_vars:
+                    # Keep track of any ComputedVar that depends on this Var
+                    super().__getattribute__("computed_var_dependencies").setdefault(
+                        name, set()
+                    ).add(requesting_attribute_name)
         inherited_vars = {
             **super().__getattribute__("inherited_vars"),
             **super().__getattribute__("inherited_backend_vars"),
@@ -503,6 +525,7 @@ class State(Base, ABC):
 
         if types.is_backend_variable(name):
             self.backend_vars.__setitem__(name, value)
+            self.dirty_vars.add(name)
             self.mark_dirty()
             return
 
@@ -620,6 +643,28 @@ class State(Base, ABC):
         # Return the state update.
         return StateUpdate(delta=delta, events=events)
 
+    def _dirty_computed_vars(self, from_vars: Optional[Set[str]] = None) -> Set[str]:
+        """Get ComputedVars that need to be recomputed based on dirty_vars.
+
+        Args:
+            from_vars: find ComputedVar that depend on this set of vars. If unspecified, will use the dirty_vars.
+
+        Returns:
+            Set of computed vars to include in the delta.
+        """
+        dirty_computed_vars = set(
+            cvar
+            for dirty_var in from_vars or self.dirty_vars
+            for cvar in self.computed_vars
+            if cvar in self.computed_var_dependencies.get(dirty_var, set())
+        )
+        if dirty_computed_vars:
+            # recursive call to catch computed vars that depend on computed vars
+            return dirty_computed_vars | self._dirty_computed_vars(
+                from_vars=dirty_computed_vars
+            )
+        return dirty_computed_vars
+
     def get_delta(self) -> Delta:
         """Get the delta for the state.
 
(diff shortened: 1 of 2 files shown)


Alek99 (approved):
Nice Nikhil tested it look good!

PR #843, opened 2023-04-19
hydrate_middleware: always emit a full dict() when hydrating

### LINKED ISSUE #841: pc.cond is briefly showing what it shouldn't be showing
**Describe the bug**
When I use pc.cond to conditionally render something the part that it should not be rendering is displayed for a split second (sometimes longer) before the part that should be rendered appears.

**To Reproduce**
If you run the crm project in the pynecone-examples repo and log in to the crm and press F5 you will see that a page with the heading "Welcome to Pyneknown!" briefly before the crm page displays.

**Expected behavior**
Only the element that should be visible according to the pc.cond condition should be displayed.

**Screenshots**
See the attached video of what happens when I refresh the page while signed into the crm example project.

 - Python Version: 3.10
 - Pynecone Version: 0.1.24
 - OS: Ubuntu 22.10
 - Browser (Optional):

https://user-images.githubusercontent.com/110377310/232963233-559cb49a-783f-49fb-b016-839241aa7b1b.mp4

### LINKED ISSUE #842: When `on_load` is used, full initial state is not sent to the client
**Describe the bug**
When `on_load` is used, the initial hydrated state sent to the app is incomplete.

**To Reproduce**

```python
import pynecone as pc


class State(pc.State):
    value1: int = 0
    value2: str = "default from State"

    def increment_value1(self):
        self.value1 = self.value1 + 1

    def on_load(self):
        self.value2 = "set from on_load handler"


def index() -> pc.Component:
    return pc.center(
        pc.vstack(
            pc.text("value1 = ", State.value1),
            pc.text("value2 = ", State.value2),
            pc.button("Increment value1", on_click=State.increment_value1),
        ),
        padding_top="10%",
    )


app = pc.App(state=State)
app.compile()
```

1. Load the app; notice that `value2 = ` shows "default from State" briefly before changing to "set from on_load handler".
2. Click "increment value1" a few times; notice that `value1 = ` reflects the incremented value
3. Refresh the page

**Expected behavior**

Expect that `value1 = X` is displayed after refreshing the page, where X was the same value before refreshing (actual: `value1 = 0` is shown regardless of what X was before).

Note that `State.value1` _is_ still set correctly by clicking `increment value1` and finding that `value1 = X+1` is shown, where X was the same value as before the refresh.

**Screenshots**
https://user-images.githubusercontent.com/1524005/232967069-d7811b13-5048-4d8a-9a8c-935f0c81fb8f.mov

 - Python Version: 3.11.2
 - Pynecone Version: 1.0.24
 - OS: macOS 13.2.1
 - Browser (Optional): Firefox, Safari
 
**Additional context**
The problem appears to be in [these 2 lines](https://github.com/pynecone-io/pynecone/blob/a52145e52a7eb85b8484424ffcb0afe9c70ff6c2/pynecone/middleware/hydrate_middleware.py#L51-L52)

masenf:
Ensure the client has access to the full state, even when on_load handlers are provided.

Update tests to assert that the first StateUpdate emitted by preprocess is the full original dict from the state without any changes applied.

Add new test to assert that middleware only includes the original dict from the state when no on_load events are defined.

Fix #842

------------------

State.is_hydrated: client-side bool to check is backend state has arrived

* `hydrate_middleware.HydrateMiddleware` registers a new Var on the base State object: `is_hydrated`.
    
* during `preprocess` of hydrate events, additional `StateUpdate` with the payload `{"is_hydrated": true}` is emitted after normal hydrate.
    
the client side is able to render alternate content based on whether the initial hydrate event has returned via `State.is_hydrated`.
    
 fix #841

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

### DIFF
--- a/pynecone/middleware/hydrate_middleware.py
+++ b/pynecone/middleware/hydrate_middleware.py
@@ -38,19 +38,25 @@ class HydrateMiddleware(Middleware):
             else:
                 load_event = None
 
+            updates = []
+
+            # first get the initial state
+            delta = format.format_state({state.get_name(): state.dict()})
+            if delta:
+                updates.append(StateUpdate(delta=delta))
+
+            # then apply changes from on_load event handlers on top of that
             if load_event:
                 if not isinstance(load_event, List):
                     load_event = [load_event]
-                updates = []
                 for single_event in load_event:
                     updates.append(
                         await self.execute_load_event(
                             state, single_event, event.token, event.payload
                         )
                     )
-                return updates
-            delta = format.format_state({state.get_name(): state.dict()})
-            return StateUpdate(delta=delta) if delta else None
+
+            return updates
 
     async def execute_load_event(
         self, state: State, load_event: EventHandler, token: str, payload: Dict
--- a/pynecone/middleware/hydrate_middleware.py
+++ b/pynecone/middleware/hydrate_middleware.py
@@ -13,6 +13,12 @@ if TYPE_CHECKING:
     from pynecone.app import App
 
 
+IS_HYDRATED = "is_hydrated"
+
+
+State.add_var(IS_HYDRATED, type_=bool, default_value=False)
+
+
 class HydrateMiddleware(Middleware):
     """Middleware to handle initial app hydration."""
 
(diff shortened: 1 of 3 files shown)


picklelo (approved):
looks good to me, thanks for adding!

PR #846, opened 2023-04-19
Apply convert_handlers_to_fns for substates

csh3695:
Previously, functions of substates could not be called as normal functions during runtime.  

This bug fix recursively registers each substate's event handler function as an attribute so that it can be called as a normal function during runtime.

### All Submissions:

### Type of change


picklelo (approved):
good catch!

PR #852, opened 2023-04-21
state: track substate ComputedVar that depends on parent Var 

### LINKED ISSUE #839: Substate computed vars are not recalculated when Parent var changes
**Describe the bug**

Partly described as part of #830, but this is technically a separate issue.

When a child state has a computed variable that depends on a parent variable, the child is not recomputed when the parent state changes.

**To Reproduce**

```python
import time

import pynecone as pc


class State(pc.State):
    should_take_a_long_time: bool = True
    filter_value: str = ""

    def toggle_should_take_a_long_time(self) -> None:
        self.should_take_a_long_time = not self.should_take_a_long_time


class FilterState(State):
    @pc.var
    def filtered_items(self) -> list[str]:
        if self.filter_value:
            print(f"FILTERING on {self.filter_value}")
            if self.should_take_a_long_time:
                # simulate especially slow operation
                time.sleep(0.1 * len(self.filter_value))
            return [self.filter_value] * 3
        return []


class ControlledInputState(State):
    input_value: str = ""

    def set_filter_value(self) -> None:
        self.filter_value = self.input_value


def index() -> pc.Component:
    return pc.center(
        pc.vstack(
            pc.input(placeholder="filter by", value=ControlledInputState.input_value, on_change=ControlledInputState.set_input_value),
            pc.button("Speed: ", pc.cond(State.should_take_a_long_time, "Slow", "Fast"), on_click=State.toggle_should_take_a_long_time),
            pc.button("Filter", on_click=ControlledInputState.set_filter_value),
            pc.text("Filtered items: ", pc.foreach(FilterState.filtered_items, pc.box)),
            spacing="1.5em",
            font_size="2em",
        ),
        padding_top="10%",
    )


app = pc.App(state=State)
app.add_page(index)
app.compile()
```

**Expected behavior**

Expect, when `ControlledInputState.set_filter_value` sets `State.filter_value` that the ComputedVar `FilterState.filtered_items` is recalculated and reflects the change.

Instead nothing happens when `State.filter_value` is modified.

 - Python Version: 3.11.2
 - Pynecone Version: 1.0.24
 - OS: macOS 13.2.1
 - Browser (Optional): Safari + Firefox
 
**Additional context**

I have a fix for this one as well, will clean it up and submit it after the fix for #830 goes in.

masenf:
Follow on from #840 and #917, that introduces substate dependence determination at `__init__` time so that each state tracks which variables its substates depend on. When marking a state dirty, if any dirty vars are dependencies of substate ComputedVar, then pass the dirty var down to the substate and also mark it dirty.

Fix #839 

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

Edit 2023-04-23: rebased after 840 merged, diff should be clean now and ready for review!
Edit 2023-05-03: rebased after #917 merged, diff is now clean and ready for review!

Thanks for taking my previous patch! 🙏


picklelo:
@masenf I put up a PR to speed up the dependency management: https://github.com/pynecone-io/pynecone/pull/864/files

Once that's merged, can you rebase this logic into the new format? Appreciate all your contributions!!

masenf:
Now that computed vars dependencies are tracked all the way to the base state, there's no need to track explicit substate var dependencies...can just push all of the `dirty_vars` down into each substate and let the substate determine which vars need to be recomputed.

This might still result in calculating some _parent_ computed var multiple times, especially if lots of other variables depend on it. Would need to poke around and see if computed var values get cached already or not. The problem is solvable because the nature of substate tree means parent computed vars will always be recalculated before substate is called.

masenf:
@picklelo i think we can get away with removing `dirty_substate` tracking altogether!

pynecone/state.py
```diff
@@ -706,32 +703,24 @@ def get_delta(self) -> Delta:
         if len(subdelta) > 0:
             delta[self.get_full_name()] = subdelta
 
-        # Recursively find the substate deltas.
-        substates = self.substates
-        for substate in self.dirty_substates:
-            delta.update(substates[substate].get_delta())
+        # Recursively find all substate deltas.
+        for substate in self.substates.values():
```
picklelo, line 707: I think we may want to keep the `dirty_substates` prop. This will work, but as the substate tree grows very big, this may end up having some performance impact as we will call `get_delta()` for every substate in the tree.

picklelo (commented):
Nice! I left one comment - I think the `dirty_substates` prop may be useful for performance reasons.

I have this PR #872  that temporarily disables the computed var dependency checks until we have the static checks from #870 so we don't have any breaking changes.

We can either merge this PR in before #870 with the checks disabled for now, or wait until we have that, either way should be okay.

masenf:
The issue that I was having with dirty_substates was how to know when a grandchild that depended on a parent var needs to update?

I was passing all of the dirty vars into each substate (to pass into their substates) because the child state doesn't know what the grandchild computed var dependencies are without checking (and it was getting complicated).

I ended up simplifying by removing dirty_substates because my approach basically put all substates in the dirty set, so it made more sense up exhaustively iterate, and have each state only responsible for its own computed var deps.

I'll try to rework this with more perf consideration in mind, but I don't want to do it by making the code excessively complicated, unless a clear perf benefit is noted.

picklelo:
@masenf I just tested it out on our [pcweb](https://github.com/pynecone-io/pcweb) repo since it has a huge state tree, and the process time for state updates went from around `0.0010` to `0.0026` so I think there is some perf impacts.

The `mark_dirty` method recursively calls `self.parent_state.mark_dirty()` so the grandchild state should mark all the states up to the parent dirty. Can we add the grandchild computed var in the parent var dependencies - and when the parent var updates we can mark the grandchild as dirty?

masenf:
I'm working on a "deep state" test that has 1000+ state nodes for doing deterministic perf testing.

I'm wondering if the extra time on pcweb is due to inefficiency in the algorithm, or the fact that substate delta was just not recomputed at all before this change.

picklelo:
Yeah I don't think substate computed vars that only depended on parent vars were computed before. If the performance impact is too large, maybe this limitation is okay since it may be a rare use case. But if there's a clean solution would be good to include.

masenf:
@picklelo the latest update reduces the process time by an order of magnitude over my previous patch. still not ready to go in (docstrings, tests, etc), but I wanted to post my progress thus far. The main wins comes from calculating the explicit var -> substate dependencies _after_ tracking during `__init__` and pushing those relationships into each parent state. Then when `mark_dirty` is called, it only has to mark substates dirty that actually depend on the changed vars, which makes a significant difference for variables that only have 1 deep dependency.

One of the things that's going to really hurt us in real world perf is recalculating the ComputedVar everytime it's accessed by some dependency. For high-level, expensive computed vars, this could make things _really_ slow if a lot of substates depend on it.

masenf:
I'm pretty happy with how this one turned out. With this PR (including #916 and #917), there appears to be no regression processing state updates in the `pcweb` project!

Give it a shot and let me know what you think.

pynecone/state.py
```diff
@@ -377,6 +397,10 @@ def add_var(cls, name: str, type_: Any, default_value: Any = None):
         cls.base_vars.update({name: var})
         cls.vars.update({name: var})
 
+        # let substates know about the new variable
+        for substate_class in cls.__subclasses__():
```
picklelo, line 401: nit: we can use `cls.get_substates()` here

pynecone/state.py
```diff
@@ -724,14 +748,27 @@ def get_delta(self) -> Delta:
 
     def mark_dirty(self):
         """Mark the substate and all parent states as dirty."""
-        if self.parent_state is not None:
+        state_name = self.get_name()
+        if (
+            self.parent_state is not None
+            and state_name not in self.parent_state.dirty_substates
+        ):
             self.parent_state.dirty_substates.add(self.get_name())
```
picklelo, line 756: We can just use `state_name` here

picklelo (approved):
nice! Just some nits we can clean up in a follow up

PR #856, opened 2023-04-22
Modify some methods in state.py to prevent being overwritten 

### LINKED ISSUE #834: Make methods in state.py private
Currently some methods in the `State` class are not private, and it prevents us from adding event handlers with certain names. For example, you cannot add an event handler names `process`, `preprocess`, or `postprocess`. We should convert the methods to be e.g. `_process`

alan707:
### Description

- closes #834 
- Prevents some of the methods in `state.py` from being overwritten if someone calls event handler built-ins
- Fixes unit tests

### All Submissions:

### Type of change

- [x] Breaking change

### DIFF
--- a/pynecone/app.py
+++ b/pynecone/app.py
@@ -455,7 +455,7 @@ async def process(
 
     # Apply the event to the state.
     if updates is None:
-        updates = [await state.process(event)]
+        updates = [await state._process(event)]
         app.state_manager.set_state(event.token, state)
 
     # Postprocess the event.
@@ -534,7 +534,7 @@ def upload(app: App):
             name=handler,
             payload={handler_upload_param[0]: files},
         )
-        update = await state.process(event)
+        update = await state._process(event)
         return update
 
     return upload_file
--- a/pynecone/middleware/hydrate_middleware.py
+++ b/pynecone/middleware/hydrate_middleware.py
@@ -76,6 +76,6 @@ class HydrateMiddleware(Middleware):
                 "The value of state cannot be None when processing an on-load event."
             )
 
-        return await state.process_event(
+        return await state._process_event(
             handler=load_event, state=ex_state, payload=payload, token=token
         )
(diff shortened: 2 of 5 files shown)


masenf:
> Bug fix (non-breaking change which fixes an issue)

Changing a previously public method into a private method **_IS_ a breaking change**. The following code that I submitted to the `pynecone-examples` repo last week breaks with this change:

https://github.com/pynecone-io/pynecone-examples/blob/cc28c87ce9ce6017d722fbf8fbc96650318f8a87/chatroom/chatroom/chatroom.py#L94

Is `State._process` actually becoming a private method? As in indicating it's not safe to call it from outside the class? If so, then I think an alternative API is needed to allow out-of-band event handling.

picklelo:
@masenf The intention isn't necessarily to make it private - but to allow people to define event handlers named `process`, etc. Currently if someone makes an event handler named `process` it overrides this method and breaks the app. I was thinking we add `_` to all the builtin State methods to avoid this confusion. Open to other suggestions though, but need to avoid this confusion.

alan707:
Hi, @masenf 👋! @picklelo's explanation is aligned with what I understood from the issue. It sounded like the request was to prevent those methods from getting overwritten unbeknownst to the person that makes those changes.  I think a "dunder" (double under-score) would make them _more private_ but I don't think that was the intention.   

You are correct these are breaking changes, so for your maybe we can consider adding tests for the example apps to prevent breaking changes from merging.  Thoughts?

masenf:
> consider adding tests for the example apps

I think this is a great idea.

As for this PR, I think it could go ahead as is. Thanks @picklelo for the explanation. Ultimately I'd rather have a real API for dispatching out of band events, rather than calling into State anyway. It does seem like almost too low level for most people to hook, so having the leading under does make sense.

Thanks for putting this in!

picklelo (approved):

PR #864, opened 2023-04-25
Speed up computed var dependency tracking

picklelo:
Changed to do the tracking up front during `__init__` to improve performance during runtime.


masenf (commented):
Not sure what exactly the problem is, but the follow test case fails with this change:

```python
def test_child_state():
    class MainState(State):
       v: int = 2

    class ChildState(MainState):
        @ComputedVar
        def rendered_var(self):
            return self.v

    ms = MainState()
    cs = ms.get_substate(ChildState.get_name())
    assert ms.v == 2
    assert cs.v == 2
    assert cs.rendered_var == 2
```

-----------------------------------

```console
____________________________________________________________________ test_child_state _____________________________________________________________________

    def test_child_state():
        class MainState(State):
           v: int = 2
    
        class ChildState(MainState):
            @ComputedVar
            def rendered_var(self):
                return self.v
    
>       ms = MainState()

tests/test_state.py:825: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
pynecone/state.py:93: in __init__
    self.substates[substate.get_name()] = substate().set(parent_state=self)
pynecone/state.py:111: in __init__
    self.__getattribute__(cvar)
pynecone/state.py:530: in __getattribute__
    return super().__getattribute__(name)
tests/test_state.py:823: in rendered_var
    return self.v
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <[AttributeError("'NoneType' object has no attribute 'v'") raised in repr()] ChildState object at 0x137761550>, name = 'v'

    def __getattribute__(self, name: str) -> Any:
        """Get the state var.
    
        If the var is inherited, get the var from the parent state.
    
        If the Var is a dependent of a ComputedVar, track this status in computed_var_dependencies.
    
        Args:
            name: The name of the var.
    
        Returns:
            The value of the var.
        """
        try:
            # Check if tracking is enabled.
            if super().__getattribute__("track_vars"):
                # Get the non-computed vars.
                all_vars = {
                    **super().__getattribute__("vars"),
                    **super().__getattribute__("backend_vars"),
                }
                # Add the var to the tracked vars.
                if name in all_vars:
                    super().__getattribute__("tracked_vars").add(name)
        except AttributeError:
            pass
    
        inherited_vars = {
            **super().__getattribute__("inherited_vars"),
            **super().__getattribute__("inherited_backend_vars"),
        }
        if name in inherited_vars:
>           return getattr(super().__getattribute__("parent_state"), name)
E           AttributeError: 'NoneType' object has no attribute 'v'

pynecone/state.py:527: AttributeError
```

I think it has something to do with [this line](https://github.com/pynecone-io/pynecone/pull/864/files#diff-cd99a17b23a92f255b06f6811a1dbf6db64cf86404193878dfda96cafee4d34dR93); it's setting the `parent_state` attribute _after_ `__init__`, so the state tracking breaks when it has to access an inherited var.

picklelo:
@masenf nice catch, you're right it's because the `parent_state` wasn't being set until after. I've updated the PR to handle this test case - just need to merge in #866 first to fix the build

Alek99 (approved):
nice lgtm

Issue #870, opened 2023-04-25
Use static analysis to determine computed var deps


picklelo:
Currently we calculate the computed vars dependencies using a bit of a hack where we track which base vars are accessed: https://github.com/pynecone-io/pynecone/blob/f019e0e55a858d9942fde31e883278ea36aeeaef/pynecone/state.py#L105

This can sometimes miss cases such as conditionals within the computed vars.

We should use static analysis and the `ast` module to do this more solidly.

masenf:
What is the timeframe on this? How do you plan to account for long chains of dependent vars (through multiple computed vars)?

picklelo:
Not sure on the timeframe, but I think we need an approach like this to catch some cases.

Or we could do a hybrid where we keep the current approach and use static analysis just to detect conditionals within the computed var functions, which I think is the main issue. And if there are if statements maybe always mark the computed var as dirty.

PR #872, opened 2023-04-26
Disable computed var deps

picklelo:
Disable computed var dependency checking - will re-enable with #870


Alek99 (approved):
Lg excited for the static analysis coming soon

PR #887, opened 2023-04-26
Install dependencies in prod mode

picklelo:


Alek99 (approved):

pynecone/utils/build.py
```diff
@@ -126,6 +126,12 @@ def setup_frontend(root: Path):
     Args:
         root: root path of the project.
     """
+    # Initialize the web directory if it doesn't exist.
+    web_dir = prerequisites.create_web_directory(root)
```
milochen0418, line 130: When use call pc init, the web folder will be there.
Because pc init will do 
```
        prerequisites.install_bun()
        prerequisites.initialize_web_directory()
```

Why we still need to use this?
`web_dir = prerequisites.create_web_directory(root)`

PR #899, opened 2023-04-28
Fix event handler calls to handlers

picklelo:


Alek99 (approved):

masenf:
@picklelo in case you didn't see it, looks like this change [regressed an example in pcweb](https://github.com/pynecone-io/pcweb/issues/62)

PR #903, opened 2023-04-28
Update Dockerfile example

picklelo:


r0b2g1t:
Hey I would not recommend this Dockerfile approach. It is less secure (running Pynecone as root) and the image size will be blown up by a lot of unnecessary software.

r0b2g1t:
The Dockerfile is missing basic functions like the default entrypoint of the image. which is called when the image is started without command, as present in the current example. Likewise, Pynecone will launch the project in dev mode.  Likewise, Pynecone will launch the project in dev mode. In addition, an Ubuntu base image is not necessary, Python-slim is sufficient and is by far smaller than the result of the build. I would not recommend doing an image build based on this Dockerfile, as it gives the impression that it can be used productively that way.

r0b2g1t (changes requested):
See comments before.

Alek99:
I see nice fix will test this today.

PR #916, opened 2023-04-30
Analyze computed var dependencies

### LINKED ISSUE #870: Use static analysis to determine computed var deps

masenf:
Instead of tracking vars accessed by a `ComputedVar` at `State.__init__` time, disassemble the `ComputedVar` getter function and iterate over all bytecode instructions to find attribute access on `self` (even indirect via helper method).

This approach was taken over `ast` to allow for dynamically generated State objects not defined in a file/source can also be used.

Closes #870.

### All Submissions:

### Type of change

  - it's possible that some State is depending on always recalculating ComputedVar, if these functions are not properly using `self.x` to access `x`, then they might not be updated when `x` changes.

### Changes To Core Features:

### DIFF
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -72,13 +72,7 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
     router_data: Dict[str, Any] = {}
 
     # Mapping of var name to set of computed variables that depend on it
-    computed_var_dependencies: Dict[str, Set[str]] = {}
-
-    # Whether to track accessed vars.
-    track_vars: bool = False
-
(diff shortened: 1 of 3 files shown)


masenf:
Merged as part of #917

PR #917, opened 2023-04-30
Cache ComputedVar

masenf:
Building upon #916 ([👉**view only the diff from 916**](https://github.com/pynecone-io/pynecone/pull/917/files/18d16e4ee246bc6208a2884dbfc1a8d29a02b1e9..53f3232d0539134c1574464f5858fb356f35c45d)), the purpose of this PR is to avoid recalculating potentially expensive `ComputedVar` values unless their dependencies have actually changed.

If many substates depend on an expensive `ComputedVar` in a parent state, then this ensures that it is only recalculated once per dirtying of dependencies.

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

### DIFF
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -72,7 +72,7 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
     router_data: Dict[str, Any] = {}
 
     # Mapping of var name to set of computed variables that depend on it
-    computed_var_dependencies: Dict[str, Set[str]] = defaultdict(set)
+    computed_var_dependencies: Dict[str, Set[str]] = {}
 
     def __init__(self, *args, parent_state: Optional[State] = None, **kwargs):
         """Initialize the state.
@@ -95,14 +95,15 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
             fn.__qualname__ = event_handler.fn.__qualname__  # type: ignore
             setattr(self, name, fn)
 
-        # Initialize the mutable fields.
-        self._init_mutable_fields()
-
         # Initialize computed vars dependencies.
+        self.computed_var_dependencies = defaultdict(set)
         for cvar_name, cvar in self.computed_vars.items():
             # Add the dependencies.
             for var in cvar.deps():
                 self.computed_var_dependencies[var].add(cvar_name)
+
+        # Initialize the mutable fields.
+        self._init_mutable_fields()
 
     def _init_mutable_fields(self):
         """Initialize mutable fields.
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -645,6 +645,18 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
         # Return the state update.
         return StateUpdate(delta=delta, events=events)
 
+    def _mark_dirty_computed_vars(self) -> None:
+        """Mark ComputedVars that need to be recalculated based on dirty_vars."""
+        dirty_vars = self.dirty_vars
+        while dirty_vars:
+            calc_vars, dirty_vars = dirty_vars, set()
+            for cvar in self._dirty_computed_vars(from_vars=calc_vars):
+                self.dirty_vars.add(cvar)
+                dirty_vars.add(cvar)
+                actual_var = self.computed_vars.get(cvar)
+                if actual_var:
+                    actual_var.mark_dirty(instance=self)
+
     def _dirty_computed_vars(self, from_vars: Optional[Set[str]] = None) -> Set[str]:
         """Determine ComputedVars that need to be recalculated based on the given vars.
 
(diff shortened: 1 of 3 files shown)


pynecone/state.py
```diff
@@ -105,18 +99,10 @@ def __init__(self, *args, parent_state: Optional[State] = None, **kwargs):
...
-        self.computed_var_dependencies = defaultdict(set)
-        for cvar in self.computed_vars:
-            self.tracked_vars = set()
-
-            # Enable tracking and get the computed var.
-            self.track_vars = True
-            self.__getattribute__(cvar)
-            self.track_vars = False
-
+        for cvar_name, cvar in self.computed_vars.items():
```
picklelo, line 102: I think we may need to keep the line `self.computed_var_dependencies = defaultdict(set)` - because of Python weirdness when we set the initial value to a mutable type like list/dict it becomes shared by all the subclasses. We ran into this issue a couple times

picklelo (commented):
Awesome! One small comment but looks good initially. I'll review this more thoroughly and try to include in the following release.

pynecone/state.py
masenf: _that_ probably explains the weirdness i was hitting w.r.t. accessing that value in a class context vs an instance context. will rework and see if i can add a test for this.

masenf: @picklelo i moved the `defaultdict` assignment back into `__init__`, so that each instance gets a copy, rather than sharing the class copy...

However to make it work, I had to also move the `_init_mutable_values` call _after_ `computed_var_dependencies` is assigned, because `_init_mutable_vars` might call `__setattr__`, which will call `mark_dirty`, which attempts to look up any dirty vars in the `computed_var_dependencies`, which won't work, unless it's been assigned as a default dict... turns out _that_ was the problem I was hitting before.

After moving the `_init_mutable_values` call, there's no problem with assigning `computed_var_dependencies` in `__init__`.

picklelo (approved):
Tested this out on `pcweb` and seems to work really well - awesome work! Much cleaner and more performant calculating the dependencies upfront like this.

pynecone/var.py
```diff
@@ -801,6 +806,84 @@ def name(self) -> str:
...
+        self_is_top_of_stack = False
+        for instruction in dis.get_instructions(obj):
+            if instruction.opname == "LOAD_FAST" and instruction.argval == self_name:
+                self_is_top_of_stack = True
+                continue
+            if self_is_top_of_stack and instruction.opname == "LOAD_ATTR":
+                d.add(instruction.argval)
+            elif self_is_top_of_stack and instruction.opname == "LOAD_METHOD":
+                d.update(self.deps(obj=getattr(self.__objclass__, instruction.argval)))
+            self_is_top_of_stack = False
```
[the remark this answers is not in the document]
picklelo: Thanks for sharing will read more about it

ElijahAhianyo:
@masenf This  PR seems to error with this code:

```python
def dynamic_index():
    return pc.text("hello")


app = pc.App(state=pc.State)
app.add_page(catchall_index, route="path/[var_arg]")
app.compile()

```


```shell
  File "/Users/eli/Documents/work/Pynecone-io/pynecone-forks/elijah/pynecone/pynecone/compiler/compiler.py", line 205, in compile_page
    code = _compile_page(component, state)
  File "/Users/eli/Documents/work/Pynecone-io/pynecone-forks/elijah/pynecone/pynecone/compiler/compiler.py", line 81, in _compile_page
    state=utils.compile_state(state),
          ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/eli/Documents/work/Pynecone-io/pynecone-forks/elijah/pynecone/pynecone/compiler/utils.py", line 104, in compile_state
    initial_state = state().dict()
                    ^^^^^^^
  File "/Users/eli/Documents/work/Pynecone-io/pynecone-forks/elijah/pynecone/pynecone/state.py", line 103, in __init__
    for var in cvar.deps():
               ^^^^^^^^^^^
  File "/Users/eli/Documents/work/Pynecone-io/pynecone-forks/elijah/pynecone/pynecone/var.py", line 873, in deps
    d.update(self.deps(obj=getattr(self.__objclass__, instruction.argval)))
                                   ^^^^^^^^^^^^^^^^^
AttributeError: 'ComputedVar' object has no attribute '__objclass__'. Did you mean: '__class__'?

```

masenf:
@ElijahAhianyo  thanks for the report, will fix up promptly

masenf:
Issue has been identified. Fix is simple, just doing some due diligence to add a test case and see if there are other parts of the code base that add ComputedVar directly to a state class.

masenf:
@ElijahAhianyo fix is [here](https://github.com/pynecone-io/pynecone/pull/952)

PR #941, opened 2023-05-04
Improve Get Attribute error message for computed vars

ElijahAhianyo:
Throw attribute error with custom message when an attribute a non-existent attribute is accessed from a computed var with right annotation or when a computed var is wrongly annotated(eg. returning an object instance in a computed var annotated as a str). 
Also throws a `TypeError` when the annotation of a computed var is not provided or annotated with `Typing.Any`

### DIFF
--- a/pynecone/var.py
+++ b/pynecone/var.py
@@ -273,26 +273,32 @@ class Var(ABC):
             The var attribute.
 
         Raises:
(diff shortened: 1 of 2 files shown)


picklelo (approved):
Nice check and message

PR #956, opened 2023-05-06
Trigger on_load when router completes navigation

### LINKED ISSUE #609: Is there an on_load event handler for dynamic routes?
Dynamic routes are very useful and I've been using them well.

However, I would like to have a specific function (event handler) triggered when navigating to a dynamic route. Is there any way to do this?

For example, let's say I defined a dynamic route like this

```
app.add_page(post, route="/posts/[post_id]", on_load=PostState.init)
```

When written like this, we expect it to behave like this

- When we navigate to `/posts/1`, we can call `PostState.init()` to initialize the vars in the state.
- When we move to `/posts/2`, we similarly call `PostState.init()` to initialize the vars in State.

However, for now, the on_load event is only called once in the above scenario, i.e. we don't call `PostState.init()` when we navigate to `/posts/2`.

I don't think I'm calling the on_load event for every dynamic route.

Is there any way to fix this?

masenf:
* Move `router_data` to a backend var on the root state, allowing for more natural change tracking and avoid the need to recursively copy router_data into substates.
* Only assign `router_data` if it has changed, not for every request. This allows ComputedVar that depends on router_data to be properly cached and recalculated when the router data actually changes.
* Use the `routeChangeComplete` event to trigger a state re-hydrate when the page changes
  * Existing page `on_load` mechanism is used to handle the event server side.

Fix #609

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

### DIFF
--- a/pynecone/app.py
+++ b/pynecone/app.py
@@ -428,8 +428,8 @@ async def process(
     state = app.state_manager.get_state(event.token)
 
     # Add request data to the state.
-    state.router_data = event.router_data
-    state.router_data.update(
+    router_data = event.router_data
+    router_data.update(
         {
             constants.RouteVar.QUERY: format.format_query_params(event.router_data),
             constants.RouteVar.CLIENT_TOKEN: event.token,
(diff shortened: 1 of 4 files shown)


masenf:
Converting to draft until #958 can be resolved, since this PR depends on backend_vars.

PR #959, opened 2023-05-06
Per-instance backend variables

### LINKED ISSUE #958: _backend variables are not correctly separated for each session
**Describe the bug**
When opening a new tab, a new session state should be created, but it seems that backend variables end up being shared between states.

**To Reproduce**
 ```
class State(pc.State):
    _test: str = "test"

    def test(self):
        print(self._test)
        if self._test == "test":
            self._test = "test2"


@pc.route()
def index() -> pc.Component:
    return pc.button(
        "Check _backend var",
        on_click=State.test,
    )


app = pc.App(state=State)
app.compile()
```
Steps:
1. Run snippet and open page in a tab
2. Click on the button (console should show 'test' )
3. Open a new tab.
4. Click on the button (console should show 'test' but show 'test2' instead)

**Expected behavior**
Backend variables correctly separated for each state.

 - Python Version: 3.10
 - Pynecone Version: main branch

masenf:
Create a copy of the default `backend_vars` from the State class and assign it to the instance during `__init__`.

Fixes #958

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

### DIFF
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import asyncio
+import copy
 import functools
 import traceback
 from abc import ABC
@@ -74,6 +75,9 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
     # Mapping of var name to set of computed variables that depend on it
     computed_var_dependencies: Dict[str, Set[str]] = {}
 
+    # Per-instance copy of backend variable values
+    _backend_vars: Dict[str, Any] = {}
+
     def __init__(self, *args, parent_state: Optional[State] = None, **kwargs):
         """Initialize the state.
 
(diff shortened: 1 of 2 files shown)


masenf:
@Lendemor `pyright` and `ruff` are kind of fighting each other here.

ruff says not to use `setattr` or `getattr` with a string literal attribute name, okay fair.

but pyright says not to overwrite a instance attribute marked as a `ClassVar`, which also makes good sense.

So trying to think of the best way to proceed with this approach, but for the moment at least seems stalled.

Opened the PR here, in case anyone had other ideas or wanted to comment on this diff without continuing to add stuff to the issue discussion.

masenf:
a resolution was reached by using an explicit `_backend_vars` field for the instance copy of the backend variables that doesn't shadow the name of the `ClassVar`. I think there might be other places in the State class where that is happening as well, but will save that problem for another day.

This is now ready for review.

Alek99 (approved):
Lgtm. Also tested with some more nested ds and works well.

```python 

class State(pc.State):
   _test: str = "test"

   _test_data  ={
         "test": "test",   
         "more": ["test", "test2"]
   }

   def test(self):
        print(self._test)
        print(self._test_data)
        self._test = random.randint(0, 100)
        self._test_data["test"] = random.randint(0, 100)
        self._test_data["more"].append(random.randint(0, 100))


@pc.route()
def index() -> pc.Component:
   return pc.button(
       "Check _backend var",
      
 ```

PR #991, opened 2023-05-10
Update README.md

mrteey:

Description:

It's super annoying to copy a command and have to remove $ sign before you run it, this update removes the $ sign from the README.md file to allow users to easily copy and run commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Pynecone requires the following to get started:
 -   [Node.js 16.6.0+](https://nodejs.org/en/) (Don't worry, you won’t have to write any JavaScript!)
 
 ```
-$ pip install pynecone
+pip install pynecone
 ```
 
 ## 🥳 2. Create your first app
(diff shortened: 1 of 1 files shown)


picklelo (approved):
I think we had that originally so that it's clear these are bash commands, but I think it will still be clear

mrteey:
> I think we had that originally so that it's clear these are bash commands, but I think it will still be clear

It will be, most people coming to GitHub already know about these things.

README.md
```diff
@@ -22,7 +22,7 @@ Pynecone requires the following to get started:
 -   [Node.js 16.6.0+](https://nodejs.org/en/) (Don't worry, you won’t have to write any JavaScript!)
 
 ```
```
masenf, line 24: Might add `sh` after the backticks as a hint to readers and the highlighter that this is shell snippet.

masenf (approved):
Much easier to copy paste without the dollar signs

PR #1000, opened 2023-05-11
@pc.cached_var: explicit opt-in for ComputedVar tracking

masenf:
Re-enable caching, but disable it by default. Users must opt-in to variable caching/dependency tracking, by using the new `@pc.cached_var` decorator.

* Add a `cache` boolean property to `ComputedVar`. If it is set to `False`, then all caching considerations are ignored. Additionally, any ComputedVar in a state with `cache=False` are considered "always dirty" and will be recalculated in `dict()` and `get_delta()`, even if their dependencies haven't changed.
* Any ComputedVar with `cache=True` that depends on an uncached variable will essentially become uncached and will be recalculated whenever the dependent var is recalculated (so, every time).
* Add a new `pc.cached_var` decorator which works just like `pc.var`, except sets `cache=True`

🎉 Hooray, PR number 1000!

### All Submissions:

### Type of change

**Really need to update pcweb to describe this new state caching functionality, expect a PR this weekend.**

### New Feature Submission:

### Changes To Core Features:

### DIFF
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -688,22 +688,29 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
         # Return the state update.
         return StateUpdate(delta=delta, events=events)
 
+    def _always_dirty_computed_vars(self) -> Set[str]:
+        """The set of ComputedVars that always need to be recalculated.
+
+        Returns:
+            Set of all ComputedVar in this state where cache=False
+        """
+        return set(
+            cvar_name
+            for cvar_name, cvar in self.computed_vars.items()
+            if not cvar.cache
+        )
+
     def _mark_dirty_computed_vars(self) -> None:
         """Mark ComputedVars that need to be recalculated based on dirty_vars."""
-        # Mark all ComputedVars as dirty.
-        for cvar in self.computed_vars.values():
-            cvar.mark_dirty(instance=self)
-
-        # TODO: Uncomment the actual implementation below.
-        # dirty_vars = self.dirty_vars
-        # while dirty_vars:
-        #     calc_vars, dirty_vars = dirty_vars, set()
-        #     for cvar in self._dirty_computed_vars(from_vars=calc_vars):
-        #         self.dirty_vars.add(cvar)
-        #         dirty_vars.add(cvar)
-        #         actual_var = self.computed_vars.get(cvar)
-        #         if actual_var:
-        #             actual_var.mark_dirty(instance=self)
+        dirty_vars = self.dirty_vars
+        while dirty_vars:
+            calc_vars, dirty_vars = dirty_vars, set()
+            for cvar in self._dirty_computed_vars(from_vars=calc_vars):
+                self.dirty_vars.add(cvar)
+                dirty_vars.add(cvar)
+                actual_var = self.computed_vars.get(cvar)
+                if actual_var:
+                    actual_var.mark_dirty(instance=self)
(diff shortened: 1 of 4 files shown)


masenf:
And yes, it works with the clock example before the modifications I made in https://github.com/pynecone-io/pynecone-examples/pull/64

pynecone/state.py
```diff
@@ -730,11 +735,18 @@ def get_delta(self) -> Delta:
...
+        # Apply dirty variables down into substates
+        self.dirty_vars.update(self._always_dirty_computed_vars())
+        self.mark_dirty()
+
+        # Return the dirty vars for this instance, any cached/dependent computed vars,
+        # and always dirty computed vars (cache=False)
+        delta_vars = (
+            self.dirty_vars.intersection(self.base_vars)
+            .union(self._dirty_computed_vars())
+            .union(self._always_dirty_computed_vars())
```
picklelo, line 747: Since you updated the `self.dirty_vars` already on line 739, do we need to add the `always_dirty_computed_vars()` again here?

pynecone/vars.py
```diff
@@ -801,6 +801,9 @@ def setter(state: State, value: Any):
 class ComputedVar(Var, property):
     """A field with computed getters."""
 
+    # Whether to track dependencies and cache computed values
+    cache: bool = False
```
picklelo, line 805: nice!

pynecone/state.py
```diff
@@ -730,11 +735,18 @@ def get_delta(self) -> Delta:
         """
         delta = {}
 
-        # Return the dirty vars and dependent computed vars
-        self._mark_dirty_computed_vars()
-        delta_vars = self.dirty_vars.intersection(self.base_vars).union(
-            self._dirty_computed_vars()
+        # Apply dirty variables down into substates
+        self.dirty_vars.update(self._always_dirty_computed_vars())
```
picklelo, line 739: Maybe we can create a function like

```python
def get_dirty_vars(self):
    return {**self.dirty_vars, **self._always_dirty_computed_vars()}
```

and that way we never have to add the always dirty vars to the vars that get cleaned up

picklelo (approved):
Looks good, this will be an awesome feature! Had one suggestion to potentially simplify in a follow up.

Also - do you think we should update the API to be `pc.var(cache=True)` rather than introducing a new variable? It may require some fancy Python checks since we should still support `pc.var` rather than converting those to `pc.var()`.

pynecone/state.py
masenf: Yes because delta vars starts by intersect dirty_vars with base_vars, which does NOT include computed vars. So then we union back in the dirty computed vars that belong to this instance (not inherited)

masenf:
> Also - do you think we should update the API to be `pc.var(cache=True)` rather than introducing a new variable?

I tried this at first and it even worked pretty okay, but from a typing perspective, it was kind of a nightmare, having the function return either a ComputedVar or a callable that returns ComputedVar. I tried to get typing.overload to work, but it was getting kind of hairy fast.

I also considered making CachedComputedVar as a subclass of ComputedVar and keeping all of the cache/dep logic over there, but this also made things more complicated when an always computed var depended on an always computed var in a parent state, there wasn't a good way of passing this information down into the substates without the deps information.

So there are still further improvements that can be made, but i think we're in a good spot for the moment.

Thanks again for taking the patch.

PR #1001, opened 2023-05-11
Unindent jinja template code

picklelo:
The indenting in the Jinja templates was causing issues when formatting multiline strings such as with `pc.code_block`s as the inner content was getting indented also.

The output after this is much uglier, but the code is formatted correctly.

We can either fix the indenting in a follow up, or use a tool like `prettier` to just format the JS afterwards (though this has a big performance cost, so we shouldn't do it always).

Also cleaned up the layout of some files in the code.


Alek99 (approved):
lgtm

PR #1002, opened 2023-05-11
Bun version validation:

### LINKED ISSUE #1014: Bun lock file incompatible with older bun versions
Current generated bun lock file is incompatible with previous versions. This errors out silently and as a result, dependencies are not installed.  You'd typically see an error `next: command not found` on `pc run`.

ElijahAhianyo:
There are incompatibilities of lock files between bun version. As such, this PR raises the minimum bun version to 0.5.9 and also shows a prompt on app initialization or run asking users permission to remove unsupported bun versions to install supported version.

Fixes #1014


picklelo (approved):
thanks for the fix!

PR #1005, opened 2023-05-11
Update upload event to use event queue

### LINKED ISSUE #944: Can't use an event chain with anything besides handle_upload()
**Describe the bug**
When I try to add another item to the event chain list for an upload button I get an error.

**To Reproduce**
```
import pynecone as pc
from typing import List

class State(pc.State):

    img: list[str] = []
    uploading: bool = False

    async def handle_upload(self, files: List[pc.UploadFile]):

        for file in files:
            outfile = f".web/public/{file.filename}"

            # Save the file.
            with open(outfile, "wb") as file_object:

            # Update the img var.

def index():
    """The main view."""
    return pc.vstack(

        pc.upload(
            pc.button("Select File"),
            pc.text("Drag and drop files here or click to select files"),
            border="1px dotted black",
            padding="20em",
            multiple=False
        ),
        pc.button(
            "Upload",
            on_click=[
                lambda: State.set_uploading(True),
                lambda: State.handle_upload(pc.upload_files()),
                lambda: State.set_uploading(False),
            ],
        ),
        pc.foreach(State.img, lambda img: pc.image(src=img)),
    )

app = pc.App(state=State)
app.add_page(index, title="Upload")
app.compile()

TypeError: State.handle_upload() missing 1 required positional argument: 'files'
```

**Expected behavior**
I should be able to add functions to the event chain without getting an error.

 - Python Version: 3.11
 - Pynecone Version: 0.1.28
 - OS: Ubuntu 23.04

ElijahAhianyo:
This PR changes the upload event logic to use the event queue. This allows chaining of events. However, this exposes another bug where the processing flag is not being set correctly due to some race conditions. There should be a follow up PR to fix this. 

fixes #944

### DIFF
--- a/pynecone/.templates/web/utils/state.js
+++ b/pynecone/.templates/web/utils/state.js
@@ -2,7 +2,9 @@
 import axios from "axios";
 import io from "socket.io-client";
 import JSON5 from "json5";
+import config from "../pynecone.json"
 
+const UPLOAD = config.uploadUrl;
 // Global variable to hold the token.
 let token;
 
@@ -92,7 +94,7 @@ export const applyEvent = async (event, router, socket) => {
 
   if (event.name == "_set_value") {
     event.payload.ref.current.value = event.payload.value;
-    return false; 
+    return false;
   }
 
   // Send the event to the server.
@@ -105,6 +107,24 @@ export const applyEvent = async (event, router, socket) => {
 
   return false;
 };
+
+/**
+ * Process an event off the event queue.
+ * @param queue_event The current event
+ * @param state The state with the event queue.
+ * @param setResult The function to set the result.
+ */
+export const applyRestEvent = async (
+  queue_event,
+  state,
+  setResult,
+) => {
+  if (queue_event.handler == "uploadFiles") {
+    await uploadFiles(state, setResult, queue_event.name, UPLOAD)
+  }
+
+}
+
 
 /**
  * Process an event off the event queue.
@@ -132,18 +152,28 @@ export const updateState = async (
   setResult({ ...result, processing: true });
 
   // Pop the next event off the queue and apply it.
-  const event = state.events.shift();
-
+  const queue_event = state.events.shift();
   // Set new events to avoid reprocessing the same event.
   setState({ ...state, events: state.events });
 
-  // Apply the event.
-  const eventSent = await applyEvent(event, router, socket);
-  if (!eventSent) {
-    // If no event was sent, set processing to false and return.
-    setResult({ ...state, processing: false });
+  // Process events with handlers via REST and all others via websockets.
+  if (queue_event.handler) {
+
+    await applyRestEvent(queue_event, state, setResult)
+
+
   }
+  else {
+    const eventSent = await applyEvent(queue_event, router, socket);
+    if (!eventSent) {
+      // If no event was sent, set processing to false and return.
+      setResult({ ...state, processing: false });
+    }
+  }
+
+
 };
+
 
 /**
  * Connect to a websocket and set the handlers.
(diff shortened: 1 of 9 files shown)


pynecone/.templates/web/utils/state.js
```diff
@@ -132,19 +152,29 @@ export const updateState = async (
...
+  const queue_event = state.events.shift();
   // Set new events to avoid reprocessing the same event.
   setState({ ...state, events: state.events });
 
   // Apply the event.
-  const eventSent = await applyEvent(event, router, socket);
-  if (!eventSent) {
-    // If no event was sent, set processing to false and return.
-    setResult({ ...state, processing: false });
+  if (queue_event.use_websocket == false) {
```
picklelo, line 160: nit: we can do `if (!queue_event.use_websocket)`

pynecone/.templates/web/utils/state.js
```diff
@@ -132,19 +152,29 @@ export const updateState = async (
...
+
+    await applyRestEvent(queue_event, state, setResult)
+
+
   }
+  else {
+    const eventSent = await applyEvent(queue_event, router, socket);
+    if (!eventSent) {
+      // If no event was sent, set processing to false and return.
+      setResult({ ...state, processing: false });
```
picklelo, line 170: I think we should apply this `processing: false` flag for both types of events to keep all the logic for that flag in this function

pynecone/components/tags/tag.py
```diff
@@ -77,10 +77,7 @@ def format_prop(
         elif isinstance(prop, EventChain):
             local_args = ",".join(([str(a) for a in prop.events[0].local_args]))
 
-            if len(prop.events) == 1 and prop.events[0].upload:
-                # Special case for upload events.
-                event = format.format_upload_event(prop.events[0])
-            elif prop.full_control:
+            if prop.full_control:
```
picklelo, line 80: NIce we are able to simplify here

pynecone/compiler/utils.py
```diff
@@ -104,7 +104,9 @@ def compile_state(state: Type[State]) -> Dict:
     initial_state = state().dict()
     initial_state.update(
         {
-            "events": [{"name": get_hydrate_event(state)}],
+            "events": [
+                {"name": get_hydrate_event(state), "use_websocket": True, "handler": ""}
```
picklelo, line 108: If we use a default value, then we don't have to set the extra args here

pynecone/pc.py
```diff
@@ -83,6 +83,9 @@ def run(
     frontend_port = get_config().port if port is None else port
     backend_port = get_config().backend_port if backend_port is None else backend_port
 
+    # set the upload url in pynecone.json file
+    build.set_pynecone_upload_endpoint()
```
picklelo, line 87: I like this idea - we currently compile many constants right into the pages file and pass them into functions in `state.js` (event the event url). I think we add more constants here. Maybe we create a new file though instead of using the `pynecone.json`

picklelo (commented):
Left some comments - have some ideas on how we can simplify things let's discuss tomorrow

pynecone/pc.py
ElijahAhianyo: created a follow up issue to handle this here #1012

pynecone/.templates/web/utils/state.js
ElijahAhianyo: Yeah got it, I could probably do all the refactor in the follow up ticket #1012

pynecone/.templates/web/utils/state.js
```diff
@@ -246,10 +271,12 @@ export const uploadFiles = async (
  * Create an event object.
  * @param name The name of the event.
  * @param payload The payload of the event.
+ * @param use_websocket Whether the event uses websocket.
```
picklelo, line 274: Remove extra arg

pynecone/utils/format.py
```diff
@@ -294,21 +294,8 @@ def format_event(event_spec: EventSpec) -> str:
...
-        event_spec: The event to format.
-
-    Returns:
-        The compiled event.
-    """
-    state, name = get_event_handler_parts(event_spec.handler)
-    parent_state = state.split(".")[0]
-    return f'uploadFiles({parent_state}, {constants.RESULT}, set{constants.RESULT.capitalize()}, {parent_state}.files, "{state}.{name}",UPLOAD)'
+    quote = '"'
+    return f"E(\"{format_event_handler(event_spec.handler)}\", {wrap(args, '{')}, {wrap(event_spec.client_handler_name, quote) if event_spec.client_handler_name else ''})"
```
picklelo, line 298: nit: we can use `",".join(...)` for the args to avoid the trailing comma

picklelo (approved):
Awesome! Some nits that we can do in a follow up

PR #1019, opened 2023-05-14
move add ref to component level render

sheldonchiu:
### All Submissions:

### Type of change

The prop `ref` should not be only limited to component of type input.
Moving the logic to component level render, so other components can use this prop


picklelo (approved):
good call, thanks for adding

PR #1031, opened 2023-05-15
Fix event default call

picklelo:
Fix issue where `e.preventDefault()` was being called incorrectly.

Also allow event handlers to return `pc.set_value` calls by adding a dictionary from id to component ref.


Alek99 (approved):
👍

Issue #1012, opened 2023-05-12
Move environment variables to a .json file


ElijahAhianyo:
This is a follow up on #1005 to store all environment variables in a json file

PR #1051, opened 2023-05-18
Compiler constants refactor

ElijahAhianyo:
Moved compiled constants into env.json file in the .web folder

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

fixes [#1012](https://github.com/pynecone-io/pynecone/issues/1012)

### DIFF
--- a/pynecone/components/layout/foreach.py
+++ b/pynecone/components/layout/foreach.py
@@ -1,7 +1,7 @@
 """Create a list of components from an iterable."""
 from __future__ import annotations
 
-from typing import Any, Callable, Dict, List, Set, Tuple, Union
+from typing import Any, Callable, Iterable
 
 from pynecone.components.component import Component
 from pynecone.components.layout.fragment import Fragment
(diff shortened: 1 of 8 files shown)


pynecone/.templates/jinja/web/pages/base_page.js.jinja2
```diff
@@ -4,6 +4,7 @@
 {% for module in imports%}
   {{- utils.get_import(module) }}
 {% endfor %}
+import env from  "{{env_file}}"
```
picklelo, line 7: I think we can get rid of this - the constants should only be needed in the `state.js` file now

pynecone/components/layout/foreach.py
```diff
@@ -13,15 +13,13 @@ class Foreach(Component):
     """A component that takes in an iterable and a render function and renders a list of components."""
 
     # The iterable to create components from.
-    iterable: Var[Union[List, Dict, Tuple, Set]]
+    iterable: Var[Iterable]
```
picklelo, line 16: nice!

picklelo (commented):
Thanks for this change!

I think we don't need the `.env` inside the component files anymore. We can import it in `state.js` instead so that file will have all the constants. Currently we are passing in the constants to the state.js files which is why we needed it here, but we can remove that now.

ElijahAhianyo:
> Thanks for this change!
> 
> I think we don't need the `.env` inside the component files anymore. We can import it in `state.js` instead so that file will have all the constants. Currently we are passing in the constants to the state.js files which is why we needed it here, but we can remove that now.

Yeah that makes sense. Will update this accordingly

picklelo (approved):
awesome, thanks for this update!

Issue #1055, opened 2023-05-19
Unhandled Runtime Error preventDefault is not a function.


TaiJuWu:
**Describe the bug**
Unhandled Runtime Error
TypeError: (0,_utils_state__WEBPACK_IMPORTED_MODULE_3__.preventDefault) is not a function. (In '(0,_utils_state__WEBPACK_IMPORTED_MODULE_3__.preventDefault)(_e)', '(0,_utils_state__WEBPACK_IMPORTED_MODULE_3__.preventDefault)' is undefined)
Call Stack
dispatchEvent

**To Reproduce**
 - Code/Link to Repo:
 ```python
import pynecone as pc

class ExampleState(pc.State):

    # The colors to cycle through.
    colors = ["black", "red", "green", "blue", "purple"]

    # The index of the current color.
    index = 0

    def next_color(self):
        """Cycle to the next color."""
        self.index = (self.index + 1) % len(self.colors)

    @pc.var
    def color(self):
        return self.colors[self.index]


def index():
    return pc.heading(
        "Welcome to Pynecone!",
        on_click=ExampleState.next_color,
        color=ExampleState.color,
        _hover={"cursor": "pointer"},
    )
app = pc.App(state=ExampleState)
app.add_page(index)
app.compile()
```

**Expected behavior**

**Screenshots**
<img width="1072" alt="image" src="https://github.com/pynecone-io/pynecone/assets/33004323/3c7076e3-79cc-4628-b1a7-8a8850c6b042">

** Specifics (please complete the following information):**
 - Python Version: 3.11
 - Pynecone Version: 0.1.31
 - OS: Ubuntu 20.04
 - nodejs version: 18.16.0
 - Browser (Optional):
 
**Additional context**

iron3oxide:
Getting the same error atm

iron3oxide:
Code:
```python
"""Welcome to Pynecone! This file outlines the steps to create a basic app."""
from pcconfig import config

import pynecone as pc

docs_url = "https://pynecone.io/docs/getting-started/introduction"
filename = f"{config.app_name}/{config.app_name}.py"


class State(pc.State):
    """The app state."""

    greeting = "Welcome to Pynecone!"
    count = 0

    def test(self):
        self.count += 1
        self.greeting = f"test {self.count}"


def index() -> pc.Component:
    return pc.center(
        pc.vstack(
            pc.heading(State.greeting, font_size="2em"),
            pc.box(
                "Get started by editing ",
[...]

Issue #1034, opened 2023-05-16
NumberInput is broken: TypeError: _e.preventDefault is not a function


masenf:
**Describe the bug**

`NumberInput` doesn't work with `on_change` anymore.

**To Reproduce**

```python
import pynecone as pc

class State(pc.State):
    v: int


def index() -> pc.Component:
    return pc.center(
        pc.number_input(
            value=State.v,
            on_change=State.set_v,
        ),
        padding_top="10%",
    )


app = pc.App(state=State)
app.add_page(index)
app.compile()
```

**Expected behavior**

`v` is updated as the user types a valid number

**Screenshots**

<img width="730" alt="image" src="https://github.com/pynecone-io/pynecone/assets/1524005/29edf6b2-e5d7-494e-81cd-049e88bf7ce6">

```
Unhandled Runtime Error
TypeError: _e.preventDefault is not a function
```

```
pages/index.js (23:11) @ preventDefault

  21 | const Event = (events, _e) => {
  22 |     if (_e) {
> 23 |       _e.preventDefault();
     |         ^
  24 |     }
  25 |     setState({
  26 |       ...state,
```

** Specifics (please complete the following information):**
 - Python Version: 3.11
 - Pynecone Version: 98f1b30735701af3f97a0b7b894eb8c03ab290fc 
 - OS: macOS 3.3.1
 - Browser (Optional): Firefox, Safari
 
**Additional context**

@picklelo Seems introduced by a recent change #1031.

With 0.1.29, it is working as expected.

picklelo:
@masenf thanks for catching this! Just put up a fix

PR #1063, opened 2023-05-21
utils/state.js: check whether event.preventDefault is undefined

masenf:
often events will inherit from `Event`, which defines `preventDefault`, in these cases `preventDefault` will be inherited and `hasOwnProperty('preventDefault')` will return false

### All Submissions:

### Type of change

### Changes To Core Features:

# Example

(Note `pc init` is required to pick up `utils/state.js` changes!)

```python
import pynecone as pc


class State(pc.State):
    foo: int = 42
    label: str = ""

    def on_load(self):
        print("Page loaded")
        self.label = ""

    def on_submit(self):
        print(f"Form submitted: {self.foo}")
        self.label = f"The answer is {self.foo}"
        self.foo = int(self.foo) + 1



def index() -> pc.Component:
    return pc.center(
        pc.vstack(
            pc.cond(
                State.label,
                pc.heading(State.label),
                pc.box(),
            ),
            pc.form(
                pc.number_input(value=State.foo, on_change=State.set_foo),
                pc.button("Submit", type_="submit"),
                on_submit=State.on_submit,
            ),
        ),
        padding_top="10%",
    )


app = pc.App(state=State)
app.compile()
```

Before the fix, this code doesn't really work, because submitting the form causes a page load, which rehydrates the state and breaks `pc.redirect` and other constructs.

After the fix, each time the form is submitted, the label is updated and no page reload is observed.

### DIFF
--- a/pynecone/.templates/web/utils/state.js
+++ b/pynecone/.templates/web/utils/state.js
@@ -285,7 +285,7 @@ export const isTrue = (val) => {
  * @param event
  */
 export const preventDefault = (event) => {
-  if (event && event.hasOwnProperty("preventDefault")) {
+  if (event && event.preventDefault) {
     event.preventDefault();
   }
 };


masenf:
Related to #1034 and #1036

ElijahAhianyo:
possible fix for #1055 as well

picklelo (approved):
Awesome thank you!

PR #1078, opened 2023-05-24
Backend host cli arg

### LINKED ISSUE #1017: Backend host manually set in dev mode
To make the backend work when accessing from a remote machine in dev mode, it was necessary to manually set the backend IP.

In pcconfig.py, first line after import pynecone as pc:

pc.constants.BACKEND_HOST = 'xxx.xxx.xxx.xxx'   # host IP

ElijahAhianyo:
This PR adds a `backend-host` cli argument on `pc run` for specifying the host to run the backend on. If you dont specify this, the default `0.0.0.0` will be used.

fixes #1017

### DIFF
--- a/pynecone/config.py
+++ b/pynecone/config.py
@@ -134,6 +134,9 @@ class Config(Base):
     # The frontend port.
     backend_port: str = constants.BACKEND_PORT
 
+    # The backend host.
+    backend_host: str = constants.BACKEND_HOST
+
     # The backend API url.
     api_url: str = constants.API_URL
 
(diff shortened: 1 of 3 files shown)


picklelo (approved):
lgtm

milochen0418:
Hi @ElijahAhianyo 
Do I misunderstand how to use  `--backend-port`. 

I try to `pc run  --backend-port=8088`  in pc 0.1.32. 
But it does not work because the front-end's code is still connecting to 8000 by default.
- https://github.com/pynecone-io/pynecone-examples/issues/93

Do I use it in the wrong way?

PR #1082, opened 2023-05-25
Handle multi-select

picklelo:
```python
import pynecone as pc


class State(pc.State):
    """The app state."""

    selected: list[str] = []


def index():
    """The main view."""
    return pc.center(
        pc.select(["option1", "option2", "option3"], is_multi=True, on_change=State.set_selected),
        pc.vstack(
            pc.heading("Selected" + State.selected.to_string()),
        ),
        padding_y="5em",
        font_size="2em",
        text_align="center",
    )


app = pc.App(state=State)
app.add_page(index)
app.compile()
```

### DIFF
--- a/pynecone/components/component.py
+++ b/pynecone/components/component.py
@@ -104,6 +104,11 @@ class Component(Base, ABC):
         initial_kwargs = {
             "id": kwargs.get("id"),
             "children": kwargs.get("children", []),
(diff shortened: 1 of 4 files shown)


Lendemor (approved):
Tested with pcweb, looks good to me.

PR #1104, opened 2023-05-30
add support for .env configuration

### LINKED ISSUE #1064: Add support for loading configuration from the environment
`pc` should look for its configuration in env vars and use those before using the `pcconfig.py` file or the defaults from `pynecone/constants.py`. Adding in support for using env vars for feeding configuration into pynecone will make using containers easier. 

I would prefix the configuration vars in  `pynecone/constants.py` that make sense for running in production with `PYNECONE_`  and if that is set in the environment, use that over what is set in the `pcconfig.py` file. 

For example I want to be able to set my `API_URL` as an environment variable when running Pynecone inside a container. If the above is implemented, we can then set `PYNECONE_API_URL=api.mycoolsite.biz` at container runtime and we can then use Traefik, Caddy, or Nginx to proxy requests to the container.

ElijahAhianyo:
In this PR you can add a .env file to your config class. 

```python


class ExpConfig(pc.Config):
    env_path = "envs/.env"

config = ExpConfig(
    app_name="exp",
    db_url="sqlite:///pynecone.db",
    env=pc.Env.DEV,
)
```
or as key work argument:

```python

class ExpConfig(pc.Config):
    pass

config = ExpConfig(
    app_name="exp",
    db_url="sqlite:///pynecone.db",
    env=pc.Env.DEV,
    env_path = "envs/.env"
)
```
By default, pynecone looks for a .env file($PROJECT_ROOT/.env) in your root directory. So if you dont specify the path but provide a .env file, it will be loaded into the os env. 
Also by default, the .env file overides os env variables. In order of precedence .env >> commandline args >> os env >> config file. However, to change this behavior/order, set `override_os_envs` to `False`.  Setting `override_os_envs=False` changes the precedence order to command line args >> os env >> .env file >> config file.

This PR also contains a breaking change: `--port` cmd line arg is renamed to `--frontend-port`

### All Submissions:

### Type of change

### Changes To Core Features:

fixes #1064


pynecone/config.py
```diff
@@ -198,6 +207,33 @@ def __init__(self, *args, **kwargs):
 
         super().__init__(*args, **kwargs)
 
+        # set overriden class attribute values as os env variables to avoid losing them
+        for key, value in dict(self).items():
+            key = key.upper() if key != "port" else "FRONTEND_PORT"
```
picklelo, line 212: Maybe we can rename the `--port` flag to `--frontend-port` to minimize confusion / having special cases. Would be a breaking change though

pynecone/constants.py
```diff
@@ -4,9 +4,32 @@
 import re
 from enum import Enum
 from types import SimpleNamespace
+from typing import Type
 
 import pkg_resources
 
+
+def get_value(key, default=None, type_: Type = str) -> Type:
```
picklelo, line 12: Can we add type annotations for `key` and `default`

pynecone/constants.py
```diff
@@ -4,9 +4,32 @@
...
+    """Get the value for the constant.
+    Obtain os env value and cast non-string types into
+    their original types.
+
+    Args:
+        key: constant name.
+        default: default value if key doesn't exist.
+        type_: the type of the constant
```
picklelo, line 20: nit: add newline between args/return

pynecone/constants.py
```diff
@@ -306,5 +333,6 @@ class RouteRegex(SimpleNamespace):
 TOGGLE_COLOR_MODE = "toggleColorMode"
 
 # Server socket configuration variables
-CORS_ALLOWED_ORIGINS = "*"
+# CORS_ALLOWED_ORIGINS = "*"
```
picklelo, line 336: remove this?

pynecone/utils/build.py
```diff
@@ -58,6 +58,14 @@ def set_environment_variables():
     )
 
 
+def set_os_env(**kwargs):
```
picklelo, line 61: Add docstrings for `kwargs`

picklelo (commented):
Awesome work, just some minor comments and I think some updates to the unit tests to get the build passing

pynecone/config.py
```diff
@@ -129,9 +132,9 @@ class Config(Base):
     username: Optional[str] = None
 
     # The frontend port.
-    port: str = constants.FRONTEND_PORT
+    frontend_port: str = constants.FRONTEND_PORT
```
picklelo, line 135: Now I'm thinking maybe we can keep both `port` and `frontend_port` - and throw a deprecation warning on the old one. Just so we don't break anyone's app with the next release.

pynecone/utils/exec.py
```diff
@@ -96,7 +96,7 @@ def run_frontend(
 
     # Run the frontend in development mode.
     console.rule("[bold green]App Running")
-    os.environ["PORT"] = get_config().port if port is None else port
+    os.environ["PORT"] = get_config().frontend_port if port is None else port
```
picklelo, line 99: We should also make sure this matches the config

picklelo (approved):
Looks good! We can merge this in and do a follow up for the comments

milochen0418:
Thanks a lot for this nice feature!

PR #1111, opened 2023-05-31
UI connection warning:

### LINKED ISSUE #971: Show warning if can't connect to api_url
When deploying an app if you don't update the `api_url` in your `pcconfig.py` currently the app doesn't work and you have to go in the Network tab to see the error. We should have a way to throw a UI warning when the backend isn't connected.

ElijahAhianyo:
This Pr adds a UI warning when server is unreachable
Screenshot 2023-05-31 at 3 34 45 PM

You also have the option of modifying the component that gets displayed.

```python

app = pc.App(state=State, connect_error_component=pc.connection_banner(
    pc.box(
        pc.text(
            "cannot connect to server. This is edited",
            bg="black",
            color="white",
        ),
        textAlign="center",
    )
))
app.add_page(index)
app.compile()
```

To turn this feature/warning off, you can set the `check_connection` flag to False:

```python

app = pc.App(state=State, check_connection=False)
app.add_page(index)
app.compile()
```
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

closes #971

### DIFF
--- a/pynecone/.templates/jinja/web/pages/index.js.jinja2
+++ b/pynecone/.templates/jinja/web/pages/index.js.jinja2
@@ -11,6 +11,7 @@
 export default function Component() {
   const [{{state_name}}, {{state_name|react_setter}}] = useState({{initial_state|json_dumps}})
   const [{{const.result}}, {{const.result|react_setter}}] = useState({{const.initial_result|json_dumps}})
+  const [notConnected, setNotConnected] = useState(false)
   const {{const.router}} = useRouter()
   const {{const.socket}} = useRef(null)
   const { isReady } = {{const.router}}
@@ -35,7 +36,7 @@ export default function Component() {
       return;
     }
     if (!{{const.socket}}.current) {
-      connect({{const.socket}}, {{state_name}}, {{state_name|react_setter}}, {{const.result}}, {{const.result|react_setter}}, {{const.router}}, {{transports}})
+      connect({{const.socket}}, {{state_name}}, {{state_name|react_setter}}, {{const.result}}, {{const.result|react_setter}}, {{const.router}}, {{transports}}, setNotConnected)
     }
     const update = async () => {
       if ({{const.result}}.{{const.state}} != null){
@@ -70,7 +71,21 @@ export default function Component() {
   {% endfor %}
 
   return (
+  <Fragment>
+      <Fragment>
+        {notConnected ? (
+          <Box sx={% raw %}{{"text-align": "center"}}{% endraw %}>
+            <Text sx={% raw %}{{ "bg": "red", "color": "white"}}{% endraw %}>
+              {`cannot connect to server. check if server is reachable.`}
+              </Text>
+          </Box>
+        ) : (
+          <Fragment>
+          </Fragment>
+        )}
+      </Fragment>
     {{utils.render(render, indent_width=0)}}
+    </Fragment>
   )
 }
 {% endblock %}
--- a/pynecone/.templates/jinja/web/pages/index.js.jinja2
+++ b/pynecone/.templates/jinja/web/pages/index.js.jinja2
@@ -74,7 +74,7 @@ export default function Component() {
   <Fragment>
       <Fragment>
         {notConnected ? (
-          <Box sx={% raw %}{{"text-align": "center"}}{% endraw %}>
+          <Box sx={% raw %}{{"textAlign": "center"}}{% endraw %}>
             <Text sx={% raw %}{{ "bg": "red", "color": "white"}}{% endraw %}>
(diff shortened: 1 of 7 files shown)


masenf (commented):
would it be possible to implement this as a component that we could add and style to fit our app? the current implementation seems kind of hard coded and fixed

pynecone/.templates/jinja/web/pages/index.js.jinja2
```diff
@@ -70,7 +71,21 @@ export default function Component() {
   {% endfor %}
 
   return (
+  <Fragment>
+      <Fragment>
+        {notConnected ? (
+          <Box sx={% raw %}{{"textAlign": "center"}}{% endraw %}>
```
picklelo, line 77: I think we can define this as a Pynecone component - and allow the user to override it as an argument to `pc.App`

picklelo (commented):
Awesome! I think we just need to update it so the warning is written in Pynecone so it can be customized

pynecone/.templates/web/utils/state.js
```diff
@@ -193,9 +193,11 @@ export const connect = async (
   result,
   setResult,
   router,
-  transports
+  transports,
+  setNotConnected
 ) => {
   // Get backend URL object from the endpoint
+  console.log("event url", EVENTURL);
```
picklelo, line 200: leftover debug

pynecone/app.py
```diff
@@ -65,6 +66,12 @@ class App(Base):
     # List of event handlers to trigger when a page loads.
     load_events: Dict[str, List[EventHandler]] = {}
 
+    # The component to render if there is a connection error to the server.
+    connect_error_component: Optional[Component] = None
+
+    # Whether to render component if server cant be reached.
```
picklelo, line 72: Maybe we can combine this with the above? If it is set to `None` we don't check the connection since there's nothing to display

pynecone/compiler/compiler.py
```diff
@@ -62,16 +67,20 @@ def _compile_theme(theme: dict) -> str:
...
 
     Args:
         component: The component to compile.
         state: The app state.
+        connect_error_component: The component to render on sever connection error.
 
     Returns:
         The compiled component.
     """
+    # breakpoint()
```
picklelo, line 83: extra debug

picklelo (commented):
nice! just a couple leftover statements to cleanup

pynecone/.templates/web/utils/state.js
ElijahAhianyo: ah nice catch

pynecone/.templates/jinja/web/pages/index.js.jinja2
```diff
@@ -11,6 +11,7 @@
 export default function Component() {
   const [{{state_name}}, {{state_name|react_setter}}] = useState({{initial_state|json_dumps}})
   const [{{const.result}}, {{const.result|react_setter}}] = useState({{const.initial_result|json_dumps}})
+  const [notConnected, setNotConnected] = useState(false)
```
masenf, line 14: Instead of capturing a bool here, this state could store the last connection error.

Then

```
connection_error = BaseVar("connectionError", type_="str", is_local=True)
```

And any page could reference this var

```python
pc.cond(
    connection_error != "",
    error_component(connection_error),
)
```

pynecone/.templates/web/utils/state.js
```diff
@@ -207,6 +208,11 @@ export const connect = async (
   // Once the socket is open, hydrate the page.
   socket.current.on("connect", () => {
     updateState(state, setState, result, setResult, router, socket.current);
+    setNotConnected(false)
+  });
+
+  socket.current.on('connect_error', (error) => {
```
masenf, line 214: I think it would be nice to stash this `error` in the component state and then export a `BaseVar` that references the state value. Then the default connection error banner could render as a `pc.cond` on that var (and it could optionally render the actual error message).

With this approach, the page template wouldn't need special handling to render the error component, since the `connectionError` could be referenced in the compiled page.

masenf (changes requested):
Having _less_ special case rendering logic in the jinja templates will increase maintainability and customization from the python side. Where possible, the client side state (not the State dict, but any other `useState` hook call) should be exposed to the pynecone compiler via `BaseVar` declarations that can be used like any other State var when constructing the frontend page.

pynecone/.templates/web/utils/state.js
picklelo: Great idea, we will add this in a follow-up

picklelo (approved):
awesome

PR #1126, opened 2023-06-03
Add Date and DateTime Pickers 

### LINKED ISSUE #237: Add Datepicker Component (pc.datepicker)

aidanjrauscher:
# Add Date and DateTime Pickers 

This PR implements a `datepicker` and `datetimepicker` component by inheriting the Input component and using the appropriate `type` attributes. 

I opted to use base html for these components as I felt the available React and Chakra libraries introduced unnecessary complexity, and the default styling for the `datepicker` and `datetimepicker` elements looks decent out of the box. 

Note: given these components are implemented as inputs, their value will be a string. 

closes #237 

P.S. Would love to be considered for the Software Engineering Intern role (if still open) ;)

-----

### All Submissions:

### Type of change

- [] Bug fix (non-breaking change which fixes an issue)
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [] This change requires a documentation update

### New Feature Submission:

### Changes To Core Features:


masenf (commented):
Not sure why these would need to be separate components when the same effect can be achieved by passing `pc.input(..., type_="datetime-local")`

aidanjrauscher:
> Not sure why these would need to be separate components when the same effect can be achieved by passing `pc.input(..., type_="datetime-local")`

Yes, the implementation is simple, but it’s a feature that has been discussed multiple times. It’s also the same method used for the email and password components.

masenf:
Okay, makes sense. Down the road this will let the project swap out the default components for something more specialized without downstream users having to change their code.

pynecone/components/__init__.py
```diff
@@ -91,6 +91,8 @@
 checkbox = Checkbox.create
 checkbox_group = CheckboxGroup.create
 copy_to_clipboard = CopyToClipboard.create
+datepicker = DatePicker.create
```
picklelo, line 94: Can we call this `date_picker` and `date_time_picker`(to match the capitalization of the classes)

pynecone/components/forms/datepicker.py
```diff
@@ -0,0 +1,11 @@
+"""A date input component."""
```
picklelo, line 1: Can we also change the filenames to `date_picker.py` and `date_time_picker.py`

picklelo (commented):
Thanks for adding! Just some comments about naming to keep it consistent with the other components

pynecone/components/__init__.py
aidanjrauscher: changed

pynecone/components/forms/datepicker.py
aidanjrauscher: changed in commit [7fb7ebf](https://github.com/pynecone-io/pynecone/pull/1126/commits/7fb7ebfd1234348d18cbab80253c213e2be71aca)

picklelo (approved):
thanks for the update!

PR #1127, opened 2023-06-03
fix hover trigger

jiangmiemie:
that is my test 

https://github.com/pynecone-io/pynecone/assets/69161207/05d95153-d988-4a0a-a523-452b588366fa

### DIFF
--- a/pynecone/components/overlay/popover.py
+++ b/pynecone/components/overlay/popover.py
@@ -108,8 +108,8 @@ class Popover(ChakraComponent):
         Returns:
             The popover component.
         """
-        if trigger == 'click' or trigger =='hover':
-            props.update({'trigger': trigger})
+#         if trigger == 'click' or trigger =='hover':
+#             props.update({'trigger': trigger})
             
         if len(children) == 0:
             contents = []
--- a/pynecone/components/overlay/popover.py
+++ b/pynecone/components/overlay/popover.py
@@ -108,8 +108,8 @@ class Popover(ChakraComponent):
         Returns:
             The popover component.
         """
-#         if trigger == 'click' or trigger =='hover':
-#             props.update({'trigger': trigger})
+        if trigger == "click" or trigger == "hover":
+            props.update({"trigger": trigger})
             
         if len(children) == 0:
             contents = []
--- a/pynecone/components/overlay/popover.py
+++ b/pynecone/components/overlay/popover.py
@@ -108,8 +108,8 @@ class Popover(ChakraComponent):
         Returns:
             The popover component.
         """
-        if trigger == "click" or trigger == "hover":
-            props.update({"trigger": trigger})
+        if props.get("trigger_comp") == "click" or props.get("trigger_comp") == "hover":
+            props.update({"trigger": props.get("trigger_comp")})
 
         if len(children) == 0:
             contents = []
--- a/pynecone/components/overlay/popover.py
+++ b/pynecone/components/overlay/popover.py
@@ -108,8 +108,13 @@ class Popover(ChakraComponent):
         Returns:
             The popover component.
         """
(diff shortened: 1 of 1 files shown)


jiangmiemie:
image

jiangmiemie:
image
It seems that there is a problem with our workflow. In order to confirm that it is a problem with the workflow, I just uploaded a comment @Alek99

jiangmiemie:
Due to the incorrect format, I submitted it repeatedly many times, I am very sorry, this is a fix, hope to deal with it as soon as possible

pynecone/components/overlay/popover.py
```diff
@@ -108,6 +108,9 @@ def create(
         Returns:
             The popover component.
         """
+        if trigger == "click" or trigger == "hover":
```
picklelo, line 111: I think the expected type of `trigger` is `pc.Component` because we use it to create the component below. We need to add type annotations for this method to make this more clear. I'm thinking we can change the name of the argument `trigger` to `trigger_comp` or something like that here, so it doesn't get override in the `props`.

picklelo (commented):
I think this will work, but may break the functionality of the convenience initializer. I think we can just rename the argument name in the `create` method

pynecone/components/overlay/popover.py
jiangmiemie: agree

picklelo: @jiangmiemie Do you want to update this PR to rename the argument? We can merge it in after

jiangmiemie:
```
app.add_page(
    pc.center(
        pc.vstack(
            pc.heading("SCP: Site Roleplay Tools", font_size="2em"),
            pc.box("Get started by choosing an option below."),
            pc.button_group(
                pc.popover(
                    pc.popover_trigger(
                        pc.button("914 Tools", size="lg"),
                    ),
                    pc.popover_content(pc.popover_body("Test")),
                    trigger_comp="hover",
                ),
                pc.button("XP Calculator", size="lg"),
                pc.button("Example", size="lg", on_click=pc.toggle_color_mode),
                space="1em",
            ),
            spacing="1.5em",
            font_size="2em",
        )
    ),
    "test",
)
```

pynecone/components/overlay/popover.py
```diff
@@ -108,6 +108,9 @@ def create(
         Returns:
             The popover component.
         """
+        if props.get("trigger_comp") == "click" or props.get("trigger_comp") == "hover":
```
picklelo, line 111: Sorry I actually meant we should change line 90 (and maybe the following ones also) to be `trigger_comp`, `header_comp`, etc. So it's clear that those are components that are passed in, while the normal `trigger` prop should remain untouched. Then we don't need this special logic here to check for "click" or "hover". Does that make sense?

jiangmiemie:
```
app.add_page(
    pc.center(
        pc.vstack(
            pc.heading("SCP: Site Roleplay Tools", font_size="2em"),
            pc.box("Get started by choosing an option below."),
            pc.button_group(
                pc.popover(
                    pc.popover_trigger(
                        pc.button("914 Tools", size="lg"),
                    ),
                    pc.popover_content(pc.popover_body("Test")),
                    trigger="hover",
                ),
                pc.button("XP Calculator", size="lg"),
                pc.button("Example", size="lg", on_click=pc.toggle_color_mode),
                space="1em",
            ),
            spacing="1.5em",
            font_size="2em",
        )
    ),
    "test",
)
```

it is test

pynecone/components/overlay/popover.py
jiangmiemie: oh, i konw your mind ~ i will fix it now

jiangmiemie:
I'm not very good at darglint , I think he should fixed correctly ...

PR #1149, opened 2023-06-06
Download file

### LINKED ISSUE #909: pc.download_files()
We have pc.upload_files() for uploading files. I think we should also have pc.download_files() for downloading files. I don't know of a clean way to create a download link in pynecone. We should have one. #795

TaiJuWu:
### All Submissions:

- [X] Have you followed the guidelines stated in [CONTRIBUTING.md](https://github.com/pynecone-io/pynecone/blob/main/CONTRIBUTING.md) file?
- [X] Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/pynecone-io/pynecone/pulls ) for the desired changed?

### Type of change

- [X] This change requires a documentation update

### New Feature Submission:

### Changes To Core Features:

Support backend download feature
Example code:
```python
app = pc.App(state=State)
app.add_page(index)
app.add_download_file("/test", "ws/pynecone/download/test.JPG")
app.compile()
```
You can download file from localhost:8000/test and localhost:3000/test

    b. Describe your changes.

Use fastAPI to download file.

close #909

### DIFF
--- a/pynecone/app.py
+++ b/pynecone/app.py
@@ -87,6 +87,9 @@ class App(Base):
     # The component to render if there is a connection error to the server.
     connect_error_component: Optional[Component] = None
 
+    # Check backend router:
+    backend_router = set()
+
     def __init__(self, *args, **kwargs):
         """Initialize the app.
 
@@ -345,9 +348,16 @@ class App(Base):
 
         Returns:
             Respone a file from server.
-        """
 
-        @self.api.get(f"/download{download_url}")
+        Raises:
+            ValueError: download_url already is used.
+        """
+        if download_url in self.backend_router:
+            raise ValueError(f"{download_url} already is used.")
+
+        self.backend_router.add(download_url)
+
+        @self.api.get(f"{download_url}")
         async def response():
             return FileResponse(
                 path=file_path,
--- a/pynecone/app.py
+++ b/pynecone/app.py
@@ -30,6 +30,7 @@ from pynecone.compiler import compiler
 from pynecone.compiler import utils as compiler_utils
 from pynecone.components.component import Component, ComponentStyle
 from pynecone.components.layout.fragment import Fragment
+from pynecone.config import get_config
 from pynecone.event import Event, EventHandler, EventSpec, redirect
 from pynecone.middleware import HydrateMiddleware, Middleware
 from pynecone.model import Model
--- a/pynecone/app.py
+++ b/pynecone/app.py
@@ -17,6 +17,7 @@ from typing import (
 
 from fastapi import FastAPI, UploadFile
 from fastapi.middleware import cors
+from fastapi.responses import FileResponse
 from socketio import ASGIApp, AsyncNamespace, AsyncServer
 from starlette_admin.contrib.sqla.admin import Admin
 from starlette_admin.contrib.sqla.view import ModelView
@@ -334,6 +335,27 @@ class App(Base):
         if route == "":
             route = constants.INDEX_ROUTE
         return self.load_events.get(route, [])
+
+    def add_download_file(self, download_url: str, file_path: str):
+        """Download file from host by downloa_durl.
+
+        Args:
+            download_url: The route to get the file.
+            file_path: The file from host.
+
+        Returns:
+            Respone a file from server.
+        """
+
+        @self.api.get(f"/download{download_url}")
+        async def response():
+            return FileResponse(
+                path=file_path,
+                filename=os.path.split(file_path)[1],
+                media_type="text/mp4",
+            )
+
+        return
 
     def _check_routes_conflict(self, new_route: str):
         """Verify if there is any conflict between the new route and any existing route.
--- a/pynecone/app.py
+++ b/pynecone/app.py
@@ -340,7 +340,7 @@ class App(Base):
         return self.load_events.get(route, [])
 
     def add_download_file(
-        self, download_url: str, file_path: str, media_type: str = "text/mp4"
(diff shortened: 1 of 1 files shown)


TaiJuWu:
Is there method to check download file feature working correctly?

pynecone/app.py
```diff
@@ -316,6 +321,34 @@ def get_load_events(self, route: str) -> List[EventHandler]:
...
+            raise ValueError(f"{download_url} already is used.")
+
+        self.backend_router.add(download_url)
+
+        @self.api.get(f"{download_url}")
+        async def response():
+            return FileResponse(
+                path=file_path,
+                filename=os.path.split(file_path)[1],
+                media_type="text/mp4",
```
masenf, line 347: this would need to be parametrized and plumbed through

pynecone/app.py
```diff
@@ -316,6 +321,34 @@ def get_load_events(self, route: str) -> List[EventHandler]:
...
+        if download_url in self.backend_router:
+            raise ValueError(f"{download_url} already is used.")
+
+        self.backend_router.add(download_url)
+
+        @self.api.get(f"{download_url}")
+        async def response():
+            return FileResponse(
+                path=file_path,
+                filename=os.path.split(file_path)[1],
```
masenf, line 346: why `[1]`? Seems like you would want `[-1]` to grab the right-most element of the list. At any rate, I'd prefer to see `pathlib.Path(file_path).name` here instead.

pynecone/app.py
```diff
@@ -316,6 +321,34 @@ def get_load_events(self, route: str) -> List[EventHandler]:
...
+            download_url: The route to get the file.
+            file_path: The file from host.
+
+        Returns:
+            Respone a file from server.
+
+        Raises:
+            ValueError: download_url already is used.
+        """
+        if download_url in self.backend_router:
```
masenf, line 337: probably also need to check in `self.pages` for practical purposes.

masenf (changes requested):
it needs to be really clear that the download files are on the backend server (port 8080 unless changed).

there should be a component that makes it easy to root-relative link to a downloadable file, by dereferencing the `api_url` config value. as it is now, there's no reliable way for the frontend to link to a downloadable file.

it might be nicer to leverage the `static_files` feature of `FastAPI`.

TaiJuWu:
> it needs to be really clear that the download files are on the backend server (port 8080 unless changed).
> 
> there should be a component that makes it easy to root-relative link to a downloadable file, by dereferencing the `api_url` config value. as it is now, there's no reliable way for the frontend to link to a downloadable file.
> 
> it might be nicer to leverage the `static_files` feature of `FastAPI`.



> it needs to be really clear that the download files are on the backend server (port 8080 unless changed).
> 
> there should be a component that makes it easy to root-relative link to a downloadable file, by dereferencing the `api_url` config value. as it is now, there's no reliable way for the frontend to link to a downloadable file.
> 
> it might be nicer to leverage the `static_files` feature of `FastAPI`.

I have some questions about your suggestion.

1. Im not sure your 8080 is default backend server(8000) or not?

2. I use the method of api.get to download the file, so it will follow the setting of the config file to mount the donwload file url with api_url(default 8000).

3. I agree to there is not frontend to link download file but Im not sure it is necessary or not. What do you think?

4. Is fastAPI static file https://fastapi.tiangolo.com/tutorial/static-files/  ?

pynecone/app.py
TaiJuWu: pages seems to be used for frontend so I don't check this part.
If Im wrong, please tell me.

pynecone/app.py
TaiJuWu: I agree with you. Thanks for your suggestion.

masenf:
1. Yea I typo'd, default is 8000
2. I think this is fine
3. The issue arises when the frontend code wants to link to a downloadable file, there isn't a easy way for frontend code to know what URL prefix is needed to access backend.
4. Yep that's the API I was referring to. My understanding is that serving files via static_files would be more efficient for the server.

pynecone/app.py
masenf: i think a common pattern for deployments is to use a reverse proxy that forwards requests to either frontend or backend based on the path, so there's a possibility of conflict in these cases. maybe it's not a big deal though.

TaiJuWu:
> * Yea I typo'd, default is 8000
> * I think this is fine
> * The issue arises when the frontend code wants to link to a downloadable file, there isn't a easy way for frontend code to know what URL prefix is needed to access backend.
> * Yep that's the API I was referring to. My understanding is that serving files via static_files would be more efficient for the server.

I have tried to automatically create backend download file by frontend component but it's very difficult so I give up and only implement the backend download file.

Static files seem not working because it will only show image/video on browsers instead of download file.

pynecone/app.py
TaiJuWu: I used self.add _page to redirect frontend to backend and there will check self.page.

Justin900429:
Hi, I have a few concerns regarding the new download support.

Firstly, it seems that the default media type is set to `mp4`. However, in your example, you didn't specify the media type and instead returned an image. I'm wondering if this could have any impact on the content file in the response header.

```python
# Provide the media type correctly
app.add_download_file("/test", "ws/pynecone/download/test.JPG", media_type="image/jpeg")
```

Secondly, I suggest considering removing the default value for the `media_type` parameter. By doing so, users will be required to explicitly specify the media type, which can help prevent any unintended side effects.

These adjustments should enhance clarity and ensure that users are aware of the media type being used and can make a conscious decision based on their specific requirements.

PR #1155, opened 2023-06-07
Support Tailwind out of the box

### LINKED ISSUE #141: Tailwind support
Would it be possible to add Tailwind in the base project? It's a very popular framework for styling and it is compatible with chakra: https://github.com/chakra-ui/chakra-ui/issues/634

Here's a quick example on how it could be used:

```py
def index():
    return pc.center(
        pc.vstack(
            pc.heading("Hello world!", class_name="text-xl"),
        ),
        class_name="pt-8",
    )
```

kbrgl:
Closes #141. This change adds Tailwind support to Pynecone out of the box by installing Tailwind in the default template and adding a `tailwind` key in the config. An empty dict simply enables Tailwind; the dict can be filled in to configure *any* Tailwind options except for `content`, which we set ourselves to avoid a leak in the abstraction (since we have to set it to a Next.js-specific path). Strings passed for `plugins` are automatically wrapped with `require(...)`.

An example app has been added at https://github.com/pynecone-io/pynecone-examples/pull/112.

### All Submissions:

- [x] Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/pynecone-io/pynecone/pulls ) for the desired changes?

### Type of change

### New Feature Submission:


pynecone/compiler/compiler.py
```diff
@@ -125,6 +125,22 @@ def _compile_components(components: Set[CustomComponent]) -> str:
     )
 
 
+def _compile_tailwind(
+    config: dict,
```
picklelo, line 129: I think the type annotation needs to be `Dict` to support 3.7, we can fix in a follow up

pynecone/compiler/compiler.py
```diff
@@ -239,6 +255,26 @@ def compile_components(components: Set[CustomComponent]):
     return output_path, code
 
 
+@write_output
+def compile_tailwind(
+    config: dict,
```
picklelo, line 260: here also

picklelo (approved):
nice job getting this working

PR #1156, opened 2023-06-07
Fix upload error when no files are set

picklelo:


Alek99 (approved):
Tested works well 👍

milochen0418:
@picklelo  After testing it from this branch. It works. :D  Thanks~~~

PR #1157, opened 2023-06-07
Fix processing flag for generator events

picklelo:
Currently the clock app has a race condition when trying to switch the time zone while it is running. This PR should fix this.


Lendemor (approved):
Looks good to me.

PR #1177, opened 2023-06-09
Fix pc.html using state vars

picklelo:
Wrap the entire page in a `pc.fragment` so that the meta-information is added into the fragment.

Test code:

```python
"""Welcome to Pynecone! This file create a counter app."""
import pynecone as pc


class State(pc.State):
    """The app state."""

    html: str = "<div>hello</div>"

def index() -> pc.Component:
    return pc.html(State.html)


app = pc.App(state=State)
app.add_page(index, title="Counter")
app.compile()
```


Lendemor (approved):
Lgtm

PR #1178, opened 2023-06-09
Fix pc.link with no href

picklelo:
Some of the example apps use a `pc.link` without an `href` (instead they bind to an `on_click` handler). This PR handles those use cases.


Alek99 (approved):

milochen0418:
From the crm example, the link disappears after using pc 0.1.33. 
Detail can see the above issue.
- https://github.com/pynecone-io/pynecone-examples/issues/105

PR #1194, opened 2023-06-15
allow custom redis client for a more flexible setup

sheldonchiu:
### All Submissions:

### Type of change

### New Feature Submission:

Allow users to bring their own Redis client if they need a more complex connection setup

### DIFF
--- a/pynecone/app.py
+++ b/pynecone/app.py
@@ -102,7 +102,10 @@ class App(Base):
         self.middleware.append(HydrateMiddleware())
 
         # Set up the state manager.
-        self.state_manager.setup(state=self.state)
+
+        self.state_manager.setup(
+            state=self.state, redis_client=kwargs.get("redis_client", None)
+        )
 
         # Set up the API.
         self.api = FastAPI()
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -900,14 +900,17 @@ class StateManager(Base):
     # The redis client to use.
     redis: Optional[Redis] = None
 
-    def setup(self, state: Type[State]):
+    def setup(self, state: Type[State], redis_client: Optional[Redis] = None):
         """Set up the state manager.
 
         Args:
             state: The state class to use.
+            redis_client: optional redis client to use for state storage
         """
         self.state = state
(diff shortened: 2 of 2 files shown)


pynecone/app.py
```diff
@@ -102,7 +102,10 @@ def __init__(self, *args, **kwargs):
         self.middleware.append(HydrateMiddleware())
 
         # Set up the state manager.
-        self.state_manager.setup(state=self.state)
+
+        self.state_manager.setup(
+            state=self.state, redis_client=kwargs.get("redis_client", None)
```
picklelo, line 107: Should we add the `redis_client` to the `pcconfig` instead? I think it makes more sense as a config field there.

ozeranskii: it may be a good idea to pass cache_client instead of redis_client. To get rid of the direct dependency on Redis and use any available cache server. For example: RedisCache, MemoryCache, DummyCache.

picklelo: Yes agreed, we should abstract that part out

ozeranskii: I can finish this PR, if there are no objections. @picklelo @sheldonchiu

ozeranskii: > > I can finish this PR, if there are no objections. @picklelo @sheldonchiu
> 
> sure, but my understand is that `pcconfig` should be a static setting, and adding this could break the init method

I'll think of a better way and propose a solution

masenf: Please check #1676, there has been a major refactoring to how the StateManager works to accommodate alternative backing stores for state. Specifically, StateManager has become an abstract class with StateManagerMemory and StateManagerRedis implementing it. Also the redis implementation has moved to asyncio only to interop with the uvicorn ASGI server while providing non-blocking locking of state.

If you still want to work on this issue, highly advise to either wait till that PR merges, or base your work upon it.

masenf: @ozeranskii heads up that #1676 has merged to `main`.

ozeranskii: @masenf it seems this PR is no longer relevant and can be closed?

PR #1198, opened 2023-06-16
Fix blocking events

picklelo:
Fixes the issue of yielding with blocking events by using async tasks to emit socket updates.


Alek99 (approved):
👍

PR #1219, opened 2023-06-20
Fix pc.py frontend setup

picklelo:
Fixes the issue where exporting fails when starting the backend server at the same time by doing the export first.


ElijahAhianyo (approved):
lgtm

PR #1221, opened 2023-06-20
Get Cookies

ElijahAhianyo:
This PR adds a feature that allows you to get browser cookies.

```python
import pynecone as pc

class State(pc.State):
    val: str = "placeholder"

    def set_data(self):
        # set name cookie data
        pc.set_cookie("name", "reflex")
        # get name cookie data
        self.val = self.get_cookies()["name"]

def index():
    return pc.vstack(
        pc.text(State.val),
        pc.button("Cookie", on_click=State.set_data),
    )

app = pc.App(state=State)
app.add_page(index)
app.compile()
```

### DIFF
--- a/pynecone/state.py
+++ b/pynecone/state.py
@@ -490,6 +490,22 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
         """
         return self.router_data.get(constants.RouteVar.QUERY, {})
 
+    def get_cookies(self) -> Dict[str, str]:
+        """Obtain the cookies of the client stored in the browser.
+
+        Returns:
+                The dict of cookies.
+        """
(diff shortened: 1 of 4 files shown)


pynecone/state.py
```diff
@@ -490,6 +490,22 @@ def get_query_params(self) -> Dict[str, str]:
         """
         return self.router_data.get(constants.RouteVar.QUERY, {})
 
+    def get_cookies(self) -> Dict[str, str]:
+        """Obtain the cookies of the client stored in the browser.
+
+        Returns:
+                The dict of cookies.
+        """
+        headers = self.get_headers().get(constants.RouteVar.COOKIE)
```
picklelo, line 499: Should we call this variable `cookies`?

picklelo (approved):
sweet

PR #1222, opened 2023-06-21
Local storage clear and remove

ElijahAhianyo:
This PR adds features to :
- clear local storage
- remove item from local storage

To clear all items in local storage:
```python 
...
def index():
    return pc.vstack(
        pc.text(State.val),
        pc.button("clear", on_click=pc.clear_local_storage()),
    )
...


```

To remove specific item(using key) from local storage:

```python
...
def index():
    return pc.vstack(
        pc.text(State.val),
        pc.button("remove", on_click=pc.remove_local_storage("key")),
    )
...

```

### DIFF
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -131,6 +131,16 @@ export const applyEvent = async (event, router, socket) => {
     return false;
   }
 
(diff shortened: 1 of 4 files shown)


picklelo (dismissed):

picklelo (approved):
looks good to me

PR #1223, opened 2023-06-21
Remove cookies

ElijahAhianyo:
This PR adds feature to remove cookies:

```python
...
def index():
    return pc.vstack(
        pc.text(State.val),
        pc.button("Cookie", on_click=pc.set_cookie("name", "reflex")),
        pc.button("remove", on_click=pc.remove_cookie("name")),
    )
...

```

remove with options:

```python
  ...
def index():
    return pc.vstack(
        pc.text(State.val),
        pc.button("Cookie", on_click=pc.set_cookie("name", "reflex")),
        pc.button("remove", on_click=pc.remove_cookie(
                                                        "name",
                                                        options={
                                                            "path": "/",
                                                            "domain": "example.com",
                                                            "secure": True,
                                                            "sameSite": "strict",
                                                        }
                                                    )
                  ),
    )

```

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -5,6 +5,7 @@ import asyncio
 import copy
 import functools
 import inspect
+import json
 import traceback
 import urllib.parse
 from abc import ABC
@@ -505,9 +506,9 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
         for pair in cookie_pairs:
             key, value = pair[0].strip(), urllib.parse.unquote(pair[1].strip())
             try:
-                # convert non-string data types to the actual types.
-                value = eval(value)
-            except Exception:
+                # cast non-string values to the actual types.
+                value = json.loads(value)
+            except json.JSONDecodeError:
                 pass
             finally:
                 cookie_dict[key] = value
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -6,6 +6,7 @@ import copy
 import functools
 import inspect
 import traceback
+import urllib.parse
 from abc import ABC
 from collections import defaultdict
 from typing import (
@@ -351,7 +352,7 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
         """Initialize a variable.
 
         Args:
-            prop (BaseVar): The variable to initialize
+            prop: The variable to initialize
 
         Raises:
             TypeError: if the variable has an incorrect type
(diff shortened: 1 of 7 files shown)


pynecone/event.py
```diff
@@ -254,6 +254,24 @@ def set_cookie(key: str, value: str) -> EventSpec:
     )
 
 
+def remove_cookie(key: str, options: Dict = {}) -> EventSpec:  # noqa: B006
```
picklelo, line 257: Can we annotate the dict also e.g. `Dict[str, str]`

pynecone/event.py
```diff
@@ -254,6 +254,24 @@ def set_cookie(key: str, value: str) -> EventSpec:
...
+    """Remove a cookie on the frontend.
+
+    Args:
+        key (str): The key identifying the cookie to be removed.
```
picklelo, line 261: Don't need to put types in the docstrings since we use type signatures

pynecone/state.py
```diff
@@ -496,15 +497,21 @@ def get_cookies(self) -> Dict[str, str]:
...
+        cookie_dict = {}
+        cookies = self.get_headers().get(constants.RouteVar.COOKIE, "").split(";")
+
+        cookie_pairs = [cookie.split("=") for cookie in cookies if cookie]
+
+        # parse all dict and list values
+        for pair in cookie_pairs:
+            key, value = pair[0].strip(), urllib.parse.unquote(pair[1].strip())
+            try:
+                value = eval(value)
```
picklelo, line 509: Could we add a comment explaining why we need this line

picklelo (commented):
nice! just some small comments

reflex/state.py
```diff
@@ -496,15 +497,21 @@ def get_cookies(self) -> Dict[str, str]:
...
+        cookie_dict = {}
+        cookies = self.get_headers().get(constants.RouteVar.COOKIE, "").split(";")
+
+        cookie_pairs = [cookie.split("=") for cookie in cookies if cookie]
+
+        for pair in cookie_pairs:
+            key, value = pair[0].strip(), urllib.parse.unquote(pair[1].strip())
+            try:
+                # convert non-string data types to the actual types.
+                value = eval(value)
```
picklelo, line 509: should we use json.parse here? Or would there be non-json types here

picklelo (commented):
looks good, just one question we can discuss offline if we need to update

picklelo (approved):
looks good to me!

PR #1225, opened 2023-06-21
Update readme example

Alek99:
Updates readme with new yield functionality.


README.md
```diff
@@ -168,33 +162,32 @@ class State(pc.State):
...
+    yield
+    response = openai.Image.create(prompt=self.prompt, n=1, size="1024x1024")
+    self.image_url = response["data"][0]["url"]
+    self.processing, self.complete = False, True
 ```
 
 Within the state, we define functions called event handlers that change the state vars. Event handlers are the way that we can modify the state in Pynecone. They can be called in response to user actions, such as clicking a button or typing in a text box. These actions are called events.
 
-Our DALL·E. app has two event handlers, `process_image` to indicate that the image is being generated, and `get_image`, which calls the OpenAI API.
+Our DALL·E. app has an event handlers, `get_image` to which get this image from the openai api. Using yield in the middle of an event handler will cause the UI to update. Otherwise the UI will update at the end of the event handler.
```
picklelo, line 190: "an event handlers" -> "an event handler". Also can we put `yield` in backticks and format it as "OpenAI API"

picklelo (commented):
just a formatting comment

picklelo (approved):

PR #1227, opened 2023-06-22
fix upload file not set state before return

sheldonchiu:
### All Submissions:

### Type of change

Fix issue with upload_file does not run set_state before return.


picklelo (approved):
great catch! we will soon abstract this all out so we don't have to repeat the logic twice

PR #1234, opened 2023-06-23
Get Local Storage

ElijahAhianyo:
This PR adds a feature to get items in local storage.

Get item with key:
```python
class State(pc.State):

    def print_data(self, value: str):
       # value is the value of the provided key in local storage.
        print(f"value: {value}")


def index():
    return pc.vstack(
        pc.text(State.val),
        pc.button("click here", on_click=State.print_data(pc.get_local_storage("key")))
    )

```

Provide no args to get all items in local storage:

```python
class State(pc.State):

    def print_data(self, storage: dict):
       # storage is a dict of local storage items.
        print(f"storage: {storage}")


def index():
    return pc.vstack(
        pc.text(State.val),
        pc.button("click here", on_click=State.print_data(pc.get_local_storage()))
    )

```

NB: Key type should be a `str` or  `Var` with type `str`

### DIFF
--- a/pynecone/.templates/web/utils/state.js
+++ b/pynecone/.templates/web/utils/state.js
@@ -76,6 +76,23 @@ export const applyDelta = (state, delta) => {
   }
 };
 
+
+/**
+ * Get all local storage items in a key-value object.
+ * @returns object of items in local storage.
+ */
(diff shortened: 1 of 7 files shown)


Lendemor (approved):
Looks good to me.

PR #1248, opened 2023-06-26
Refactored project dependencies

### LINKED ISSUE #1240: Get Rid of Plotly/Redis Dependency
Plotly and Redis should not be a requirements of Pynecone.

palashcode:
### All Submissions:

### Type of change

### New Feature Submission:

Refactored project dependencies: Moved 'plotly' and 'redis' dependencies from regular dependencies to dev dependencies using Poetry. This change ensures that these dependencies are only required for development purposes, reducing the overall project dependencies.
closes #1240


pyproject.toml
```diff
@@ -60,6 +58,8 @@ pandas = [
 ]
 asynctest = "^0.13.0"
 pre-commit = {version = "^3.2.1", python = ">=3.8,<4.0"}
+plotly = "^5.15.0"
```
picklelo, line 61: I think we may need to edit the sections within the code that import these modules, and put some sort of `try/except` around them. Otherwise users will get an error when they try to run reflex.

picklelo (commented):
Thanks for the cleanup! Agreed that these should become optional dependencies.

We need to test that the package will still work without these installed. I believe for example in `utils/format.py` we import plotly at the top, so importing that entire module will fail now.

We can maybe move the imports for plotly and redid to the parts of the code where they are actually used, so users who don't need them won't hit the imports.

Also looks like we need to rebase since we updated a dependency lately.

masenf:
i think using extras is better here is better dev dependencies... because `redis` and `plotly` are needed to deploy production apps that use these libraries, and they wouldn't want all of the other dev dependencies.

https://python-poetry.org/docs/pyproject/#extras

palashcode:
Hi @masenf and @picklelo,
I created a fresh PR here(https://github.com/pynecone-io/pynecone/pull/1295) for this making the changes suggested by you. Can you please review it and suggest improvements.
### Made following changes
- Made plotly and redis optional and extra dependecies.
- Added try/except in the code where plotly and redis are used to throw error and prompt user to install these extra dependencies.
- Did a rebase on to the main branch.

masenf:
Closing in favor of #1816, which will address the plotly side of things and start moving us in this direction.

PR #1252, opened 2023-06-26
exclude .web folder from uvicorn file sync due to increased CPU load

### LINKED ISSUE #1251: Increased CPU usage in dev mode
**Describe the bug**
I've noticed that when I run any `pynecone` (`reflex`) project, it always consumes some CPU.
I understand that in dev mode, the server monitors files which have been changed, but usually it does not consume as much CPU time, it just few percents but anyway that was suspicious to me.
Take a look at first screenshot.
> by the way on macOS 36 does mean 3.6%.

So I did some investigation and here is what I found.

I've run `py-spy` profiler, and I've found that most of the time the server is busy doing this 
> Function: should_restart (uvicorn/supervisors/statreload.py:32) (784 samples, 95.38%)

Since `reflex` under the hood use `uvicorn` I started looking in this direction.
Googling the issue I found such source code [uvicorn StatReload](https://github.com/encode/uvicorn/blob/ffa5b1ac96b10976ed0e092a0bc1dd5526101356/uvicorn/supervisors/statreload.py#L25).
As we can see, there is a hint that `uvicorn` has option `--reload-exclude` and to enable this we got to have `watchfiles` package.
Then I've found other thread discussing the issue [GitHub discussion](https://github.com/encode/uvicorn/issues/338).

As we can see from `reflex` source code [link](https://github.com/pynecone-io/pynecone/blob/1dfe76c4a3996edf3d4b7f00b6e0cc978e9942c4/reflex/utils/exec.py#L132C12-L132C12) it run `uvicorn` just with option `--reload`.

According to `uvicorn` documentation [doc link](https://www.uvicorn.org/settings/#reloading-with-watchfiles) using this flag will force `uvicorn` to look for all python files that might have changed in the current folder and it all subfolders.

And here is the issue!
When `reflex` compiles some project, it creates a `.web` directory.
Let just check how many files are there `find .web/ -type f | wc -l`. In my case for empty project I got *30763*, quite big amount of files to scan.

So as we already know, `uvicorn` has option `--reload-exclude` and to enable this functionality we need to add one more package.

After applying the change, finally CPU is totally cool.
Take a look at the second screenshot.
I've already made changes and will push PR after the issue will be saved.

**To Reproduce**
- create an empty folder
- init empty project with `reflex init`
- run project with `reflex run`

**Expected behavior**
Open any system monitor and find python CPU usage. You will notice that even in case when the web server is idling it consume some quite high CPU as for doing nothing. Take a look at the screenshot below. Also, you may notice increased disk usage.

**Screenshots**
> I incorrectly highlighted the process on the screenshot, we need a python process not kernel one

CPU usage before fix cpu before

CPU usage after fix cpu after

 - Python Version: 3.11.4
 - Reflex Version: 0.1.34
 - OS: macos Ventrura 13.3.1
 - Browser (Optional): FF 115

maks1m:
Exclude `.web` folder from uvicorn file sync due to increased CPU load and redundant scanning of thousands of files in that folder;

Add package `watchfiles` to enable `uvicorn` use option `--reload-exclude`.

### Type of change

`closes #1251`


picklelo (dismissed):
nice catch!

picklelo (commented):
I think you need to run `poetry install` to update the lock file

picklelo (approved):

PR #1265, opened 2023-06-28
fix pc export

sheldonchiu:
All Submissions:

Have you followed the guidelines stated in [CONTRIBUTING.md](https://github.com/pynecone-io/pynecone/blob/main/CONTRIBUTING.md) file?

    Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/pynecone-io/pynecone/pulls) for the desired changed?

Type of change

    Bug fix (non-breaking change which fixes an issue)

app.compile() was not run since app.py was not imported. bun run export will fail due to missing js files

### DIFF
--- a/reflex/reflex.py
+++ b/reflex/reflex.py
@@ -229,6 +229,7 @@ def export(
 
     if frontend:
         build.setup_frontend(Path.cwd())
+        prerequisites.get_app()
 
     build.export_app(
         backend=backend,


picklelo (approved):
nice catch, we should add a comment describing why this line is necessary so we don't forget in the future

masenf (approved):

PR #1271, opened 2023-06-28
Fix missing stdout in event handlers

picklelo:
We shouldn't redirect stdout for the backend process.


masenf (approved):
works for me; thanks!

PR #1272, opened 2023-06-29
link: clear `as` property if link has no href

### LINKED ISSUE #105: crm example: The link Log Out is disappear (0.1.33)
From the following pictures, we know that the Link disappears after using pc 0.1.33
This is **0.1.32**
Screenshot 2023-06-11 at 5 03 15 PM
This is **0.1.33**
Screenshot 2023-06-11 at 5 01 04 PM

```
My Testing Environment
        OS Darwin 22.4.0
        Pynecone 0.1.33
        Python 3.11.3
        Node v16.8.0
        Bun 0.5.9
```

masenf:
The component is _already_ a Link, so rendering it as a Link doesn't make sense, and also doesn't work because `"Link"` is not the same as `{Link}`.

Fix pynecone-io/pynecone-examples#105 pynecone-io/pynecone-examples#110

Follow-up for #1178

### DIFF
--- a/reflex/components/navigation/link.py
+++ b/reflex/components/navigation/link.py
@@ -51,7 +51,7 @@ class Link(ChakraComponent):
             raise ValueError("Link without a child will not display")
         elif href is None and len(children):
             # Don't use a NextLink if there is no href.
-            props["as_"] = "Link"
+            props["as_"] = ""
         if href:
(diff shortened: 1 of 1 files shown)


Alek99 (approved):
lgtm

PR #1278, opened 2023-06-29
Bugfix: Some Pages are not Generated:

### LINKED ISSUE #1264: Some pages are not generated on reflex run
running pcweb should be enough to repro this. Found the root cause to be multiprocessing logic in adding pages: https://github.com/pynecone-io/pynecone/pull/1199

ElijahAhianyo:
Fix the issue with some pages not being generated.

fixes #1264

### DIFF
--- a/reflex/utils/path_ops.py
+++ b/reflex/utils/path_ops.py
@@ -71,8 +71,7 @@ def mkdir(path: str):
     Args:
         path: The path to the directory.
     """
-    if not os.path.exists(path):
-        os.makedirs(path)
+    os.makedirs(path, exist_ok=True)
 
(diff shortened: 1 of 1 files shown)


picklelo (approved):

masenf (approved):
excellent. have you audited any other places in the code where we're introducing race conditions like this for no reason?

ElijahAhianyo:
> excellent. have you audited any other places in the code where we're introducing race conditions like this for no reason?

Yeah, this should be good, Ive created an issue #1279  as a follow to unit-test the compile logic. Doesn't appear to be any tests for this atm.

PR #1281, opened 2023-06-30
fix silent compilation error

Lendemor:
Resolve a problem where compilation fail silently instead of raising error properly.
Also resolve Prism conflict between code_block and markdown

### All Submissions:

### Type of change

### Changes To Core Features:

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -466,21 +466,26 @@ class App(Base):
         # Compile the pages.
         custom_components = set()
         thread_pool = ThreadPool()
+        compil_results = []
         for route, component in self.pages.items():
             component.add_style(self.style)
-            thread_pool.apply_async(
-                compiler.compile_page,
-                args=(
-                    route,
-                    component,
-                    self.state,
-                    self.connect_error_component,
-                ),
+            compil_results.append(
+                thread_pool.apply_async(
+                    compiler.compile_page,
+                    args=(
+                        route,
+                        component,
+                        self.state,
+                        self.connect_error_component,
+                    ),
+                )
             )
             # Add the custom components from the page to the set.
             custom_components |= component.get_custom_components()
         thread_pool.close()
         thread_pool.join()
+        for r in compil_results:
+            r.get()
 
         # Compile the custom components.
         compiler.compile_components(custom_components)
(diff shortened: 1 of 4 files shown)


reflex/app.py
```diff
@@ -466,21 +466,26 @@ def compile(self):
         # Compile the pages.
         custom_components = set()
         thread_pool = ThreadPool()
+        compil_results = []
```
picklelo, line 469: should this be `compile_results`

reflex/app.py
```diff
@@ -466,21 +466,26 @@ def compile(self):
...
+                        self.state,
+                        self.connect_error_component,
+                    ),
+                )
             )
             # Add the custom components from the page to the set.
             custom_components |= component.get_custom_components()
         thread_pool.close()
         thread_pool.join()
+        for r in compil_results:
```
picklelo, line 487: Can we add a comment saying why we need this line

reflex/compiler/compiler.py
```diff
@@ -78,11 +78,24 @@ def _compile_page(
...
+
     Returns:
         The compiled component.
     """
     # Merge the default imports with the app-specific imports.
     imports = utils.merge_imports(DEFAULT_IMPORTS, component.get_imports())
+    # Make sure that the same Tag is not used in multiple import.
+    used_tags = {}
+    for lib, imps in imports.items():
+        for iv in imps:
```
picklelo, line 92: Can we use full names for the variables, e.g. `for import in imps` and below, `import_name = `, to make it a bit more readable

reflex/compiler/compiler.py
```diff
@@ -78,11 +78,24 @@ def _compile_page(
...
 
+    Raises:
+        ValueError: if a conflict on "tag/alias" is detected for an import.
+
     Returns:
         The compiled component.
     """
     # Merge the default imports with the app-specific imports.
     imports = utils.merge_imports(DEFAULT_IMPORTS, component.get_imports())
+    # Make sure that the same Tag is not used in multiple import.
```
picklelo, line 89: Nice check! We could also move this into the `merge_imports` function. Or maybe add a function `utils.validate_imports` to keep this function cleaner

picklelo (approved):
thanks for the updates!

PR #1284, opened 2023-06-30
Add default styles to markdown

picklelo:
Add default padding to the typography elements for better default appearance.


Alek99 (approved):
👍

PR #1292, opened 2023-07-01
fix the link toward docs in default apps

Lendemor:
template for new apps now point to `reflex.dev` instead of `pynecone.io`


Alek99 (commented):
👍

Alek99 (approved):

PR #1295, opened 2023-07-01
Refactored project dependencies 'plotly' and 'redis'

### LINKED ISSUE #1240: Get Rid of Plotly/Redis Dependency

palashcode:
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

Refactored project dependencies:
- Moved 'plotly' and 'redis' dependencies from regular dependencies to extra dependencies using Poetry. This change ensures that these dependencies are only installed when required, reducing the overall project dependencies.
- Added try-except to capture the error when the user needs the package and it is required to be installed.
- To install these extra dependencies when required, you can use the following command:
```poetry install -E plotly -E redis```

closes #1240

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -26,7 +26,7 @@ trio = ["trio (<0.22)"]
 name = "async-timeout"
 version = "4.0.2"
 description = "Timeout context manager for asyncio programs"
-optional = false
+optional = true
 python-versions = ">=3.6"
 files = [
     {file = "async-timeout-4.0.2.tar.gz", hash = "sha256:2163e1640ddb52b7a8c80d0a67a08587e5d245cc9c553a74a847056bc2976b15"},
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by Lendemor** [comment]
This change assume that all users will use `poetry`, which may not be the case.
We are using it for the project itself, but we shouldn't force it on the end users of the package.

**Comment by palashcode** [comment]
Do you mean error should be this 
```
raise ImportError(
            "plotly is not installed\n"
        ) from e
```
Instead of this
```
raise ImportError(
            "plotly is not installed\n" "install it using 'poetry install -E plotly'"
        ) from e
```
And then used can install it using pip or something else also.

masenf:
Part of my concern with this change is that it increases friction for new users and might be breaking workflows for existing users. Although i do agree generally about trying to reduce the footprint of reflex, particularly for advanced use cases, we want to keep the `pip install reflex` step very simple.

Because extras are always extra, there's no python packaging mechanism that I'm aware of to have the base package (`reflex`) be the complete installation, other than creating a second (`reflex-core`) package and then just treating the `reflex` name as a metapackage that installs all of the `reflex-core` with all extras, and allowing customization of extras for `reflex-core` in advanced use case.

I expect to put some more time into this problem and PR tomorrow, thanks for your patience and contribution.

masenf (changes requested):
i'm hesitating on this for a few reasons:

* suggesting either `pip` or `poetry` to install the dependencies may be wrong, depending on the project. Shouldn't we suggest installing the `reflex[redis]` or `reflex[plotly]` extras instead?
* duplication on the import checking; i would prefer some kind of utility function that we can call to have the "X is not installed, do Y to install X" in one place instead of scattered throughout the codebase

Otherwise i think we can take this in, since plotly and redis functionality are currently not core to any built in workflow. If you're using them, then you would know about it and can adjust accordingly.

masenf:
Close in favor of #1295

PR #1305, opened 2023-07-05
vars: unbox EventHandler and functools.partial for dep analysis

### LINKED ISSUE #1303: AttributeError: 'EventHandler' object has no attribute '__code__' when cached_var calls event handler
**Describe the bug**

When an `rx.cached_var` function calls an EventHandler, the dependency analysis breaks when attempting to look up the function code for the event handler. It should be dereferencing `.fn` and analyzing its code.

**To Reproduce**
```python
import reflex as rx


class State(rx.State):
    def handle(self):
        print("handle")

    @rx.cached_var
    def foo(self) -> str:
        self.handle()
        return "foo"


app = rx.App(state=State)

@app.add_page
def index() -> rx.Component:
    return rx.text("Welcome to ", State.foo)

app.compile()
```

**Expected behavior**
The event handler is called while calculating the cached var.

 - Python Version: 3.11.4
 - Reflex Version: 0.2.0
 - OS: macOS 13.4.1
 
**Additional context**
[Discord thread 🧵](https://discord.com/channels/1029853095527727165/1061874061250150441/1126218871905468518)

masenf:
When calculating the variable dependencies of a cached_var, reach into objects with .func or .fn attributes and perform analysis on those.

Fix #1303

### DIFF
--- a/reflex/vars.py
+++ b/reflex/vars.py
@@ -869,10 +869,18 @@ class ComputedVar(Var, property):
                 obj = cast(FunctionType, self.fget)
             else:
                 return set()
-        if not obj.__code__.co_varnames:
+        with contextlib.suppress(AttributeError):
+            # unbox functools.partial
+            obj = cast(FunctionType, obj.func)  # type: ignore
+        with contextlib.suppress(AttributeError):
+            # unbox EventHandler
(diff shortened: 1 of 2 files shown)


Alek99 (approved):
👍

PR #1306, opened 2023-07-05
WIP:upload using websockets

ElijahAhianyo:
This PR is a WIP aimed at performing uploads via websockets(initially done through REST)

### DIFF
--- a/reflex/event.py
+++ b/reflex/event.py
@@ -25,6 +25,8 @@ class Event(Base):
     # The event payload.
     payload: Dict[str, Any] = {}
 
+    type_: Optional[str] = None
+
 
 class EventHandler(Base):
     """An event handler responds to an event to update the state."""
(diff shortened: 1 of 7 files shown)


reflex/event.py
```diff
@@ -25,6 +25,8 @@ class Event(Base):
     # The event payload.
     payload: Dict[str, Any] = {}
 
+    type_: Optional[str] = None
```
ElijahAhianyo, line 28: Have this here as a means to identify what event is an upload event. Is this a good enough approach? Are there better alternatives? The name seems a bit misleading though. What's a better one (assuming this is the right approach, of course)

masenf (commented):
my concern currently is that a large upload could block UI interactivity if other events cannot be processed until the upload is complete.

secondary concern is that the b64 -> websocket -> b64 -> disk is not optimized like nginx or other proxy servers are at efficiently getting the data to disk without extra slow memory copies and processing steps.

more robust performance characterization is needed

PR #1307, opened 2023-07-06
component: do not create ref for dynamic id

### LINKED ISSUE #1302: Can't use id-based anchor link in foreach loop
**Describe the bug**

If you use the `id` field of a component within a `foreach` loop, you get an error. This prevents you from taking advantage of the anchor linking functionality introduced in #469.

**To Reproduce**

* Create a list of objects that have an ID field
* Create a foreach loop with a component within it
* Set the `id` field of the component to the ID of objects in the list
* You will get an error

**Expected behavior**

You should be able to set the `id` on components within a `foreach` loop so you can anchor link to items generated by the loop.

**Specifics:**
 - Python Version: 3.11.4
 - OS: Mac
 - Browser: Chrome
 
**Additional context**

The specifics of my case are:

- I have a references page
- The page is generated dynamically at /references/{entity_one}/{entity_two} to provide references for the relationship between the two entities specified
- I generate the references using a `foreach` loop over a list of `Reference` objects
- I want to anchor link directly to references, using the format /references/{entity_one}/{entity_two}#reference_id
- However, if I add any component within the foreach loop and set its `id` value to reference.reference_id, I get an error (a "hard navigation" type of error specifically)
- My solution was to use the `html` component to create an `<a name="reference.reference_id"></a>` anchor

masenf:
allow dynamic id (BaseVar) to be specified (from state value or foreach), but do not create a react reference for it, since that must be known at compile time.

if a react ref is needed, then specify a string _at compile time_, not from state.

fix #1302

### DIFF
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -553,7 +553,8 @@ class Component(Base, ABC):
         Returns:
             The ref name.
         """
-        if self.id is None:
+        # do not create a ref if the id is dynamic or unspecified
+        if self.id is None or isinstance(self.id, BaseVar):
             return None
         return format.format_ref(self.id)
 


picklelo (approved):
nice catch

PR #1313, opened 2023-07-06
state.js: accumulate incoming events

### LINKED ISSUE #1309: Some events are missed when multiple handlers are chained using yield
There seem to be some race conditions on the client that causes some events to be missed. To repro:

```python

class State(rx.State):
    x: int = 0

    def perform(self):
        yield self.add10
        yield self.add10
        yield self.add10

    def add1(self):
        self.x += 1

    def add10(self):
        self.x += 10


def index():
    return rx.center(
        rx.vstack(
            rx.text(State.x),
            rx.button(
                "Click here",
                on_click=State.perform
            ),
        )
    )


app = rx.App(state=State)
app.add_page(index)
app.compile()
```
If you run the example above, you'll notice the inconsistencies in additions

masenf:
If the backend yields multiple events from a single event handler, ensure that each event gets appended to the pending result before being transferred to the state. Previously each incoming event would overwrite any other pending event(s) leading to racy and unpredictable behavior when yielding events from the backend in quick succession.

Fixes #1309

Bonus Fix: when setting `processing: false` after client-side events (like _console), spread from `result`, not `state` to avoid accumulating duplicate queued events. I think this was originally an undiscovered bug that was exposed by the main fix.

### DIFF
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -234,7 +234,7 @@ export const processEvent = async (
 
   // If no event was sent, set processing to false.
   if (!eventSent) {
-    setResult({ ...state, final: true, processing: false });
+    setResult({ ...result, final: true, processing: false });
   }
 };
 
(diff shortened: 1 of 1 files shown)


ElijahAhianyo (approved):
looks great!

PR #1316, opened 2023-07-07
parameter for turning off nextJS compression

Lendemor:
### Description

add a `bool` parameter called next_compression allowing to turn off the NextJS Gzip compression

The compression is still turned on by default for better performance.

### Checklist

### Type of change

### New Feature Submission:

### DIFF
--- a/reflex/.templates/web/next.config.js
+++ b/reflex/.templates/web/next.config.js
@@ -1,3 +1,4 @@
 module.exports = {
   reactStrictMode: true,
+  compress: true,
 };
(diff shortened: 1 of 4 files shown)


masenf (commented):
what is the use case for turning this off? i'm trying to understand who would use this setting

Lendemor:
The use case is when you have a reverse proxy (Nginx, Apache, others...) between your nextjs serveur and your visitors.

If you want the compression to be done by the reverse proxy instead of NextJS, you have to turn it off in `next.config.json`

(see the `Compression` paragraph in my PR on self-hosting in the docs)

It's a niche parameter, but it's necessary if any user want some level of control over how the nextJS server is running.

(We could eventually generalize it by passing a dict of NextJS parameters rather than specific ones like that.)

PR #1321, opened 2023-07-10
rx.Model: automigrate using alembic

masenf:
Provide the means to automatically create schema migration scripts using alembic. This allows mostly seamless changes to models to take effect without losing data. It also provides fine-grain control over the migrations that can be tested and validated in development environments and easily played out for production databases.

How to use it:
* install `alembic` with downstream app requirements
* run `python -m alembic init` once within the same directory as `rxconfig.py`
* running the app will trigger alembic to upgrade the database to the latest `head` revision
* future changes to data models with attempt to automatically generate schema migration scripts inside `alembic/versions` and then apply them to upgrade the database
* to avoid automatic generation, [Create a Migration Script](https://alembic.sqlalchemy.org/en/latest/tutorial.html#create-a-migration-script) manually via `python -m alembic revision "apply some change"`

How to not use it:
* do not have an `alembic.ini` in your app directory, and alembic will not be used.

### Warnings

Alembic users MUST be aware of [What Does Autogenerate Detect and What Does it NOT Detect](https://alembic.sqlalchemy.org/en/latest/autogenerate.html#what-does-autogenerate-detect-and-what-does-it-not-detect).

Specifically, **table renames** and **column renames** are treated as a DROP / CREATE, which is a data _destructive_ operation. Although renames may be accommodated non-destructively by hand-creation or hand-editing of the generated migration script, in most cases, column and table renames should be avoided.

### New feature
- [ ] Doc updates
- [x] Unit tests

# Testing

Automated test cases run through a series of basic scenarios and validations. I've also been testing with a sample app where modifications are repeatedly made to a specific model.

## Scenarios

* new database
  * initialize alembic before first run
  * initialize alembic after first run
* table add / drop
* column add / drop
* change column type


reflex/model.py
```diff
@@ -75,6 +93,154 @@ def get_db_engine():
...
+            True - indicating the process was successful
+            None - indicating the process was skipped
+        """
+        if not has_alembic or not Path(constants.ALEMBIC_CONFIG).exists():
+            return
+
+        with cls.get_db_engine().connect() as connection:
+            cls._alembic_upgrade(connection=connection)
+            changes_detected = cls._alembic_autogenerate(connection=connection)
+            if changes_detected:
```
picklelo, line 239: Maybe we can give the user a prompt here to confirm if they want to migrate

picklelo (approved):
Looks good to me!

reflex/model.py
danik292: Or make a setting If the user wants to update automatically or ask

picklelo: @danik292 We have a follow up #1336 where we will add this as a config option

PR #1322, opened 2023-07-11
Computed vars tuple and str indexing support

### LINKED ISSUE #1325: Computed vars Tuple support
base vars support tuples but computed vars dont

ElijahAhianyo:
added support for tuples. This should work now:

```python
...
    @rx.var
    def elements(self) -> Tuple[str, str]:
        return "hello", "world"
...

```
indexing str vars should also work:

```python

class State(rx.State):
      value: str = "hello world"
   
def index():
     return rx.vstack(
     rx.text(State.value[0]),
     rx.text(State.value[3:-1])
)

```

fixes #1325

### DIFF
--- a/reflex/vars.py
+++ b/reflex/vars.py
@@ -17,6 +17,7 @@ from typing import (
     Optional,
     Set,
     Type,
+    Tuple,
     Union,
     _GenericAlias,  # type: ignore
     cast,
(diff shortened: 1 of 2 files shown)


picklelo (approved):
looks good to me

Issue #1087, opened 2023-05-26
The cursor always show in wrong place when every time we typing in TextArea . 


milochen0418:
**Describe the bug**
This is a sample code from stack overflow. 
And the problem we meet is that Textarea cannot for us to type. 

Setting the value and on_change looks like the only way to support the typing and the setting value by function. 
And the cursor is just put in the wrong place... So we cannot support type well... 

```python
"""Welcome to Pynecone! This file outlines the steps to create a basic app."""
from pcconfig import config

import pynecone as pc

class State(pc.State):
    q1_response: str
    q1_c0: bool = False
    q1_c1: bool = False
    q1_c2: bool = False
    def add_response(self, *args):
        text = ''
        if self.q1_c0:
            text += q.options[0][1] + '\n'
        if self.q1_c1:
            text += q.options[1][1] + '\n'
        if self.q1_c2:
            text += q.options[2][1] + '\n'
        self.q1_response = text.strip()
    def test_get_value(self):
        return "GDD"
    def the_key_down(self,*args):
        print("the_key_down")
        self.q1_response = self.q1_response + "A" 


class Question:
    def __init__(self):
        self.q = 'This is the question'
        self.options = [
            ('None', 'Nothing'),
            ('Answer 1', 'The long answer 1.'),
            ('Answer 2', 'The long answer 2.'),
        ]
q = Question()
def index():
    return pc.container(
        pc.text(q.q, font_size="1.5em"),
        pc.vstack(
            pc.checkbox(q.options[0][0] + ' - ' + q.options[0][1], on_change=State.set_q1_c0),
            pc.checkbox(q.options[1][0] + ' - ' + q.options[1][1], on_change=State.set_q1_c1),
            pc.checkbox(q.options[2][0] + ' - ' + q.options[2][1], on_change=State.set_q1_c2),
            pc.button("Add", on_click=State.add_response, bg="blue", color="white"),
            pc.text_area(
                value = State.q1_response,
                placeholder = 'The long answer will go here',
                on_change = State.set_q1_response,
                rows = 8,
            ),
        ),
    )   
app = pc.App(state=State)
app.add_page(index)
app.compile()
```
**To Reproduce**
 - Code/Link to Repo:

**Expected behavior**

**Screenshots**

** Specifics (please complete the following information):**
 - Python Version: 3.11
 - Pynecone Version: 0.1.31
 - OS: Mac 
 - Browser (Optional): chrome
 
**Additional context**

[...]

Issue #673, opened 2023-03-14
Input example w/ Clear button loses characters


ShakataGaNai:
**Describe the bug**
When you control the text of a field, the server does not save the text fast enough.

**To Reproduce**
* https://pynecone.io/docs/library/forms/input
* Use the 3rd input box, the one with a clear button.
* If you type fast enough, it saves *some* but not everything of what you type

**Expected behavior**
I'd expect it to save everything I type. 

**Screenshots**
https://youtu.be/k9_lFsbLUB4

**Specifics (please complete the following information):**
 - Python Version: 
 - Pynecone Version:
 - OS: macOS 13.1
 - Browser (Optional): Firefox 110.0.1 (64-bit)
 
**Additional context**
I've tested this locally, with a very simple single-page pynecone app (less than 100 lines total), and this seems to work as you'd expect it. Possibly because the help docs are bigger and more complicated? But it seems like maybe it is overkill to need to `set_text` and build the var/function... just to have a clear button?

ShakataGaNai:
Did some more research, and I found the on_key_down / on_key_up input example is more reliably slow. This is me typing "asdf" and backspace (repeatedly). https://youtu.be/t-zgIu6tqcw - This was recorded in Brave Browser, but I also tested Firefox & Chrome. Also tested in private mode with all extensions disabled with the same result. 

Initially I was experimenting with figuring out how to capture enter and only enter. Since it seems to capture any meta keys, including backspace.

Alek99:
Got it thanks for taking a look, I'll do some profiling on this tomorrow can post what I find it should be fast. Are you just using `pc run` or `pc run --env prod`. Dev mode is slower than prod

ShakataGaNai:
Locally I ran `pc run`, no prod. But the videos were from the official pynecone.io docs.

masenf:
While investigating this issue, I found that the [fix](https://github.com/pynecone-io/pynecone/pull/700/files) for #632 aims to alleviate the sort of problem seen here, but I haven't been able to get it working yet.

As an interim solution for having nice form inputs, I ended up wrapping [`react-debounce-input`](https://www.npmjs.com/package/react-debounce-input) to great effect. The benefits are less websocket roundtrips and automatic blur/enter triggering on onChange.

Here is my sample code, which allows comparing the direct control and debounced control behavior. I don't love the `get_custom_code`, but posting here in case it helps anyone with this approach, or maybe someone has a better idea.

```python
"""A debounced input example"""
[...]

PR #1327, opened 2023-07-11
Add debounce_input based on react-debounce-input

masenf:
The debounce_input component can help reduce roundtrips when using fully controlled input elements. This is especially critical in situations of high latency and also fixes the issue where the cursor moves to the end of the input while typing.

Addresses #673 and #1087

Originally available as https://github.com/trivial-intelligence/reflex-debounce-input

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
     rev: v1.1.313
     hooks:
     - id: pyright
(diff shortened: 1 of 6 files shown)


picklelo (approved):
nice! We can apply debouncing to slider also, I think those are the only two components that really need it (maybe color-picker if we bring that to the core). This test harness is great

PR #1333, opened 2023-07-12
Strict type checking for indexing with vars

### LINKED ISSUE #1331: No strict type checking for indexing with vars
This code shouldnt be allowed but it compiles into a js code that doesnt fail either:

```python
class State(rx.State):
    d: list[int] = [1, 2, 3]

    @rx.var
    def elements(self) -> dict[str, str]:
        return {"name": "reflex"}


def index():
    return rx.center(
        rx.text(State.elements[State.d]),
    )
```

ElijahAhianyo:
closes #1331

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,15 +37,15 @@ jobs:
         virtualenvs-create: true
         virtualenvs-in-project: true
 
-#    - name: cache deps
-#      id: cache-deps
-#      uses: actions/cache@v2
-#      with:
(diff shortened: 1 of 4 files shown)


ElijahAhianyo:
Keeping this as a draft till this PR #1322 get's merged. Will need to rebase and add tests for tuples

masenf (approved):

picklelo (approved):
Awesome, thanks for adding all the unit tests too

PR #1336, opened 2023-07-12
reflex db migrate CLI and associated config

masenf:
# Docs: https://github.com/reflex-dev/reflex-web/pull/136

# CLI

## `reflex db init`

Initialize alembic, run the first `makemigrations`, then apply migrations.

When an app attempts to call `reflex.model.get_engine()` and there is no alembic config file, a warning will be printed instructing the user to first run this command.

## `reflex db migrate`

There is now a dedicated command, `reflex db migrate` which initiates `Model.migrate`

## `reflex db makemigrations`

When alembic is used, autogenerate revisions between the current database schema and the sqlmodel metadata. If the current database is not already up to date, then it must first be updated to the `head` revision before additional revisions can be generated.

Autogenerate only happens when the user requests it, or when initializing the database. Because autogenerated migrations are best-effort and often require manual tweaks, it doesn't make sense to generate and use schema migrations by default except for the initial schema generation (which is easy to derive).

# Workflow

NOTE: this is currently a bit out of date with the latest revisions, will update once all of the changes are in.

1. create a new reflex app

```console
masenf@minicone ~ % mkdir my-alembic-app && cd my-alembic-app
masenf@minicone my-alembic-app % python3 -m venv .venv && . ./.venv/bin/activate
(.venv) masenf@minicone my-alembic-app % python3 -m pip install ~/code/pynecone-io/reflex
(.venv) masenf@minicone my-alembic-app % reflex init
──────────────────────────────────────────────────── Initializing my_alembic_app ─────────────────────────────────────────────────────
[12:58:08] Initializing the web directory.                                                                               console.py:30
           Initializing the app directory.                                                                               console.py:30
           Finished Initializing: my_alembic_app                                                                         console.py:30
```

2. add a data model to `my_alembic_app/my_alembic_app.py`

```python
class T(rx.Model, table=True):
    x: str = "x"
```

3. makemigrations before initializing alembic?

```console
(.venv) masenf@minicone my-alembic-app % reflex db makemigrations
Database is not initialized. Run reflex db init first.
```

4. initialize alembic, and attempt to run the app

```console
(.venv) masenf@minicone my-alembic-app % reflex db init
  Creating directory '/Users/masenf/code/scratch/my-alembic-app/alembic' ...  done
  Creating directory '/Users/masenf/code/scratch/my-alembic-app/alembic/versions' ...  done
  Generating /Users/masenf/code/scratch/my-alembic-app/alembic/script.py.mako ...  done
  Generating /Users/masenf/code/scratch/my-alembic-app/alembic/env.py ...  done
  Generating /Users/masenf/code/scratch/my-alembic-app/alembic/README ...  done
  Generating /Users/masenf/code/scratch/my-alembic-app/alembic.ini ...  done
  Please edit configuration/connection/logging settings in '/Users/masenf/code/scratch/my-alembic-app/alembic.ini' before proceeding.
  Generating /Users/masenf/code/scratch/my-alembic-app/alembic/versions/7337dcab68c0_.py ...  done
(.venv) masenf@minicone my-alembic-app % reflex run
──────────────────────────────────────────────────────── Starting Reflex App ─────────────────────────────────────────────────────────
──────────────────────────────────────────────────── Installing frontend packages ────────────────────────────────────────────────────
bun install v0.6.9 (2ab797ff)
──────────────────────────────────────────────────────────── App Running ─────────────────────────────────────────────────────────────
App running at: http://localhost:3000
[Updating App] Applying changes and refreshing. Time: 2023-07-15 15:24:50.118794
```

`reflex db init` automatically created an alembic revision script with the schema and applied it to the database

```python
"""empty message

Revision ID: 7337dcab68c0
Revises: 
Create Date: 2023-07-15 15:24:23.377266

"""
from alembic import op
import sqlalchemy as sa
import sqlmodel

# revision identifiers, used by Alembic.
revision = '7337dcab68c0'
down_revision = None
branch_labels = None
depends_on = None


def upgrade() -> None:
    # ### commands auto generated by Alembic - please adjust! ###
    op.create_table('t',
    sa.Column('id', sa.Integer(), nullable=False),
    sa.Column('x', sqlmodel.sql.sqltypes.AutoString(), nullable=False),
    sa.PrimaryKeyConstraint('id')
    )
    # ### end Alembic commands ###


def downgrade() -> None:
    # ### commands auto generated by Alembic - please adjust! ###
    op.drop_table('t')
    # ### end Alembic commands ###
```

4a. BONUS: initialize the schema against a second database

```console
(.venv) masenf@minicone my-alembic-app % DB_URL=sqlite:///second.db reflex db migrate
(.venv) masenf@minicone my-alembic-app % sqlite3 second.db <<< 'PRAGMA table_info(t)'
0|id|INTEGER|1||1
1|x|VARCHAR|1||0
```

5. Add a field to the model

```python
class T(rx.Model, table=True):
    x: str = "x"
    y: str = "y"
```

6. Run the app again

```console
(.venv) masenf@minicone my-alembic-app % reflex run                                  
──────────────────────────────────────────────────────── Starting Reflex App ─────────────────────────────────────────────────────────
Detected database schema changes. Run reflex db makemigrations to generate migration scripts.
```

reflex used alembic to detect schema changes between the current database and the sqlmodel metadata, but lets the app attempt to run anyway.

7. Explicitly create migration script

```console
(.venv) masenf@minicone my-alembic-app % reflex db makemigrations            
  Generating /Users/masenf/code/scratch/my-alembic-app/alembic/versions/3a02a76485bf_.py ...  done
```

The generated script adds the column:

```python
"""empty message

Revision ID: 3a02a76485bf
Revises: dbb07d44141c
Create Date: 2023-07-13 13:07:46.660884

"""
from alembic import op
import sqlalchemy as sa
import sqlmodel

# revision identifiers, used by Alembic.
revision = '3a02a76485bf'
down_revision = 'dbb07d44141c'
branch_labels = None
depends_on = None


def upgrade() -> None:
    # ### commands auto generated by Alembic - please adjust! ###
    op.add_column('t', sa.Column('y', sqlmodel.sql.sqltypes.AutoString(), server_default=sa.text("'y'"), nullable=False))
    # ### end Alembic commands ###


def downgrade() -> None:
    # ### commands auto generated by Alembic - please adjust! ###
    op.drop_column('t', 'y')
    # ### end Alembic commands ###
```

8. Run the app again before apply the migration script

```console
(.venv) masenf@minicone my-alembic-app % reflex run         
──────────────────────────────────────────────────────── Starting Reflex App ─────────────────────────────────────────────────────────
Target database is not up to date. Run reflex db migrate to update database.
```

Additionally, notice that subsequent `makemigrations` will not work until the current database is brought to the latest revision:

```console
(.venv) masenf@minicone my-alembic-app % reflex db makemigrations             
...
alembic.util.exc.CommandError: Target database is not up to date.
```

9. Perform the migration and run the app again

```console
(.venv) masenf@minicone my-alembic-app % reflex db migrate && reflex run                           
──────────────────────────────────────────────────────── Starting Reflex App ─────────────────────────────────────────────────────────
──────────────────────────────────────────────────── Installing frontend packages ────────────────────────────────────────────────────
bun install v0.6.9 (2ab797ff)
──────────────────────────────────────────────────────────── App Running ─────────────────────────────────────────────────────────────
App running at: http://localhost:3000
```

9a. BONUS apply migrations to another database

```console
(.venv) masenf@minicone my-alembic-app % DB_URL=sqlite:///second.db reflex db migrate
(.venv) masenf@minicone my-alembic-app % sqlite3 second.db <<< 'PRAGMA table_info(t)'
0|id|INTEGER|1||1
1|x|VARCHAR|1||0
2|y|VARCHAR|1|'y'|0
```


reflex/app.py
```diff
@@ -444,7 +444,7 @@ def compile(self):
         config = get_config()
 
         # Update models during hot reload.
-        if config.db_url is not None and not Model.automigrate():
+        if config.db_url is not None and config.db_migrate and not Model.migrate():
```
picklelo, line 447: Maybe we can wrap up this condition into a function - `if should_migrate():`

reflex/config.py
```diff
@@ -203,6 +206,19 @@ class Config(Base):
     # Timeout when launching the gunicorn server.
     timeout: int = constants.TIMEOUT
 
+    # These fields are neither saved nor read from the environment
```
picklelo, line 209: I wonder if eventually we should support all of these as environment fields also for consistency

reflex/model.py
```diff
@@ -132,15 +132,23 @@ def _alembic_render_item(
...
+    def alembic_autogenerate(
+        cls,
+        connection: sqlalchemy.engine.Connection,
+        write_migration_scripts: bool = True,
+    ) -> bool:
         """Generate migration scripts for alembic-detectable changes.
 
         Args:
             connection: sqlalchemy connection to use when detecting changes
+            write_migration_scripts: if True, write autogenerated revisions to script directory
```
picklelo, line 144: nit: can we start the comments with capital letters and end with periods for consistency - `If True, ... directory.`

reflex/model.py
```diff
@@ -184,7 +192,7 @@ def run_autogenerate(rev, context):
             upgrade_ops = revision_context.generated_revisions[-1].upgrade_ops
             if upgrade_ops is not None:
                 changes_detected = bool(upgrade_ops.ops)
-        if changes_detected:
+        if changes_detected and write_migration_scripts:
```
picklelo, line 195: Below this line, can we just make `revision_context.generate_scripts()` generate all the scripts without having to iterate? Or are there places where we need the `_script` value?

reflex/reflex.py
```diff
@@ -78,6 +78,11 @@ def run(
     frontend_port: str = typer.Option(None, help="Specify a different frontend port."),
     backend_port: str = typer.Option(None, help="Specify a different backend port."),
     backend_host: str = typer.Option(None, help="Specify the backend host."),
+    db_migrate: bool = typer.Option(
```
picklelo, line 81: Since we have the `db` subcommand now, it may be cleaner to only have one way of migrating and not support auto migrate in the `run` command anymore.

reflex/reflex.py
```diff
@@ -124,13 +129,20 @@ def run(
         )
         raise typer.Exit()
 
+    # Update db migration settings
+    if not db_migrate:
+        get_config().db_migrate = False
```
picklelo, line 134: Maybe we don't need the `db_migrate` in the config anymore. Just have them explicitly do migrations.

reflex/reflex.py
```diff
@@ -228,6 +240,8 @@ def export(
     console.rule("[bold]Compiling production app and preparing for export.")
 
     if frontend:
+        # never migrate the database during export
```
picklelo, line 243: same nit: Can we start all the comments with capital letters and end with periods everywhere

reflex/reflex.py
```diff
@@ -255,6 +269,38 @@ def export(
         )
 
 
+db_cli = typer.Typer()
+
+
+@db_cli.command()
+def migrate():
+    """Create or update database schema based on app models or existing migration scripts."""
```
picklelo, line 277: Does this require that `makemigrations` is run first, and then will apply those?

reflex/reflex.py
```diff
@@ -255,6 +269,38 @@ def export(
...
+    # explicitly running `reflex db migrate` overrides config file value
+    get_config().db_migrate = True
+    prerequisites.get_app()
+    prerequisites.check_schema_up_to_date()
+
+
+@db_cli.command()
+def makemigrations():
```
picklelo, line 285: Maybe we can name these two commands something like `make-migrations` and `apply-migrations` for clarity?

tests/test_config.py
```diff
@@ -50,7 +50,7 @@ def config_none_db_url_values(base_config_values):
     """
     base_config_values["db_url"] = "None"
     yield base_config_values
-    os.environ.pop("DB_URL")
+    os.environ.pop("DB_URL", None)
```
picklelo, line 53: Why do we need to `pop` here instead of `get`

masenf: We are removing the value from the current environment. This change just ensures it's a no-op if the variable is not defined.

Previously, the code was always setting DB_URL in the environment so the tests didn't notice.

reflex/reflex.py
masenf: Yeah, I think that might be a simpler mental model:

Initialize like before on blank database, otherwise print the error message and tell the user to explicitly migrate.

reflex/reflex.py
siddhantgoel, line 285: It would be nice to have this command accept a `--message` argument so we're able to name migration files for readability. As a user, I'm not sure if I recall not giving any of my migration files a name.

reflex/reflex.py
```diff
@@ -255,6 +269,38 @@ def export(
...
+    """Create autogenerated alembic migration scripts."""
+    if not model.has_alembic:
+        console.print(
+            "[red]Alembic is not installed. Run [bold]pip install alembic[/bold] first."
+        )
+        return
+    if not Path(constants.ALEMBIC_CONFIG).exists():
+        console.print(
+            "[red]Alembic is not initialized. Run [bold]python -m alembic init alembic[/bold] first."
+        )
```
siddhantgoel, line 295: Is there a specific reason for not putting this behind a `reflex db init`?

reflex/reflex.py
```diff
@@ -255,6 +269,38 @@ def export(
...
+        return
+    get_config().db_migrate = False
```
siddhantgoel, line 297: Off-topic: it would be nice to have a `set_config` function for cases like this (eg. `set_config("db_migrate", False)`).

reflex/reflex.py
```diff
@@ -255,6 +269,38 @@ def export(
...
+    prerequisites.get_app()
+    prerequisites.check_schema_up_to_date()
+
+
+@db_cli.command()
+def makemigrations():
+    """Create autogenerated alembic migration scripts."""
+    if not model.has_alembic:
+        console.print(
+            "[red]Alembic is not installed. Run [bold]pip install alembic[/bold] first."
```
siddhantgoel, line 289: I'm not sure if we should prompt the user to specifically use pip. They might be using something else eg. poetry or pipenv. Maybe we could have a more generic message here?

reflex/reflex.py
masenf: this naming convention was designed to mirror django; i can make it whatever you want, so let me know.

reflex/app.py
masenf: this bit was removed entirely

reflex/config.py
masenf: i dropped this commit for now. i think the whole Config handling needs a rethink before too long, but that was too much of a refactor for this PR

reflex/reflex.py
masenf: updated in the latest commit

reflex/reflex.py
masenf: yes, `makemigrations` has to run first.

the `reflex db init` commit will initialize alembic, run `makemigrations` and then perform the migration. so after `init` the database is completely ready to go without further commands, until the schema changes.

reflex/reflex.py
masenf: also a great idea, and that's been done in the latest commit.

if you're using the db built into reflex, you will be using alembic. otherwise you can bring your own database, since the db layer has been pulled out into its own CLI and never does migrations by default anymore.

reflex/reflex.py
masenf: alembic was made part of the requirements, and will be present whenever reflex is installed.

down the road, we might pull the db stuff into an optional extra, and alembic will be part of that extra (think `pip install reflex[db]`)

reflex/model.py
masenf: i ended up turning this into a `tuple()` collapse. the problem is that `generate_scripts` is an iterator, so we need to iterate through it or the scripts don't actually get written out.

reflex/reflex.py
picklelo: Can we move the docstring into the `help` kwarg - currently running `reflex db --help` cuts off the docstring for this method

reflex/reflex.py
```diff
@@ -255,6 +257,51 @@ def export(
...
+    """Create or update database schema based on app models or existing migration scripts."""
+    prerequisites.get_app()
+    if not prerequisites.check_db_initialized():
+        return
+    model.Model.migrate()
+    prerequisites.check_schema_up_to_date()
+
+
+@db_cli.command()
+def makemigrations(
```
picklelo, line 291: If I do the following, I get an error:

1. `reflex db init`
2. `rm reflex.db` (For example, if someone else clones the repo, they won't have the `.db` file)
3. `reflex db makemigrations` leads to `alembic.util.exc.CommandError: Target database is not up to date.`

If I run `reflex db migrate` first it will recreate the `.db` file. Maybe we should have a message if they try to run `makemigrations` without an existing db, telling them to run migrate first.

picklelo (approved):
Awesome work, tested with a few migrations and worked well for me.

Just one comment about potentially adding a custom error message / warning. We can add it in a follow up

reflex/reflex.py
picklelo: True, that would be cleaner as we do this in a few places

picklelo (approved):

reflex/reflex.py
masenf: filed https://github.com/reflex-dev/reflex/issues/1373 to capture this as part of a larger refactor!

PR #1339, opened 2023-07-13
Yield with uploads

ElijahAhianyo:
This PR should have uploads working with `yield`. The code below should work:

```python
import reflex as rx


class State(rx.State):
    img : list[str] = []
    val : str = "new"

    async def handle_upload(self, files : list[rx.UploadFile]):
        for file in files:
            outfile = f".web/public/{file.filename}"

            # Save the file.
            with open(outfile, "wb") as file_object:

            # Update the img var.

        yield State.set_value("Finished uploading")

    def set_value(self, val: str):
        self.val = val

    def upload_and_update(self, files : list[rx.UploadFile] ):

        yield State.set_value("uploading value")
        yield State.handle_upload(files)


def index():
    return rx.center(
        rx.vstack(
            rx.text(State.val),
            rx.upload(
                rx.button("Select File"),
                rx.text("Drag and drop files here or click to select files"),
                border="1px dotted black",
                padding="20em",
            ),
            rx.button(
                "Upload",
                on_click=lambda: State.handle_upload(rx.upload_files())
            ),
            rx.foreach(State.img, lambda img: rx.image(src=img)),
        )
    )


app = rx.App(state=State)
app.add_page(index)
app.compile()


```

### DIFF
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -323,20 +323,7 @@ export const uploadFiles = async (state, setResult, handler) => {
   }
 
   // Send the file to the server.
-  await axios.post(UPLOADURL, formdata, headers).then((response) => {
-    // Apply the delta and set the result.
-    const update = response.data;
-    applyDelta(state, update.delta);
-
-    // Set processing to false and return.
-    setResult({
-      state: state,
-      events: update.events,
-      final: true,
-      processing: false,
-    });
-  });
-
+  await axios.post(UPLOADURL, formdata, headers);
   return true;
 };
 
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -338,12 +338,11 @@ export const uploadFiles = async (state, setResult, handler) => {
           console.log(error.request);
         } else {
           // Something happened in setting up the request that triggered an Error
-          console.log('Error', error.message);
+          console.log(error.message);
         }
         return false;
       }
     )
-  //  return true;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


reflex/app.py
```diff
@@ -586,6 +590,8 @@ async def upload_file(files: List[UploadFile]):
             assert file.filename is not None
             file.filename = file.filename.split(":")[-1]
 
+        # get the current session ID
+        sid = app.session_map[token]
         # Get the state for the session.
         state = app.state_manager.get_state(token)
```
masenf, line 596: i thought the state knew the `sid` already...

```
sid = state.get_sid()
```

reflex/.templates/web/utils/state.js
```diff
@@ -323,20 +323,7 @@ export const uploadFiles = async (state, setResult, handler) => {
...
-    // Set processing to false and return.
-    setResult({
-      state: state,
-      events: update.events,
-      final: true,
-      processing: false,
-    });
-  });
-
+  await axios.post(UPLOADURL, formdata, headers);
```
masenf, line 326: i know we didn't have it before... but we should probably catch errors here 🙏

reflex/app.py
ElijahAhianyo: ah you're absolutely right, that should make the session mapping redundant.

masenf (approved):

masenf (approved):
Anything else on this one?

ElijahAhianyo:
> Anything else on this one?

Just rebased this on main, should be good to go

PR #1350, opened 2023-07-17
Added Custom Styles To Markdown

### LINKED ISSUE #1283: Support custom styles for rx.markdown
Currently the Markdown component cannot be custom-styled.

We should support custom styles similar to https://react.email/docs/components/markdown

bsarthak16:
Fixes #1283

**Changes Done**

- Added Custom Style to Markdown as asked in the issue

- Used https://react.email/docs/components/markdown for inspiration of changes as asked in the issue

### DIFF
--- a/reflex/components/typography/markdown.py
+++ b/reflex/components/typography/markdown.py
@@ -63,11 +63,11 @@ class Markdown(Component):
             ._render()
             .add_props(
                 components={
-                    "h1": "{({node, ...props}) => <Heading size='2xl' paddingY='0.5em' {...props} />}",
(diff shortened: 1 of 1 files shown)


masenf (changes requested):
I don't understand why these heading tags are given hardcoded weird colors... 🤔

The intention of #1283 is to allow the user to pass arbitrary styles for each element rendered as markdown. There should be some processing of a user-supplied style dict that maps the styles onto each component to be rendered.

User should be able to supply a dict like

```python
rx.markdown(
    markdown_content,
    markdown_style={
          "h1": { "color": "red" },
          "h2": { "color": "blue" },
          "codeInline": { "background": "grey" },
    },
)
```

Issue #1346, opened 2023-07-16
Set an event handler for a button so that it will call browser-side JavaScript code.


milochen0418:
**Describe the bug**

### Feature 

I know how to write simple html/javascript code and put them in the WebView of mobile 
so that iOS native apps and Android native apps can communicate to the webpage in the WebView. 

I am evaluating an outsourcing case to make iOS/Android mobile app. 
I will put the reflex web app in the WebView in the iOS/Android mobile app. 

- Scenario 1 
Still can show ID as QRCode from the native app when mobile's internet  is offline 
When mobile is in the offline status, there is a button in the web app, and the user can click it 
to trigger the function in the iOS/android native app to show the QRCode for some string.  
This string will be some special ID of a VIP member in the group.  
We hope they can still show their ID to check even if their mobile cannot connect to the internet.

- Scenario 2
mobile's internet is online in this scenario. 
When the user wants to scan QRCode, they can click the button in the web app. 
Click this button will call browser-side javascript code to trigger iOS/android native app
to show the camera to scan QRCode. 

When iOS/Android native app gets the scan result as a string from QRCode, 
it can send the string to the web app in the WebView by the API route. 

**To Reproduce**
 - Code/Link to Repo:

**Expected behavior**

**Screenshots**

**Specifics (please complete the following information):**
 - Python Version:
 - Reflex Version:
 - OS: 
 - Browser (Optional):
 
**Additional context**

picklelo:
We can definitely support this - just have to think of the correct API design
1) How should the user define their custom JS functions
2) How to reference those functions from an event trigger

Will think more about this, feel free to post any ideas in this thread.

milochen0418:
@picklelo 


### Final Result  (We expected from Idea )

Q1: How should the user define their custom JS functions
My answer is set all of your functions in `script_tags` when add_page. 

Q2: How to reference those functions from an event trigger
My answer is to use 
`yield rx.browser_js_call("{'__func__':'test_func', 'myname':'Milo', 'myaction':'Jump'  }")`
or `yield rx.browser_js_call(some_dict_obj )`





I did some experiments on the current code.
This is my plan. 

### experiments
I can set the script_tags attribute in add_page  
So that all of the functions we want to call can be set in the script 

In the script, I also put the following code to override the `window.alert("")` function 
```javascript
window.alert = function(msg){console.log(msg);};
```
When I use the following function in the event handler and trigger it
```python
yield rx.window_alert("ABCDE")
```
[...]

PR #1355, opened 2023-07-17
Expose Script component from next/script

### LINKED ISSUE #273: Allow custom JS code
The user should be able to give a string of custom JS code that will be compiled to the top of the page.

### LINKED ISSUE #357: Server side event to run Javascript and get return value
We are working on a new component to add customize script to compile output, but we also need a way to run Javascript code to access Cookie and LocalStorage ( in low-level, for more user friendly way, check #356 ).

Because run Javascript in client-side is not related to App state, we can place them in the package's namespace, for example, get the result of 1+1:

```python
result = pc.run_script("1 + 1")
```

Get Cookies string:

```python
cookies_string = pc.run_script("document.cookie")
```

masenf:
## Expose `Script`

The `Script` component allow inline or remotely loaded scripts as specified by the user. It provides event handlers that can call into the backend when the script is loaded/ready, or encounters an error. This component can be conditionally rendered to execute javascript in response to server side events. The `rx.client_side` special event is added to allow calling user-defined javascript functions in response to component events.

```python
import reflex as rx


class State(rx.State):
    """The app state."""
    should_log_to_console: bool = False

    def on_ready(self):
        print(f"READY (should_log_to_console={self.should_log_to_console})")


console_logging = """
function do_log(v) {
    console.log("hello javascript", v)
}"""


def index():
    return rx.fragment(
        rx.script(console_logging, on_ready=State.on_ready),
        rx.cond(
            State.should_log_to_console,
            rx.script("do_log('hey')"),
        ),
        rx.button("On", on_click=lambda: State.set_should_log_to_console(True)),
        rx.button("Off", on_click=lambda: State.set_should_log_to_console(False)),
        rx.button("Do it", on_click=rx.client_side("do_log(args)")),
    )


app = rx.App(state=State)
app.add_page(index)
app.compile()
```

## [Rename reflex.components.base.document.Script as NextScript](https://github.com/reflex-dev/reflex/commit/b6950e8a6edb6e3a116c60a5251bb75d03575eb8)

`NextScript` is a special component that includes the scripts needed for nextjs itself. Rename accordingly to disambiguate the `Script` component from `next/script` (which facilitates execution of user supplied javascript).

This changes `rx.script` from `reflex.components.base.link.ScriptTag` to `reflex.components.base.script.Script` because the existing plain `ScriptTag` cannot usefully appear in a component definition since it is just a bare `<script>` tag and nextjs doesn't know how/when to render it.

address #1346 
fix #357 
fix #273

### DIFF
--- a/reflex/components/base/script.py
+++ b/reflex/components/base/script.py
@@ -12,8 +12,11 @@ from reflex.vars import BaseVar, Var
 class Script(Component):
     """Next.js script component.
 
-    Note that this component differs from reflex.components.base.document.Script
+    Note that this component differs from reflex.components.base.document.NextScript
     in that it is intended for use with custom and user-defined scripts.
+
+    It also differs from reflex.components.base.link.ScriptTag, which is the plain
+    HTML <script> tag which does not work when rendering a component.
     """
 
     library = "next/script"
--- a/reflex/components/base/script.py
+++ b/reflex/components/base/script.py
@@ -2,10 +2,10 @@
 
 https://nextjs.org/docs/app/api-reference/components/script
 """
-from typing import Optional, Set
+from typing import Set
 
 from reflex.components.component import Component
-from reflex.event import EventChain, EventHandler
+from reflex.event import EventChain
 from reflex.vars import BaseVar, Var
 
 
@@ -21,19 +21,10 @@ class Script(Component):
     is_default = True
 
     # Required unless inline script is used
-    src: Optional[Var[str]] = None
+    src: Var[str]
 
     # When the script will execute: afterInteractive | beforeInteractive | lazyOnload
     strategy: Var[str] = "afterInteractive"  # type: ignore
-
-    # Execute code after the script has finished loading.
-    on_load: Optional[EventHandler] = None
-
-    # Execute code after the script has finished loading and every time the component is mounted.
-    on_ready: Optional[EventHandler] = None
-
-    # Execute code if the script fails to load.
-    on_error: Optional[EventHandler] = None
 
     @classmethod
     def create(cls, *children, **props) -> Component:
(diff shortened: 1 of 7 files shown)


reflex/components/base/script.py
```diff
@@ -0,0 +1,80 @@
...
+    Note that this component differs from reflex.components.base.document.Script
+    in that it is intended for use with custom and user-defined scripts.
+    """
+
+    library = "next/script"
+    tag = "Script"
+    is_default = True
+
+    # Required unless inline script is used
+    src: Optional[Var[str]] = None
```
picklelo, line 24: We don't need to use `Optional` or set the default value to `None`, as that is done automatically for all props.

Can just do:
```python
src: Var[str]
strategy: Var[str]
...
```

reflex/components/base/script.py
```diff
@@ -0,0 +1,80 @@
...
+    def get_triggers(self) -> Set[str]:
+        """Get the event triggers for the component.
+
+        Returns:
+            The event triggers.
+        """
+        return super().get_triggers() | {"on_load", "on_ready", "on_error"}
+
+
+def client_side(javascript_code) -> Var[EventChain]:
```
picklelo, line 68: Nice! So if someone defines a `rx.script` with some functions, do you know if those functions would automatically become available here?

picklelo (commented):
Looks good, just a comment on formatting the props

reflex/components/base/script.py
masenf: Yes that's exactly right. Either in-line or in a separate js file, all defined functions declared with rx.script can be used with rx.client_side as event handlers.

reflex/components/base/script.py
masenf: Okay thanks, good to know.

picklelo (approved):
looks good to me

PR #1360, opened 2023-07-18
Lendemor/fix datatable rendering

Lendemor:
### All Submissions:

### Type of change

### Changes To Core Features:

-  [x] Have you successfully ran tests with your changes locally?

Fix a rendering problem where using computed var as data and fixed columns didn't work.
Separated the logic between dataframe and native type.

### DIFF
--- a/reflex/components/datadisplay/datatable.py
+++ b/reflex/components/datadisplay/datatable.py
@@ -61,14 +61,12 @@ class DataTable(Gridjs):
         # render pandas dataframes.
         if isinstance(data, ComputedVar) and data.type_ == Any:
             raise ValueError(
-                "Annotation of the computed var assigned to the data field"
-                " should be provided."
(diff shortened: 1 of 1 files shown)


masenf (commented):
Seems like the `if types.is_dataframe(type(self.data)):` section isn't really doing anything anymore? is it still necessary?

Also the `if isinstance(self.columns, Var):` seems like it would always be true if `self.data` is a dataframe, and overwrite the `self.columns` assignment being made on line 104.

I think nested the conditionals might improve readability:

```python
if isinstance(self.data, Var)
    if types.is_dataframe(self.data.type_):
        ...
    else:
        ...
        if isinstance(self.columns, Var):
            ...
```

admittedly i'm unfamiliar with this code, but it doesn't look quite right to me, even if it does work, i think it could be simplified.

masenf (approved):
much nicer! 🎉

PR #1369, opened 2023-07-19
[Fixes #1351] print SQL execution statements if SQLALCHEMY_ECHO constant is set for debugging

### LINKED ISSUE #1351: Expose SQLALCHEMY_ECHO like sqlalchemy-flask
For debugging, it should be simple to set this common environment variable to allow creation of the engine with `echo=True`

SlamChillz:
Implements the exact behavior similar to `SQLALCHEMY_ECHO` config parameter in `flask-sqlalchemy`. Defaults to `False`. Ideally this shouldn't be set in `production` environment, so this constant is respected in `dev` mode only. This PR closes #1351

### DIFF
--- a/reflex/model.py
+++ b/reflex/model.py
@@ -40,9 +40,12 @@ def get_engine(url: Optional[str] = None):
         console.print(
             "[red]Database is not initialized, run [bold]reflex db init[/bold] first."
         )
+    echo_db_query = False
+    if conf.env == constants.Env.DEV and constants.SQLALCHEMY_ECHO:
+        echo_db_query = True
     return sqlmodel.create_engine(
         url,
-        echo=False,
+        echo=echo_db_query,
         connect_args={"check_same_thread": False} if conf.admin_dash else {},
     )
 
--- a/reflex/constants.py
+++ b/reflex/constants.py
@@ -124,6 +124,8 @@ RUN_BACKEND_PROD_WINDOWS = f"uvicorn --timeout-keep-alive {TIMEOUT}".split()
 # Socket.IO web server
 PING_INTERVAL = 25
 PING_TIMEOUT = 120
+# flag to make the engine print all the SQL statements it executes
+SQLALCHEMY_ECHO = get_value("SQLALCHEMY_ECHO", False, type_=bool)
 
 # Compiler variables.
 # The extension for compiled Javascript files.


reflex/model.py
```diff
@@ -40,9 +40,12 @@ def get_engine(url: Optional[str] = None):
         console.print(
             "[red]Database is not initialized, run [bold]reflex db init[/bold] first."
         )
+    echo_db_query = False
+    if conf.env == constants.Env.DEV and constants.SQLALCHEMY_ECHO:
```
masenf, line 44: in my testing the check for dev mode is not really working, and the reason is because `reflex run --env prod` sets dev mode in the main process, but then it spins up a second process to start the prod backend, and that process reads the config from the file (and does not realize that it should be in prod mode).

i'm not going to block on this, because our configuration needs to be refactored anyway, to protect this scenario, but just calling it out here as it was noticed during testing.

masenf (approved):
thanks for your contribution 🙏 we will merge it for 0.2.2 since 0.2.1 is about to release today and has already been thoroughly tested.

reflex/model.py
masenf: Filed #1373 to track this problem. No code change is needed here at this time.

SlamChillz: > in my testing the check for dev mode is not really working, and the reason is because `reflex run --env prod` sets prod mode in the main process, but then it spins up a second process to start the prod backend, and that process reads the config from the file (and does not realize that it should be in prod mode).
> 
> i'm not going to block on this, because our configuration needs to be refactored anyway, to protect this scenario, but just calling it out here as it was noticed during testing.

Thank You @masenf. I perfectly understand what you’ve highlighted. At the moment the configurations aren’t synchronised across all processes, reason for this issue.

SlamChillz: > Filed #1373 to track this problem. No code change is needed here at this time.

I know is more technical to fix, but I am willing to lay my hands on it. So, please kindly assign this issue to me.

masenf: @SlamChillz I left a note over in the other issue describing the requirements. I think if you comment on that issue I'll be able to assign you, but you're not showing up in my list of choices currently.

masenf (approved):
approving again to trigger integration test workflows

PR #1370, opened 2023-07-19
Synchronize Event Namespace

### LINKED ISSUE #1342: Websocket namespace routing fails when not root directory
**Describe the bug**
The websocket event namespaces in the frontend and backend can get out of sync when api_url is set to a non-root directory such as `https://www.domain.com/api/` even if everything is correctly proxied via nginx or similar. Note, this occurs AFTER the websocket connection is established and once an event occurs.

**To Reproduce**
- have the backend running at `https://www.domain.com/api` or 'http://localhost/api' or similar
- setup the config as such:

```
config = pc.Config(
...
    api_url="https://www.domain.com/api",
)
```
**Expected behavior**
The frontend and backend should be in sync in regards to websocket event namespace. Either the frontend needs changes to send `/event/` even if api_url is set to something non-standard or the backend should be changed to expect `api_path + /event/`

**Screenshots**
Logs showing the websocket connection can successfully establish, but once its established the requested event handler is out of sync with that is hardcoded into the `pc.App.__init()`.
```
DEBUG:    > HTTP/1.1 101 Switching Protocols
DEBUG:    > Upgrade: websocket
DEBUG:    > Connection: Upgrade
DEBUG:    > Sec-WebSocket-Accept: ktcaYTD1OHahZUTx095a4PE4AMc=
DEBUG:    > Sec-WebSocket-Extensions: permessage-deflate
DEBUG:    > date: Thu, 13 Jul 2023 03:16:43 GMT
DEBUG:    > server: uvicorn
INFO:     connection open
DEBUG:    = connection is OPEN
DEBUG:    > TEXT '0{"sid":"...","upgrades":[],"p...0,"pingInterval":25000}' [87 bytes]
DEBUG:    < TEXT '40/api/event,' [13 bytes]
DEBUG:    > TEXT '44/api/event,"Unable to connect"' [32 bytes]
DEBUG:    < CLOSE 1005 (no status code [internal]) [0 bytes]
DEBUG:    = connection is CLOSING
DEBUG:    > CLOSE 1005 (no status code [internal]) [0 bytes]
DEBUG:    x half-closing TCP connection
DEBUG:    = connection is CLOSED
INFO:     connection closed
```

**Additional context**

For my specific case, these changes seemed to work by changing the namespace on the backend, but likely could make some adjustments to the frontend to fix this alternatively:
``` 
# Configure app with main state class
app = pc.App(state=State)

# Fix for the FastAPI OpenAPI/docs page
app.api.root_path = "/api"

# Fix for websocket namespace
event_namespace = pc.app.EventNamespace("/api/event", app)
app.sio.register_namespace(event_namespace)
```

ElijahAhianyo:
The web socket event namespace is hardcoded as `/event`, this means if your `api_url` contains a path to a resource or location (eg. `http://example.com/api`) there's a desynchronization btn the event URL (which becomes `ws://example.com/api/event`) and the namespace for the web socket (which remains `/event`). This PR synchronizes the event URL and the namespace implicitly by inferring the namespace from the event URL. It also adds an attribute(`event_namespace`) to the `Config` class which can be overridden.

fixes #1342

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -97,6 +97,9 @@ class App(Base):
         Args:
             *args: Args to initialize the app with.
             **kwargs: Kwargs to initialize the app with.
+
+        Raises:
+            ValueError: If the event namespace is not provided in the config.
         """
         super().__init__(*args, **kwargs)
 
@@ -129,7 +132,10 @@ class App(Base):
         # Create the socket app. Note event endpoint constant replaces the default 'socket.io' path.
         self.socket_app = ASGIApp(self.sio, socketio_path="")
         namespace = config.get_event_namespace()
-        assert namespace
+
+        if not namespace:
+            raise ValueError("event namespace must be provided in the config.")
+
         # Create the event namespace and attach the main app. Not related to any paths.
         self.event_namespace = EventNamespace(namespace, self)
 
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -128,9 +128,10 @@ class App(Base):
 
         # Create the socket app. Note event endpoint constant replaces the default 'socket.io' path.
         self.socket_app = ASGIApp(self.sio, socketio_path="")
-
+        namespace = config.get_event_namespace()
+        assert namespace
         # Create the event namespace and attach the main app. Not related to any paths.
-        self.event_namespace = EventNamespace("/event", self)
+        self.event_namespace = EventNamespace(namespace, self)
 
         # Register the event namespace with the socket.
         self.sio.register_namespace(self.event_namespace)
--- a/reflex/config.py
+++ b/reflex/config.py
@@ -255,8 +255,12 @@ class Config(Base):
             except AttributeError:
                 pass
 
-    def get_event_namespace(self):
-        """Get the websocket event namespace."""
+    def get_event_namespace(self) -> Optional[str]:
+        """Get the websocket event namespace.
+
+        Returns:
+            The namespace for websocket.
+        """
         if self.event_namespace:
             return f'/{self.event_namespace.strip("/")}'
 
--- a/reflex/config.py
+++ b/reflex/config.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import importlib
 import os
+import re
 import sys
 import urllib.parse
 from typing import Any, Dict, List, Optional
@@ -206,6 +207,9 @@ class Config(Base):
     # Whether to enable or disable nextJS gzip compression.
     next_compression: bool = True
 
+    # The event namespace for ws connection
+    event_namespace: Optional[str] = constants.EVENT_NAMESPACE
+
     def __init__(self, *args, **kwargs):
         """Initialize the config values.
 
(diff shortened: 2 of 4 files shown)


reflex/app.py
```diff
@@ -128,9 +128,10 @@ def __init__(self, *args, **kwargs):
 
         # Create the socket app. Note event endpoint constant replaces the default 'socket.io' path.
         self.socket_app = ASGIApp(self.sio, socketio_path="")
-
+        namespace = config.get_event_namespace()
+        assert namespace
```
masenf, line 132: `assert` is ignored when running with `python -O` and should only be used in test code. This check should raise an exception instead.

```python
if not namespace:
    raise RuntimeError("event_namespace must be provided in the config")
```

reflex/config.py
```diff
@@ -251,6 +255,19 @@ def recompute_field_values(self):
...
+        """Get the websocket event namespace.
+
+        Returns:
+            The namespace for websocket.
+        """
+        if self.event_namespace:
+            return f'/{self.event_namespace.strip("/")}'
+
+        event_url = constants.Endpoint.EVENT.get_url()
+        match = re.match(r"^\w+://[^/]+(/.*)?$", event_url)
```
masenf, line 268: suggest the use of stdlib `urllib.parse.urlsplit` here to improve readability:

```python
>>> import urllib.parse
>>> urllib.parse.urlsplit("https://foo.bar.com/api/2/event")
SplitResult(scheme='https', netloc='foo.bar.com', path='/api/2/event', query='', fragment='')
>>> urllib.parse.urlsplit("wss://example.com/this/is/my/cool/api/event")
SplitResult(scheme='wss', netloc='example.com', path='/this/is/my/cool/api/event', query='', fragment='')
```

```python
def get_event_namespace(self) -> Optional[str]:
    ...
    urlparts = urllib.parse.urlsplit(event_url)
    return urlparts.path
```

tests/test_config.py
```diff
@@ -133,3 +133,27 @@ def test_get_value(monkeypatch, key, value, expected_value_type_in_config):
...
+@pytest.mark.parametrize(
+    "kwargs, expected",
+    [
+        ({"app_name": "test_app", "api_url": "http://example.com"}, "/event"),
+        ({"app_name": "test_app", "api_url": "http://example.com/api"}, "/api/event"),
+        ({"app_name": "test_app", "event_namespace": "/event"}, "/event"),
+        ({"app_name": "test_app", "event_namespace": "event"}, "/event"),
+    ],
+)
+def test_event_name_space(mocker, kwargs, expected):
```
masenf, line 147:
```suggestion
def test_event_namespace(mocker, kwargs, expected):
```

nit for grep-ability

masenf (changes requested):

tests/test_config.py
```diff
@@ -133,3 +133,27 @@ def test_get_value(monkeypatch, key, value, expected_value_type_in_config):
...
 
     assert isinstance(casted_value, expected_value_type_in_config)
+
+
+@pytest.mark.parametrize(
+    "kwargs, expected",
+    [
+        ({"app_name": "test_app", "api_url": "http://example.com"}, "/event"),
+        ({"app_name": "test_app", "api_url": "http://example.com/api"}, "/api/event"),
+        ({"app_name": "test_app", "event_namespace": "/event"}, "/event"),
```
masenf, line 143: suggest adding a test case where `event_namespace` has a trailing slash.

reflex/config.py
ElijahAhianyo: ah nice!

ElijahAhianyo: ah nice!

masenf (approved):

PR #1375, opened 2023-07-19
state.js: do not shadow outer `state` var when updating state

### LINKED ISSUE #1329: on_change event sent twice when on_key_down is also used
**Describe the bug**
When an `input` or `text_area` has both `on_change` and `on_key_down` handlers defined (even if they do _nothing_), typing into the component causes `on_change` to be sent twice.

**To Reproduce**
```python
import reflex as rx


class State(rx.State):
    def on_key_down(self, key):
        print(f"on_key_down: {key}")

    def on_change(self, value):
        print(f"on_change: {value}")


def index() -> rx.Component:
    return rx.vstack(
        rx.input(
            placeholder="Double on_change box",
            on_change=State.on_change,
            on_key_down=State.on_key_down,
        ),
        rx.input(placeholder="Single on_change box", on_change=State.on_change),
        padding_top="5%",
    )


app = rx.App(state=State)
app.add_page(index)
app.compile()
```

 - Python Version: 3.11.4
 - Reflex Version: 0.2.0
 - OS: macOS 13.4.1
 - Browser (Optional): firefox

masenf:
avoid double events, fix #1329

### DIFF
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -222,7 +222,7 @@ export const processEvent = async (
   const event = state.events.shift();
 
   // Set new events to avoid reprocessing the same event.
-  setState(state => ({ ...state, events: state.events }));
+  setState(currentState => ({ ...currentState, events: state.events }));
 
   // Process events with handlers via REST and all others via websockets.
   let eventSent = false;


picklelo (approved):
nice catch - do we need to fix this in any other areas? I know we call setState elsewhere also

masenf:
> nice catch - do we need to fix this in any other areas?

The other call sites seemed okay, because they were not first munging the local state variable.

PR #1379, opened 2023-07-19
ConnectionModal and ConnectionBanner cleanup

masenf:
# [state: save connectError instead of notConnected boolean](https://github.com/reflex-dev/reflex/commit/390ff6fd6b352955acb8573fa6dffb9bc777411d)

Allow connection banner code to access the error message and other details about the websocket error to provide more context to the user.

# [Add ConnectionModal for displaying connection errors](https://github.com/reflex-dev/reflex/commit/eb81e555340493f0e8248bd8a45672b4fc40abc8)

* expose `connection_error` var for downstream use
* expose `has_connection_error` var for downstream use
* expose `default_connection_error` as list[str | Var] for downstream use with
  other components or styles without duplicating the error message
* include API_URL in the error message so users can easily see what the frontend
  was trying to connect to

Create `rx.connection_modal` which takes over the page and definitely does NOT
give any illusion as to the page interactivity status.

# [default template: add rx.connection_modal()](https://github.com/reflex-dev/reflex/commit/7f562b2b61ccb8734caeca4535c856e7e7c22f12)

show off the connection status modal dialog in the default template to help users determine when their backend is disconnected.

fix https://github.com/reflex-dev/reflex/issues/1364

image


reflex/.templates/apps/default/default.py
```diff
@@ -15,6 +15,7 @@ class State(rx.State):
 
 def index() -> rx.Component:
     return rx.fragment(
+        rx.connection_modal(),  # displays error when backend cannot connect
```
picklelo, line 18: Should we enable this in all apps by default? And have them specify if they want this in the config?

picklelo (commented):
Code looks good, just wondering what the best way for the user to configure this would be. I'm thinking we should have this enabled by default, and let them turn it off.

masenf:
Rebased on the new refactored event loop.

reflex/app.py
```diff
@@ -269,6 +275,31 @@ def add_middleware(self, middleware: Middleware, index: Optional[int] = None):
         else:
             self.middleware.insert(index, middleware)
 
+    @staticmethod
+    def _generate_component(component: Component | ComponentCallable) -> Component:
```
picklelo, line 279: We can move this out of this class also, to a utils file

picklelo (approved):

reflex/app.py
masenf: I'll do it for the client side routing PR, since I also need to expand the 404 page over there before I pass it to wait_for_redir

reflex/components/overlay/banner.py
```diff
@@ -1,11 +1,41 @@
...
+    Returns:
+        The default connection error message.
+    """
+    from reflex.config import get_config
+
+    return [
+        "Cannot connect to server: ",
+        connection_error,
+        ". Check if server is reachable at ",
+        get_config().api_url or "<API_URL not set>",
```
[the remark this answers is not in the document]
masenf: Yeah and maybe link it to /ping?

PR #1382, opened 2023-07-19
Fix sitemap output dir

picklelo:
Fix the output dir of the sitemap


Alek99 (dismissed):
Waiting for changes as discussed

Alek99 (approved):
This works for me.

- Note we need to add in the docs if you want a sitemap url a deploy_url needs to be provided.

PR #1383, opened 2023-07-20
Disable fully controlled Input and TextArea

masenf:
Due to event ordering issues, fully controlled inputs do not work without some client-side buffer to maintain the value while the state is settling.

To avoid user frustration, fail during compile time when attemping to use a fully controlled input, and suggest the use of debounce_input if such a control is desired, to avoid dropped events and broken typing.

Add additional integration test for one-way bound inputs.

### DIFF
--- a/integration/test_input.py
+++ b/integration/test_input.py
@@ -20,12 +20,17 @@ def FullyControlledInput():
 
     @app.add_page
     def index():
-        return rx.debounce_input(
-            rx.input(
+        return rx.fragment(
+            rx.debounce_input(
+                rx.input(
+                    on_change=State.set_text, id="debounce_input_input"  # type: ignore
+                ),
                 value=State.text,
-                on_change=State.set_text,  # type: ignore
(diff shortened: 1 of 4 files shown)


picklelo (approved):
Sorry for the delay on this one, looks good. In a follow up, should we rename `rx.debounce_input` to `rx.debounce` and have it work with other components such as slider also?

masenf:
> should we rename `rx.debounce_input` to `rx.debounce` and have it work with other components such as slider also

I think already works with slider, checkbox, and other components that have `value` and `on_change` props that work how you would expect.

As for the name, I went with `debounce_input` because the library is `react-debounce-input` and the component name is `DebounceInput`, so I didn't want to needlessly rename it (even though the python module is `reflex.components.forms.debounce`). I'm fine with changing the name if you think it's warranted, but that was my reasoning.

Working on the docs for this component today, so i'll include the slider example and other more exotic ones as well.

PR #1389, opened 2023-07-20
Fix foreach index type

picklelo:
Make sure the type of the index is `int` to make type checking happy.

### DIFF
--- a/reflex/components/tags/iter_tag.py
+++ b/reflex/components/tags/iter_tag.py
@@ -5,7 +5,7 @@ import inspect
 from typing import TYPE_CHECKING, Callable, List
 
 from reflex.components.tags.tag import Tag
-from reflex.vars import Var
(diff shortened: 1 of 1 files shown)


reflex/components/tags/iter_tag.py
```diff
@@ -30,9 +30,11 @@ def get_index_var() -> Var:
...
-        assert index is not None
-        return index
+        return BaseVar(
+            name=INDEX_VAR,
+            type_=int,
+            is_local=True,
+        )
 
     @staticmethod
     def get_index_var_arg() -> Var:
```
Lendemor, line 40: Shouldn't we change the type annotation if we are now returning BaseVar instead of Var?

masenf (approved):
🚢

masenf (approved):
integrate me!

PR #1399, opened 2023-07-21
Allow custom Tailwind content

### LINKED ISSUE #1378: Tailwind support custom contents
Currently when passing in `contents` to the tailwind config it throws the error

```
    dict(**config.tailwind, content=constants.TAILWIND_CONTENT)
TypeError: dict() got multiple values for keyword argument 'content'
```

We should allow this to be overridden (or merged with the default content)

kbrgl:
Closes #1378.

If a user supplies a content key, then it replaces the default. I think this works better than merging the user-supplied content with the default content. If a user wants to merge the default content as well, they can import `constants.TAILWIND_CONTENT` and add that manually. This keeps the abstraction barrier intact while making the API explicit and predictable.

Here’s how it looks for a user to merge the default Tailwind content with their own contents:

```python
TailwindConfig(
    ...,
    tailwind={
        "content": rx.constants.TAILWIND_CONTENT + ["custom_content.py"],
    },
)
```

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -503,14 +503,11 @@ class App(Base):
         compile_results.append(compiler.compile_theme(self.style))
 
         # Compile the Tailwind config.
-        config.tailwind["content"] = config.tailwind.get(
-            "content", constants.TAILWIND_CONTENT
-        )
-        compile_results.append(
-            compiler.compile_tailwind(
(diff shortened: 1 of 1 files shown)


picklelo (approved):
nice, thanks for fixing

PR #1404, opened 2023-07-22
feature: Auto install node by nvm on Linux

### LINKED ISSUE #1381: Automatically install NodeJS
Just like we install Bun for the user, we should automatically install NodeJS if it is not available. This should make the onboarding/setup much easier and remove one prerequisite.

TaiJuWu:
### All Submissions:

- [X] Have you followed the guidelines stated in [CONTRIBUTING.md](https://github.com/pynecone-io/pynecone/blob/main/CONTRIBUTING.md) file?
- [X] Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/pynecone-io/pynecone/pulls ) for the desired changed?

### Type of change

### New Feature Submission:

### Changes To Core Features:

Auto install node by using nvm package but the feature **only support on Linux right now**.
    b. Describe your changes.

related #1381

### DIFF
--- a/reflex/utils/prerequisites.py
+++ b/reflex/utils/prerequisites.py
@@ -281,7 +281,7 @@ def install_node():
         FileNotFoundError: if unzip or curl packages are not found.
         Exit: if installation failed
     """
-    if platform.system() !="Windows":
+    if platform.system() != "Windows":
         # Only install if bun is not already installed.
         console.log("Installing nvm...")
 
--- a/reflex/utils/prerequisites.py
+++ b/reflex/utils/prerequisites.py
@@ -73,17 +73,8 @@ def get_package_manager() -> str:
 
     Raises:
         FileNotFoundError: If bun or npm is not installed.
-        Exit: If the app directory is invalid.
-
     """
     config = get_config()
-
-    # Check that the node version is valid.
-    if not check_node_version():
-        console.print(
-            f"[red]Node.js version {constants.MIN_NODE_VERSION} or higher is required to run Reflex."
-        )
-        raise typer.Exit()
 
     # On Windows, we use npm instead of bun.
     if platform.system() == "Windows" or config.disable_bun:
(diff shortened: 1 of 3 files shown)


TaiJuWu:
I can't remove nvm package on my localhost.
Please check the feature available before merge.

picklelo (commented):
This is awesome!, With a few tweaks I got it working on my Mac also. I will merge this later this weekend and include my follow-on.

picklelo (approved):
Nice, I will add a follow up to enable on Mac also. I also created #1407 to clean up where we store the dependencies, let me know what you think.

reflex/utils/prerequisites.py
```diff
@@ -276,6 +267,47 @@ def remove_existing_bun_installation():
...
+    if platform.system() == "Linux":
+        # Only install if bun is not already installed.
+        console.log("Installing nvm...")
+
+        # Check if curl is installed
+        curl_path = path_ops.which("curl")
+        if curl_path is None:
+            raise FileNotFoundError("Reflex requires curl to be installed.")
+
+        result = subprocess.run(constants.INSTALL_NVM, shell=True)
```
ElijahAhianyo, line 293: we should check here if nvm is already installed to avoid duplicate installation.

ElijahAhianyo: Also, there are some edge cases I think arent being handled here:

- If nvm is already installed, there might be a chance that there could be multiple versions of node installed. We need to check if there's a supported node version(`nvm ls`). If there is, we should switch to that version(`nvm use <version>`). If not, we can go ahead and install the required version and switch to that version(very important as the version after installing still remains the previous version until switched)

- We also need a node version check on `reflex run` as the app will still compile even if the node version does not meet the requirements

reflex/utils/prerequisites.py
```diff
@@ -276,6 +267,47 @@ def remove_existing_bun_installation():
...
+
+        if result.returncode != 0:
+            raise typer.Exit(code=result.returncode)
+
+        console.log("Installing node...")
+        result = subprocess.run(constants.INSTALL_NODE, shell=True)
```
ElijahAhianyo, line 299: After installing node, it would be nice to reload the shell configuration files (`.zshrc`, `.bashrc`) as the node command will not be immediately available until the shell is restarted. This means reflex will always try to install nvm and node on `reflex init` until shell is restarted

reflex/utils/prerequisites.py
```diff
@@ -276,6 +267,47 @@ def remove_existing_bun_installation():
...
+def install_node():
+    """Install nvm and nodejs onto the user's system.
+
+
+    Raises:
+        FileNotFoundError: if unzip or curl packages are not found.
+        Exit: if installation failed
+    """
+    if platform.system() == "Linux":
+        # Only install if bun is not already installed.
```
ElijahAhianyo, line 285: nit: replace bun with node

reflex/utils/prerequisites.py
```diff
@@ -276,6 +267,47 @@ def remove_existing_bun_installation():
         path_ops.rm(os.path.expandvars(constants.BUN_ROOT_PATH))
 
 
+def validate_and_install_node():
+    """Validate nodejs have install or not."""
+    if not check_node_version():
+        install_node()
```
ElijahAhianyo, line 273: Just like we do for bun, we should also prompt users whether or not to install node

reflex/utils/prerequisites.py
picklelo: In a follow up we will move everything to a `~/.reflex` folder so the user won't have to manage their own node or bun dependencies.

picklelo (approved):
I think you just need to run `black` to format the code, we can merge this in then!

picklelo (approved):

Issue #1254, opened 2023-06-27
env=pc.Env.PROD, no --env, no env_path, pc run as dev


evandeng2009:
**Describe the bug**
env=pc.Env.PROD, no --env, no env_path or .env, pc run still as dev, which means env=pc.Env.PROD is not effective.

**To Reproduce**
 - Code/Link to Repo:

**Expected behavior**

**Screenshots**

** Specifics (please complete the following information):**
 - Python Version:3.11
 - Reflex Version:0.1.34
 - OS: Windows
 - Browser (Optional):
 
**Additional context**

PR #1415, opened 2023-07-24
[Fixes #1373] exact config values set in the main process can be seen and used by the subprocesses

### LINKED ISSUE #1373: [Config Refactor] need a way to set_config in the main process and apply config values to subprocesses
**Describe the bug**
When the reflex cli starts the app, it sets some config values by calling `get_config()` and setting the values directly in memory, then it spins up two separate processes, one for the frontend and one for the backend. These subprocesses cannot generally see the updated config values, unless they happen to also be in recognizable environment variables.

**To Reproduce**

```python
from rxconfig import config

import reflex as rx


class State(rx.State):
    pass


def index() -> rx.Component:
    return rx.fragment(
        rx.vstack(
            rx.heading("I'm running in ", config.env, " mode!", font_size="2em"),
            padding_top="10%",
        ),
    )


app = rx.App(state=State)
app.add_page(index)
app.compile()
```

**Expected behavior**
I expect the app to be able to read the actual `config.env` value.

Instead, it says "dev" unless `rxconfig.py` is updated with `env=Env.PROD`.

Even when running with `reflex run --env prod`, it still says "dev mode" because the app backend is running in a subprocess that reads the config file directly, so setting the env mode in the main process is invisible to the subprocess.

SlamChillz:
This refactor synchronizes configuration values between the main and sub processes. Also, as requested by @masenf , variables starting with `RX_` now takes precedence i.e `API_URL` works today and will still work, but if `RX_API_URL` is set, then `API_URL` will be ignored. This PR closes #1373

### DIFF
--- a/reflex/config.py
+++ b/reflex/config.py
@@ -223,7 +223,15 @@ class Config(Base):
 
         super().__init__(*args, **kwargs)
 
-        # set overriden class attribute values as os env variables to avoid losing them
+        # Load config values from env_path and override os_env if override_os_env is True
+        if self.env_path is not None and os.path.isfile(self.env_path):
+            load_dotenv(self.env_path, override=self.override_os_envs)  # type: ignore
+            # Recompute constants after loading env variables
+            importlib.reload(constants)
+            # Recompute instance attributes
+            self.recompute_field_values()
+
+        # Set class attributes not present in os_env in os_env
         for key, value in dict(self).items():
             key = key.upper()
             if (
@@ -231,16 +239,16 @@ class Config(Base):
                 or key in os.environ
                 or (value is None and key != "DB_URL")
             ):
+                if key in os.environ and key == "ENV":
+                    env_values: Dict[str, constants.Env] = {
+                        "Env.DEV": constants.Env.DEV,
+                        "Env.PROD": constants.Env.PROD,
+                    }
+                    setattr(self, key.lower(), env_values[str(os.environ.get(key))])
                 continue
             os.environ[key] = str(value)
-
-        # Avoid overriding if env_path is not provided or does not exist
-        if self.env_path is not None and os.path.isfile(self.env_path):
-            load_dotenv(self.env_path, override=self.override_os_envs)  # type: ignore
-            # Recompute constants after loading env variables
-            importlib.reload(constants)
-            # Recompute instance attributes
-            self.recompute_field_values()
+        # Recompute constants to use the new values in os env
+        importlib.reload(constants)
 
     def recompute_field_values(self):
         """Recompute instance field values to reflect new values after reloading
--- a/reflex/reflex.py
+++ b/reflex/reflex.py
@@ -68,7 +68,7 @@ def init(
 @cli.command()
 def run(
     env: constants.Env = typer.Option(
-        get_config().env, help="The environment to run the app in."
+        constants.Env.DEV, help="The environment to run the app in."
     ),
     frontend: bool = typer.Option(
         False, "--frontend-only", help="Execute only frontend."
@@ -88,7 +88,9 @@ def run(
         )
     # Set ports as os env variables to take precedence over config and
     # .env variables(if override_os_envs flag in config is set to False).
+    # print(env.value)
     build.set_os_env(
+        env=str(env),
         frontend_port=frontend_port,
         backend_port=backend_port,
         backend_host=backend_host,
--- a/reflex/constants.py
+++ b/reflex/constants.py
@@ -26,7 +26,7 @@ def get_value(key: str, default: Any = None, type_: Type = str) -> Type:
     Returns:
         the value of the constant in its designated type
     """
-    value = os.getenv(key, default)
+    value = os.getenv(f"RX_{key}", os.getenv(key, default))
     try:
         if value and type_ != str:
             value = eval(value)


reflex/config.py
```diff
@@ -223,24 +223,32 @@ def __init__(self, *args, **kwargs):
...
             key = key.upper()
             if (
                 key.startswith("_")
                 or key in os.environ
                 or (value is None and key != "DB_URL")
             ):
+                if key in os.environ and key == "ENV":
+                    env_values: Dict[str, constants.Env] = {
+                        "Env.DEV": constants.Env.DEV,
+                        "Env.PROD": constants.Env.PROD,
```
masenf, line 245: i think we should just expect the user to pass "dev" or "prod" here (the value of the enum item), not the dot-qualified string.

reflex/config.py
```diff
@@ -223,24 +223,32 @@ def __init__(self, *args, **kwargs):
...
+        # Load config values from env_path and override os_env if override_os_env is True
+        if self.env_path is not None and os.path.isfile(self.env_path):
+            load_dotenv(self.env_path, override=self.override_os_envs)  # type: ignore
+            # Recompute constants after loading env variables
+            importlib.reload(constants)
+            # Recompute instance attributes
+            self.recompute_field_values()
+
+        # Set class attributes not present in os_env in os_env
         for key, value in dict(self).items():
```
masenf, line 235: i dont think this code needs to set values in the environment that were actually provided in the initializer of the config.

* if they come from the `env_path` file, then they will already be in the environment, so no need to set
* if they are set after `Config` initialization then they will need to be applied to the environment in `__setattr__`
* otherwise the values are provided when initializing `Config`, and subsequent `get_config` calls in a subprocess will import `rxconfig.py`

The only values that really need to be put into the environment are those set _after_ initialization.

reflex/reflex.py
```diff
@@ -68,7 +68,7 @@ def init(
 @cli.command()
 def run(
     env: constants.Env = typer.Option(
-        get_config().env, help="The environment to run the app in."
+        constants.Env.DEV, help="The environment to run the app in."
```
masenf, line 71: This change regresses #1254

masenf (changes requested):
i'm running into some issues just running the config tests:

```console
poetry run pytest tests/test_config.py
```

This seems to expose some assumptions in the proposed change that certain variables will be set external (and other tests apparently do set them, but we cannot rely on this always being the case).

I also wrote an additional test case that sets values on the config, and then reads them from a subprocess, none of these are passing with this change:

```python
def get_config_subprocess(key) -> Any:
    """Subprocess function for use with multiprocessing.
    
    Args:
        key: The config key to look up.
        
    Returns:
        The value of the config key."""
    return getattr(get_config(), key)


@pytest.mark.parametrize(
    ("config_key", "config_value"),
    [
        ("username", "foo"),
        ("telemetry_enabled", False),
        ("api_url", "https://foo.bar.baz"),
        ("timeout", 42),
    ],
)
def test_config_subprocess(config_key, config_value):
    """Test that config values set in parent process are recognized in child process.
    
    Args:
        config_key: the key to set on Config
        config_value: the value to set on Config (should be same in subprocess)
    """

    # set the value in the parent process
    setattr(get_config(), config_key, config_value)
    with multiprocessing.Pool() as pool:
        # it should be the same in the child process
        assert pool.apply(get_config_subprocess, args=(config_key,)) == config_value
```

The config refactor needs to generally support setting any (serializable) config value in the main process and then having that value readable automatically in a subprocess.

I'll be on vacation the next week, but I'll try to keep some attention on this PR. Thanks for submitting.

SlamChillz:
> i'm running into some issues just running the config tests:
> 
> ```
> poetry run pytest tests/test_config.py
> ```
> 
> This seems to expose some assumptions in the proposed change that certain variables will be set external (and other tests apparently do set them, but we cannot rely on this always being the case).
> 
> I also wrote an additional test case that sets values on the config, and then reads them from a subprocess, none of these are passing with this change:
> 
> ```python
> def get_config_subprocess(key) -> Any:
>     """Subprocess function for use with multiprocessing.
>     
>     Args:
>         key: The config key to look up.
>         
>     Returns:
>         The value of the config key."""
>     return getattr(get_config(), key)
> 
> 
> @pytest.mark.parametrize(
>     ("config_key", "config_value"),
>     [
>         ("username", "foo"),
>         ("telemetry_enabled", False),
>         ("api_url", "https://foo.bar.baz"),
>         ("timeout", 42),
>     ],
> )
> def test_config_subprocess(config_key, config_value):
>     """Test that config values set in parent process are recognized in child process.
>     
>     Args:
>         config_key: the key to set on Config
>         config_value: the value to set on Config (should be same in subprocess)
>     """
> 
>     # set the value in the parent process
>     setattr(get_config(), config_key, config_value)
>     with multiprocessing.Pool() as pool:
>         # it should be the same in the child process
>         assert pool.apply(get_config_subprocess, args=(config_key,)) == config_value
> ```
> 
> The config refactor needs to generally support setting any (serializable) config value in the main process and then having that value readable automatically in a subprocess.
> 
> I'll be on vacation the next week, but I'll try to keep some attention on this PR. Thanks for submitting.

This is quite interesting. I will take a closer look at it again.

reflex/config.py
SlamChillz: > i think we should just expect the user to pass "dev" or "prod" here (the value of the enum item), not the dot-qualified string.

Right, will do

reflex/config.py
SlamChillz: > i dont think this code needs to set values in the environment that were actually provided in the initializer of the config.
> 
> * if they come from the `env_path` file, then they will already be in the environment, so no need to set
> * if they are set after `Config` initialization then they will need to be applied to the environment in `__setattr__`
> * otherwise the values are provided when initializing `Config`, and subsequent `get_config` calls in a subprocess will import `rxconfig.py`
> 
> The only values that really need to be put into the environment are those set _after_ initialization.

oh! I did not quite think of it like this. I am surely going to attempt to correct this.

masenf:
closing for now, since the config/env var handling has been refactored since this patch was proposed. Please feel free to rebase and re-open if you feel this patch is still needed.

PR #1437, opened 2023-07-27
update templates to remove deprecation warning

Lendemor:
follow up to #1361 

Remove the `state` `kwargs` in the templates so that new Apps do not throw warning


ElijahAhianyo (approved):
Nice

PR #1447, opened 2023-07-28
update behaviour for wrong state passed as argument of rx.App

Lendemor:
### All Submissions:

### Type of change
- [x] Non-breaking internal behavior change.

### Description

Change the warning printed in case of wrong State passed to `rx.App`

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -108,6 +108,7 @@ class App(Base):
         """
         super().__init__(*args, **kwargs)
         state_subclasses = State.__subclasses__()
+        inferred_state = state_subclasses[-1]
         is_testing_env = constants.PYTEST_CURRENT_TEST in os.environ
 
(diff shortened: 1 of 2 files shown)


reflex/app.py
```diff
@@ -108,6 +108,7 @@ def __init__(self, *args, **kwargs):
         """
         super().__init__(*args, **kwargs)
         state_subclasses = State.__subclasses__()
+        inferred_state = state_subclasses[-1]
```
picklelo, line 111: Why do we index `-1` here? Is it because of the `DefaultState` class? I think we should do index `1` in that case. Also, we should throw an exception if `len(state_subclasses) > 1` (or 2 if that's the expected)

picklelo: Nvm, we have the multiple subclasses check right below, missed that

picklelo (approved):
looks good!

PR #1454, opened 2023-07-28
Improve prerequisites check

picklelo:
Make sure the app is initialized before running `reflex run` or `reflex export` commands.


Alek99 (approved):
lgtm

PR #1455, opened 2023-07-28
CI: factor out env setup as composite action

jackie-pc:
Main goals:
- Factor out "setup python" / "install poetry" // "poetry install" into a single place (rather than duplicated in every job.  The mechanic here is [Composite Actions](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action).
- This new composite action `setup_build_env` will have a refreshed caching implementation. It should be correct as of right now - if caching doesn't work in future, we can focus on this one place to fix.

* Caching - cache poetry installation, cache project deps installed using "poetry install" (pyproject.toml format deps only).  Note this matches existing caching behavior.  I.e. if we are integration testing against the counter app, we don't cache the latter's requirements correctly today anyway.  Fixing this can be a future exercise.

Why now?
We want to beef up CI infra in various dimensions in near future.  Such as including Windows, MacOS in a first class way, or adding UI based tests, etc.  We need to deduplicate some code now to keep things more maintainable over time.

### DIFF
--- a/.github/actions/setup_build_env/action.yml
+++ b/.github/actions/setup_build_env/action.yml
@@ -1,0 +1,87 @@
+# Assumes:
+# - CWD is the root directory of your project (e.g. reflex/)
+# - You have a poetry.lock file in the root directory of your project
+# - You have a pyproject.toml file in the root directory of your project
+name: 'Setup Reflex build environment'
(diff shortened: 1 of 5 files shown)


picklelo (approved):
lgtm

PR #1457, opened 2023-07-28
CI: separate pre-commit github action job

jackie-pc:
Depends on #[1455](https://github.com/reflex-dev/reflex/pull/1455)

Pre-commit does not need to run in a matrix or on multiple OSes.
    
But unit tests DO. Separate it out. We will be reusing the unit test workflow (`build.yml`) for multiple OSes soon.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,28 @@
+name: pre-commit
+
+on:
+  pull_request:
+    branches: [main] # TODO remove jackie-py-setup
+  push:
+    # Note even though this job is called "pre-commit" and runs "pre-commit", this job will run
+    # also POST-commit on main also!  In case there are mishandled merge conflicts / bad auto-resolves
+    # when merging into main branch.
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,28 @@
...
+          # running vs. one version of Python is OK
+          # i.e. ruff, black, etc.
+          python-version: 3.11
+          run-poetry-install: true
+          shell: bash
+          create-venv-at-path: .venv
+      # TODO pre-commit related stuff can be cached too (not a bottleneck yet)
+      - run: |
+          poetry run pip install pre-commit
+          poetry run pre-commit run --all-files
```
jackie-pc, line 28: Note: here we will run the same stuff as developers would run locally when they commit code, so respect [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

picklelo (approved):
awesome, looks good

PR #1458, opened 2023-07-28
Remove curl and parallelize node/bun install

picklelo:
Use `httpx` to install nvm and bun instead of curl, removing one dependency. Also parallelize the process to install nvm/bun during `reflex init`

### DIFF
--- a/reflex/utils/prerequisites.py
+++ b/reflex/utils/prerequisites.py
@@ -314,7 +314,14 @@ def install_node():
 
     # Install node.
     print("installing node")
-    result = subprocess.run(["bash", "-c", f". {constants.NVM_DIR}/nvm.sh && nvm install {constants.NODE_VERSION}"], env=env)
+    result = subprocess.run(
+        [
+            "bash",
+            "-c",
+            f". {constants.NVM_DIR}/nvm.sh && nvm install {constants.NODE_VERSION}",
+        ],
+        env=env,
+    )
     if result.returncode != 0:
         raise typer.Exit(code=result.returncode)
 
--- a/reflex/utils/prerequisites.py
+++ b/reflex/utils/prerequisites.py
@@ -9,12 +9,14 @@ import platform
 import re
 import subprocess
 import sys
+import tempfile
 from datetime import datetime
 from fileinput import FileInput
 from pathlib import Path
 from types import ModuleType
 from typing import Optional
 
+import httpx
 import typer
 from alembic.util.exc import CommandError
 from packaging import version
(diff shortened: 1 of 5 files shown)


jackie-pc:
Possible to base this PR on the other "package.json" PR, so we see what changed?

reflex/utils/prerequisites.py
```diff
@@ -319,27 +336,35 @@ def install_bun():
...
-            raise FileNotFoundError("Reflex requires unzip to be installed.")
-
-        result = subprocess.run(constants.INSTALL_BUN, shell=True)
+    # Skip if bun is already installed.
+    if os.path.exists(constants.BUN_PATH):
+        return
 
-        if result.returncode != 0:
-            raise typer.Exit(code=result.returncode)
+    # Check if unzip is installed
```
jackie-pc, line 345: We should create an issue to remove unzip as a requirement as well.

reflex/utils/prerequisites.py
```diff
@@ -319,27 +336,35 @@ def install_bun():
...
+    unzip_path = path_ops.which("unzip")
+    if unzip_path is None:
+        raise FileNotFoundError("Reflex requires unzip to be installed.")
+
+    # Get the bun install script.
+    response = httpx.get(constants.BUN_INSTALL_URL)
```
jackie-pc, line 351: We can potentially dedupe this further.... `def download_and_run()`:  same for both nvm and bun.

reflex/utils/prerequisites.py
```diff
@@ -281,31 +284,45 @@ def install_node():
...
-    console.log("Installing node...")
-    result = subprocess.run(constants.INSTALL_NODE, shell=True)
-
+    # Install node.
+    print("installing node")
+    result = subprocess.run(
+        [
+            "bash",
+            "-c",
+            f". {constants.NVM_DIR}/nvm.sh && nvm install {constants.NODE_VERSION}",
```
jackie-pc, line 322: Add comment that we have to do this "bash -c" business because we need to source in `. nvm.sh`?

jackie-pc:
We can verify curl is gone as a dependency by doing an `apt remove` [here](https://github.com/reflex-dev/reflex/blob/afb15109db0f0e3b040a88b5a37f49b00f6dd699/integration/init-test/Dockerfile#L18C5-L18C5).  Or we can just not use curl (e.g. by using wget instead, etc) within the Dockerfile itself.

It's fine to also file a new issue for this part IMO.  Benefit is mainly to ensure we catch any future addition of this dependency and that there are no other indirect deps on curl (e.g. within nvm.sh, etc)

reflex/utils/prerequisites.py
picklelo: Agreed, created this ticket #1459

reflex/utils/prerequisites.py
picklelo: added a comment

reflex/utils/prerequisites.py
picklelo: Great idea, factored this out to a function

jackie-pc (dismissed):

Alek99 (approved):

PR #1460, opened 2023-07-28
Full OS matrix builds for unit and integration tests (Linux, Mac, Windows)

jackie-pc:
This PR enables CI across all three supported OSes: Linux, Mac, Windows.  For Mac and Windows, this represents a 0 to 1 improvement in automated test coverage.

It is very important to minimize (or avoid) any code duplication in order to support all three OSes.  Most code changes in this PR were made with this goal in mind.  E.g.

* Use Github Actions OS matrix builds, running the SAME workflow definition across all OSes.
* Make whatever changes necessary to make the point above possible.  E.g. leverage Window's "Git Bash" shell, rewriting certain test utils in Python (instead of from POSIX-focused shell tools).

In the new world, workflows have been refactored as follows:
* `integration_tests.yml` workflow: this includes OS matrix jobs for "counter" and "reflex-web" integration tests.
* `unit_tests.yml` workflow: this includes a single OS matrix job for running unit tests.

There were also a couple of drive-by bug fixes.  They came up only as a side-effect of debugging Windows test failures.  The bugs and their fixes were obvious.  However it's unclear yet **how** exactly they affected the Windows tests in particular.  Either way, fixing them seems to correlate with getting Windows build green.

Note in terms of speed, running all these tests in Windows are significantly slower than on Linux and Mac.  Most of the perf penalty is likely related to general known poor perf of Windows VMs in Github Actions.  Not much we can do about this.  We do not run WSL (might be faster), and that is by design - we *want* to test in a pure non-WSL Windows OS.   If we want to include WSL coverage in future, it should be "in addition to", and not instead of vanilla Windows coverage.


.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,99 @@
+name: integration-tests
+
+on:
+  push:
+    branches: [ main ]
+    paths-ignore:
+      - '**/*.md'
```
jackie-pc, line 7: Let's not run tests on docs.

If we need any linting on docs specifically, we should create a separate workflow with different trigger conditions (e.g. if *.md's changed, run "md-lint" workflow)

masenf (approved):
ship it broken! 🚢 💥

picklelo (approved):

PR #1462, opened 2023-07-29
Add unified logging

picklelo:
Add a unified logger that respects the initial log level.

### DIFF
--- a/reflex/constants.py
+++ b/reflex/constants.py
@@ -1,4 +1,5 @@
 """Constants used throughout the package."""
+from __future__ import annotations
 
 import os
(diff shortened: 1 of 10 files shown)


Alek99 (dismissed):
Lgtm tested and different levels work

(Looks like pytest isn't passing)

Alek99 (commented):
Pytest

reflex/constants.py
Alek99, line 1: Error, Critical sre hiding start up displays

Alek99 (approved):
✅

Issue #1285, opened 2023-06-30
`reflex export` exit code and log messages are inaccurate


masenf:
`reflex export` reports that backend and frontend were exported unconditionally; even when `--backend-only` or `--frontend-only` are specified, and even when the export process itself fails. The command even returns exit code 0, despite the export failing.

martinxu9:
tracking this internally as REF-264

PR #1468, opened 2023-07-31
[WIP] Clean up config

picklelo:
Clean up config variables and CLI commands.

### DIFF
--- a/reflex/reflex.py
+++ b/reflex/reflex.py
@@ -97,10 +97,10 @@ def run(
         constants.Env.DEV, help="The environment to run the app in."
     ),
     frontend_only: bool = typer.Option(
-        False, "frontend-only", help="Execute only frontend."
+        False, "--frontend-only", help="Execute only frontend.", show_default=False
     ),
     backend_only: bool = typer.Option(
-        False, "backend-only", help="Execute only backend."
+        False, "--backend-only", help="Execute only backend.", show_default=False
     ),
     frontend_port: str = typer.Option(
         constants.FRONTEND_PORT,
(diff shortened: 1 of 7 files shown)


reflex/reflex.py
```diff
@@ -234,50 +187,76 @@ def export(
...
-        zip=zipping,
+        backend=export_backend,
+        frontend=export_frontend,
         deploy_url=config.deploy_url,
     )
 
     # Post a telemetry event.
     telemetry.send("export", config.telemetry_enabled)
+    console.success(
+        """App exported to[bold]backend.zip[/bold] and [bold]frontend.zip[bold]."""
```
masenf, line 216: i think we should keep track of output files and not say we exported a particular file if we didn't (#1285)

```python
outputs = set()
if export_backend:
    if zipping:
        outputs.add("backend.zip")
    else:
        outputs.add("app")
if export_frontend:
    if zipping:
        outputs.add("frontend.zip")
    else:
        outputs.add(".web/_static")
```
...
```python
console.success(
    "App exported to " + " and ".join(f"[bold]{filename}[/bold]" for filename in outputs)
)
```

reflex/reflex.py
```diff
@@ -168,63 +155,29 @@ def run(
...
-        httpx.put(frontend, data=f)  # type: ignore
-
-    with open(constants.BACKEND_ZIP, "rb") as f:
-        httpx.put(backend, data=f)  # type: ignore
-
-
 @cli.command()
 def export(
-    zipping: bool = typer.Option(
-        True, "--no-zip", help="Disable zip for backend and frontend exports."
```
masenf, line 221: is there a reason `--no-zip` is going away? it's very convenient for docker container builds

reflex/reflex.py
picklelo: Good call, will change this

reflex/reflex.py
picklelo: I'll add this back

masenf:
Fixed via #1591

PR #1478, opened 2023-07-31
check_initialized: skip REFLEX_DIR check for backend only

### LINKED ISSUE #1477: `reflex run --backend-only` should not require `reflex init`
**Describe the bug**
Running the backend should NOT require reflex init (or bun or nvm).

**To Reproduce**

Use the `counter` app from `reflex-examples`, add the following files:

```Dockerfile
FROM python:3.11
COPY . /app
WORKDIR /app
RUN python3 -m pip install -r requirements.txt
CMD reflex run --env prod --backend-only
```

```console
docker build -t counter .
docker run -it counter
```

**Expected behavior**
Expect the backend of the app to run, which doesn't require init, compiling, or bun/nvm.

Instead, it says `Error: The app is not initialized. Run reflex init first.`

 - Python Version: 3.11
 - Reflex Version: 0.2.3

The issue is introduced by this new check: https://github.com/reflex-dev/reflex/blob/9cbe02e3fe8aa034f544cbc9ca77bbff5142d6f0/reflex/utils/prerequisites.py#L396

masenf:
when running the python backend only, the REFLEX_DIR does not need to exist because there is no bun/nvm dependency for the backend.

fix #1477

adds a test to `in_docker_test_script` that attempts to export the backend only before init, which passes with this fix

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 ### **✨ Performant, customizable web apps in pure Python. Deploy in seconds. ✨**
 [![PyPI version](https://badge.fury.io/py/reflex.svg)](https://badge.fury.io/py/reflex)
-![tests](https://github.com/pynecone-io/pynecone/actions/workflows/build.yml/badge.svg)
+![tests](https://github.com/pynecone-io/pynecone/actions/workflows/integration.yml/badge.svg)
 ![versions](https://img.shields.io/pypi/pyversions/reflex.svg)
 [![Documentaiton](https://img.shields.io/badge/Documentation%20-Introduction%20-%20%23007ec6)](https://reflex.dev/docs/getting-started/introduction)
 [![Discord](https://img.shields.io/discord/1029853095527727165?color=%237289da&label=Discord)](https://discord.gg/T5WSbC2YtQ)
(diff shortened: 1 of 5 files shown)


integration/init-test/in_docker_test_script.sh
```diff
@@ -11,4 +11,4 @@ echo "Installing reflex from local repo code"
 cd /reflex-repo
 poetry install
 echo "Running reflex init in test project dir"
-poetry run /bin/bash -c "cd ~/hello && reflex init"
\ No newline at end of file
+poetry run /bin/bash -c "cd ~/hello && reflex init && rm -rf ~/.reflex .web && reflex export --backend-only"
```
jackie-pc, line 14: The internal logic depends on platform (like windows vs Linux etc.). Would it make sense to also inject this into the example/counter or reflex-web integration tests?  I.e.

Prior to running the normal `reflex init` in those integration tests, just try to `export --backend-only` (and possibly `reflex run --backend-only`?  That way we gain full matrix coverage for this behavior.

jackie-pc (dismissed):
🚢 it!

picklelo (approved):

Issue #1479, opened 2023-07-31
Clean up Dynamic Imports


picklelo:
For example in `components/lib/react_player.py` we do the dynamic import as part of the `_get_custom_code`. This should be moved to the `_get_imports` function.

```python
    def _get_imports(self):
        return {"next/dynamic": {ImportVar(tag="dynamic", is_default=True)}}
```

PR #1480, opened 2023-07-31
Fix Nextjs Dynamic Import

Alek99:
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

If multiple components that imported a dynamic component were on the same page, it would error. This fixes that by moving next/dynamic to _get_imports

    b. Describe your changes.

Move nextdynamic to _get_imports

closes: #1479

### DIFF
--- a/reflex/components/graphing/plotly.py
+++ b/reflex/components/graphing/plotly.py
@@ -4,12 +4,12 @@ from typing import Dict
 
 from plotly.graph_objects import Figure
 
-from reflex.components.component import Component
+from reflex.components.component import NoSSRComponent
 from reflex.components.tags import Tag
-from reflex.vars import ImportVar, Var
+from reflex.vars import Var
 
 
-class PlotlyLib(Component):
+class PlotlyLib(NoSSRComponent):
     """A component that wraps a plotly lib."""
 
     library = "react-plotly.js"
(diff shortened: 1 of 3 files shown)


reflex/components/graphing/plotly.py
```diff
@@ -36,10 +36,10 @@ class Plotly(PlotlyLib):
     use_resize_handler: Var[bool]
 
     def _get_imports(self):
-        return {}
+        return {"next/dynamic": {ImportVar(tag="dynamic", is_default=True)}}
```
masenf, line 39: seems like this could be exposed as a reusable name

```python
DYNAMIC_IMPORT = {"next/dynamic": {ImportVar(tag="dynamic", is_default=True)}}
```

if the "no ssr" import functionality could be baked into the existing `ImportVar` compiler mechanism, then that might be even better... but a quick look shows that is not trivial as it's currently written.

So maybe an implementation like

```python
class NoSSRComponent(Component):
    def _get_imports(self):
        return {"next/dynamic": {ImportVar(tag="dynamic", is_default=True)}}

    def _get_custom_code(self) -> str:
        return f"const {self.tag} = dynamic(() => import('{self.library}'), { ssr: false });"
```

masenf (commented):
looks like this already merged, so for your consideration

reflex/components/graphing/plotly.py
Alek99: Good point thanks

reflex/components/libs/react_player.py
```diff
@@ -44,12 +41,3 @@ class ReactPlayerComponent(Component):
...
     # Set the height of the player: ex:640px
     height: Var[str]
-
-    def _get_imports(self) -> Optional[imports.ImportDict]:
-        return {}
-
-    def _get_custom_code(self) -> Optional[str]:
-        return """
-import dynamic from "next/dynamic";
-const ReactPlayer = dynamic(() => import("react-player/lazy"), { ssr: false });
```
masenf, line 54: need to set `library = "react-player/lazy"` on line 14 now

masenf (changes requested):

masenf (approved):

PR #1484, opened 2023-08-01
Wrap Input and TextArea with DebounceInput for full control

masenf:
When the caller supplies a `value` prop as a state `Var` and also provides an `on_change` handler, automatically wrap it in a `DebounceInput` component with zero timeout to avoid typing jank.

Update `DebounceInput` component to allow arbitrarily nested hierarchies of `DebounceInput`, making it simpler to reuse and override `DebounceInput` props, even when auto-created components are in the tree.

Update integration test to use implicit fully controlled input (like a naive user).

Add component-level tests for debounce to validate prop combining behavior.

### DIFF
--- a/integration/test_input.py
+++ b/integration/test_input.py
@@ -10,7 +10,7 @@ from reflex.testing import AppHarness
 
 
 def FullyControlledInput():
-    """App using a fully controlled input with debounce wrapper."""
+    """App using a fully controlled input with implicit debounce wrapper."""
     import reflex as rx
 
     class State(rx.State):
@@ -21,12 +21,10 @@ def FullyControlledInput():
     @app.add_page
     def index():
         return rx.fragment(
-            rx.debounce_input(
-                rx.input(
-                    on_change=State.set_text, id="debounce_input_input"  # type: ignore
-                ),
+            rx.input(
+                id="debounce_input_input",
+                on_change=State.set_text,  # type: ignore
                 value=State.text,
-                debounce_timeout=0,
             ),
             rx.input(value=State.text, id="value_input"),
             rx.input(on_change=State.set_text, id="on_change_input"),  # type: ignore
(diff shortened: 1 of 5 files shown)


reflex/components/forms/debounce.py
```diff
@@ -78,3 +79,29 @@ def props_not_none(c: Component) -> dict[str, Any]:
     """
     cdict = {a: getattr(c, a) for a in c.get_props() if getattr(c, a, None) is not None}
     return cdict
+
+
+def _collect_first_child_and_props(c: Component) -> tuple[Component, dict[str, Any]]:
```
picklelo, line 84: I think we need to use `Tuple` for Python 3.7 support

integration/test_input.py
```diff
@@ -21,12 +21,10 @@ class State(rx.State):
     @app.add_page
     def index():
         return rx.fragment(
-            rx.debounce_input(
-                rx.input(
-                    on_change=State.set_text, id="debounce_input_input"  # type: ignore
-                ),
+            rx.input(
```
picklelo, line 24: I think we don't need the `id` anymore

reflex/components/forms/debounce.py
masenf: this file uses `from __future__ import annotations`, which enables PEP 563 behavior (even in 3.7), allowing these annotations to work.

i can change them if it matters, but the tests are passing on 3.7 and AFAIK `typing.Tuple` and `typing.Dict` are on their way out once 3.9 goes EoL, so if we can support the newer style now, with no drawbacks, i think that reduces our future burden to fix.

picklelo: Ah okay got it, we should convert all the code to use this then it's much nicer

picklelo (approved):

PR #1489, opened 2023-08-01
Cache unchanged pages during compilation

picklelo:
First pass at caching pages to avoid recompilation.

Strategy:
During compilation, calculate checksums for all the components (NOTE: the hashing function is not perfect yet, and needs to be improved)
If the checksum matches the previous page, skip compilation.
Otherwise, continue with compilation.
At the end, write the new checksums to a file to preserve caching between runs (and hot reloads)


masenf (changes requested):
Can we close this one for now, and re-pick this up in a different way as part of "Immutable Components"? REF-1343, REF-1344

PR #1492, opened 2023-08-01
Simplify rxconfig template

picklelo:
Removed all the default arguments to simplify the template.


masenf (approved):
do we also need to subclass `rx.Config`? or can we just instantiate it

```python
config = rx.Config(
    app_name="{{ app_name }}",
)
```

picklelo:
@masenf someone added the subclass so that they could add additional fields to the config, but I agree it would be cleaner without it.

PR #1493, opened 2023-08-01
fail faster in integration test

jackie-pc:
Checking if a pid exists in a portable way (across OSes) was not immediately apparent before to me (@jackie-pc), so was left out of the big CI merge.

This is the quick follow up to do this properly (so we don't wait 10 min) if `reflex run` already died early.

### DIFF
--- a/scripts/integration.sh
+++ b/scripts/integration.sh
@@ -10,11 +10,20 @@ export PYTHONUNBUFFERED=1
 # Start the server in the background
 reflex run --loglevel debug --env "$2" & pid=$!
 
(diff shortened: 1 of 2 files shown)


masenf (approved):
yeah, killing processes is hard, especially when they fork.

here's what i'm doing in the `AppHarness` to handle it for reference https://github.com/reflex-dev/reflex/blob/fb7e2b6a0caccd47135a7fec074d90659fdbe928/reflex/testing.py#L227-L246

but i think what you're doing here is fine as well.

PR #1520, opened 2023-08-04
update Dockerfile example

masenf:
with optional compose + caddy support for TLS termination

Parking this draft here, since i have to shift gears today and might not get back to it until next week.

### DIFF
--- a/docker-example/Dockerfile
+++ b/docker-example/Dockerfile
@@ -1,20 +1,27 @@
 # Stage 1: init
 FROM python:3.11 as init
+
 # Pass `--build-arg API_URL=http://app.example.com:8000` during build
 ARG API_URL
+
 # Copy local context to `/app` inside container (see .dockerignore)
 WORKDIR /app
 COPY . .
+
 # Reflex will install bun, nvm, and node to `$HOME/.reflex` (/app/.reflex)
 ENV HOME=/app
+
 # Create virtualenv which will be copied into final container
 ENV VIRTUAL_ENV=/app/.venv
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 RUN python3 -m venv $VIRTUAL_ENV
+
 # Install app requirements and reflex inside virtualenv
 RUN pip install -r requirements.txt
+
 # Deploy templates and prepare app
 RUN reflex init
+
 # Export static copy of frontend to /app/.web/_static (and pre-install frontend packages)
 RUN reflex export --frontend-only --no-zip
 
--- a/docker-example/Dockerfile
+++ b/docker-example/Dockerfile
@@ -1,49 +1,31 @@
-FROM python:3.11-slim as base
-
-RUN adduser --disabled-password reflex
-
-
-FROM base as build
-
+# Stage 1: init
+FROM python:3.11 as init
+# Pass `--build-arg API_URL=http://app.example.com:8000` during build
+ARG API_URL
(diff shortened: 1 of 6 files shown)


docker-example/README.md
```diff
@@ -1,37 +1,64 @@
...
+## Build Reflex Container Image
+
+To build your container image run the following command:
+
+```bash
+docker build -t reflex-app:latest . --build-arg API_URL=http://app.example.com:8000
+```
 
-The `rxconfig.py` includes the configuration of your Reflex service. Edit the file like the following configuration. If you want to use a custom database you can set the endpoint in this file. Ensure that `api_url` points to the publicly accessible hostname where the container will be running.
+Ensure that `API_URL` is set to the publicly accessible hostname or IP where the app
```
picklelo, line 20: If they don't set this, should we default to use localhost:8000 ?

docker-example/Dockerfile
```diff
@@ -1,49 +1,31 @@
...
+FROM python:3.11 as init
+# Pass `--build-arg API_URL=http://app.example.com:8000` during build
+ARG API_URL
+# Copy local context to `/app` inside container (see .dockerignore)
 WORKDIR /app
-ENV VIRTUAL_ENV=/app/venv
-RUN python3 -m venv $VIRTUAL_ENV
-ENV PATH="$VIRTUAL_ENV/bin:$PATH"
-
 COPY . .
```
picklelo, line 7: Do we need to for example exclude anything in the COPY (like the .web directory?)

docker-example/README.md
```diff
@@ -1,37 +1,64 @@
...
-```python
-import reflex as rx
+## Start Container Service
 
-config = rx.Config(
-    app_name="app",
-    api_url="http://app.example.com:8000",
-    db_url="sqlite:///reflex.db",
-)
+Finally, you can start your Reflex container service as follows:
```
picklelo, line 25: Maybe we should add a note that it will take some time for the services to show up at ports 3000 8000. Or can we make the docker container forward the logs? Since there's no feedback it's hard to know when it has started

masenf: Or we could take the `-d` off the command here and then the logs would just stream to the console. More experienced docker users would know that `-d` could be added (along with some health checks) to background the container and wait for it to come up.

I got some feedback from another user on discord that we probably shouldn't be recommending `-d` in the example command.

docker-example/README.md
masenf: if the user doesn't provide `API_URL`, then the reflex default will be used, so whatever is in `rxconfig.py` or `http://localhost:8000`.

Do you think I should explicitly mention that in this section?

docker-example/Dockerfile
masenf: No, in fact, `.web` is in the `.dockerignore` file because we specifically do NOT want to copy the potentially huge `.web` directory into the build context. 

This `COPY . .` command will take everything in the same directory of the `Dockerfile`, except stuff in `.dockerignore`, so it should sweep up everything needed to build the app inside the container, which is going to be better, for example if the node_modules include compiled code and the build system and container architecture differ.

masenf (commented):
Thanks for the feedback.

docker-example/README.md
picklelo: Ah, yeah good call taking off the `-d` they can add that back if they want

picklelo (approved):

Issue #1422, opened 2023-07-25
Event handler chaining is non-deterministic.


stuz5000:
To reproduce `reflex run`:
```
import reflex as rx

class State2(rx.State):
    text: str

    def f1(self):
        print("f1")
    def f2(self, arg):
        print(f"f2 {arg}")

    def onClick(self):
        print("---")
        yield State2.f1()
        yield State2.f1()

        yield State2.f2(123)  
        yield State2.f1()

def index():
    return rx.box(
        rx.button("Confirm", on_click=State2.onClick),
    )

app = rx.App(state=State2)
app.add_page(index)
app.compile()
```

Smash the click button many time and see the following:
```
[Updating App] Applying changes and refreshing. Time: 2023-07-25 14:39:08.065674
---
f1
f1
---
f1
f1
---
f1
f1
---
f1
f1
f1
---
f1
f2 123
---
f1
f1
---
f1
f1
---
f1
---
f1
f1
```

Four bugs:
- f1 is never executed 3 times
- sometimes f2 is executed, sometimes not
- sometimes f1 is just executed once before f2, sometimes twice
- sometimes f1 is executed just once, and nothing else

Behavior is similarly broken with just
```
    def onClick(self):
        print("---")
        yield State2.f1()
        yield State2.f2(123)  
```
... where f2 sometimes do not execute.

Switching all methods to `async def` - no change in behavior.

Event handler arguments makes no difference. Fails as above:
```
class State2(rx.State):
    async def f1(self):
        print("f1")

    async def f11(self):
        print("f11")

    async def onClick(self):
        print("---")
        yield State2.f1()
        yield State2.f11()
```
---------------
reflex==0.2.0
python 3.10

stuz5000:
The following seems to work fine:

```
class State2(rx.State):
    text: str

    async def f1(self):
        print("f1")
    async def f11(self):
        print("f11")
    async def f2(self, arg):
        print(f"f2 {arg}")

    async def onClick(self):
        print("---")
        return [State2.f1(), State2.f1(), State2.f2(123), State2.f1()]
```

Lendemor:
I'm investigating this issue.

What is worthy to note is that yielding an event will not execute it immediately, but rather append it at the end of the event queue, and will be executed after current event handler (and other event in the loop) have finished running.

The following code is deterministic : 
```python
class State(rx.State):
    def f1(self):
        print("f1")

    def f2(self, value):
        print(f"f2 {value}")

[...]

PR #1523, opened 2023-08-04
Event queue no longer part of reactive state

masenf:
Do not manage incoming websocket events via React's state system.

Reactive state should never have side effects, and by definition the incoming websocket event handler and event queuing is a side effect outside of react, which introduces subtle incorrect data races.

Instead, manage the event queue completely outside of the state, and only interact with it during useEffect hook, which allows side effects since it occurs after render time.

# Testing

Some tests for yielding and returning State and client-side events have been added, but unfortunately these tests still do not reliably fail without the proposed changes (although they will fail eventually). I'd like some more deterministic test cases that really bring out the dropped events behavior that we see in #1422

### DIFF
--- a/integration/test_event_chain.py
+++ b/integration/test_event_chain.py
@@ -280,11 +280,11 @@ def test_event_chain_click(event_chain, driver, button_id, exp_event_order):
     btn.click()
     if "redirect" in button_id:
         # wait a bit longer if we're redirecting
-        time.sleep(0.5)
+        time.sleep(1)
     if "many_events" in button_id:
         # wait a bit longer if we have loads of events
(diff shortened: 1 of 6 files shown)


masenf:
Moving this one back to draft state; needs a rethink.

Essentially, we should be leveraging `useReducer` to batch updates to the state from event handlers to avoid updating the state multiple times as a `useEffect`. The event queue can still be managed out of band, but the `applyDelta` call needs to become a reducer function to let react handle the queuing and application of state changes.

masenf:
closing. will combine with the useReducer PR and cleanup

PR #1538, opened 2023-08-07
Run AppHarness selenium integration tests in CI

masenf:
# `conftest.py`

Prepare the integration tests to run via github actions:

* use pyvirtualdisplay to create a virtual xvfb display to run the browser in non-headless mode
* if any test fails (and $SCREENSHOT_DIR is set), grab the virtual X display and save it to a file named after the failed test case.

# `integration_app_harness.yml`

New workflow executes the python tests in `integration/` using latest ubuntu and python 3.11.

* install testing deps: `poetry run pip install pyvirtualdisplay pillow`
* upload any screenshots at the end of the test

# Other

The timing in `test_input.py` was bumped up to make the test have fewer false positives.

# Example Screenshots

The following failures were induced in earlier runs to demonstrate screenshot functionality:

https://github.com/reflex-dev/reflex/actions/runs/5789798397
image

integration_test_input
integration_test_server_side_event

### DIFF
--- a/reflex/testing.py
+++ b/reflex/testing.py
@@ -155,6 +155,8 @@ class AppHarness:
                 )
                 self.app_module_path.write_text(source_code)
         with chdir(self.app_path):
+            # ensure config is reloaded when testing different app
+            reflex.config.get_config(reload=True)
             self.app_module = reflex.utils.prerequisites.get_app()
         self.app_instance = self.app_module.app
 
--- a/.github/workflows/integration_app_harness.yml
+++ b/.github/workflows/integration_app_harness.yml
@@ -1,0 +1,37 @@
+name: integration-app-harness
+
+on:
+  push:
+    branches: [ "main" ]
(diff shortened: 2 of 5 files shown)


reflex/testing.py
```diff
@@ -155,6 +155,8 @@ def _initialize_app(self):
                 )
                 self.app_module_path.write_text(source_code)
         with chdir(self.app_path):
+            # ensure config is reloaded when testing different app
+            reflex.config.get_config(reload=True)
```
picklelo, line 159: I wonder if we should have a way to specify the config, so multiple configs can exist together and we can avoid the forced re-import

picklelo (approved):
awesome!

reflex/testing.py
masenf: we definitely should. i think that can be handled during our upcoming config refactor

PR #1541, opened 2023-08-08
fix missing on_load parameter in custom_404

Lendemor:
### All Submissions:

### Type of change

### Changes To Core Features:

### Description

Add the missing on_load kwargs to the custom 404 method, allowing a default redirect when a page is unknown.

Also make component optional (default to a rx.fragment()) since defining one is not needed for redirects.

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -404,14 +404,14 @@ class App(Base):
 
     def add_custom_404_page(
         self,
-        component=None,
-        title=None,
-        image=None,
-        description=None,
+        component: Optional[Union[Component, ComponentCallable]],
+        title: str = constants.TITLE_404,
+        image: str = constants.FAVICON_404,
+        description: str = constants.DESCRIPTION_404,
         on_load: Optional[
             Union[EventHandler, EventSpec, List[Union[EventHandler, EventSpec]]]
         ] = None,
-        meta=constants.DEFAULT_META_LIST,
+        meta: List[Dict] = constants.DEFAULT_META_LIST,
     ):
         """Define a custom 404 page for any url having no match.
 
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -427,11 +427,7 @@ class App(Base):
             meta: The metadata of the page.
         """
         self.add_page(
-            component=(
-                component
-                if isinstance(component, Component)
-                else (component() if component else Fragment.create())
-            ),
+            component=component if component else Fragment.create(),
             route=constants.SLUG_404,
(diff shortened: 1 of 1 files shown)


reflex/app.py
```diff
@@ -404,10 +404,13 @@ def _check_routes_conflict(self, new_route: str):
 
     def add_custom_404_page(
         self,
-        component,
+        component=None,
```
ElijahAhianyo, line 407: nit: will be nice to have  annotations for all params for consistency and readability

masenf (approved):

reflex/app.py
```diff
@@ -436,8 +444,17 @@ def add_custom_404_page(
...
-        froute = format.format_route
-        self.pages[froute(constants.SLUG_404)] = component
+        # format 404 route
+        route = format.format_route(constants.SLUG_404)
+
+        # add the page
+        self.pages[route] = component
+
+        # Add the load events.
+        if on_load:
```
picklelo, line 454: We should refactor this with the `add_page` method so we're not duplicating code. Is there any reason this method can't just call `add_page`?

picklelo (approved):
nice!

reflex/app.py
```diff
@@ -404,11 +404,14 @@ def _check_routes_conflict(self, new_route: str):
 
     def add_custom_404_page(
         self,
-        component,
-        title=None,
-        image=None,
-        description=None,
-        meta=constants.DEFAULT_META_LIST,
+        component: Optional[Union[Component, ComponentCallable]] = None,
```
picklelo, line 407: Why is the component optional if they're adding a page?

reflex/app.py
```diff
@@ -420,25 +423,23 @@ def add_custom_404_page(
...
-        compiler_utils.add_meta(
-            component,
-            title=title,
-            image=image,
-            description=description,
+        self.add_page(
+            component=(
+                component
+                if isinstance(component, Component)
+                else (component() if component else Fragment.create())
```
picklelo, line 433: This check is already done in `add_page`, so we don't need to do it here actually.

reflex/app.py
Lendemor: If we setup a redirect, the actual page to render doesn't matter (and will be set to default as a rx.fragment()

reflex/app.py
Lendemor: We need to for the case where component is None, maybe I can simplify it here.

picklelo (approved):

PR #1543, opened 2023-08-08
Event Handlers should not shadow base state methods

### LINKED ISSUE #1424: Event handlers shadow base state methods
Currently for example you cannot have an event handler named `reset` because the State class already has this method. We should modify the base class method names, and throw errors if anything gets shadowed as it will break functionality

ElijahAhianyo:
With this PR an error is thrown when an event handler shadows a builtin state method. Also renames some builtin methods

closes #1424

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -247,12 +247,13 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
             NameError: When an event handler shadows an inbuilt state method.
         """
         overridden_methods = set()
+        state_base_functions = cls._get_base_functions()
         for name, method in inspect.getmembers(cls, inspect.isfunction):
             # Check if the method is overridden and not a dunder method
             if (
                 not name.startswith("__")
-                and method.__name__ in dir(State)
-                and getattr(State, method.__name__) != method
+                and method.__name__ in state_base_functions
+                and state_base_functions[method.__name__] != method
             ):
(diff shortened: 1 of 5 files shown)


reflex/state.py
```diff
@@ -444,6 +452,19 @@ def _set_default_value(cls, prop: BaseVar):
             field.required = False
             field.default = default_value
 
+    @staticmethod
+    def get_base_functions() -> List[str]:
```
picklelo, line 456: Should we make this an underscore method too?

picklelo (commented):
nice changes, looks like we need to just fix the unit tests

reflex/state.py
```diff
@@ -611,17 +648,17 @@ def __setattr__(self, name: str, value: Any):
...
         if name == constants.ROUTER_DATA:
             self.dirty_vars.add(name)
-            self.mark_dirty()
+            self._mark_dirty()
             # propagate router_data updates down the state tree
             for substate in self.substates.values():
                 setattr(substate, name, value)
 
-    def reset(self):
+    def _reset(self):
```
picklelo, line 661: I know some people are using the `.reset()` function in their apps already. Should we maybe keep this one as is for now? We throw a warning anyways, so it can make sure people don't accidentally override it

picklelo (commented):
looks good just one comment

picklelo (approved):
lgtm

PR #1552, opened 2023-08-09
Reinstate bun path

### LINKED ISSUE #1470: `v0.2.3` breaks `bun_path` without warning
**Describe the bug**  
As of Reflex version `v0.2.3`, the support for specifying a custom `bun_path` has been dropped, breaking compatibility with non-FHS compliant distros like Guix System and NixOS. When trying to run Reflex, it fails due to this missing support, even when manual workarounds are attempted, such as manually creating the `bun` path or touching an executable `bun`.

**To Reproduce**  
1. On a Guix System or NixOS (or other non-FHS compliant distros), upgrade Reflex to `v0.2.3`.
2. Try to run Reflex with a custom `bun_path`.
3. Observe the errors indicating the absence of the `bun` executable in the specified path, even if the path and executable are manually created.

**Expected behavior**  
Reflex should respect the custom `bun_path` and allow users to use their own Bun installation. In environments where Reflex's default method of managing Bun isn't feasible, users should be able to provide their own path, disabling built-in checks and installations, allowing Reflex to just use what's provided without guarantees.

 - Python Version: 3.11
 - Reflex Version: 0.2.3
 - OS: Guix System and NixOS (non-FHS compliant distros)
 
**Additional context**  
The removal of `bun_path` support impacts users of non-FHS compliant distros significantly. Reflex's decision to manage its own Bun installation and dependencies makes it incompatible with these systems due to their unique file system hierarchy and dependency management. It's crucial for Reflex to allow flexibility for users in such environments to specify custom paths for dependencies. This used to work in version `v0.2.2` and its removal in `v0.2.3` has rendered Reflex unusable for a segment of its user base.

For logs and additional info, please begin reading from issue https://github.com/reflex-dev/reflex/issues/574#issuecomment-1657737762

cc @ElijahAhianyo @Alek99

ElijahAhianyo:
We took out the bun_path config param in `v0.2.3` and this introduced regression in non-fhs systems. This PR reinstates the `bun_path` param allowing users with non-FHS systems to specify a custom path to a bun executable. It also adds validation checks to ensure the bun exec is compatible with reflex.

 **Note that this only applies to non-fhs systems, for all other OS, reflex handles installation of these dependencies under the hood**

closes #1470

### DIFF
--- a/reflex/config.py
+++ b/reflex/config.py
@@ -169,7 +169,7 @@ class Config(Base):
     frontend_packages: List[str] = []
 
     # The bun path
-    bun_path: Optional[str] = constants.BUN_PATH
+    bun_path: str = constants.BUN_PATH
 
     # The Admin Dash.
     admin_dash: Optional[AdminDash] = None
--- a/reflex/config.py
+++ b/reflex/config.py
@@ -168,6 +168,9 @@ class Config(Base):
     # Additional frontend packages to install.
     frontend_packages: List[str] = []
 
+    # The bun path
+    bun_path: Optional[str] = constants.BUN_PATH
+
     # The Admin Dash.
     admin_dash: Optional[AdminDash] = None
 
--- a/reflex/constants.py
+++ b/reflex/constants.py
@@ -65,6 +65,8 @@ JINJA_TEMPLATE_DIR = os.path.join(TEMPLATE_DIR, "jinja")
 # Bun config.
 # The Bun version.
 BUN_VERSION = "0.7.0"
+# Min Bun Version
+MIN_BUN_VERSION = "0.7.0"
 # The directory to store the bun.
 BUN_ROOT_PATH = os.path.join(REFLEX_DIR, ".bun")
 # Default bun path.
--- a/reflex/constants.py
+++ b/reflex/constants.py
@@ -67,8 +67,10 @@ JINJA_TEMPLATE_DIR = os.path.join(TEMPLATE_DIR, "jinja")
 BUN_VERSION = "0.7.0"
 # The directory to store the bun.
 BUN_ROOT_PATH = os.path.join(REFLEX_DIR, ".bun")
+# Default bun path.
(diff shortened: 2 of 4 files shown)


reflex/config.py
```diff
@@ -168,6 +168,9 @@ class Config(Base):
     # Additional frontend packages to install.
     frontend_packages: List[str] = []
 
+    # The bun path
+    bun_path: Optional[str] = constants.BUN_PATH
```
masenf, line 172:
```suggestion
    bun_path: str = constants.BUN_PATH
```

i think this is all that's need to fix the pre-commit errors

masenf (approved):
approve once CI is passing

reflex/constants.py
```diff
@@ -65,10 +65,14 @@ def get_value(key: str, default: Any = None, type_: Type = str) -> Type:
 # Bun config.
 # The Bun version.
 BUN_VERSION = "0.7.0"
+# Min Bun Version
+MIN_BUN_VERSION = "0.7.0"
```
ElijahAhianyo, line 69: We still need a floor bun version for custom bun path use cases. If we increase the current bun version, users won't necessarily need to update bun versions(which could be annoying)

masenf (approved):

reflex/utils/prerequisites.py
```diff
@@ -435,12 +435,45 @@ def is_latest_template() -> bool:
...
+        if not bun_version:
+            console.error(
+                f"Failed to obtain bun version. Make sure the specified bun path in your config is correct."
+            )
+            raise typer.Exit(1)
+        elif bun_version < version.parse(constants.MIN_BUN_VERSION):
+            console.error(
+                f"Reflex requires bun version {constants.BUN_VERSION}+ to run. If you specified a bun path in your "
+                f"config,"
+                f"make sure to specify one that meets the requirements."
```
masenf, line 458: it might be nice for debugging to tell the user what bun version we detected

picklelo (approved):
Looks good to me, I'll rebase my config PR on top of this

PR #1599, opened 2023-08-15
Move initial state to separate file

masenf:
REF-23

# Write initialState and initialEvents to `utils/context.js`

Factor out initial state into its own file to avoid duplication

# EventLoopProvider context encapsulates the event loop

The `EventLoopProvider` is defined at the app level now and maintains the websocket connection and state outside of the page lifecycle, resulting in better navigation performance between pages and allowing arbitrary nested child components to access the state and event loop without affecting the render/dirty state of their parents.

Currently all pages use `StateContext` and `EventContext` so nothing really changes, however this structure will enable further optimizations down the road

# Remove `is_dynamic` hack for `routeChangeComplete` event

Now that the websocket isn't being reset on every static page navigation event, we rely on `routeChangeComplete` to trigger page hydration when navigating via links for all pages.

### DIFF
--- a/reflex/compiler/compiler.py
+++ b/reflex/compiler/compiler.py
@@ -6,7 +6,6 @@ from typing import List, Set, Tuple, Type
 from reflex import constants
 from reflex.compiler import templates, utils
 from reflex.components.component import Component, ComponentStyle, CustomComponent
-from reflex.route import get_route_args
 from reflex.state import State
 from reflex.utils import imports
 from reflex.vars import ImportVar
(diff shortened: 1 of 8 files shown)


reflex/compiler/compiler.py
```diff
@@ -18,6 +17,7 @@
         ImportVar(tag="useEffect"),
         ImportVar(tag="useRef"),
         ImportVar(tag="useState"),
```
picklelo, line 19: We can remove `useState` now

picklelo (approved):
This is so nice! Everything works for me, and the page outputs are much cleaner 😄

PR #1566, opened 2023-08-10
Automatic Install FNM and Node for Windows:

ElijahAhianyo:
Related to #1381 This PR installs [fnm ](https://github.com/Schniz/fnm)package and Node automatically on windows. On Windows, A `reflex` dir is created in `C:/Users/<username>/Appdata/Local/` to manage any reflex dependency

PR #1606, opened 2023-08-16
Fnm and node for POSIX

### LINKED ISSUE #1585: Use Fnm for node installation on POSIX
As a follow up to #1566 , we should use [fnm ](https://github.com/Schniz/fnm) for node installation on Linux and MacOs

ElijahAhianyo:
## Description

This pull request aims to enhance the installation process of Node.js by replacing the previous usage of [NVM](https://github.com/nvm-sh/nvm) with [FNM](https://github.com/Schniz/fnm). This change not only introduces cross-platform compatibility but also leverages FNM's capability to generate release binaries for Linux, macOS, and Windows distributions. Furthermore, the installation of dependencies has been streamlined to align with platform conventions.

## Changes Made

- Replaced NVM-based Node.js installation with FNM.
- Introduced cross-platform compatibility by adopting FNM.
- Enhanced Windows support by previously merged PR (v0.2.5) that installs Node.js using FNM.
- Adjusted the installation of dependencies to platform-specific directories:
    - Windows: `C:/Users/<username>/AppData/Local/reflex`
    - macOS: `~/Library/Application Support/reflex`
    - Linux: `~/.local/share/reflex`

## Additional Context

The motivation behind this PR is to provide a unified and reliable installation experience for Node.js across different operating systems. The transition from NVM to FNM enables seamless management of Node.js versions while addressing Windows compatibility limitations that existed with NVM. The adoption of platform-specific directories ensures clean separation of dependencies, enhancing organization and system integration.

It's worth noting that FNM generates release binaries for Linux, macOS, and Windows distributions, which aligns perfectly with the multi-platform approach of this project.

This PR builds upon the previous Windows-related PR #1566 (merged for v0.2.5), and its successful testing and merging indicate that this solution is ready for wider adoption.

Closes #1585


ElijahAhianyo:
should be reviewed/merged after #1566

reflex/utils/prerequisites.py
```diff
@@ -265,75 +266,73 @@ def download_and_run(url: str, *args, show_status: bool = False, **env):
...
+    """Install fnm and nodejs for use by Reflex.
     Independent of any existing system installations.
     """
-    if constants.IS_WINDOWS:
-        path_ops.mkdir(constants.FNM_DIR)
-        if not os.path.exists(constants.FNM_EXE):
-            download_and_extract_fnm_zip(constants.FNM_WINDOWS_INSTALL_URL)
+    if not constants.FNM_FILENAME:
+        # fnm only support Linux, macOS and Windows distros.
+        return
```
masenf, line 308: so what will happen in this case? can the user install and run their own copy of `node` in these environments? asking because it looks like we set `NODE_BIN_PATH` in an FNM-specific way

masenf (approved):

reflex/constants.py
```diff
@@ -18,6 +19,28 @@
...
+    elif platform_os == "Darwin":
+        return "fnm-macos"
+    elif platform_os == "Linux":
+        machine = platform.machine()
+        if machine == "arm" or machine.startswith("armv7"):
+            return "fnm-arm32"
+        elif machine.startswith("aarch") or machine.startswith("armv8"):
+            return "fnm-arm64"
+        return "fnm-linux"
+    return
```
picklelo, line 41: Make this `return None` explicitly

reflex/constants.py
```diff
@@ -18,6 +19,28 @@
 IS_WINDOWS = platform.system() == "Windows"
 
 
+def get_fnm_name() -> Optional[str]:
```
picklelo, line 22: I think this function can be moved to `prerequisites.py`

reflex/constants.py
```diff
@@ -18,6 +19,28 @@
...
+    """Get the appropriate fnm executable name based on the current platform.
+
+    Returns:
+        str: The fnm executable name for the current platform.
```
picklelo, line 26: Don't need to add type annotations in the docstrings as we use type hints.

reflex/constants.py
```diff
@@ -28,18 +51,13 @@
 # The directory to store reflex dependencies.
 REFLEX_DIR = (
     # on windows, we use C:/Users/<username>/AppData/Local/reflex.
+    # on macOS, we use ~/Library/Application Support/reflex.
+    # on linux, we use ~/.local/share/reflex.
```
picklelo, line 55: nice, this is clean!

reflex/constants.py
```diff
@@ -28,18 +51,13 @@
...
-        os.path.join(
-            "$HOME",
-            f".{MODULE_NAME}",
-        ),
-    )
 )
 # The root directory of the reflex library.
 ROOT_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
-# The name of the assets directory.
+# The name of the assets' directory.
```
picklelo, line 60: Extra change?

reflex/constants.py
```diff
@@ -56,46 +74,39 @@
...
 FNM_VERSION = "1.35.1"
 # The Node version.
 NODE_VERSION = "18.17.0"
 # The minimum required node version.
 NODE_VERSION_MIN = "16.8.0"
-# The directory to store nvm.
-NVM_DIR = os.path.join(REFLEX_DIR, ".nvm")
 # The directory to store fnm.
 FNM_DIR = os.path.join(REFLEX_DIR, "fnm")
+FNM_FILENAME = get_fnm_name()
```
picklelo, line 92: Can be moved to `prerequisites.py`

picklelo (commented):
Functionality works great on my Mac :) just some small comments

reflex/constants.py
ElijahAhianyo: The `FNM_FILENAME` is needed to construct the URL to the release binary (`FNM_INSTALL_URL`). Do you mean to move that as well (i.e move `get_fnm_name`, `FNM_FILE_NAME` and `FNM_INSTALL_URL` to prerequisites)?

reflex/utils/prerequisites.py
ElijahAhianyo: Thats a good question, Previously(when users had to install node themselves), we checked for node on the host machine (`which node` or use the node command and check for errors) and used that. We can do that as well as expose a config knob (just like we do for bun) to allow users specify custom paths to node binaries, but Im open to suggestions. What do you think? 
cc: @masenf @picklelo @Lendemor @jackie-pc

reflex/constants.py
```diff
@@ -18,6 +19,28 @@
...
+
+    Returns:
+        str: The fnm executable name for the current platform.
+    """
+    platform_os = platform.system()
+
+    if platform_os == "Windows":
+        return "fnm-windows"
+    elif platform_os == "Darwin":
+        return "fnm-macos"
```
jackie-pc, line 33: Why is there no ARM vs Intel distinction for Windows and Mac?

reflex/utils/prerequisites.py
jackie-pc: the `which` solution sounds reasonable to me.  Should we log something (even if debug level) when we fall through to such cases.  I.e. either we use `which node` or straight give up.  Note this should be a pretty rare case and so it's worth logging something.  I.e. we are talking about an environment that is NOT Windows, Linux or Mac (roughly)

reflex/utils/prerequisites.py
```diff
@@ -265,75 +266,73 @@ def download_and_run(url: str, *args, show_status: bool = False, **env):
...
     Raises:
         Exit: If an error occurs while downloading or extracting the FNM zip.
     """
-    # TODO: make this OS agnostic
     # Download the zip file
+    url = constants.FNM_INSTALL_URL
     console.debug(f"Downloading {url}")
-    fnm_zip_file = f"{constants.FNM_DIR}\\fnm_windows.zip"
-    # Function to download and extract the FNM zip release
+    fnm_zip_file = os.sep.join([constants.FNM_DIR, f"{constants.FNM_FILENAME}.zip"])
```
jackie-pc, line 278: Why not [os.path.join](https://docs.python.org/3/library/os.path.html#os.path.join)?  That's what we are joining right? (paths)

reflex/constants.py
masenf: i would assume macos is using fat binaries and windows is probably falling back to x86 emulation... but yeah, would definitely be good to test. i have windows arm, but i don't have x86 mac to test on.

reflex/utils/exec.py
```diff
@@ -72,13 +91,28 @@ def run_frontend_prod(
...
+            "Could not find NPM package manager. Make sure you have node installed."
+        )
+        raise typer.Exit(1)
 
+    if not prerequisites.check_node_version():
+        node_version = prerequisites.get_node_version()
+        console.error(
+            f"Reflex requires node version {constants.NODE_VERSION_MIN} or higher to run, but the detected version is {node_version}",
+        )
+        raise typer.Exit(1)
```
masenf, line 112: factor out this duplicated code block

ElijahAhianyo: fixed

masenf (approved):

PR #1607, opened 2023-08-16
Incrementally Add New Packages

Alek99:
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

```
Reduces initial package size from ~490mb to < 200mb. Instead of installing all node packages at the beginning, we only add the ones the user has referenced in their Reflex app.

Positives: Reduces initial start up time and node modules significantly
Negatives: Incrementally installing node modules will pause the page from reloading ~5-10 seconds when a new comment library is referenced.

```


reflex/.templates/web/package.json
```diff
@@ -15,28 +15,15 @@
...
     "focus-visible": "^5.2.0",
-    "framer-motion": "^10.12.4",
-    "gridjs": "^6.0.6",
-    "gridjs-react": "^6.0.1",
     "json5": "^2.2.3",
     "next": "^13.3.1",
     "next-sitemap": "^4.1.8",
-    "plotly.js": "^2.22.0",
     "react": "^18.2.0",
     "react-debounce-input": "^3.3.0",
```
masenf, line 22:
```suggestion
```

this one can probably go too, since we name it as a `library` here: https://github.com/reflex-dev/reflex/blob/0be5d670ed82f7b47e0c0b1fb17052933e060eda/reflex/components/forms/debounce.py#L11-L20

reflex/app.py
```diff
@@ -498,11 +520,18 @@ def compile(self):
...
+                all_imports.update(component.get_imports())
+
                 # Add the custom components from the page to the set.
                 custom_components |= component.get_custom_components()
+
         thread_pool.close()
         thread_pool.join()
 
+        # install frontend packages
+        self.get_frontend_packages(all_imports)
```
masenf, line 533: is there a reason to not just do this once at the end, after the custom component imports have been catalogued at well?

reflex/utils/exec.py
```diff
@@ -29,18 +73,38 @@ def run_process_and_launch_url(
...
+            )
+        if process.stdout:
+            for line in process.stdout:
+                if "ready started server on" in line:
+                    if first_run:
+                        url = line.split("url: ")[-1].strip()
+                        console.print(f"App running at: [bold green]{url}")
+                        first_run = False
+                    else:
+                        console.print(f"New packages detected updating app...")
```
masenf, line 95:
```suggestion
                        console.print("New packages detected updating app...")
```

i thought `ruff` fixed "useless 'f'-string"... 🤔

reflex/utils/exec.py
```diff
@@ -29,18 +73,38 @@ def run_process_and_launch_url(
...
+                else:
+                    console.debug(line)
+                    new_hash = detect_package_change(json_file_path)
+                    if new_hash != last_hash:
+                        last_hash = new_hash
+                        kill(process.pid)
+                        process = None
+                        break
```
masenf, line 103:
```suggestion
                        break  # for line in process.stdout
```

reflex/utils/exec.py
```diff
@@ -29,18 +73,38 @@ def run_process_and_launch_url(
...
+        if process is not None:
+            break
```
masenf, line 105:
```suggestion
            break  # while True
```

reflex/utils/exec.py
```diff
@@ -29,18 +73,38 @@ def run_process_and_launch_url(
...
+
+    print("App stopped")
```
masenf, line 107: we already log `[11:48:49] Reflex app stopped.` in `console.py:81`; so is this needed?

masenf (changes requested):
i don't see anywhere that version pinning is being handled; is that pushed to a future PR?

are we planning to take this for 0.2.5?

masenf:
I think we should hold off on this one until the next release because we don't want to lose our current library pins in `package.json`. With the new approach as currently written, we could end up bringing in new and possibly incompatible packages and the user has no way to opt-out of this, other than falling back to the previous version of reflex.

So lets get the library version pinning worked out, figure out how to handle conflicts between library pins, and test the heck out of this for the next next release.

Lendemor:
Need the changes in #1665 to pass test I think

reflex/compiler/utils.py
```diff
@@ -84,13 +88,23 @@ def compile_imports(imports: imports.ImportDict) -> List[dict]:
...
             assert rest is not None and len(rest) > 0, "No fields to import."
             for module in sorted(rest):
                 import_dicts.append(get_import_dict(module))
             continue
 
+        # remove the version before rendering the package imports
+        if lib.startswith("@") and lib.count("@") == 2:
+            lib = f"@{lib.split('@')[1]}"
+        elif not lib.startswith("@") and lib.count("@") == 1:
+            lib = lib.split("@")[0]
```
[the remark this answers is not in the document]
Lendemor: Thank for the suggestion, I like it!

reflex/utils/prerequisites.py
```diff
@@ -367,25 +367,32 @@ def install_bun():
...
-    packages = get_config().frontend_packages
+    # Install the custom packages, if any.
     if len(packages) > 0:
         process = processes.new_process(
             [get_install_package_manager(), "add", *packages],
             cwd=constants.WEB_DIR,
             shell=constants.IS_WINDOWS,
         )
-        processes.show_status("Installing custom frontend packages", process)
+        # processes.show_status("Installing custom frontend packages", process)
```
masenf, line 395:
```suggestion
        processes.show_status("Installing custom frontend packages", process)
```

This line is causing a race condition! Because the `bun add ...` command doesn't have `run=True` set, it starts the command and returns the running process object immediately. Usually bun is pretty fast, but if it's still installing packages after this function returns, then occasionally node will hit a compilation error due to missing imports.

`processes.show_status` is a blocking call that waits for the command to exit, so we previously were waiting here, but with this line commented out, the code is essentially starting the process and not waiting for it to finish.

masenf (commented):
Found the race condition; will push updated commit.

reflex/app.py
```diff
@@ -483,6 +483,27 @@ def setup_admin_dash(self):
 
             admin.mount_to(self.api)
 
+    def get_frontend_packages(self, imports: Dict[str, str]):
+        """Gets the frontend packages to be installed and filters out the unnecessary ones.
+
+        Args:
+            imports (Dict[str, str]): A dictionary containing the imports used in the current page.
```
picklelo, line 490: Don't need type annotations in the docstrings

reflex/utils/exec.py
```diff
@@ -25,20 +28,78 @@ def start_watching_assets_folder(root):
     asset_watch.start()
 
 
+def detect_package_change(json_file_path: str):
```
picklelo, line 31: add return type annotation

picklelo (dismissed):
Awesome, working well for me!

masenf (approved):

PR #1609, opened 2023-08-16
format_dict: only unescape quotes when removing outer quotes

masenf:
If the outer quotes are removed, then it's safe and necessary to remove the escapes on the inner quotes. Otherwise they should be left alone or they will break the double-quote terminated string.

Only remove double quote escapes if we are for sure removing outer quotes with a curly brace; a string containing the pattern `}"` will _never_ match this regex because the quote following the curly will always be escaped by the JSON encoder as `}\"`. So our regex here will only ever match a curly followed by JSON encoder's closing double quote (which is always unescaped).

This allows `rx.html` strings to contain double quotes and generally fixes weird quoting issue described in REF-226.

### DIFF
--- a/reflex/utils/format.py
+++ b/reflex/utils/format.py
@@ -473,10 +473,12 @@ def format_dict(prop: ComponentStyle) -> str:
     # Dump the dict to a string.
     fprop = json_dumps(prop)
 
+    def unescape_double_quotes_in_var(m: re.Match) -> str:
+        # Since the outer quotes are removed, the inner escaped quotes must be unescaped.
+        return re.sub('\\\\"', '"', m.group(1))
+
     # This substitution is necessary to unwrap var values.
-    fprop = re.sub('"{', "", fprop)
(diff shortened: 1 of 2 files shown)


masenf:
This blows up reflex-web with the following error

```
Server Error

Error: Invalid tag: {({node, ...props}) => <Text {...props}  />}
```

and that's all it gives... 🤔

masenf:
> This blows up reflex-web with the following error

The problem was exposed in the `<ReactMarkdown>` component, which was taking a big dictionary of props with Vars interspersed. This was being formatted as a long string, which itself had curly braces within it, so the moment the regex hit the closing curly brace `}`, it stopped matching and broke the formatter.

This was fixed by instead doing a non-greedy match on _anything_ between `"{` and `}"`, as long as the opening quote-brace was not preceeded by an escape character.

The latest change also switches to `re.VERBOSE` mode and adds comments for each element.

picklelo (approved):
awesome, great fix!

PR #1614, opened 2023-08-17
Check the yield/return from user defined event handlers are valid

martinxu9:
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

    b. Describe your changes.


reflex/app.py
```diff
@@ -707,6 +707,7 @@ def on_disconnect(self, sid):
         pass
 
     async def on_event(self, sid, data):
+        # TODO: should this be front-end websocket events below?
```
masenf, line 710: nit: either put this comment above the function declaration or after the docstring, not between

reflex/state.py
```diff
@@ -738,13 +738,43 @@ async def _process(self, event: Event) -> AsyncIterator[StateUpdate]:
...
+            )
+            or (
+                not isinstance(events, list)
+                and isinstance(events, (EventHandler, EventSpec))
+            )
+        ):
+            return events
+
+        raise TypeError(
+            f"Your handler {handler.fn.__qualname__} must only return/yield: None, Events or other EventHandlers"
```
masenf, line 768:
```suggestion
            f"Your handler {handler.fn.__qualname__} must only return/yield: None, Events or other EventHandlers referenced by their class (not using `self`)"
```

maybe give a bit more hint to users about how to reference other handlers.

reflex/state.py
```diff
@@ -738,13 +738,43 @@ async def _process(self, event: Event) -> AsyncIterator[StateUpdate]:
...
+        Raises:
+            TypeError: If any of the events are not valid.
+
+        Returns:
+            The events as they are if valid.
+        """
+        if (
+            events is None
+            or (
+                isinstance(events, list)
```
masenf, line 757:
```suggestion
                isinstance(events, (list, tuple))
```

i think it's valid for users to return / yield a tuple of events as well as a list.

to avoid the explicit `isinstance` check, you might structure it something like

```python
if events is None or isinstance(events, (EventHandler, EventSpec)):
    return events

try:
    if all(isinstance(e, (EventHandler, EventSpec)) for e in events):
        return events
except TypeError:
    pass  # events is not iterable

raise TypeError(...)
```

tests/test_state.py
```diff
@@ -1205,3 +1205,30 @@ def reset(self):
         err.value.args[0]
         == f"The event handler name `reset` shadows a builtin State method; use a different name instead"
     )
+
+
+@pytest.mark.asyncio
+async def test_state_with_invalid_yield():
```
masenf, line 1211:
```suggestion
def test_state_with_invalid_yield():
```

only needs to be `async` if you are awaiting within the test.

masenf (changes requested):

reflex/app.py
martinxu9: my bad, forgot to take this out. 🤦🏼‍♂️

martinxu9: https://github.com/reflex-dev/reflex/pull/1614/commits/b5c695b4057ef4f810015098257dcd71f0541fc7

reflex/state.py
martinxu9: this is a good idea!

martinxu9: I guess essentially an iterable of the valid types

martinxu9: how do you feel about check for events being `Iterable`?

masenf: is it possible to check for something being `Iterable` in a reliable way without calling `iter` on it?

martinxu9: I think check for being `Iterable` looks for the definition of the iter magic function. I tried defining a class with that function defined, then `isinstance` returns True. That seems to be the same as iterate inside a try/except block?

martinxu9: I realized after commenting. If it's a user defined class, the `__iter__` could be implemented wrong and still get `isinstance` to return True. 

I'll change it to the try/except way. That's definitely more reliable

masenf (approved):
🚢

Issue #822, opened 2023-04-14
File upload selected files preview


ElijahAhianyo:
currently, the dropzone doesn't show a preview of the selected files before they are uploaded. We should be able to show a preview

amcraig:
> I have found a workaround for this issue. It's hacky but it works.
> 
> ![image](https://user-images.githubusercontent.com/110377310/260053161-0018aea0-be76-46aa-865a-6493369539a2.png)
> 
> ```
> import reflex as rx
> 
> class State(rx.State):
>     img: list[str] = []
>     val: str = "new"
>     uploading: bool = False
> 
>     async def handle_upload(self, files: list[rx.UploadFile]):
>         for file in files:
>             upload_data = await file.read()
>             outfile = f"assets/{file.filename}"
> 
>             # Save the file.
>             with open(outfile, "wb") as file_object:
>                 file_object.write(upload_data)
> 
>             # Update the img var.
>             self.img.append(file.filename)
> 
>         self.uploading = False
>         self.val = "Finished uploading"
> 
>     async def start_vars(self):
>         self.uploading = True
>         self.val = "Uploading files"
> 
>     async def clear_state(self):
>         self.reset()
> 
> def index():
>     return rx.center(
>         rx.script("""
> setInterval(function() {
>   var elements = document.querySelectorAll('.chakra-input.css-1kp110w');
>   elements.forEach(function(element) {
>     element.style.display = 'flex';
>   });
> }, 500);
> """),
>         rx.vstack(
>             rx.text(State.val),
>             rx.upload( 
>                 border="1px dotted black",
>                 padding="1em",
>                 height="4.5em",
>                 no_drag=True,
>                 no_click=True,
>             ),
>             rx.cond(
>                 State.uploading,
>                 rx.button(
>                     is_loading=True,
>                     is_disabled=True,
>                 ),
>                 rx.button(
>                     "Upload",
>                     on_click=[
>                         State.start_vars(),
>                         State.handle_upload(rx.upload_files()),
>                     ],
>                 ),
>             ),
>             rx.button(
>                 "Reset",
>                 on_click=[
>                     State.clear_state(),
>                 ],
>             ),
>             rx.foreach(State.img, lambda img: rx.image(src=img)),
>         ),
>     )
> 
> app = rx.App(state=State)
> app.add_page(index)
> app.compile()
> ```

Is this all your code? I don't see a way how you are updating the upload box with the file name.

jq6l43d1:
[...]

PR #1616, opened 2023-08-17
Handle file uploads with component-local state

masenf:
* create component-local `useState` specifically to hold pending file uploads
  -> do not store `File` objects in the main backend state
* pass component-local `files` as `uploadFile` payload
  -> uploadFile handler in state.js no longer depends on state
* expose `selected_files` for frontend rendering (fix https://github.com/reflex-dev/reflex/issues/822)
* update `test_upload.py` to check `selected_files` in frontend

REF-234

### DIFF
--- a/integration/test_upload.py
+++ b/integration/test_upload.py
@@ -126,3 +126,49 @@ def test_upload_file(tmp_path, upload_file: AppHarness, driver):
     # check that the selected files are displayed
     selected_files = driver.find_element(By.ID, "selected_files")
     assert selected_files.text == exp_name
+
+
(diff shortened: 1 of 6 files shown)


picklelo (approved):
Works for me!

As a follow up - should we clear the current `selected_files` when the user finishes an upload?

masenf:
> should we clear the current `selected_files` when the user finishes an upload

the code we had before did not, but i think it'd be a nice thing to do; navigating away from the page would also accomplish the same. i was also thinking that we might expose something like

```python
clear_selected_files = BaseVar(name="_e => setFiles((files) => [])", type_=EventChain)
```

to allow downstream users to have that functionality as an event handler too.

filed REF-252 to track this.

PR #1623, opened 2023-08-18
Update .gitignore to ignore coverage.* files

martinxu9:


.gitignore
```diff
@@ -4,5 +4,5 @@ dist/*
 examples/
 .idea
 .vscode
-.coverage
-venv
\ No newline at end of file
+.coverage*
```
masenf, line 7:
```suggestion
.coverage
.coverage.*
```

probably don't want ignore `.coveragerc`, if we end up using that someday [[1]](https://coverage.readthedocs.io/en/latest/config.html?highlight=.coveragerc)

masenf (approved):

masenf (approved):

PR #1627, opened 2023-08-18
Add prop `debounce_timeout` to Input and TextArea default to 50

martinxu9:
## Summary
- Add prop `debounce_timeout` to input and textarea, default to 50 ms.

## Tests
- [x] Tested the 50ms default with Chat tutorial. With local dev, did not observe lags and loss of characters while attempting to type really fast in the input box.
- [x] CI green.

### DIFF
--- a/reflex/components/forms/input.py
+++ b/reflex/components/forms/input.py
@@ -50,6 +50,9 @@ class Input(ChakraComponent):
     # "lg" | "md" | "sm" | "xs"
     size: Var[str]
 
+    # Time in milliseconds to wait between end of input and triggering on_change
+    debounce_timeout: Var[int]
+
     def _get_imports(self) -> imports.ImportDict:
         return imports.merge_imports(
             super()._get_imports(),
@@ -84,7 +87,9 @@ class Input(ChakraComponent):
         if isinstance(props.get("value"), Var) and props.get("on_change"):
             # create a debounced input if the user requests full control to avoid typing jank
             return DebounceInput.create(
-                super().create(*children, **props), debounce_timeout=0
+                super().create(*children, **props),
+                # Currently default to 50ms, which appears to be a good balance
+                debounce_timeout=props.get("debounce_timeout", 50),
             )
         return super().create(*children, **props)
 
--- a/reflex/components/forms/textarea.py
+++ b/reflex/components/forms/textarea.py
@@ -43,6 +43,9 @@ class TextArea(ChakraComponent):
     # "outline" | "filled" | "flushed" | "unstyled"
     variant: Var[str]
 
+    # Time in milliseconds to wait between end of input and triggering on_change
+    debounce_timeout: Var[int]
+
     def get_controlled_triggers(self) -> Dict[str, Var]:
         """Get the event triggers that pass the component's value to the handler.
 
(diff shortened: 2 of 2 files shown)


picklelo (approved):
nice!

reflex/components/forms/input.py
```diff
@@ -50,6 +50,9 @@ class Input(ChakraComponent):
     # "lg" | "md" | "sm" | "xs"
     size: Var[str]
 
+    # Time in milliseconds to wait between end of input and triggering on_change
+    debounce_timeout: Var[int]
```
masenf, line 54: i probably would have just defined this in the `create` method, since it's not a real prop of `Input` or `TextArea`. how does the generated code end up looking?

martinxu9: @masenf  If it's not a prop only an arg to `create` method, is it still exposed to the user through docs? I intended to keep that as sort of a last resort knob in case our 50ms doesn't work with certain user setups.
The generated code is like this (part of the chat tutorial with some new code on how to clear the text):

```
</Box>
  <HStack>
  <DebounceInput debounceTimeout={50} element={Input} onChange={_e => Event([E("state.chatapp_state.set_question", {value:_e.target.value})], _e)} placeholder={`Ask a question`} sx={{"borderWidth": "1px", "padding": "1em", "boxShadow": "rgba(0, 0, 0, 0.15) 0px 2px 8px"}} type={`text`} value={state.chatapp_state.question}/>
  <Button onClick={_e => Event([E("state.chatapp_state.answer2", {})], _e)} sx={{"bg": "#CEFFEE", "boxShadow": "rgba(0, 0, 0, 0.15) 0px 2px 8px"}}>
  {`Ask`}
</Button>
</HStack>
</Container>
</Center>
</Box>
```

masenf: Does the generated code look okay for non-controlled inputs that set debounce_timeout though?

I'm concerned that the Input component would get rendered with an unknown prop that might raise a console warning.

masenf: @martinxu9 this is the code that concerns me

```python
import reflex as rx


class State(rx.State):
    def on_change(self, value):
        print(value)


def index() -> rx.Component:
    return rx.vstack(
        rx.input(debounce_timeout=300, on_change=State.on_change),
        padding_top="10%",
    )


app = rx.App()
app.add_page(index)
app.compile()
```

and it generates this code

```javascript
<Input debounceTimeout={300} onChange={_e => Event([E("state.on_change", {value:_e.target.value})], _e)} type={`text`}/>
```

which throws this console error (and doesn't actually debounce the input)

```console
Warning: React does not recognize the `debounceTimeout` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `debouncetimeout` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

This is why i think it's a bad idea to define `debounce_timeout` as a prop on `Input`, because it's _not_ a prop on `Input`. If it was passed to the `create` constructor instead, then the logic could be changed to

```python
if (isinstance(props.get("value"), Var) and props.get("on_change")) or props.get("debounce_timeout") is not None:
    debounce_timeout = props.pop("debounce_timeout", 50)
    return DebounceInput.create(
        super().create(*children, **props), debounce_timeout=debounce_timeout
    )
```
(and note here, explicitly `pop`ing `debounce_timeout` from `props`, so it's not used when creating the element.)

martinxu9: makes sense. Thanks for catching this.

PR #1636, opened 2023-08-20
Implement `on_mount` and `on_unmount` for all components.

masenf:
# All components expose `on_mount` and `on_unmount` triggers

These are tied directly into the react lifecycle hook `useEffect`. The `on_mount` is called as the main part of the effect, with `on_unmount` chain being called as the cleanup function. The ordering of when these are called for different components is non-deterministic. In dev mode, these triggers will fire twice, because react calls all `useEffect` hooks twice in StrictMode.

REF-231

# Expose `format_var` and `format_event_chain`

Previously, these were internal to `Tag.format_props`, but exposing them here allows them to be reused in other contexts.

Partial fix for REF-209, but leaving open until the `Tag` code can be refactored to use the new methods.

# Split `_get_hooks` into `_get_hooks_internal`

Make it easier to downstream components to override `_get_hooks` without having to worry about screwing up Reflex's ref and on_mount handling behavior.


reflex/components/component.py
```diff
@@ -279,7 +279,11 @@ def get_triggers(self) -> Set[str]:
         Returns:
             The event triggers.
         """
-        return EVENT_TRIGGERS | set(self.get_controlled_triggers())
+        return (
+            EVENT_TRIGGERS
+            | set(self.get_controlled_triggers())
+            | set((constants.ON_MOUNT, constants.ON_UNMOUNT))
```
picklelo, line 285: Shouldn't we just add these to the default `EVENT_TRIGGERS` ?

reflex/utils/format.py
```diff
@@ -182,6 +183,24 @@ def format_string(string: str) -> str:
     return string
 
 
+def format_var(var: Var) -> str:
```
picklelo, line 186: Shouldn't this be the same as the Var.__str__ method? Should we move this code there? This was the intention of that method.

reflex/utils/format.py
```diff
@@ -311,6 +330,46 @@ def format_event(event_spec: EventSpec) -> str:
...
+    if isinstance(event_chain, Var):
+        from reflex.event import EventChain
+
+        if event_chain.type_ is not EventChain:
+            raise ValueError(f"Invalid event chain: {event_chain}")
+        return "".join(
+            [
+                "(() => {",
+                format_var(event_chain),
+                f"; preventDefault({format_var(event_arg)})" if event_arg else "",
```
picklelo, line 358: The `Event` function in `state.js` should also handle the `preventDefault` logic I thought?

picklelo (commented):
Works great for me, some minor comments.

I think we should replace the `on_load` handler with this once it's merged, and deprecate that way. One less thing to handle in the hydration. (Or are there any limitations to that?)

reflex/utils/format.py
masenf: if we're formatting an arbitrary var as an event chain, there's no guarantee that `Event` gets called at all; it could be anything.

picklelo:
We also need to update `reflex-web` to add docs for these new event triggers to pass the integration tests.

reflex/components/component.py
masenf: i went back and forth on this, and ultimately did it this way because these aren't "real" event triggers that come from browser events, but are synthetic events that reflex handles.

reflex/utils/format.py
masenf: hmm, it's a little different. i cribbed this from `Tag.format_prop`, so i assumed it must be doing something special. i'll see if i can combine them without breaking tests, so we can just use `str(var)` instead.

masenf:
> We also need to update reflex-web to add docs for these new event triggers to pass the integration tests.

I meant to get to it today, but it got away from me. Definitely will have to land that before this can come in.

PR #1640, opened 2023-08-21
state: _init_event_handlers recursively

masenf:
Allow event handlers defined in parent states to be directly called from substates.

Previously, only event handlers defined on the substate could be called on the substate. A workaround for queueing event handlers defined in a parent state was to explicitly `return`/`yield` the event handler/spec (`return State.handlerA()`)

```python
import reflex as rx


class State(rx.State):
    def handlerA(self):
        print("handlerA")


class Substate(State):
    def handlerB(self):
        self.handlerA()  # does NOT work before this change
        print("handlerB")
        self.handlerC()
        return State.handlerA()  # does work

    def handlerC(self):
        print("handlerC")


def index() -> rx.Component:
    return rx.vstack(
        rx.button("click", on_click=Substate.handlerB),
        padding_top="10%",
    )


app = rx.App()
app.add_page(index)
app.compile()
```

Before this change, the output is

```console
handlerB
handlerC
handlerA
```

After this change, the output is

```console
handlerA
handlerB
handlerC
handlerA
```

Reported in [discord](https://discord.com/channels/1029853095527727165/1061874061250150441/1143022149657960509) by community member **Samsam**

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -106,11 +106,7 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
         for substate in self.get_substates():
             self.substates[substate.get_name()] = substate(parent_state=self)
         # Convert the event handlers to functions.
-        for name, event_handler in self.event_handlers.items():
-            fn = functools.partial(event_handler.fn, self)
-            fn.__module__ = event_handler.fn.__module__  # type: ignore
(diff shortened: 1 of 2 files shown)


picklelo (approved):
nice!

PR #1645, opened 2023-08-21
remove prop `debounce_timeout` from input and textarea

### LINKED ISSUE #1641: Remove `debounce_timeout` prop from `Input` and `TextArea` components
See https://github.com/reflex-dev/reflex/pull/1627#discussion_r1299594001 for detailed explanation and suggested fix.

----------------

`debounce_timeout` isn't a real prop on `Input` or `TextArea` and should **never** be rendered into the JSX for these chakra components.

Instead, this value should be handled specially in the `Input.create` constructor when determining whether to use a real `Input` or `DebounceInput` wrapped `Input`

danik292:
### All Submissions:

### Type of change

fixes #1641

### DIFF
--- a/reflex/components/forms/input.py
+++ b/reflex/components/forms/input.py
@@ -81,8 +81,10 @@ class Input(ChakraComponent):
         Returns:
             The component.
         """
-        if (isinstance(props.get("value"), Var) and props.get("on_change")) or props.get("debounce_timeout") is not None:
-            # Currently default to 50ms, which appears to be a good balance 
+        if (
+            isinstance(props.get("value"), Var) and props.get("on_change")
+        ) or "debounce_timeout" in props:
+            # Currently default to 50ms, which appears to be a good balance
             debounce_timeout = props.pop("debounce_timeout", 50)
             # create a debounced input if the user requests full control to avoid typing jank
             return DebounceInput.create(
--- a/reflex/components/forms/input.py
+++ b/reflex/components/forms/input.py
@@ -81,7 +81,9 @@ class Input(ChakraComponent):
         Returns:
             The component.
         """
-        if isinstance(props.get("value"), Var) and props.get("on_change"):
+        if (isinstance(props.get("value"), Var) and props.get("on_change")) or props.get("debounce_timeout") is not None:
+            # Currently default to 50ms, which appears to be a good balance 
+            debounce_timeout = props.pop("debounce_timeout", 50)
             # create a debounced input if the user requests full control to avoid typing jank
             return DebounceInput.create(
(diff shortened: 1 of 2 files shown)


reflex/components/forms/input.py
```diff
@@ -87,9 +84,7 @@ def create(cls, *children, **props) -> Component:
         if isinstance(props.get("value"), Var) and props.get("on_change"):
```
masenf, line 84:
```suggestion
        if (isinstance(props.get("value"), Var) and props.get("on_change")) or props.get("debounce_timeout") is not None:
            # Currently default to 50ms, which appears to be a good balance 
            debounce_timeout = props.pop("debounce_timeout", 50)
```

reflex/components/forms/input.py
```diff
@@ -87,9 +84,7 @@ def create(cls, *children, **props) -> Component:
...
             # create a debounced input if the user requests full control to avoid typing jank
             return DebounceInput.create(
-                super().create(*children, **props),
-                # Currently default to 50ms, which appears to be a good balance
-                debounce_timeout=props.get("debounce_timeout", 50),
+               super().create(*children, **props), debounce_timeout=0
```
masenf, line 87:
```suggestion
               super().create(*children, **props), debounce_timeout=debounce_timeout
```

masenf (changes requested):
we still want to respect `debounce_timeout` if the caller provides it; it just shouldn't be part of the actual props for the component, since it is only recognized by the `DebounceInput`

danik292:
reqested chenges done So you meybe can marge. 👍

masenf (changes requested):
> reqested chenges done

can you apply similar change to the `TextArea` as well 🙏 

Definitely appreciate the quick response, this was one of two things blocking our release today!

danik292:
@masenf  how to add commit here without new PR

I am using VS code and github desktop.

masenf:
@danik292 yeah you can just `git pull` your branch to bring in the changes made on the web interface, then make your commit locally and `git push` to your `main` branch

masenf:
@danik292 hmm, still not seeing changes here. are you having trouble?

danik292:
@masenf i was taking bath.

PS. Are your reqeested changes same for second file same??

masenf:
> Are your reqeested changes same for second file

yeah, since they both use the same structure, they need symmetrical interfaces.

masenf (approved):
@danik292 i just had to push the change here, this issue is blocking the release of 0.2.5 so we need to just get it in.

picklelo (approved):

Issue #1588, opened 2023-08-14
Better Message Menu_Item


Alek99:
(Note when we switch component libraries, this may no longer be an issue putting here so we don't forget to check)
https://discord.com/channels/1029853095527727165/1140692663536996442/1140699922732175450

Problem: Menu_Item error when it has children.
```python
import reflex as rx


class State(rx.State):
    pass

def index():
    return rx.menu(
        rx.menu_button("this is a menu"),
        rx.menu_list(
            rx.menu_item("this is a regular item"),
            rx.menu_item(rx.button("this is a button in a menu item"))
        )
    )

app = rx.App()
app.add_page(index)
app.compile()
```

![](https://cdn.discordapp.com/attachments/1061874061250150441/1140530514403541022/image.png)

Catch this at compile time and throw a better error message if this persists

PR #1647, opened 2023-08-21
Validate component children

### LINKED ISSUE #1648: Valid children of component
We currently have the `invalid_children` prop to validate invalid components, we also need a valid-only prop field (`valid_children`)

ElijahAhianyo:
This PR allows a component to explicitly state only allowed child components

usage:
```python
         class TestComponent5(Component):
              tag = "RandomComponent"
      
              invalid_children: List[str] = ["Table"]
      
              valid_children: List[str] = ["Text", "Box"]

```

closes #1648 #1588

### DIFF
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -107,9 +107,10 @@ class Component(Base, ABC):
             TypeError: If an invalid prop is passed.
         """
         # Set the id and children initially.
+        children = kwargs.get("children", [])
         initial_kwargs = {
             "id": kwargs.get("id"),
-            "children": kwargs.get("children", []),
+            "children": children,
             **{
                 prop: Var.create(kwargs[prop])
                 for prop in self.get_initial_props()
@@ -117,6 +118,8 @@ class Component(Base, ABC):
             },
         }
         super().__init__(**initial_kwargs)
+
+        self._validate_component_children(children)
 
         # Get the component fields, triggers, and props.
         fields = self.get_fields()
@@ -385,6 +388,7 @@ class Component(Base, ABC):
             else Bare.create(contents=Var.create(child, is_string=True))
             for child in children
         ]
+
         return cls(children=children, **props)
 
     def _add_style(self, style):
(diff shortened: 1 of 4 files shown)


danik292 (approved):

Lendemor (approved):
Looks good to me.

masenf (approved):
can you also add

* `Text` and `Center` as invalid children of `Text`
* `Tr` as the only valid child of `Thead`, `Tbody`, and `Tfoot`
* `Thead`, `Tbody`, and `Tfoot` as the only valid children of `Table`

This should help us reduce incidence of hydrations errors from silly mistakes

reflex/components/component.py
```diff
@@ -435,30 +443,44 @@ def render(self) -> Dict:
             ),
             autofocus=self.autofocus,
         )
-        self._validate_component_children(
```
picklelo, line 438: should we keep the check here? Only reason is because if they somehow added children dynamically after creating their component, then we wouldn't catch it. Though I don't think I've seen people really do that...

reflex/components/component.py
```diff
@@ -435,30 +443,44 @@ def render(self) -> Dict:
...
-        if not self.invalid_children:
+        if not self.invalid_children and not self.valid_children:
             return
-        for child in children:
-            name = child["name"]
-            if name in self.invalid_children:
+
+        comp_name = type(self).__name__
+
+        def validate_invalid_child(child_name):
```
picklelo, line 460: nit: we could inline these functions ore define them outside of the function. I've noticed before in Python that defining functions within tight loops is often a performance issue, and these checks will be called many times per compile.

picklelo (approved):
Looks good! Some comments but I don't think they're critical, we can come back to them in a follow up

reflex/components/component.py
```diff
@@ -66,6 +66,10 @@ class Component(Base, ABC):
 
     # components that cannot be children
     invalid_children: List[str] = []
+
+    # components that are only allowed as children
+    valid_children: List[str] = []
```
picklelo, line 71: I'm also thinking in a follow up if these should maybe be `List[Type[Component]]` rather than `str`

PR #1649, opened 2023-08-21
Allow decorators on event handlers

picklelo:
TODO: add test code here


picklelo:
This PR is not needed - decorators are working currently already.

PR #1650, opened 2023-08-21
Use DebounceInput wrapper for fully controlled Editable

martinxu9:
## Summary
- When it is fully controlled or user specifies `debounce_time`, wrap Editable inside DebounceInput. Similar to Input and TextArea.
- Add UT for Editable.
## Tests
- [x] New UT for Editable pass.
- [x] CI green.

### DIFF
--- a/reflex/components/forms/editable.py
+++ b/reflex/components/forms/editable.py
@@ -56,10 +56,11 @@ class Editable(ChakraComponent):
             isinstance(props.get("value"), Var) and props.get("on_change")
         ) or "debounce_timeout" in props:
             # Create a debounced input if the user requests full control to avoid typing jank
+            # Currently default to 50ms, which appears to be a good balance
+            debounce_timeout = props.pop("debounce_timeout", 50)
             return DebounceInput.create(
                 super().create(*children, **props),
-                # Currently default to 50ms, which appears to be a good balance
-                debounce_timeout=props.pop("debounce_timeout", 50),
+                debounce_timeout=debounce_timeout,
             )
         return super().create(*children, **props)
 
--- a/reflex/components/forms/editable.py
+++ b/reflex/components/forms/editable.py
@@ -2,6 +2,8 @@
 
 from typing import Dict
 
+from reflex.components.component import Component
+from reflex.components.forms.debounce import DebounceInput
 from reflex.components.libs.chakra import ChakraComponent
 from reflex.event import EVENT_ARG
 from reflex.vars import Var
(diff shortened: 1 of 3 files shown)


reflex/components/forms/editable.py
```diff
@@ -36,6 +38,31 @@ class Editable(ChakraComponent):
...
+            The component.
+        """
+        if (
+            isinstance(props.get("value"), Var) and props.get("on_change")
+        ) or "debounce_timeout" in props:
+            # Create a debounced input if the user requests full control to avoid typing jank
+            return DebounceInput.create(
+                super().create(*children, **props),
+                # Currently default to 50ms, which appears to be a good balance
+                debounce_timeout=props.pop("debounce_timeout", 50),
```
masenf, line 62: have to `pop` the prop before passing off to `super().create` on line 60

reflex/components/forms/editable.py
```diff
@@ -36,6 +38,31 @@ class Editable(ChakraComponent):
     # The initial value of the Editable in both edit and preview mode.
     default_value: Var[str]
 
+    # Time in milliseconds to wait between end of input and triggering on_change
+    debounce_timeout: Var[int]
```
masenf, line 42:
```suggestion
```

should avoid defining this as an actual prop on `Editable`, since the `Editable` component doesn't recognize it.

masenf (changes requested):

reflex/components/forms/editable.py
martinxu9: Thanks for making this suggestion again. I just looked at how DebounceInput is a wrapper of an equivalent react component, I get what you're saying about real props. Now this is what I think. As a non-react developer, I'm thinking about the behavior I want with my component rather than if a certain arg maps to a legit prop in the react component's or not. Personally I wouldn't go out of my way to find that out if I'm using reflex. I don't feel it needs to be a prop here. What I want to do is since we are not 100% confident on the default 50ms, there should be the option to give users. Of course if we just add a non-prop argument in the `create` method, it can create debounced wrappers, but do we expose docs on that, or we pretty much just tell users about the props.

reflex/components/forms/editable.py
martinxu9: oh oops, this happens after the super.create, I'll fix it.

martinxu9: fixed in https://github.com/reflex-dev/reflex/pull/1650/commits/80ce5fb85e003d5872043d5fb943bd77b77c6833

reflex/components/forms/editable.py
masenf: > Of course if we just add a non-prop argument in the create method, it can create debounced wrappers, but do we expose docs on that, or we pretty much just tell users about the props.

There are many components in reflex that special case kwargs to the `create` function that are not actual props, so there's precedent for this type of structure.

See some examples in
* reflex/components/datadisplay/code.py
* reflex/components/datadisplay/table.py
* reflex/components/disclosure/tabs.py
* reflex/components/forms/slider.py

I think how we document it is important. Currently those other components with special handling i think just provide examples of use, without fully documenting the special cases, which is problematic.

At least on the `Input`, `TextArea` and `Editable` docs, we can mention the `debounce_timeout` parameter and link to the `DebounceInput` docs for a full explanation.

martinxu9: That sounds good!

martinxu9: removed here https://github.com/reflex-dev/reflex/pull/1650/commits/18ff4992ef065c32a657b50bb01c141c88010ad4, l'm also working on updating the docs/tutorial, will add it there.

masenf (approved):

PR #1652, opened 2023-08-21
ReflexList: reassign field on `insert`

masenf:
Seems this operation was just missing, cannot think of a good reason why it was not included.


picklelo (approved):
good catch

PR #1653, opened 2023-08-22
Run backend from the main thread in dev mode

masenf:
This allows debuggers to hook into the uvicorn subprocess and enable breakpoints in the state and event handlers.

### DIFF
--- a/reflex/utils/processes.py
+++ b/reflex/utils/processes.py
@@ -146,7 +146,7 @@ def new_process(args, run: bool = False, show_logs: bool = False, **kwargs):
 
 
 @contextlib.contextmanager
-def run_concurrently(
+def run_concurrently_context(
     *fns: Union[Callable, Tuple]
 ) -> Generator[list[futures.Future], None, None]:
     """Run functions concurrently in a thread pool.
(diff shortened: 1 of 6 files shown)


reflex/utils/processes.py
```diff
@@ -145,21 +145,34 @@ def new_process(args, run: bool = False, show_logs: bool = False, **kwargs):
...
     Args:
         *fns: The functions to run.
+
+    Yields:
+        The futures for the functions.
     """
     # Convert the functions to tuples.
     fns = [fn if isinstance(fn, tuple) else (fn,) for fn in fns]  # type: ignore
 
+    # If no functions are provided, yield an empty list and return.
```
picklelo, line 163: nit: this can be the first line in the function

reflex/utils/processes.py
```diff
@@ -145,21 +145,34 @@ def new_process(args, run: bool = False, show_logs: bool = False, **kwargs):
     return fn(args, **kwargs)
 
 
-def run_concurrently(*fns: Union[Callable, Tuple]):
+@contextlib.contextmanager
+def run_concurrently(
+    *fns: Union[Callable, Tuple]
```
picklelo, line 150: Should we add a param indicating if they want to return a generator? I think in the other cases, it will make them do an `with` and `pass`. Maybe the generator option should be the opt-in with the param?

picklelo (commented):
This seems to break the message printed on ctrl+c about the Reflex app stopping

reflex/utils/processes.py
masenf: i don't think it's possible to have a function that returns normally OR is a generator; you kind of have to pick one or the other.

what i can do though is rename this function as `run_concurrently_context`, and then make a new function with the same name `run_concurrently` that just does the `with ...: pass` internally.

what do you think of that?

personally I don't have a problem if this function is always a generator, but keeping the original name as a normal function would result in less collateral changes to unrelated parts of the code.

masenf (commented):
I changed the `signal.signal` handler to `atexit.register` and it seems to work more reliably for printing the message.

reflex/utils/processes.py
picklelo: I think this works great

picklelo (approved):
Looks good to me!

PR #1660, opened 2023-08-22
add type conversion for int,float in built-in setters

### LINKED ISSUE #336: number_input component's on_change event handler sets the state variable as string
**Describe the bug**
Looks like the on_change event handler of number_input component sets the state variable as string instead of int. 

```
min_yield <class 'int'> 50
wait  - compiling /...
event - compiled client and server successfully in 270 ms (853 modules)
min_yield <class 'int'> 50
min_yield <class 'int'> 50
<<<< The value is updated in the UI
min_yield <class 'str'> 51
  File "[REDACTED]app.py", line 36, in df
    (df['y'] > self.min_yield / 10)
TypeError: unsupported operand type(s) for /: 'str' and 'int'
```

**To Reproduce**
 - Code/Link to Repo:

**Expected behavior**
The event handler should set the state variable as int.

**Screenshots**

 - Python Version: 3.9.13
 - Pynecone Version: 0.1.13
 - OS: MacOS
 - Browser (Optional): Firefox
 
**Additional context**

Lendemor:
### All Submissions:

### Type of change

### Changes To Core Features:

### Description

Convert value to the expected type for built-in setter generated by Reflex.

If you have:
```
class State(rx.State):
    foo: int = 0

def index():
    return rx.number_input(value=State.foo, on_change=State.set_foo)
```

The value given to the setter `set_foo` is usually `str`, so the setter will attempt to cast to `int`

Closes #336

### DIFF
--- a/reflex/vars.py
+++ b/reflex/vars.py
@@ -839,6 +839,8 @@ class BaseVar(Var, Base):
                     console.warn(
                         f"{self.name}: Failed conversion of {value} to '{self.type_.__name__}'. Value not set.",
                     )
+            else:
+                setattr(state, self.name, value)
 
         setter.__qualname__ = self.get_setter_name()
 


reflex/vars.py
```diff
@@ -831,7 +831,16 @@ def setter(state: State, value: Any):
                 state: The state within which we add the setter function.
                 value: The value to set.
             """
-            setattr(state, self.name, value)
+            if self.type_ in [int, float]:
```
masenf, line 834:
```suggestion
            if self.type_ in [int, float, bool]:
```

`bool` too maybe?

masenf (approved):

PR #1662, opened 2023-08-22
model: only pass "check_same_thread" arg for sqlite database

### LINKED ISSUE #1661: sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) invalid dsn: invalid connection option "check_same_thread"
**Describe the bug**
Using a non-sqlite database fails.

**To Reproduce**
Use any non-sqlite database with reflex

 - Python Version: 3.11.4
 - Reflex Version: 0.2.5
 - OS: -
 - Browser (Optional): -
 
**Additional context**
From discord: https://discord.com/channels/1029853095527727165/1061874061250150441/1143574299425386671

masenf:
Fix #1661

### DIFF
--- a/reflex/model.py
+++ b/reflex/model.py
@@ -43,12 +43,9 @@ def get_engine(url: Optional[str] = None):
         )
     # Print the SQL queries if the log level is INFO or lower.
     echo_db_query = os.environ.get("SQLALCHEMY_ECHO") == "True"
-    return sqlmodel.create_engine(
-        url,
-        echo=echo_db_query,
-        # Needed for the admin dash.
-        connect_args={"check_same_thread": False},
-    )
+    # Needed for the admin dash on sqlite.
+    connect_args = {"check_same_thread": False} if url.startswith("sqlite") else {}
+    return sqlmodel.create_engine(url, echo=echo_db_query, connect_args=connect_args)
 
 
 class Model(Base, sqlmodel.SQLModel):


picklelo (approved):

reflex/model.py
```diff
@@ -43,12 +43,9 @@ def get_engine(url: Optional[str] = None):
...
     # Print the SQL queries if the log level is INFO or lower.
     echo_db_query = os.environ.get("SQLALCHEMY_ECHO") == "True"
-    return sqlmodel.create_engine(
-        url,
-        echo=echo_db_query,
-        # Needed for the admin dash.
-        connect_args={"check_same_thread": False},
-    )
+    # Needed for the admin dash on sqlite.
+    connect_args = {"check_same_thread": False} if url.startswith("sqlite") else {}
```
picklelo, line 47: Just saw the original line was:
```python
connect_args={"check_same_thread": False} if conf.admin_dash else {}
```
But we don't have the admin_dash in config anymore. Should we still check that arg? I'm not sure what this param does actually.

picklelo (commented):
Just saw the original

reflex/model.py
masenf: in sqlite, you get an exception if different threads use the same connection; and admin_dash runs on a different thread, so this param ignores that check.

PR #1664, opened 2023-08-22
Remove need for User to call `rx.App.compile()`

martinxu9:
## Summary
- Rename `App.compile` to `App._compile` to indicate private methods
- Call the new `App._compile` after importing the app in the run_frontend. Do this for both reflex CLI and AppHarness.
- Remove the use of `App.compile` from code examples and tests.
- Add back `App.compile` which only prints deprecation warning, the method is no-op.
## Tests
- [ ] CI green

### DIFF
--- a/reflex/testing.py
+++ b/reflex/testing.py
@@ -158,8 +158,8 @@ class AppHarness:
             # ensure config is reloaded when testing different app
             reflex.config.get_config(reload=True)
             self.app_module = reflex.utils.prerequisites.get_app()
+            self.app_module.app._compile()
         self.app_instance = self.app_module.app
-        self.app_instance._compile()
 
     def _start_backend(self):
         if self.app_instance is None:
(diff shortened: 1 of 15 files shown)


reflex/testing.py
```diff
@@ -159,6 +159,7 @@ def _initialize_app(self):
             reflex.config.get_config(reload=True)
             self.app_module = reflex.utils.prerequisites.get_app()
         self.app_instance = self.app_module.app
+        self.app_instance._compile()
```
martinxu9, line 162: @masenf this doesn't seem to be correct or enough, any idea what I should do instead?

masenf: nest these calls under that `chdir` context manager above; all of the compiler code uses relative file paths, so if you're not in the app directory, then the files get compiled to the working dir

martinxu9: ah that makes sense. Thanks!

martinxu9:
Based on internal discussion, since we rely on uvicorn's reload capability to reload upon code change (the compile is re-run when imported) and uvicorn does not have a reload hook to append custom logic, we will keep the `App.compile` in the user code for now.

PR #1665, opened 2023-08-23
Avoid compile step when running production backend

masenf:
When executing `reflex run --env prod`, set `REFLEX_SKIP_COMPILE=yes` and then during `App.compile`, exit early if this env var is set. This avoids running the compile needlessly, since the prod backend does NOT use the compiled output for anything.

In DEV mode, we must keep the compile step, because uvicorn re-importing the app module and triggering the compile on the backend is what triggers the NextJS frontend to rebuild.

REF-160


picklelo (commented):
Somehow this is still compiling when I run `reflex run --env prod --backend-only`

picklelo (approved):
What's the reason we can't apply this to backend only in dev mode also?

Also, should we print a message similar to "Your backend is running at <backend_url>"?

We can add those in a follow up if necessary.

PR #1668, opened 2023-08-23
Refactor zipping in `reflex export`

### LINKED ISSUE #1285: `reflex export` exit code and log messages are inaccurate

### LINKED ISSUE #1506: reflex export should fail if zipping fails
We call `os.system` to create fronted.zip or backend.zip (or both).

If zip is not available, reflex export terminates with exit code 0.  This should not be the case.

We should take this opportunity to simply migrate this logic to [Python](https://docs.python.org/3/library/zipfile.html) itself.  This removes the dependency on `zip` CLI altogether and will be fully portable across OSes by design.  When we do this migration, we should ensure we propagate any errors and failures properly to `reflex` CLI invoker.

martinxu9:
## Summary
- Fixes #1506 .
- Fixes #1285 .
- Refactor zipping code in the `reflex export` command.
- Use zipfile library to make the zip instead of opening up an external sub-shell, this also unifies the code for windows and non-windows.
- Add separate progress bars for backend and frontend zip; remove the console.log as it is not consistent with rest of the CLI output.
- Slightly modified the logic what files to exclude: 
  - Previously:
    - 1) `assets*`, `frontend.zip*`, `backend.zip*` are excluded 
    - 2) `__pycache__` included
  - Now:
    - 1) `assets` folder, `frontend.zip`, `backend.zip` are excluded
    - 2) `__pycache__` excluded
## Tests
- [x] Manual testing export on the `reflex-web` and compare the files in the frontend and backend zips.
- Progress bars added for zipping: 
image

### DIFF
--- a/reflex/utils/build.py
+++ b/reflex/utils/build.py
@@ -6,8 +6,12 @@ import json
 import os
 import random
 import subprocess
+import zipfile
+from enum import Enum
 from pathlib import Path
 from typing import Optional, Union
+
+from rich.progress import MofNCompleteColumn, Progress, TimeElapsedColumn
 
 from reflex import constants
 from reflex.config import get_config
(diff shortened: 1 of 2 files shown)


reflex/utils/build.py
```diff
@@ -85,6 +89,64 @@ def generate_sitemap_config(deploy_url: str):
...
+        MofNCompleteColumn(),
+        TimeElapsedColumn(),
+    )
+    task = progress.add_task(
+        f"Zipping {component_name.value}:", total=len(files_to_zip)
+    )
+
+    with progress, zipfile.ZipFile(target, "w", zipfile.ZIP_DEFLATED) as zipf:
+        for file in files_to_zip:
+            progress.advance(task)
```
masenf, line 146:
```suggestion
            console.debug(f"{target}: {file}")
            progress.advance(task)
```

might be nice to log the files that are getting included when `--loglevel debug` is used.

masenf (dismissed):
just a comment for your consideration. otherwise looks good to me.

martinxu9:
Approve one more time please @masenf

masenf (approved):
✅

PR #1670, opened 2023-08-23
Revert "Use DebounceInput wrapper for fully controlled Editable"

martinxu9:
Reverts reflex-dev/reflex#1650

Didn't really work after trying a fully controlled case with editable.


masenf (approved):
Bummer; but it happens

martinxu9:
The approval was lost, one more time please @masenf

masenf (approved):

Issue #1671, opened 2023-08-24
rx.cond is not defined in reflex/reflex/components/__init__.py


nevdelap:
**Describe the bug**
rx.cond is not defined in reflex/reflex/components/__init__.py

**To Reproduce**
rx.cond is not defined in reflex/reflex/components/__init__.py

**Expected behavior**
mypy doesn't complain about it.

**Screenshots**
N/A

**Specifics (please complete the following information):**
 - Python Version: 3.11
 - Reflex Version: 0.2.5
 - OS: Linux
 - Browser (Optional):
 
**Additional context**
I'll put up a PR.

PR #1672, opened 2023-08-24
Fix rx.cond is not defined in reflex/reflex/components/__init__.py (#1671)

nevdelap:
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

### Other

I think that file needs sorting, but I didn't because it would make my one liner hard to review.

### DIFF
--- a/reflex/components/__init__.py
+++ b/reflex/components/__init__.py
@@ -157,6 +157,7 @@ voronoi = Voronoi.create
 box = Box.create
 center = Center.create
 circle = Circle.create
+cond = Cond.create
 container = Container.create
 flex = Flex.create
 foreach = Foreach.create


reflex/components/__init__.py
```diff
@@ -157,6 +157,7 @@
 box = Box.create
 center = Center.create
 circle = Circle.create
+cond = Cond.create
```
picklelo, line 160: The `cond` function is a bit unique, as it can either return `Cond.create` if the arguments are components, or something else when using conditional props. For example in the case of `rx.text(color=rx.cond(State.foo, "blue", "green"))`.

You can see the function defined here: https://github.com/reflex-dev/reflex/blob/76b8af3b42a618939c1fd7dd8af5dfaad9f5906b/reflex/components/layout/cond.py#L81

So we are importing this function already at the top when we do `from .layout import *`, so there's no need to redefine it again here.

picklelo (commented):
Thanks for adding - but I don't think this change is necessary. Is there some issue you ran into that you needed this?

reflex/components/__init__.py
nevdelap: Hi @picklelo. It is because mypy complains about it like this...

image

image

nevdelap: I like to not have mypy complain about anything. :wink:

nevdelap: I will have a look at what mypy is making of cond.py/def conf() and get back to you.

nevdelap:
Hi @picklelo. I can see that `rx.cond` is `layout.py/def cond()` as you say and that mypy is maybe making a false positive or something, so I'll close this off, and I'll put a `# type: ignore` on it until it annoys me enough to look further at what is going on. Thanks for the explanation too!

PR #1674, opened 2023-08-24
Type Validation for Var Operations and Enhanced Compatibility

ElijahAhianyo:
### Description
This pull request addresses an issue where Var operations were previously allowed between any types, which led to unintended behaviors and potential edge cases. The purpose of this pull request is to align Var operations more closely with Python's type system and enforce type safety, especially considering that the codebase involves rendering operations in their JavaScript equivalents.

### Changes Made
* Added type checks and validation to all Var operations to ensure compatibility with Python's type system.
* Refactored the code to use Python-specific type annotations and appropriate type hints.
* Created type-specific checks for different operations, preventing unintended and potentially unsafe operations.
* Introduced stricter type validation based on Python's type compatibility to prevent issues that could arise due to the permissive nature of JavaScript's type coercion.

### Motivation
In the context of our codebase, which primarily involves writing code in Python, it's important to ensure that the operations performed between State Vars are compatible with Python's type rules and restrictions.
By limiting operations to be more Python-specific, we can avoid unexpected behaviors, potential type errors, and edge cases that might arise from JavaScript's permissive type system.
This change is intended to enhance the reliability and maintainability of the codebase while maintaining consistency with Python's type-safe practices.

### Impact
This change enforces stricter type validation for Var operations, which might result in some previously valid operations being flagged as errors. However, these changes are in line with Python's type-safe philosophy and aim to prevent unexpected behaviors.
Developers who were relying on the previously permissive type handling might need to review their code to ensure compatibility with the new type restrictions.


reflex/vars.py
```diff
@@ -38,6 +38,42 @@
 # Set of unique variable names.
 USED_VARIABLES = set()
 
+# operators supported my all types.
```
picklelo, line 41: small typo: "by all types"

reflex/vars.py
```diff
@@ -38,6 +38,42 @@
...
+ALL_OPS = ["==", "!=", "!==", "==="]
+DELIMITERS = [","]
+OPERATION_MAPPING = {
```
picklelo, line 44: Can we add similar comments to these other variables describing what they are

reflex/vars.py
```diff
@@ -379,6 +427,33 @@ def operation(
...
+        """
+        if operator in ALL_OPS or operator in DELIMITERS:
+            return True
+
+        # bools are subclasses of ints
+        pair = (
+            int if operand1_type == bool else operand1_type,
+            int if operand2_type == bool else operand2_type,
+        )
+        if pair in OPERATION_MAPPING and operator in OPERATION_MAPPING[pair]:
```
picklelo, line 453: Can be simplified to 
```python
return pair in OPERATION_MAPPING and operator in OPERATION_MAPPING[pair]
```

reflex/vars.py
```diff
@@ -498,6 +573,22 @@ def __le__(self, other: Var) -> Var:
         """
         return self.compare("<=", other)
 
+    @staticmethod
+    def get_outer_container_type(type_annotation) -> Type:
```
picklelo, line 577: I believe you can use the method `types.get_base_class` which should already do this

reflex/vars.py
```diff
@@ -683,6 +792,12 @@ def __ror__(self, other: Var) -> Var:
         Returns:
             A var representing the logical or.
         """
+        other_type = other.type_ if isinstance(other, Var) else type(other)
```
picklelo, line 795: Do we have to add this logic for the other operations too? Also, might be worth to see if we can abstract this code out as it's repeated for each function.

picklelo (commented):
Nice, this approach is much cleaner!

reflex/vars.py
ElijahAhianyo: ah yes, good catch

reflex/vars.py
```diff
@@ -407,6 +447,15 @@ def operation(
             name = f"{op}{self.full_name}"
         else:
             props = (other, self) if flip else (self, other)
+            if op and not self.is_valid_operation(
```
picklelo, line 450: Can we add some high level comments to every little block of code

reflex/vars.py
```diff
@@ -407,6 +447,15 @@ def operation(
...
+                types.get_base_class(props[0].type_),
+                types.get_base_class(props[1].type_),
+                op,
+            ):
+                raise TypeError(
+                    f"Unsupported Operand type(s) for {op}: `{props[0].full_name}` of type {props[0].type_.__name__} and `{props[1].full_name}` of type {props[1].type_.__name__}"
```
picklelo, line 456: Why do we need the `.__name__` ? would it work okay without that?

reflex/vars.py
```diff
@@ -546,6 +620,12 @@ def __add__(self, other: Var) -> Var:
         Returns:
             A var representing the sum.
         """
+        other_type = other.type_ if isinstance(other, Var) else type(other)
+        if (
```
picklelo, line 624: Add comment for this block

reflex/vars.py
picklelo: This is fine actually

picklelo (commented):
Looks good, just a couple comments

integration/test_var_operations.py
```diff
@@ -0,0 +1,701 @@
...
+                VarOperationState.dict1.contains(1).to_string(), id="dict_contains"
+            ),
+        )
+
+    app.compile()
+
+
+@pytest.fixture(scope="session")
+def var_operations(tmp_path_factory) -> Generator[AppHarness, None, None]:
+    """Start FormSubmit app at tmp_path via AppHarness.
```
picklelo, line 519: small nit: need to update app name here

integration/test_var_operations.py
```diff
@@ -0,0 +1,701 @@
...
+        root=tmp_path_factory.mktemp("var_operations"),
+        app_source=VarOperations,  # type: ignore
+    ) as harness:
+        assert harness.app_instance is not None, "app is not running"
+        yield harness
+
+
+@pytest.fixture
+def driver(var_operations: AppHarness):
+    """GEt an instance of the browser open to the var operations app.
```
picklelo, line 537: Get

picklelo (commented):
Awesome, thanks for adding the extensive tests also. Should we remove the bitwise operator before merging?

picklelo (approved):
great work, looks good to me

PR #1676, opened 2023-08-24
rx.background and StateManager.modify_state provides safe exclusive access to state

masenf:
App.modify_state provides exclusive access to state AND sends a state update after exiting the async contextmanager.

## Background EventHandler example

`@rx.background` enables an EventHandler to not block the event loop!

A background task can either yield events that will update the state or enter `async with self` context block to directly munge the state.

```python
import asyncio

import reflex as rx


class State(rx.State):
    counter: int = 0
    _task_id: int = 0

    @rx.background
    async def handle_event(self):
        async with self:
            task_id = self._task_id
            self._task_id += 1
        print(f"[{task_id}] background event starting; counter = {self.counter}")
        for ix in range(5):
            await asyncio.sleep(2)
            print(f"[{task_id}] [{ix}] background event stale counter = {self.counter}")
            async with self:
                print(f"[{task_id}] [{ix}] background event bumping counter = {self.counter}")
                self.counter += 1
            yield rx.console_log(f"Background task {task_id} is on iteration {ix}")

    @rx.background
    async def handle_event_yield_only(self):
        async with self:
            task_id = self._task_id
            self._task_id += 1
        print(f"[{task_id}] background event starting; counter = {self.counter}")
        for ix in range(5):
            await asyncio.sleep(2)
            yield State.increment_arbitrary(5)
            yield rx.console_log(f"Background task {task_id} is on iteration {ix}")

    def increment(self):
        self.counter += 1

    def increment_arbitrary(self, amount: int):
        self.counter += int(amount)

    def reset_counter(self):
        self.counter = 0

    async def chain_background_task(self, direct: bool):
        if direct:
            await self.non_blocking_pause()
        else:
            # must yield / return, no direct calling of background tasks
            yield State.non_blocking_pause()

    async def blocking_pause(self):
        await asyncio.sleep(5)
        return rx.console_log("Blocking pause done")

    @rx.event.background
    async def non_blocking_pause(self):
        await asyncio.sleep(5)
        return rx.console_log("Non-blocking pause done")


def index() -> rx.Component:
    return rx.vstack(
        rx.heading(State.counter),
        rx.button("Delayed Increment", on_click=State.handle_event),
        rx.button("Yield Increment", on_click=State.handle_event_yield_only),
        rx.button("Blocking Pause", on_click=State.blocking_pause),
        rx.button("Non-Blocking Pause", on_click=State.non_blocking_pause),
        rx.button("Chain background", on_click=lambda: State.chain_background_task(False)),
        rx.button("Chain background direct", on_click=lambda: State.chain_background_task(True)),
        rx.button("Reset", on_click=State.reset_counter),
    )


app = rx.App()
app.add_page(index)
app.compile()
```

## Background task example app

```python
import asyncio
import uuid

import reflex as rx


TASKS = set()


def background_task(token) -> asyncio.Task:
    task_id = uuid.uuid4()
    async def _task():
        for ix in range(10):
            await asyncio.sleep(1)
            async with app.modify_state(token) as state:
                state.messages.append(f"[{task_id}] [{ix}] Hello to {token}")
        print(f"Task {task_id} finished")
    return asyncio.create_task(_task())


class State(rx.State):
    messages: list[str] = []
    counter: int = 0

    async def start_background_task(self):
        """Start a background task."""
        t = background_task(self.get_token())
        t.add_done_callback(TASKS.discard)
        TASKS.add(t)

    async def increment_counter(self):
        await asyncio.sleep(0.25)
        self.counter += 1

    @rx.var
    def reversed_messages(self) -> list[str]:
        return self.messages[::-1]


def index() -> rx.Component:
    return rx.fragment(
        rx.vstack(
            rx.heading(f"Counter: {State.counter}"),
            rx.button("Start background task", on_click=State.start_background_task),
            rx.button("Increment counter", on_click=State.increment_counter),
            rx.foreach(
                State.reversed_messages,
                lambda m: rx.text(m),
            ),
        ),
    )


app = rx.App()
app.add_page(index)
app.compile()
```

### Run it with Redis?

```console
docker run -d -p 6379:6379 redis
REDIS_URL=localhost:6379 reflex run
```

### Bugs

~Click Increment counter about 10 times quickly, then click Start Background Task... then it deadlocks 😔~ Fixed via [b471403](https://github.com/reflex-dev/reflex/pull/1676/commits/b471403f3a40505eea598d5f49f6befc73832a04)

# Remaining Work

- [x] redis lock expiry
- [x] unit tests for `StateProxy` and background events
- [x] end-to-end integration tests w/ redis


reflex/app.py
```diff
@@ -563,6 +568,71 @@ def compile(self):
...
+                state._clean()
+                await self.event_namespace.emit(
+                    str(constants.SocketEvent.EVENT),
+                    StateUpdate(delta=delta).json(),
+                    to=state.get_sid(),
+                )
+
+    def _process_background(
+        self, state: State, event: Event
+    ) -> Union[asyncio.Task, bool]:
```
picklelo, line 599: nit: we can use `asyncio.Task | bool` annotations now

reflex/app.py
```diff
@@ -563,6 +568,71 @@ def compile(self):
         thread_pool.close()
         thread_pool.join()
 
+    @contextlib.asynccontextmanager
+    async def modify_state(self, token: str) -> AsyncIterator[State]:
```
picklelo, line 572: This function doesn't actually modify the state right? Should we call it `get_latest_state` or something?

reflex/app.py
```diff
@@ -759,9 +827,7 @@ async def on_event(self, sid, data):
         # Process the events.
         async for update in process(self.app, event, sid, headers, client_ip):
             # Emit the event.
-            await asyncio.create_task(
-                self.emit(str(constants.SocketEvent.EVENT), update.json(), to=sid)
-            )
+            await self.emit(str(constants.SocketEvent.EVENT), update.json(), to=sid)
```
picklelo, line 830: Good cleanup, not sure why we needed that originally but all the event chain stuff seems to be working well in my testing

picklelo (commented):
Awesome, working really well for me! I think the `async with self` syntax is nice. I didn't review the `state.py` file yet, will add comments to that

reflex/app.py
masenf: the intent is that you modify the state inside the context and then it saves it to the backing store when you exit the context. it's a combination of `get_state` and `set_state`, with exclusive locking baked in.

reflex/state.py
```diff
@@ -1078,7 +1266,75 @@ def set_state(self, token: str, state: State):
...
+
+        Raises:
+            RuntimeError: If redis is not configured.
+        """
+        if self.redis is None:
+            raise RuntimeError("Redis is not configured")
+        lock_key = f"{token}_lock".encode()
+        # TODO: set key expiry
+        state_is_locked = await self.redis.setnx(lock_key, 1)
+        while not state_is_locked:
```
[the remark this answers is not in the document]
masenf: considered; now we raise a LockExpiredError if the mutator holds the lock for too long and we reject their changes so that other waiters don't get corrupted. The exception message suggests to the developer how to increase the lock_expiration timeout or use background tasks.

masenf:
This now has tests and is ready for review; sorry for the large size.

The only remaining issue is some new errors/warnings showing up when running the test in redis mode: https://github.com/reflex-dev/reflex/actions/runs/6193642837/job/16815445395

```console
Exception ignored in: <function StreamWriter.__del__ at 0x7f2b1fae5da0>
  File "/opt/hostedtoolcache/Python/3.11.5/x64/lib/python3.11/asyncio/streams.py", line 396, in __del__
    self.close()
  File "/opt/hostedtoolcache/Python/3.11.5/x64/lib/python3.11/asyncio/streams.py", line 344, in close
    return self._transport.close()
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.5/x64/lib/python3.11/asyncio/selector_events.py", line 860, in close
    self._loop.call_soon(self._call_connection_lost, None)
  File "/opt/hostedtoolcache/Python/3.11.5/x64/lib/python3.11/asyncio/base_events.py", line 761, in call_soon
    self._check_closed()
  File "/opt/hostedtoolcache/Python/3.11.5/x64/lib/python3.11/asyncio/base_events.py", line 519, in _check_closed
Error:     raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
```

and it repeats that sometimes hundreds of times; probably need to understand that before merging, but I only ever see it in CI, not locally or in real apps.

reflex/app.py
```diff
@@ -641,6 +646,67 @@ def compile(self):
...
+        substate, handler = state._get_event_handler(event)
+        if not handler.is_background:
+            return None
+
+        async def _coro():
+            if self.event_namespace is None:
+                raise RuntimeError("App has not been initialized yet.")
+
+            # Process the event.
+            async for update in state._process_event(
```
picklelo, line 693: Should we abstract this to its own function?  Similar logic is repeated in the `app._process` as well as the upload logic

reflex/state.py
```diff
@@ -167,9 +169,34 @@ def _init_event_handlers(self, state: State | None = None):
         if state is None:
             state = self
 
+        def no_chain_background_task(name, fn):
```
picklelo, line 172: Can we add docstrings for this

reflex/state.py
```diff
@@ -820,13 +851,30 @@ async def _process_event(
...
-                0: The state update after processing the event.
-                1: Whether the event is the final event.
+            StateUpdate object
         """
         # Get the function to process the event.
         fn = functools.partial(handler.fn, state)
 
+        token = self.get_token()
+
+        def as_state_update(events, final) -> StateUpdate:
```
picklelo, line 861: Should we lift this outside of the function, so that it's not redefined every time we call `_process_event`?

reflex/state.py
```diff
@@ -1011,6 +1061,170 @@ def dict(self, include_computed: bool = True, **kwargs) -> dict[str, Any]:
...
+        pass
+
+
+class ImmutableStateError(AttributeError):
+    """Raised when a background task attempts to modify state outside of context."""
+
+    pass
+
+
+class StateProxy:
```
picklelo, line 1095: Would the `wrapt.ObjectProxy` class that you introduced be useful here?

reflex/state.py
```diff
@@ -1031,6 +1245,10 @@ class StateUpdate(Base):
     final: bool = True
 
 
+class LockExpiredError(Exception):
```
picklelo, line 1248: We started a new file `reflex/utils/exceptions.py`, should we move the error definitions there or keep them?

reflex/state.py
```diff
@@ -1065,26 +1315,126 @@ def get_state(self, token: str) -> State:
...
             state: The state to set.
+            lock_id: If provided, the lock_key must be set to this value to set the state.
+
+        Raises:
+            LockExpiredError: If lock_id is provided and the lock for the token is not held by that ID.
+        """
+        if self.redis is None:
+            return
+        if lock_id is not None:
+            lock_key = f"{token}_lock".encode()
```
picklelo, line 1342: Should we make a function to format the `lock_key` from the token? Since it's used again later. Also maybe this `if` check can be moved to its own function to validate the lock.

reflex/state.py
```diff
@@ -1065,26 +1315,126 @@ def get_state(self, token: str) -> State:
...
+            RuntimeError: If redis is not configured.
+            LockExpiredError: If the lock has expired while processing the event.
         """
         if self.redis is None:
+            raise RuntimeError("Redis is not configured")
+        lock_key = f"{token}_lock".encode()
+        lock_id = uuid.uuid4().hex.encode()
+        lock_key_channel = f"__keyspace@0__:{lock_key.decode()}"
+
+        def try_get_lock():
```
picklelo, line 1372: Might be good to define this outside of the function as it will get redefined every time this is called.

reflex/state.py
```diff
@@ -1065,26 +1315,126 @@ def get_state(self, token: str) -> State:
...
+
+        state_is_locked = await try_get_lock()
+        if not state_is_locked:
+            # Missed the fast-path to get lock, subscribe for lock delete/expire events
+            await self.redis.config_set(
+                "notify-keyspace-events", self._redis_notify_keyspace_events
+            )
+            async with self.redis.pubsub() as pubsub:
+                await pubsub.psubscribe(lock_key_channel)
+                while not state_is_locked:
```
picklelo, line 1388: This block could also be refactored into a function `wait_for_lock`

reflex/state.py
```diff
@@ -1065,26 +1315,126 @@ def get_state(self, token: str) -> State:
...
+    async def modify_state(self, token: str) -> AsyncIterator[State]:
+        """Modify the state for a token while holding exclusive lock.
+
+        Args:
+            token: The token to modify the state for.
+
+        Yields:
+            The state for the token.
+        """
+        if self.redis is not None:
```
picklelo, line 1422: Later maybe we should make `RedisStateManager` just a subclass of `StateManager` so we don't need to have these special conditional checks, we can just override the methods.

picklelo (commented):
Code works well for me!

Was discussing with @Alek99 about the API and maybe we the `async with self` is not the most obvious syntax. What do you think about making it something like `async with self.get_latest()` or something like that? We can discuss more offline.

reflex/app.py
masenf: this one is sort of hard to abstract away because each call site is slightly different:

* App._process_background.<locals>._coro: uses `state._process_event` and `emit_update` directly
* app.process: uses `state._process` and yields the updates
* upload: uses `state._process` and `emit_update` directly

Might be nice to try to unify these down the road, but I think I'll defer on this PR.

masenf:
@picklelo thanks for the feedback. I addressed almost everything here: https://github.com/reflex-dev/reflex/pull/1676/files/f6ba2ab4eb855837ae517e05ffc187765bbaaa93..1093229068b6d7f5aec2917c5d08b0ad43fb4211

I also went ahead and fixed all of the `AppHarness` tests to work with redis, so now we're running with both `StateManagerMemory` and `StateManagerRedis` for better coverage. This will also ensure that future tests and features are at least considering how they interop with redis backing store. The refactoring of the `StateManager` as an abstract base class also opens the door to simpler implementation of future backing stores if we ever need or want that.

As for `async with self` I've been pondering over potentially better syntax, but i haven't found anything i particularly like. I don't really want to expose an underscore method as part of the API, but I also am reluctant to restrict downstream users from another name. And I'm a sucker for the conciseness of `async with self`. I think we should put that one to the community and see what people think.

As an aside, I had to kick the CI on reflex-web a few times; I wouldn't think it's my changes here, but the [error is weird](https://github.com/reflex-dev/reflex/actions/runs/6231962820/job/16914600046?pr=1676) and I don't really understand it.

masenf (commented):
~Further testing has revealed that this sort of breaks yielding updates in normal event handlers. Need to write a test case and investigate.~

Fixed and test case written ✔️

reflex/app.py
```diff
@@ -89,7 +97,7 @@ class App(Base):
     state: Type[State] = DefaultState
 
     # Class to manage many client states.
-    state_manager: StateManager = StateManager()
+    state_manager: StateManager = StateManagerMemory(state=DefaultState)
```
picklelo, line 100: Why do we set this if we're overriding it again in the `__init__`?

reflex/state.py
```diff
@@ -1009,31 +1218,83 @@ class StateUpdate(Base):
...
+        Yields:
+            The state for the token.
         """
-        self.state = state
-        self.redis = prerequisites.get_redis()
+        yield self.state()
 
-    def get_state(self, token: str) -> State:
+
+class StateManagerMemory(StateManager):
```
picklelo, line 1278: Awesome, this is much cleaner we can extend to other backends in the future too

picklelo (approved):
Awesome!!

reflex/app.py
masenf: I didn't want to make it `Optional` and default it to `None` because then we have to put `is not None` checks all over the place.

But if we do something like `StateManager.create(state=DefaultState)` here, _and_ redis is used, then we create an unserializable default, which pydantic doesn't like.

So the default `StateManager` is the `StateManagerMemory`, then inside `__init__`, we call `StateManager.create(...)` to update the reference to either memory or redis.

PR #1695, opened 2023-08-28
Client-side Routing (404 redirect)

masenf:
Support client-side routing of dynamic routes.

When a user requests a URL that doesn't exist on the server/filesystem and gets the 404 page, automatically try to redirect them to the requested page via `next/router` before displaying the 404 error.

* Remove 404.js entirely, implement redirect logic in `/utils/client_side_routing.js`
* Maintain `routeNotFound` state var at the App level, and plumb it through with `useContext` -- this allows the code to avoid a redirect loop when the requested url is truly a 404
* Use default nextjs 404 page, or any custom 404 page

* Update `docker-example` to run in `--backend-only` mode, using Caddy to serve the static frontend.
  * Reduce `reflex-app` container image size to <350MB (from over 1GB) 🎉


masenf:
There seems to be a bug in the client side routing code, where the hydrate occurs on the 404 page, then redirects to the _actual_ page, but skips hydrate... however skipping hydrate misses on_load events for the redirected page.

Somehow need to skip the initial hydrate on websocket connection from 404, unless we're NOT redirecting.

masenf:
Consider reimplementing `on_load` and `state.hydrate` events via `on_mount`.

reflex/.templates/web/utils/state.js
```diff
@@ -206,7 +203,17 @@ export const applyRestEvent = async (event) => {
...
-  event_queue.push(...events)
+  event_queue.push(...events.map((ev) => {
+    // Copy the event and set the token.
+    const new_event = {...ev, token: getToken()}
+    // Do NOT overwrite router_data if the event already has it set.
+    if (ev.router_data === undefined || Object.keys(ev.router_data).length === 0) {
+      new_event.router_data = (({ pathname, query, asPath }) => ({ pathname, query, asPath }))(Router)
+    }
+    return new_event
+  }))
```
masenf, line 214: consider removing this section to avoid behavior change;

do not overwrite existing router_data, but continue to set router_data right before the event is sent so that `get_current_page()` is more likely to return the actual page the user is on when the event was sent.

masenf:
With the latest changes everything is vastly simplified:

* only deal with `useClientSideRouting` on the actual 404 page
  * handle templating of the hook inside a python component, ClientSideRouting, that is included as part of `wait_for_client_redirect`, since that function is the only one that cares about `not_found`
* get rid of top level `ClientSideRoutingProvider` -- turns out that wasn't actually needed at all
* move initial hydration into `useEventLoop`, so it can get the router_data early (after websocket connect is too late, the page may have already redirected).
* move routeChangeComplete hydration back into `index.js.jinja2` where it was before (so no change here anymore)

picklelo (approved):
Awesome, this ended up being very clean :)

PR #1701, opened 2023-08-28
Fix Node Installation on ARM-based Macs with FNM

ElijahAhianyo:
## Issue
On ARM-based Macs (M1 and M2), FNM was installing the x64 binary of Node, leading to compatibility issues and flakiness in the application.

## Solution
 This PR addresses the problem by explicitly specifying the architecture (--arch=arm64) when installing Node using FNM on ARM-based Macs. Since there is currently no FNM ARM64 release binary available for macOS, FNM defaults to installing the x64 binary, causing compatibility issues when installing Node. By explicitly setting the architecture during Node installation, we ensure that the correct architecture is used, mitigating any compatibility problems.

## Changes

Introduces a conditional check to determine if the platform is an ARM-based Mac (M1/M2).
If the platform is ARM-based, the --arch=arm64 flag is included in the FNM installation command.
This change ensures that the ARM64 architecture is used for Node installation on ARM-based Macs, resolving the flakiness and compatibility issues.

Note: This fix is essential for maintaining consistent behavior and stability, especially for users on ARM-based Macs. The explicit architecture specification during Node installation resolves the root cause of the flakiness observed previously. This PR ensures that our application runs reliably across different Mac architectures.

### DIFF
--- a/reflex/utils/prerequisites.py
+++ b/reflex/utils/prerequisites.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import glob
 import json
 import os
+import platform
 import re
 import stat
 import sys
@@ -330,10 +331,15 @@ def install_node():
             [
                 constants.FNM_EXE,
                 "install",
+                "--arch=arm64"
+                if platform.system() == "Darwin"
+                and platform.machine()
+                == "arm64"  # specify arm64 arch explicitly for M1/M2
+                else "",
                 constants.NODE_VERSION,
                 "--fnm-dir",
                 constants.FNM_DIR,
-            ]
+            ],
         )
     processes.show_status("Installing node", process)
 
--- a/reflex/utils/prerequisites.py
+++ b/reflex/utils/prerequisites.py
@@ -327,15 +327,18 @@ def install_node():
         # Add execute permissions to fnm executable.
         os.chmod(constants.FNM_EXE, stat.S_IXUSR)
         # Install node.
+        # Specify arm64 arch explicitly for M1s and M2s.
+        architecture_arg = (
+            ["--arch=arm64"]
+            if platform.system() == "Darwin" and platform.machine() == "arm64"
+            else []
+        )
+
         process = processes.new_process(
             [
                 constants.FNM_EXE,
                 "install",
-                "--arch=arm64"
-                if platform.system() == "Darwin"
-                and platform.machine()
-                == "arm64"  # specify arm64 arch explicitly for M1/M2
-                else "",
+                *architecture_arg,
                 constants.NODE_VERSION,
                 "--fnm-dir",
                 constants.FNM_DIR,
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -549,13 +549,32 @@ def test_node_install_windows(tmp_path, mocker):
     download.assert_called_once()
 
 
-def test_node_install_unix(tmp_path, mocker):
+@pytest.mark.parametrize(
+    "machine, system",
+    [
+        ("x64", "Darwin"),
+        ("arm64", "Darwin"),
+        ("x64", "Windows"),
+        ("arm64", "Windows"),
+        ("armv7", "Linux"),
+        ("armv8-a", "Linux"),
+        ("armv8.1-a", "Linux"),
+        ("armv8.2-a", "Linux"),
+        ("armv8.3-a", "Linux"),
+        ("armv8.4-a", "Linux"),
+        ("aarch64", "Linux"),
+        ("aarch32", "Linux"),
+    ],
+)
+def test_node_install_unix(tmp_path, mocker, machine, system):
     fnm_root_path = tmp_path / "reflex" / "fnm"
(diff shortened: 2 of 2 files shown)


Alek99:
I am still getting the error on my Mac just tested

masenf (changes requested):
this breaks `tests/test_utils.py::test_node_install_unix` on M1 mac platform (i'm assuming github runners are x64).

suggest parametrizing and monkeypatching the platform call for this test case.

reflex/utils/prerequisites.py
```diff
@@ -326,14 +327,22 @@ def install_node():
         # Add execute permissions to fnm executable.
         os.chmod(constants.FNM_EXE, stat.S_IXUSR)
         # Install node.
+        # Specify arm64 arch explicitly for M1s and M2s.
```
realFranco, line 330: Hello, what if I have an ARM chip computer and is not a Apple product?

masenf: should already be handled for linux `aarch64`. windows also seems to work as is.

you can test with `main` and see if there is some error?

ElijahAhianyo:
> this breaks `tests/test_utils.py::test_node_install_unix` on M1 mac platform (i'm assuming github runners are x64).
> 
> suggest parametrizing and monkeypatching the platform call for this test case.

good catch. fixed it now

ElijahAhianyo:
> I am still getting the error on my Mac just tested, only about 25% of the time though
> 
> ```
> ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── App Running ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> App running at: http://localhost:3000
> Compiling:  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 1/1 0:00:00
> Starting frontend failed with exit code 1
> 
> > dev
> > next dev
> 
> - ready started server on 0.0.0.0:3000, url: http://localhost:3000
> - [36minfo[39m Using wasm build of next-swc
> - [33mwarn[39m Attempted to load @next/swc-darwin-x64, but it was not installed
> assertion failed : BasicBlock requested for unrecognized address
> (BuilderBase.h:550 block_for_offset)
>  Run with --loglevel debug  for the full log.
> ```

I see, did you remove the reflex dir (`rm -rf /User/<username>/Application Support/reflex`) before running `reflex init`?

tests/test_utils.py
```diff
@@ -549,13 +549,32 @@ class Resp(Base):
...
+        ("aarch32", "Linux"),
+    ],
+)
+def test_node_install_unix(tmp_path, mocker, machine, system):
     fnm_root_path = tmp_path / "reflex" / "fnm"
     fnm_exe = fnm_root_path / "fnm"
 
     mocker.patch("reflex.utils.prerequisites.constants.FNM_DIR", fnm_root_path)
     mocker.patch("reflex.utils.prerequisites.constants.FNM_EXE", fnm_exe)
     mocker.patch("reflex.utils.prerequisites.constants.IS_WINDOWS", False)
```
masenf, line 575: should we set this based on the mocked `system`?

masenf (approved):

reflex/utils/prerequisites.py
realFranco: Hello, thank you the reply. Are you refering to execute the test using a computer based on arm & Apple?, if so, sure I can, I have a M1 chip in hands.

masenf: The reflex dev team mostly uses M1 and M2 macs, so we have pretty good coverage on that. I have an aarch64 linux laptop (pine book pro) that i've tested on, which works. And I have an armv7 (32-bit) raspberry pi, that I've been meaning to test with, but haven't yet.

If there is some particular os/arch combination that you're concerned about, my suggestion would be to test running `pip install git+https://github.com/reflex-dev/reflex` and `reflex init` your project and see if it's working. If not, please file a bug and we will do our best to accomodate it or steer you in the right direction. 

The assumption at this point is the new `fnm`-based node bootstrap is working on windows (x86, x64, arm64), macos (x64, arm64), and linux (x64, arm64, arm32).

ltm920716: image
hello, I run ```pip install git+https://github.com/reflex-dev/reflex``` and ```reflex init```, but this bug still shows sometimes, I use mac M1

ltm920716: I find that some error operations like index overflow or database error will break reflex down and can not re-open

masenf: @ltm920716 did you remove the reflex dir (`rm -rf /User/<username>/Application Support/reflex`) before running `reflex init`

ltm920716: sorry that there is no 'Application Support' dir in this ’/User/<username>/‘ path, I wiil try to find it，thanks

ltm920716: I find that path is ```/Users/<username>/Library/Application Support```

PR #1705, opened 2023-08-29
Remove postponed annotations within component files

picklelo:
Remove postponed annotations within component files, as the forward refs were causing issues


masenf (changes requested):
i don't think this is a viable long-term solution

picklelo:
Closing this as we're not going with this approach.

PR #1708, opened 2023-08-29
add pyi_generator script

Lendemor:
### All Submissions:

### Type of change

### New Feature Submission:

### Description

Add a script to generate .pyi files which can be used by editor like VSCode, PyCharm, etc..
To suggest all the known arguments when initializing a component.


jackie-pc:
This is awesome! BTW are there any existing "pyi" generators out there already?  If there are, where do they fall short of our needs?

scripts/pyi_generator.py
```diff
@@ -0,0 +1,144 @@
...
+                definition += (
+                    f"{kwarg}: {_get_type_hint(create_spec.annotations[kwarg])}, "
+                )
+            else:
+                definition += f"{kwarg}, "
+
+        for name, value in _class.__annotations__.items():
+            if name in create_spec.kwonlyargs:
+                continue
+            definition += f"{name}: {_get_type_hint(value)}, "
```
masenf, line 67: the `Var` seem to be losing their arg; so `Var[bool]` just becomes `Var`.

Personally, I think these should be typed as `Var[bool] | bool` so that users don't get type checking errors when passing bare values that aren't `Var`-ified

masenf (commented):
So what is our strategy for these files? Will they committed or generated during the build/release process?

I also noticed that the event triggers are not generated, but we probably do want those to show up in the auto-complete list.

scripts/pyi_generator.py
```diff
@@ -0,0 +1,144 @@
...
+
+def _get_type_hint(value):
+    res = ""
+    if hasattr(value, "_name") and value._name == "Optional":
+        res = value.__args__[0].__name__
+    elif isinstance(value, str):
+        res = value
+    else:
+        res = value.__name__
+    return res
```
masenf, line 23:
```suggestion
def _get_type_hint(value):
    res = ""
    args = get_args(value)
    if args:
        res = f"{value.__name__}[{', '.join([_get_type_hint(arg) for arg in args if arg is not type(None)])}]"
        if value.__name__ == "Var":
            res = " | ".join([res] + [_get_type_hint(arg) for arg in args if arg is not type(None)])
    elif isinstance(value, str):
        res = value
    else:
        res = value.__name__
    return res
```

If we call this recursively on the args of the type annotation, then we can keep the `Var[bool]`, etc and also get the union functionality of acception `Var[bool] | bool`.

Looking through these generated pyi files though, i'm realizing we have some discrepancies, some of which need to be fixed in the main package, but at least for the hints, i think we do need default values of `None` everywhere, which means these generated annotations basically all need to be `Optional`.

Lendemor: I don't think we can use your suggestion as-is because the | is not available in python 3.7.

It's a good idea however, I'll write something along those line to make it work and to handle the defaults to `None`

Lendemor:
> This is awesome! BTW are there any existing "pyi" generators out there already? If there are, where do they fall short of our needs?

mypy provide `stubgen` which generate a stub based on a existing file.

The reason we need our own script is because we want to enrich the signature of the create method for all components.
The script scan all the components, collect the attributes and add them.

(I've actually used `stubgen` to generate `reflex.vars.pyi`, then edited it manually after it to make it work.)

masenf (approved):
excellent

PR #1713, opened 2023-08-30
Allow underscores in routes

ElijahAhianyo:
### Description
This pull request addresses the need for converting underscores to hyphens in routes, which was previously implemented. The goal of this change is to enhance the flexibility of route naming by allowing both underscores and hyphens to coexist without the need for any conversions. This PR proposes to remove the underscore-to-hyphen conversion logic, allowing developers to freely choose between these two naming conventions for routes.

### Changes Made
* Removed the underscore-to-hyphen conversion logic from route handling, enabling route names to support both underscores and hyphens directly.

### Motivation
* Allowing both underscore-separated and hyphen-separated route names provides developers with greater flexibility in choosing naming conventions that align with their preferences and practices.
* By removing the conversion, we empower developers to use the naming convention that best suits their needs, contributing to a more adaptable and user-friendly codebase.

### Impact
With this change, developers are free to use both underscores and hyphens as they see fit in route names.
The removal of the underscore-to-hyphen conversion ensures that developers can choose their preferred naming convention without the need for special handling.
Existing routes with either underscores or hyphens in their names will continue to function as expected.

### DIFF
--- a/integration/test_dynamic_routes.py
+++ b/integration/test_dynamic_routes.py
@@ -69,7 +69,7 @@ def DynamicRoute():
 
 @pytest.fixture(scope="session")
 def dynamic_route(
-    app_harness_env: Type[AppHarness], tmp_path_factory
+        app_harness_env: Type[AppHarness], tmp_path_factory
 ) -> Generator[AppHarness, None, None]:
     """Start DynamicRoute app at tmp_path via AppHarness.
 
@@ -81,8 +81,8 @@ def dynamic_route(
         running AppHarness instance
     """
     with app_harness_env.create(
-        root=tmp_path_factory.mktemp(f"dynamic_route"),
-        app_source=DynamicRoute,  # type: ignore
+            root=tmp_path_factory.mktemp(f"dynamic_route"),
+            app_source=DynamicRoute,  # type: ignore
     ) as harness:
         yield harness
 
@@ -131,7 +131,7 @@ def backend_state(dynamic_route: AppHarness, driver: WebDriver) -> State:
 
 @pytest.fixture()
 def poll_for_order(
-    dynamic_route: AppHarness, backend_state: State
+        dynamic_route: AppHarness, backend_state: State
 ) -> Callable[[list[str]], None]:
     """Poll for the order list to match the expected order.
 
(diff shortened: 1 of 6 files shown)


masenf (dismissed):
Any plans to also make upper case letters work? That seems to come up about every 2-3 weeks on discord.

ElijahAhianyo:
> Any plans to also make upper case letters work? That seems to come up about every 2-3 weeks on discord.

Yeah, could add that to this PR

reflex/utils/format.py
```diff
@@ -210,9 +210,10 @@ def format_route(route: str) -> str:
     Returns:
         The formatted route.
     """
-    # Strip the route.
     route = route.strip("/")
-    route = to_snake_case(route).replace("_", "-")
+    # Strip the route and format casing.
+    if format_case:
```
picklelo, line 215: Let's make a function called `to_kebab_case` and call it here

integration/test_event_chain.py
```diff
@@ -97,11 +97,11 @@ def click_yield_nested(self):
 
         def redirect_return_chain(self):
             self.event_order.append("redirect_return_chain")
-            yield rx.redirect("/on-load-return-chain")
+            yield rx.redirect("/on_load_return_chain")
```
picklelo, line 100: This file shouldn't need to change

picklelo (changes requested):
Looks good! I think we just need to update the integration tests in integration/test_dynamic_routes.py

integration/test_dynamic_routes.py
```diff
@@ -69,7 +69,7 @@ def redirect_page():
 
 @pytest.fixture(scope="session")
 def dynamic_route(
-    app_harness_env: Type[AppHarness], tmp_path_factory
+        app_harness_env: Type[AppHarness], tmp_path_factory
```
masenf, line 72: i don't understand why these blocks are all indented more...

ElijahAhianyo: probably as a result of running precommits (black), not typically sure why those parts are affected

picklelo (approved):
looks good!

PR #1714, opened 2023-08-30
Refactor: Move format_prop Static Method for Improved Reusability

ElijahAhianyo:
### Description

The format_prop static method, which was previously tightly coupled with the Tag class, has been moved to the format.py module to enhance its reusability. This refactor allows us to use format_prop across various parts of our codebase without direct dependencies on the Tag class.

### Changes
The format_prop static method has been relocated from the Tag class to the format.py module.
Necessary import statements have been updated to ensure that format_prop is accessible where needed.

### Motivation
This change is driven by the need for a more flexible and reusable format_prop method. By moving it to format.py, we can now utilize it in a broader context, such as other classes, modules, or future extensions of our codebase, without introducing unnecessary coupling to the Tag class.

### Impact
This change does not alter the behavior of the format_prop method itself; it simply relocates it to a more appropriate location. Existing code that relies on format_prop should continue to function as expected.

### DIFF
--- a/reflex/components/tags/tag.py
+++ b/reflex/components/tags/tag.py
@@ -2,19 +2,12 @@
 
 from __future__ import annotations
 
-import json
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple, Union
-
-from plotly.graph_objects import Figure
-from plotly.io import to_json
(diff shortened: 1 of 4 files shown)


masenf (approved):
good to me!

picklelo (approved):

PR #1723, opened 2023-08-31
Fix setting substate client storage

picklelo:
We were resetting the state when hydrating the client storage, making the tab lose state on refresh.


masenf (approved):
Wow thanks for catching this. I'm gonna blame copilot for this one 😅

masenf:
Pushed a fix for the flaky cookie test

Issue #1583, opened 2023-08-14
`--env prod` reports `App running at: http://localhost`.


nevdelap:
**Describe the bug**

When running in prod mode it reports that it is running at `http://localhost`. It's a minor thing, but `0.0.0.0` is not localhost (`127.0.0.1`), it's bind to all addresses, and it sent me up the garden path debugging for a long time when I'd simply mistyped `api_url` as `app_url` in `rxconfig.py`. :scream: 

```bash
root@droplet:~/webchat/webui# reflex run --env prod --frontend-port 80
──────────────────────────────────────────────────────────────────────────────────────────────── Starting Reflex App ────────────────────────────────────────────────────────────────────────────────────────────────
Compiling:  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 1/1 0:00:00
Creating Production Build:  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 9/9 0:00:57
──────────────────────────────────────────────────────────────────────────────────────────────────── App Running ────────────────────────────────────────────────────────────────────────────────────────────────────
App running at: http://localhost:80
Compiling:  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 1/1 0:00:00
```

**To Reproduce**

As above, run with `--env prod`.

**Expected behavior**

`App running at: http://0.0.0.0:80`

**Screenshots**

N/A

**Specifics (please complete the following information):**

 - Python Version: 3.10.12
 - Reflex Version: 0.2.4
 - OS: Linux
 - Browser (Optional): N/A
 
**Additional context**

It looks like it is coming from from Next.js logging this: `ready started server on 0.0.0.0:3000, url: http://localhost:3000`. `0.0.0.0` is not `localhost` - `127.0.0.1` so it's misleading. Next.js (if that's what is reporting that line) is telling a porky.

`exec.py` parses that and reports the wrong info with these lines.

```
if "ready started server on" in line:
    url = line.split("url: ")[-1].strip()
    console.print(f"App running at: [bold green]{url}")
```

Thanks. :pray:

nevdelap:
Add a commit to fix this problem, and to make it show the basePath if set. Is on top of #1633.

PR #1724, opened 2023-08-31
Issues 1633 Add frontend_path to config to support running multiple reflex apps off the same domain, and 1583 Show the correct info on where the site is being served.

### LINKED ISSUE #1633: Unable to override `basePath` in `next.config.js` to allow running multiple Reflex apps on the same domain because of how Next.js serves `/_next`.
**Describe the bug**

* `reflex/utils/prerequisites.py` can change `compress` to `true` or `false` based on `next_compression` in `rxconfig.js` .
* As far as I can tell you cannot override anything else.
* That `next.config.js` is in `.web` is in constants that as far as I can tell cannot be overridden.
* If I `reflex init`, add `basePath: "chat",` to the `.web/next.config.js` that it created, and run `reflex run` it will run the site on a different path, e.g. `/chat` instead of `/`, in dev mode, including `/chat/_next` instead of `/_next`. This is good.
* And if I do the same thing in a Dockerfile that runs `reflex run --env prod` it works too. This is good.

So what I want to suggest is the issue is that what I've done in those two last points is a workaround that with a couple of lines of code in `prerequisites.py` could be a supported feature.

**Note:** The reason this is needed is to change where Next.js serves `/_next`. The front-end and back-end can be served on whatever path you like on the domain (e.g. with Nginx `proxy_pass`), but it is `/_next` that is preventing running multiple sites on the same domain, as far as I've been able to figure out, because each app will have its own `/_next` that cannot both be served on that same path publicly.

(I'm new to Reflex and know nothing about Next.js other than what I've been investigating today, so maybe there's a better way that is already supported that I've not found.)

**To Reproduce**
As above.

**Expected behavior**
If `prerequisites.py` supported updating `basePath` like it updates `compress` based on `rxconfig.js` having `next_compression = "true/false"`, by having `next_base_path`, changing the Next.js `basePath` would work out of the box to allow running multiple Reflex apps on the same domain with a one line addition to `rxconfig.js` in each app like `next_base_path = "/chat",` and `next_base_path = "/someotherapp",`

It would want to do one other thing. When it reports `App running at: http://localhost:80` it would need to report `App running at: http://localhost:80/chat`. That might just happen since that message is coming from Next.js, but I don't think it did. (It's midnight as I write this and I need to go to bed. I'll update this when I've checked.)

**Screenshots**
N/A

 - Python Version: 3.11
 - Reflex Version: 0.2.4
 - OS: Linux
 - Browser (Optional): Chrome
 
**Additional context**
N/A

Thanks. 🙏

nevdelap:
### All Submissions:

- [x] Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/pynecone-io/pynecone/pulls ) for the desired changed.

### Type of change

### New Feature Submission:

I've run this, all good, except for some pre-existing issues that are in main complaining about test_config.py.

```
home 15:27:48 (0) ~> cw reflex_check
#!/bin/bash
set -eufo pipefail

readonly ESC=$'\033'"["
readonly BOLD="${ESC}1m"
readonly RESET="${ESC}0m"

echo -e "\n${BOLD}Testing.$RESET"
poetry run pytest tests --cov --no-cov-on-fail --cov-report=

echo -e "\n${BOLD}Linting.$RESET"
poetry run ruff check --fix .
poetry run pyright reflex tests
find reflex tests -name "*.py" -not -path reflex/reflex.py | xargs poetry run darglint

echo -e "\n${BOLD}Formatting.$RESET"
poetry run black reflex tests

echo -e "\n${BOLD}All good.$RESET"
```

### Changes To Core Features:

- [x] Have you added an explanation of what your changes do and why you'd like us to include them? In the issues.

Closes #1633 and #1583.


nevdelap:
After merging and resolving the conflict I've retested locally for the thing that needed resolving, and for the things the changes do, and I've rerun the tests and everything, and it's all good. I'm redeploying to my own sites on DigitalOcean.

nevdelap:
Had to discover that I needed to change `event` to `_event`, then it all worked like a bought one.

reflex/utils/exec.py
```diff
@@ -36,9 +37,13 @@ def run_process_and_launch_url(run_command: list[str]):
     )
 
     for line in processes.stream_logs("Starting frontend", process):
-        if "ready started server on" in line:
-            url = line.split("url: ")[-1].strip()
+        if match := re.search("ready started server on ([0-9.:]+)", line):
```
masenf, line 40: the walrus was introduced in python 3.8, but reflex still claims to support 3.7 although we do not test with it... this would certainly require us to drop 3.7 support in the main package.

The team has been discussing dropping 3.7 for at least a month, but we haven't made a solid decision yet. Although 3.7 went end of life 2 months ago (2023-06-27), there are still some supported linux distributions that might be shipping it.

i'm not asking you do change the code here...yet, but will discuss with the team and see if the time has come to drop 3.7 support. (Personally, I'm of the mind, that if we don't test it in CI, then we don't really support it).

reflex/utils/exec.py
```diff
@@ -36,9 +37,13 @@ def run_process_and_launch_url(run_command: list[str]):
...
+            url = match.group(1)
```
masenf, line 41:
```suggestion
            url = f"http://{match.group(1)}"
```

reflex/utils/exec.py
```diff
@@ -36,9 +37,13 @@ def run_process_and_launch_url(run_command: list[str]):
...
+            if get_config().frontend_path != "":
+                url += get_config().frontend_path
```
masenf, line 43: I'd prefer to see `urllib.parse.urljoin` here, to ensure that leading and trailing slashes are squashed.

```suggestion
                url = urljoin(url, get_config().frontend_path)
```

reflex/utils/exec.py
```diff
@@ -36,9 +37,13 @@ def run_process_and_launch_url(run_command: list[str]):
...
             console.print(f"App running at: [bold green]{url}")
+        else:
+            console.debug(line)
```
masenf, line 46: i think `process.stream_logs` already logs each line at debug level.

masenf (changes requested):
This looks really great! Just a couple of comments, and will discuss with the team about the walrus operator.

reflex/utils/exec.py
nevdelap: There are things you cannot do without the :=, but since this isn't one of them, happy to change it.

reflex/utils/exec.py
nevdelap: Done.

reflex/utils/exec.py
nevdelap: Done.

reflex/utils/exec.py
nevdelap: Done.

masenf (approved):

nevdelap:
I reran the checks and got failures that I think are caused by fed75ea, error: "NoRenderImportVar" is unknown import symbol.

nevdelap:
Actually I checked out and fed75ea and it passed, so I'll looking at who my stuff could make it complain.

nevdelap:
Looks like it is caused by 63b5fbd.

image

masenf:
Yeah that should be fixed now, if you remerge from main

nevdelap:
Hi @masenf. I hadn't run `poetry run pytest integration` since it's not mentioned in [CONTRIBUTING.md](https://github.com/reflex-dev/reflex/blob/main/CONTRIBUTING.md), but I've just run it and it worked.

image

masenf (approved):
sorry the CI was flaky, looks good!

PR #1725, opened 2023-08-31
Proper serialization for chained Event payloads

### LINKED ISSUE #1722: State variable becomes string when passed as argument to another State Event
**Describe the bug**
If you try to send a state variable as a argument to another state event the variable will be converted to a string.

**To Reproduce**

Run code, change page limit, check log.

 - Code/Link to Repo:  https://github.com/PedroDaumas/test_reflex
 - You can see the bug here https://github.com/PedroDaumas/test_reflex/blob/master/service/app/views/home/state.py#L62

**Expected behavior**
The variable should keep it's type.

**Screenshots**

 - Python Version: 3.8
 - Reflex Version: 0.2.6
 - OS: Linux
 - Browser (Optional): Chrome

masenf:
Calling an `EventHandler` returns an `EventSpec` with the function call arguments encoded as `Var`.

This works fine when formatting the `Var` into an event via `format.format_event`, however when directly creating a payload and serializing the entire `Event` to send to the client to be queued, the `fix_events` function was just taking the "name" of the `Var`, which is a json-encoded value.

In order to properly serialize the event payload, these `Var` args need to be decoded back into their python values, so that the payload as a whole can be re-serialized properly.

Implement a `Var._decode` operation, which operates (mostly) as the inverse of `Var.create`. This is an underscore method, because it actually doesn't return another `Var`, but instead the real python value that the `Var` is created from.

Fix #1722

---------------

Add/update unit and integration test cases in this area.

### DIFF
--- a/integration/test_event_chain.py
+++ b/integration/test_event_chain.py
@@ -31,6 +31,9 @@ def EventChain():
         def event_arg(self, arg):
             self.event_order.append(f"event_arg:{arg}")
 
+        def event_arg_repr_type(self, arg):
+            self.event_order.append(f"event_arg_repr:{arg!r}_{type(arg).__name__}")
+
         def event_nested_1(self):
             self.event_order.append("event_nested_1")
             yield State.event_nested_2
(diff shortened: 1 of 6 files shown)


picklelo (approved):
nice fix!

Issue #1709, opened 2023-08-29
rx.cached_var does not find dependencies inside a list comprehension


masenf:
**Describe the bug**
Using a cached var with a list comprehension does not get updated when the dependent vars change because `ComputedVar.deps` does not correctly identify dependencies inside a list comprehension.

**To Reproduce**
```python


class State(rx.State):
    _selected_images: set[str] = set()

    def toggle_image(self, image: str):
        if image in self._selected_images:
            self._selected_images.remove(image)
        else:
            self._selected_images.add(image)

    @rx.cached_var
    def image_infos(self) -> list[ImageInfo]:
        breakpoint()
        return [
            ImageInfo(image=image, selected=image in self._selected_images)
            for image in IMAGES
        ]
```

👆 `image_infos` never gets updated 😢

masenf:
WiP hacking: https://github.com/reflex-dev/reflex/tree/masenf/cached-var-comprehensions

PR #1728, opened 2023-08-31
Track var dependencies in comprehensions and nested functions

masenf:
Improve tracking of `rx.cached_var` dependencies when the function uses nested functions and list comprehensions.

Fix https://github.com/reflex-dev/reflex/issues/1709


picklelo (dismissed):
thanks for adding the unit tests :)

picklelo (approved):

PR #1729, opened 2023-08-31
state: _init_mutable_fields for backend vars as well

### LINKED ISSUE #1727: Backend vars of a state are not converted to special ReflexList/ReflexDict/ReflexSet instances
**Describe the bug**
This was probably by design, because backend vars are not present in the delta, so there's no reason to track dirtiness when they get modified...right? Well actually `rx.cached_var` tracks dependencies in backend vars as well, so that when a cached computed var depends on a backend var, it will be recomputed when the backend var is updated.

masenf:
ensure that backend var list, dict, and set are re-assigned as special reactive containers that can track modifications

fix #1727

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -127,27 +127,32 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
                             parent_state.parent_state,
                         )
 
-        # Initialize the mutable fields.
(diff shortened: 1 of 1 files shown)


picklelo (approved):
Looks good - we should add some unit tests in a follow up

PR #1739, opened 2023-09-01
ConnectionModal uses Cond for rendering

masenf:
If we let the ConnectionModal use it's `is_shown` prop, then this adds 3ms to every re-render, which is about 20% of an average render (of a tiny page)

### DIFF
--- a/reflex/components/layout/cond.py
+++ b/reflex/components/layout/cond.py
@@ -24,7 +24,10 @@ class Cond(Component):
 
     @classmethod
     def create(
-        cls, cond: Var, comp1: Component, comp2: Optional[Component]
+        cls,
+        cond: Var,
+        comp1: Component,
+        comp2: Optional[Component] = None,
     ) -> Component:
         """Create a conditional component.
 
--- a/reflex/components/overlay/banner.py
+++ b/reflex/components/overlay/banner.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 from typing import Optional
 
 from reflex.components.component import Component
-from reflex.components.layout import Box, Cond, Fragment
+from reflex.components.layout import Box, Cond
 from reflex.components.overlay.modal import Modal
 from reflex.components.typography import Text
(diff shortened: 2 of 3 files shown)


reflex/components/overlay/banner.py
```diff
@@ -63,7 +63,7 @@ def create(cls, comp: Optional[Component] = None) -> Component:
         return super().create(has_connection_error, comp, Fragment.create())  # type: ignore
 
 
-class ConnectionModal(Modal):
+class ConnectionModal(Cond):
```
picklelo, line 66: Should we just make this subclass Component and use a Cond inside the render?

reflex/components/overlay/banner.py
```diff
@@ -79,7 +79,11 @@ def create(cls, comp: Optional[Component] = None) -> Component:
...
-            header="Connection Error",
-            body=comp,
-            is_open=has_connection_error,
+            has_connection_error,
+            Modal.create(
+                header="Connection Error",
+                body=comp,
+                is_open=has_connection_error,
+            ),
+            Fragment.create(),
```
picklelo, line 88: Cond doesn't need a second argument

picklelo: Though I noticed the type annotation for `Cond.create` is wrong as we don't set a default value of None - we should fix that.

masenf: > we should fix that.

on it

reflex/components/layout/cond.py
```diff
@@ -24,7 +24,10 @@ class Cond(Component):
 
     @classmethod
     def create(
-        cls, cond: Var, comp1: Component, comp2: Optional[Component]
+        cls,
+        cond: Var,
+        comp1: Component,
+        comp2: Optional[Component] = None,
```
picklelo, line 30: nit: we can do `Component | None` now

picklelo (approved):

PR #1748, opened 2023-09-04
Reassign state Var when fields on a Base instance change

### LINKED ISSUE #1675: [REF-619] Custom Vars do not update after changing one of the value.
**Describe the bug**
When updating one of the field of a custom Var, the frontend does not receive the delta to update the state on it's side.

**To Reproduce**
```python
class User(rx.Base):
    email: str = ""
    first_name: str = ""
    last_name: str = ""

class State(rx.State):
    user: User = User(email="foo@gg.com", first_name="A", last_name="B")

    def set_user_email(self, mail):
        self.user.email = mail

@rx.page(route="/")
def idx():
    return rx.vstack(
        rx.text(f"Welcome {State.user.email}"),
        rx.input(value=State.user.email, on_change=State.set_user_email),
    )
```

The event handler trigger correctly because a refresh actually show the new value.

**Expected behavior**
The delta should be sent when updating a value of a custom var.

**Screenshots**

 - Python Version: 3.10
 - Reflex Version: 0.2.5
 - OS: Ubuntu 22.04
 - Browser (Optional): Chrome

<sub>[REF-619](https://linear.app/reflex-dev/issue/REF-619/custom-vars-do-not-update-after-changing-one-of-the-value)</sub>

masenf:
Refactor mutable type handling in State classes:
  * Create new `MutableProxy` object that wraps `list`, `dict`, `set`, and `Base`-derived objects coming from the State.
  * When certain `__magic_attrs__` are accessed on wrapped objects, and they are callable, mark the associated var field in the state as dirty.
  * When accessing attributes or items through the proxy, and the resulting object is also mutable, then also wrap it in `MutableProxy` associated with the original state and field_name.
  * When assigning a `MutableProxy` back to the state, actually assign the wrapped object.

This approach does not incur as much runtime overhead as the current implementation, since it does not create copies of objects just for state tracking and does not create copies of objects when assigning back to the state. Instead it just wraps the original mutable object.

Fixes #1675


masenf:
Yikes, this appears to approximately double the time taken by `tests/test_state.py` 😱

masenf:
So I was thinking about this one a bit more... and instead of doing a full re-assignment (and reinitialization, etc) for these mutable types, we probably should just mark them dirty in the state.

Essentially, we will add some code to `State.__getattr__` which returns an object proxy for mutable types, when any attribute is set or mutation method called, the proxy will mark the field as dirty in the state, and then pass the call through to the underlying object.

Then `State.__setattr__` will be extended to identify when attempting to assign an object proxy and just assign the object itself instead.

With this approach, we won't need specialty collection objects and we won't need slow up-front depth first traversals. We can just mark the fields appropriately at runtime when they are accessed/modified.

picklelo (approved):
nice, this solution ended up very clean :)

PR #1750, opened 2023-09-05
Var: __bool__ and __iter__ always raise a TypeError

masenf:
To help avoid programming mistakes when using Var, any attempt to use a Var in a conditional, or with short circuit operators will now raise a TypeError with the exception message suggesting the use of `rx.cond`, and bitwise operators instead.

Attempting to use a Var with a native for loop or list comprehension will also raise a TypeError suggesting the use of `rx.foreach`

----------------

Fix all test cases that were making erroneous Var assertions
  * must use `.equals()` to actually compare Var instances
  * must `._decode()` Vars to compare with python values
  * must use `if x is not None` instead of `if x` where x is a Var

### DIFF
--- a/reflex/vars.py
+++ b/reflex/vars.py
@@ -205,6 +205,27 @@ class Var(ABC):
             out = format.format_string(out)
         return out
 
+    def __bool__(self) -> bool:
+        """Raise exception if using Var in a boolean context.
+
+        Raises:
+            TypeError: when attempting to bool-ify the Var.
+        """
+        raise TypeError(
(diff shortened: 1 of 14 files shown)


reflex/vars.py
```diff
@@ -205,6 +205,27 @@ def __str__(self) -> str:
             out = format.format_string(out)
         return out
 
+    def __bool__(self) -> bool:
+        """Raise exception if using Var in a boolean context.
+
+        Raises:
+            TypeError: when attempting to bool-ify the Var.
+        """
+        raise TypeError(
```
picklelo, line 214: We can also link to the relevant portion of the docs here

picklelo (approved):
Awesome

PR #1751, opened 2023-09-05
Style props with Callable Values

ElijahAhianyo:
TODO: Add description

### DIFF
--- a/reflex/utils/format.py
+++ b/reflex/utils/format.py
@@ -9,7 +9,8 @@ import os
 import os.path as op
 import re
 import sys
-from typing import TYPE_CHECKING, Any, Type, Union
+import types as builtin_types
+from typing import TYPE_CHECKING, Any, Callable, Type, Union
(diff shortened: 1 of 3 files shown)


reflex/utils/format.py
```diff
@@ -308,6 +309,8 @@ def format_prop(
             # Dump the prop as JSON.
             prop = json_dumps(prop)
     except TypeError as e:
+        if "The style prop" in str(e):
```
masenf, line 312: String scanning exceptions is a code smell, because if the exception message gets changed or reworded, it's not obvious that this needs to be changed.

Consider defining a new exception like

```python
class InvalidStyleProp(TypeError):
    """An invalid object was passed for a style prop."""
```

And raise/catch that here.

ElijahAhianyo: got it

masenf (approved):

Issue #1756, opened 2023-09-06
Bun coredumps with Illegal instruction on older CPUs without AVX2 support


jdoss:
**Describe the bug**

I run Reflex on a bunch of older servers that do not have AVX2 support on their CPUs.  This causes Bun to coredump with an Illegal instruction error. The Bun developers created a baseline build to work around this issue https://github.com/oven-sh/bun/issues/733 and to work around this issue in Reflex I have to manually install the baseline version as part of my container build process. See my Containerfile below.

The `install.sh` provided by Bun actually detects this correctly https://github.com/oven-sh/bun/blob/7c82dc86baf3b01600b66d77737b2766ece83b03/src/cli/install.sh#L87-L99 but this can still have issues issues when deploying Reflex to production. In my case, my build servers that create the containers have AVX2 support so the Bun `install.sh` script uses the normal build, but my Reflex containers may get deployed in production to an older server that does not have AVX2 support and Reflex crashes. 

I think it would be handy to be able to control what bun build to install, normal or baseline as an environment variable in Reflex so the a user can set something like `BUN_BUILD=baseline` and Reflex will install that over the normal build. I am willing to submit a PR for this support if the project thinks it is a good idea. 

If I created a PR to support this feature, I'd move away from downloading the install.sh script and running that to install Bun. I'd just directly grab the correct bun build/version zip file from GitHub and install the Bun binary to the correct location for Reflex to use. Unfortunately the `install.sh` script from Bun doesn't have a means to override what buiild to use. It just auto detects the CPU features and installs the correct build. 

Otherwise I can keep overwriting Bun manually in my Containerfiles and hopefully this issue helps any other Reflex users work around Bun coredumping on older hardware without AVX2 support.

```
FROM python:3.11-slim as base

RUN adduser --disabled-password reflex

FROM base as build

WORKDIR /home/reflex/app
ENV VIRTUAL_ENV=/home/reflex/app/venv
RUN python3 -m venv $VIRTUAL_ENV
ENV PATH="$VIRTUAL_ENV/bin:$PATH"

COPY requirements.txt .

RUN pip install wheel \
  && pip install -r requirements.txt

COPY . .

FROM base as runtime

RUN apt-get update && apt-get install -y \
  curl \
  && curl -fsSL https://deb.nodesource.com/setup_19.x | bash - \
  && apt-get update && apt-get install -y \
  nodejs \
  unzip \
  && rm -rf /var/lib/apt/lists/*

[...]

PR #1757, opened 2023-09-06
Upgrade Bun 0.7.0 to 0.7.3

jdoss:
The 0.7.0 version core dumps when using the bun-linux-x64-baseline binary. Upgrading to 0.7.3 fixes the issues. See https://github.com/reflex-dev/reflex/issues/1756 for more context on why I have to use the baseline build of Bun.

I have tested three reflex sites with Bun 0.7.3 and everything seems to work fine :tm:

### All Submissions:

### Type of change

### DIFF
--- a/reflex/constants.py
+++ b/reflex/constants.py
@@ -69,9 +69,9 @@ JINJA_TEMPLATE_DIR = os.path.join(TEMPLATE_DIR, "jinja")
 
 # Bun config.
 # The Bun version.
-BUN_VERSION = "0.7.0"
+BUN_VERSION = "0.7.3"
 # Min Bun Version
-MIN_BUN_VERSION = "0.7.0"
+MIN_BUN_VERSION = "0.7.3"
 # The directory to store the bun.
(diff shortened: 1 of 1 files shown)


reflex/constants.py
```diff
@@ -69,9 +69,9 @@ def get_fnm_name() -> str | None:
 
 # Bun config.
 # The Bun version.
-BUN_VERSION = "0.7.0"
+BUN_VERSION = "0.7.3"
 # Min Bun Version
-MIN_BUN_VERSION = "0.7.0"
+MIN_BUN_VERSION = "0.7.3"
```
masenf, line 74: I think we should leave the `MIN_BUN_VERSION` alone, since we _can_ run on 0.7.0, it just causes issues in certain configurations. The `BUN_VERSION` determines what will get installed by reflex when it is bootstrapping.

masenf (changes requested):

reflex/constants.py
jdoss: Sounds good. Pushed a change with that feedback.

masenf (approved):

PR #1758, opened 2023-09-06
Send updated client-side storage values when hydrating

masenf:
Re-use the `initialEvents` list when hydrating on page navigation events.

Eventually, we will decouple hydrate from `on_load` and this problem will go away, but until then, since we're resetting the client storage values on the hydrate event, each hydrate event needs to include the updated client storage values, so just reuse the `initialEvents` from `context.js` when re-hydrating for page loads.

### DIFF
--- a/reflex/.templates/jinja/web/pages/index.js.jinja2
+++ b/reflex/.templates/jinja/web/pages/index.js.jinja2
@@ -25,7 +25,7 @@ export default function Component() {
 
   // Route after the initial page hydration.
   useEffect(() => {
-    const change_complete = () => Event([E('{{state_name}}.{{const.hydrate}}', {})])
+    const change_complete = () => Event(initialEvents.map((e) => ({...e})))
     {{const.router}}.events.on('routeChangeComplete', change_complete)
     return () => {
       {{const.router}}.events.off('routeChangeComplete', change_complete)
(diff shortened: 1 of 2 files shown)


reflex/.templates/jinja/web/pages/index.js.jinja2
```diff
@@ -25,7 +25,7 @@ export default function Component() {
 
   // Route after the initial page hydration.
   useEffect(() => {
-    const change_complete = () => Event([E('{{state_name}}.{{const.hydrate}}', {})])
+    const change_complete = () => Event(initialEvents.map((e) => ({...e})))
```
picklelo, line 28: Why can't this just be `() => Event(initialEvents)` ?

masenf: Because with the new client-side routing code, we only update the `router_data` in `applyEvent` if it's not already set. So if we don't spread the `initialEvents` to make a copy, then subsequent calls with hydrate with the wrong router_data.

I don't love it. But we're getting rid of it soon, so this could be considered a workaround until we fully decouple hydrate and `on_load`. (Hydrate doesn't care about the router_data, only the on_load events do).

picklelo (approved):

PR #1764, opened 2023-09-06
Move custom styles to root App file(_app.js)

ElijahAhianyo:
In this PR, we enhance maintainability and optimize the loading performance of our application. We've moved custom styles from the `<head>` tag in `_document.js` to `_app.js`. Additionally, all stylesheets and fonts are now consolidated into a root stylesheet (styles.css) at `.web/styles/styles.css`, which is imported into `_app.js`. Please note that the actual impact on performance may vary based on specific factors.

### Changes Made

- Relocated Custom Styles: Custom styles, formerly in the `<head>` tag in `_document.js`, are now in `_app.js`. This follows Next.js best practices, reducing HTTP requests during initial page load and enabling efficient style caching.

- Consolidated Stylesheets and Fonts: All stylesheets and fonts are consolidated into a root stylesheet (styles.css). This streamlines our codebase, simplifies styling management, and ensures browser caching for improved performance.

### Caching Considerations

Previously, styles in `_document.js` were not efficiently cached, except when served from a CDN. Moving styles to `_app.js` and consolidating them into a root stylesheet ensures efficient loading and caching, resulting in faster subsequent page loads.

### Performance Benefits:
- Varied Gain: The performance gain achieved by this change may vary significantly based on the size of the stylesheet. Smaller stylesheets may show minimal improvement, while larger stylesheets are likely to experience more noticeable gains.
- Reduced HTTP Requests: Consolidating styles minimizes initial HTTP requests, further optimizing performance.

- Simplified Maintenance: Centralizing styles simplifies maintenance and updates as the application evolves.

Additional Notes:

This change aligns with Next.js best practices, leveraging browser caching. It also streamlines our codebase by consolidating styles into a root stylesheet for improved maintainability.


ElijahAhianyo:
Pipeline needs https://github.com/reflex-dev/reflex-web/pull/202 merged to pass

reflex/compiler/compiler.py
```diff
@@ -116,6 +119,54 @@ def _compile_page(
...
+    def is_valid_url(url):
+        result = urlparse(url)
+        return all([result.scheme, result.netloc])
+
+    sheets = [constants.TAILWIND_ROOT_STYLE_PATH]
+    for stylesheet in stylesheets:
+        if not is_valid_url(stylesheet):
+            # check if stylesheet provided exists.
+            if not os.path.exists(
+                Path.cwd() / constants.APP_ASSETS_DIR / stylesheet.strip("/")
```
picklelo, line 160: Maybe store this in a variable, since you construct this path twice

reflex/compiler/compiler.py
```diff
@@ -116,6 +119,54 @@ def _compile_page(
...
+        stylesheets: The stylesheets to include in the root stylesheet.
+
+    Returns:
+        The compiled root stylesheet.
+
+    Raises:
+        FileNotFoundError: If a specified stylesheet in assets directory does not exist.
+    """
+
+    def is_valid_url(url):
```
picklelo, line 151: This could be moved to the `compiler/utils.py` file

reflex/compiler/compiler.py
```diff
@@ -116,6 +119,54 @@ def _compile_page(
...
+    for stylesheet in stylesheets:
+        if not is_valid_url(stylesheet):
+            # check if stylesheet provided exists.
+            if not os.path.exists(
+                Path.cwd() / constants.APP_ASSETS_DIR / stylesheet.strip("/")
+            ):
+                raise FileNotFoundError(
+                    f"The stylesheet file {Path.cwd() / constants.APP_ASSETS_DIR / stylesheet} does not exist."
+                )
+            stylesheet = "/".join(["../public", stylesheet.strip("/")])
```
picklelo, line 165: Do we need to go back a directory to `../public` because this gets compiled into the `pages/_app.js` file? Maybe it would bee better to make this `/public/` which should get it from the app root directory (I think)

reflex/compiler/utils.py
```diff
@@ -257,18 +256,14 @@ def compile_custom_component(
...
-    Args:
-        stylesheets: The list of stylesheets to include in the document root.
-
     Returns:
         The document root.
     """
-    sheets = [RawLink.create(rel="stylesheet", href=href) for href in stylesheets]
     return Html.create(
-        DocumentHead.create(*sheets),
+        DocumentHead.create(),
```
picklelo, line 266: Do we even need this now?

reflex/compiler/utils.py
```diff
@@ -324,6 +319,17 @@ def get_theme_path() -> str:
     return os.path.join(constants.WEB_UTILS_DIR, constants.THEME + constants.JS_EXT)
 
 
+def get_root_stylesheet_path() -> str:
```
picklelo, line 322: This could be defined in `constants.py`

ElijahAhianyo: Yeah I thought so but I had it here for consistency with other path functions

reflex/compiler/utils.py
ElijahAhianyo: Yes we still do, the page metadata arent loaded without it, which means the stylesheets wont render

reflex/compiler/compiler.py
ElijahAhianyo: Yeah, I just made a push to change this.

picklelo (approved):
looks good to me!

Issue #1732, opened 2023-08-31
Add .pyi files for app.py, state.py, and config.py


picklelo:
These files also need generated .pyi files.

PR #1765, opened 2023-09-06
improve pyi files generation

### LINKED ISSUE #1730: Add docstrings to .pyi files
The generated .pyi files should include the docstrings from the original class. We should also see if we can extract the component prop descriptions from the comments (using the same script we use on the reflex-web website) to improve the docstrings.

Lendemor:
### All Submissions:

### Description

Improve the generation of pyi files.

Closes #1730 and #1732


reflex/app.py
```diff
@@ -249,9 +249,9 @@ async def preprocess(self, state: State, event: Event) -> StateUpdate | None:
         """
         for middleware in self.middleware:
             if asyncio.iscoroutinefunction(middleware.preprocess):
-                out = await middleware.preprocess(app=self, state=state, event=event)
+                out = await middleware.preprocess(app=self, state=state, event=event)  # type: ignore
```
picklelo, line 252: Why are the type ignore's needed now? Can we updated the .pyi so we don't need them?

picklelo (commented):
I think we wanted to sort the props in the docstrings if possible

reflex/app.py
Lendemor: I didn't find a way, because of the self reference, pyright distinguish between the App in app.py and the one in App.pyi (they are considered different object, so self referenced doesn't pass pyright)

picklelo (approved):

PR #1769, opened 2023-09-06
be more specific with python versions - so caching can be safe

jackie-pc:
Python published minor releases 1-2 weeks ago.  Our cached python artifacts started becoming unusable sometimes.   CI logs hint at minor versions being off (cached with 3.11.5, but we use the cache on a runner with another minor version).  Theory is if we pin versions more precisely we won't get cache misalignment any more.

Litmus test is to rerun all CI jobs with this PR, multiple times, and look for lack of any flakes.


picklelo (approved):

masenf (approved):
🙌

PR #1781, opened 2023-09-07
Don't show error logs on sigint

picklelo:
Avoid showing logs when exiting the app in a Docker container

### DIFF
--- a/reflex/utils/processes.py
+++ b/reflex/utils/processes.py
@@ -217,7 +217,8 @@ def stream_logs(message: str, process: subprocess.Popen):
             logs.append(line)
             yield line
 
-    if process.returncode != 0:
+    # Check if the process failed (not printing the logs for SIGINT).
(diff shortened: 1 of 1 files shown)


reflex/utils/processes.py
```diff
@@ -217,7 +217,8 @@ def stream_logs(message: str, process: subprocess.Popen):
             logs.append(line)
             yield line
 
-    if process.returncode != 0:
+    # Check if the process failed (not printing the logs for SIGINT).
+    if process.returncode not in [0, 2]:
```
masenf, line 221:
```suggestion
    if process.returncode not in [0, -2]:
```

SIGINT is `-2`

masenf (changes requested):

masenf (approved):
tested in `python:3.11` on debian 12 host

PR #1788, opened 2023-09-08
supply default for rx.Model PK for both DB and python to work

martinxu9:
Summary
- According to https://sqlmodel.tiangolo.com/tutorial/automatic-id-none-refresh/, the primary key should be defined as Optional with default None for python to work (static type checker complains) while DB works.

Tests
- Updated UTs. They should still pass.


picklelo (approved):
good catch!

PR #1797, opened 2023-09-11
add download event

Lendemor:
### All Submissions:

### Type of change

### New Feature Submission:

### Description

Add an API for easily downloading files.

Usage is as follow:
```python
rx.button("Download file", on_click=rx.download("/uri/for/file.extension", "target_file.extension")
```

or 
```python
def export_data(self):
    ...
    # do export logic here and write to filepath
    # then
    yield rx.download(filepath, filename)
```

### DIFF
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -73,7 +73,7 @@ export const getToken = () => {
  * @param delta The delta to apply.
  */
 export const applyDelta = (state, delta) => {
-  const new_state = {...state}
+  const new_state = { ...state }
   for (const substate in delta) {
     let s = new_state;
     const path = substate.split(".").slice(1);
@@ -151,6 +151,24 @@ export const applyEvent = async (event, socket) => {
   if (event.name == "_set_clipboard") {
     const content = event.payload.content;
     navigator.clipboard.writeText(content);
+    return false;
+  }
+  if (event.name == "_download") {
+    const url = event.payload.url;
+    const filename = event.payload.filename;
+
+    // create a hidden 'a' object that will trigger the download
+    fetch(url)
+      .then(response => response.blob())
+      .then(blob => {
+        const blobUrl = URL.createObjectURL(blob);
+        const a = document.createElement('a');
+        a.href = blobUrl;
+        a.download = filename;
+        a.click();
+        URL.revokeObjectURL(blobUrl);
+      })
+      .catch(error => console.error('Error downloading file:', error));
     return false;
   }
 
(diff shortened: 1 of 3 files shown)


reflex/.templates/web/utils/state.js
```diff
@@ -153,6 +153,24 @@ export const applyEvent = async (event, socket) => {
...
+      .then(response => response.blob())
+      .then(blob => {
+        const blobUrl = URL.createObjectURL(blob);
+        const a = document.createElement('a');
+        a.href = blobUrl;
+        a.download = filename;
+        a.click();
+        URL.revokeObjectURL(blobUrl);
+      })
+      .catch(error => console.error('Error downloading file:', error));
```
masenf, line 171: This makes me a bit nervous. If the download fails, how will the user know? How can the developer determine it failed and respond accordingly?

masenf (commented):
I'm a little unclear on why this would be preferred over just using a standard link. This seems like more moving parts and harder to deal with error handling.

If I wanted to provide a download link, I would keep the href in a state Var and render a normal `<a>` tag in the frontend.

Lendemor:
> I'm a little unclear on why this would be preferred over just using a standard link. This seems like more moving parts and harder to deal with error handling.
> 
> If I wanted to provide a download link, I would keep the href in a state Var and render a normal `<a>` tag in the frontend.

This only work if the file is already available beforehand.
Using a `<a>` with a href doesn't let you execute an event handler before starting the download.

Take an example where you have a data_table and an export button.
You can see your data in the datatable, but the file doesn't exist yet (you could want to apply some filter before export etc..)
Clicking an export button will trigger an event handler where you can both do your logic and then yield the front side download event.

reflex/.templates/web/utils/state.js
Lendemor: We could probably improve the logic in there, but the developer should be able to check the file before actually yielding the download event.

masenf: if there is some client side problem, that would be invisible to the user though; it would just seem to not work, unless they check dev tools

masenf:
> Using a `<a>` with a href doesn't let you execute an event handler before starting the download.

Fair enough, but is it possible make the hidden `<a>` that gets clicked link to the actual URL and let the client browser deal with the error handling? Or is there some security restriction that requires the clicked link href to be a `URL.createObjectURL`?

reflex/.templates/web/utils/state.js
Lendemor: Good point.
I'll see what I can change :+1:

reflex/event.py
```diff
@@ -330,6 +330,29 @@ def set_clipboard(content: str) -> EventSpec:
     )
 
 
+def download(url: str, filename: str) -> EventSpec:
```
picklelo, line 333: If they don't provide a filename, maybe we can default to the last part of the path in the url? Basically give it the same filename it was saved as

picklelo (commented):
Code works well for me! Just one suggestion

reflex/event.py
Lendemor: Sure we could do that

reflex/event.py
```diff
@@ -330,6 +330,34 @@ def set_clipboard(content: str) -> EventSpec:
     )
 
 
+def download(url: str, filename: Optional[str] = None) -> EventSpec:
```
picklelo, line 333: nit: Instead of `Optional` we can do `str | None` now

picklelo (approved):
awesome, great work

masenf (commented):
Sorry didn't re-review, but the merged implementation is killer, thanks for making the changes @Lendemor

PR #1816, opened 2023-09-14
Add serializers for different var types

picklelo:
Factored out var serialization so that it's not hardcoded within the compiler. The user can now define a serializer using the `@rx.serializer` decorator.

Added serializers for Plotly, dict, and date time.


reflex/components/graphing/plotly.py
```diff
@@ -25,7 +25,7 @@ class Plotly(PlotlyLib):
     is_default = True
 
     # The figure to display. This can be a plotly figure or a plotly data json.
-    data: Var[Figure]
+    data: Var[Any]
```
masenf, line 28:
```suggestion
    data: Var["Figure"]
```

And then at the top of the file, do something like

```python
try:
    from plotly.graph_objects import Figure
except ImportError:
    Figure = Any
```

That way the type checking at least works nicer when you do have plotly installed.

reflex/components/media/image.py
```diff
@@ -60,9 +63,37 @@ def get_triggers(self) -> Set[str]:
...
-        if types.is_image(type(self.src)):
-            self.src = Var.create(format.format_image_data(self.src))  # type: ignore
+        try:
+            from PIL.Image import Image as Img
+
+            # If the src is an image, convert it to a base64 string.
+            if types._isinstance(self.src, Img):
+                self.src = Var.create(serialize(self.src))  # type: ignore
+        except ImportError:
+            pass
```
masenf, line 73: Do we need the import guard here? Doesn't seem like this block is actually using `Img`

reflex/utils/serializers.py
```diff
@@ -0,0 +1,180 @@
...
+        raise ValueError("Serializer must take a single argument.")
+
+    # Get the type of the argument.
+    type_ = type_hints[args[0]]
+
+    # Make sure the type is not already registered.
+    registered_fn = SERIALIZERS.get(type_)
+    if registered_fn is not None and registered_fn != fn:
+        raise ValueError(
+            f"Serializer for type {type_} is already registered as {registered_fn}."
```
masenf, line 47: `registered_fn.__qualname__` might be useful here, in case the name is ambiguous

reflex/utils/serializers.py
```diff
@@ -0,0 +1,180 @@
...
+def serialize(value: Any) -> str | dict | None:
+    """Serialize the value to a JSON string.
+
+    Args:
+        value: The value to serialize.
+
+    Returns:
+        The serialized value, or None if a serializer is not found.
+    """
+    global SERIALIZERS
```
masenf, line 66:
```suggestion
```

reflex/utils/serializers.py
```diff
@@ -0,0 +1,180 @@
...
+        type_: The type to check.
+
+    Returns:
+        Whether there is a serializer for the type.
+    """
+    return get_serializer(type_) is not None
+
+
+@serializer
+def serialize_dict(prop: Dict[str, Any]) -> str:
```
masenf, line 117: can the existing `format.format_dict` function just be decorated as `@serializer` and then moved into this file later? the duplication of such complex functionality makes me nervous.

reflex/components/media/image.py
masenf: i guess i was thinking we can just call `self.src = Var.create(serialize(self.src))` and not care about what `self.src` is... looks like `serialize` has a default `str` handler.

tests/utils/test_serializers.py
```diff
@@ -0,0 +1,102 @@
...
+            The serialized value.
+        """
+        return str(value)
+
+    # Initially there should be no serializer for int.
+    assert not serializers.has_serializer(int)
+    assert serializers.serialize(5) is None
+
+    # Register the serializer.
+    serializers.serializer(serialize_test)
```
masenf, line 70:
```suggestion
    assert serializers.serializer(serialize_test) is serialize_test
```

just a bit of a regression test to make sure `serializer` isn't munging the object it decorates

masenf (approved):
killer!

reflex/components/graphing/plotly.py
```diff
@@ -39,14 +43,22 @@ class Plotly(PlotlyLib):
     # If true, the graph will resize when the window is resized.
     use_resize_handler: Var[bool]
 
-    def _render(self) -> Tag:
```
picklelo, line 42: Turns out this wasn't actually doing anything... We can reinvestigate how to add this logic back in.

reflex/components/media/image.py
```diff
@@ -60,6 +61,12 @@ def get_triggers(self) -> set[str]:
         """
         return super().get_triggers() | {"on_error", "on_load"}
 
+    def _render(self) -> Tag:
+        self.src.is_string = True
+
+        # Render the table.
```
masenf, line 67:
```suggestion
        # Render the image.
```

😉

masenf (approved):

reflex/components/media/image.py
picklelo: Will fix this in the follow up

PR #1820, opened 2023-09-15
New API to define triggers

Lendemor:
### All Submissions:

### Type of change

### Changes To Core Features:

### Description

Change the API for defining the signature expected for any event triggers (uncontrolled or controlled):

```python
def get_event_triggers(self):
    return super().get_event_triggers() | { 
        "trigger_name": lambda: [], #uncontrolled 
        "trigger_name": lambda e0: [e0], #1 args 
        "trigger_name": lambda e0,e1: [e0, e1.target.value], #2 args or more
    }
```

Deprecate the following : `get_controlled_triggers`, `EVENT_ARG`, `get_triggers`, `EventChain`

Also remove the unused and not working component `CopyToClipboard`


reflex/constants.py
```diff
@@ -363,6 +363,37 @@ class RouteRegex(SimpleNamespace):
...
+    ON_MOUSE_UP = "on_mouse_up"
+    ON_SCROLL = "on_scroll"
+    ON_SUBMIT = "on_submit"
+    ON_MOUNT = "on_mount"
+    ON_UNMOUNT = "on_unmount"
+
+
 # Names of event handlers on all components mapped to useEffect
 ON_MOUNT = "on_mount"
 ON_UNMOUNT = "on_unmount"
```
masenf, line 399: probably can ditch these now

reflex/event.py
```diff
@@ -462,7 +511,9 @@ def call_event_fn(fn: Callable, arg: Var) -> list[EventSpec]:
     return events
 
 
-def get_handler_args(event_spec: EventSpec, arg: Var) -> tuple[tuple[Var, Var], ...]:
+def get_handler_args(
+    event_spec: EventSpec, arg: Var | ArgsSpec
```
masenf, line 515: does this function even use `arg`?

masenf (commented):
This looks awesome!

reflex/event.py
Lendemor: It doesn't, I just kept the signature since `arg` was passed beforehand, but we could remove it I think.

reflex/event.py
```diff
@@ -392,21 +397,62 @@ def call_event_handler(event_handler: EventHandler, arg: Var) -> EventSpec:
...
+    Args:
+        arg_spec: The spec of the args.
+
+    Returns:
+        The parsed args.
+    """
+    spec = inspect.getfullargspec(arg_spec)
+    return arg_spec(
+        *[
+            BaseVar(name=f"_{l_arg}", type_=FrontendEvent, is_local=True)
```
masenf, line 449:
```suggestion
            BaseVar(
                name=f"_{l_arg}",
                type_=spec.annotations.get(l_arg, FrontendEvent),
                is_local=True,
            )
```

We should let the user supply the correct types here:

```python
class C1State(State):
    def mock_handler(self, _e, bravo, charlie):
        pass


def test_component_event_trigger_arbitrary_args():
    class Obj(rx.Base):
        custom: int = 0

    def on_foo(
        _e,
        alpha: str,
        bravo: dict[str, Any],
        charlie: Obj,
    ):
        return [_e.target.value, bravo["nested"], charlie.custom + 42]

    class C1(Component):
        library = "/local"
        tag = "C1"

        def get_event_triggers(self) -> Dict[str, Any]:
            return {
                "on_foo": on_foo,
            }

    rendered = C1.create(on_foo=C1State.mock_handler).render()
    assert rendered["props"][0] == (
        "onFoo={(__e,_alpha,_bravo,_charlie) => addEvents("
        '[Event("c1_state.mock_handler", {_e:__e.target.value,bravo:_bravo["nested"],charlie:(_charlie.custom + 42)})], '
        "(__e,_alpha,_bravo,_charlie))}"
    )
```

masenf (changes requested):

reflex/event.py
Lendemor: I've added your suggestion, it seems to work when running a normal app, but running `pytest` fail with type errors.
(for both the Dict and custom Obj arguments)

masenf (dismissed):
Love the new API. This is great!

reflex/components/component.py
```diff
@@ -258,45 +257,93 @@ def _create_event_chain(
...
+            deprecated_triggers = {
+                trigger: lambda: [] for trigger in deprecated_triggers
+            }
+        else:
+            deprecated_triggers = {}
+
+        deprecated_controlled_triggers = self.get_controlled_triggers()
+        if deprecated_controlled_triggers:
+            console.deprecate(
+                feature_name="get_controlled_triggers ({self.__class__.__name__})",
```
picklelo, line 311: Should this be an f-string?

reflex/components/forms/checkbox.py
```diff
@@ -48,14 +49,15 @@ class Checkbox(ChakraComponent):
     # The spacing between the checkbox and its label text (0.5rem)
     spacing: Var[str]
 
-    def get_controlled_triggers(self) -> Dict[str, Var]:
+    def get_event_triggers(self) -> dict[str, Union[Var, Any]]:
```
picklelo, line 52: We could define a type `Callable[Var, List[Var]]` for the lambda type and use that instead of `Any`. Also, when would `Var` be the dictionary value type, wouldn't it always be a Callable?

reflex/utils/format.py
```diff
@@ -300,9 +301,21 @@ def format_prop(
 
         # Handle event props.
         elif isinstance(prop, EventChain):
+            if prop.args_spec is None:
```
picklelo, line 304: Maybe we can move this into a function `format_event_chain` as it's quite complex? (Can be done in a follow up)

reflex/components/component.py
```diff
@@ -220,36 +219,36 @@ def _create_event_chain(
...
 
+        wrapped = False
         # If the input is a single event handler, wrap it in a list.
         if isinstance(value, (EventHandler, EventSpec)):
+            wrapped = True
             value = [value]
 
         # If the input is a list of event handlers, create an event chain.
         if isinstance(value, List):
+            if not wrapped:
```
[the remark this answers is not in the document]
masenf: I'm good deprecating it. I think we should seek community feedback before removing it.

reflex/utils/format.py
Lendemor: This is the code for the backward compatibility of previous EVENT_ARG, we can simplify it later when we actually remove the feature.

reflex/components/forms/checkbox.py
Lendemor: I put Any because my previous definition using LambdaTypes was causing a lot of pyright errors.

Also the Union[Var, ...] is for backward compatibility until removal of deprecated features.

masenf (approved):
re-approve

picklelo (approved):

PR #1827, opened 2023-09-18
Prevent Substate class shadowing

ElijahAhianyo:
This PR aims to prevent shadowing children substate which inherently leads to unexpected behaviors since they're not exactly shadowed. With this PR the following example will throw an error:

```python
import reflex as rx

class State(rx.State):
       ...

class ChildState(State):
       ...

class ChildState(State):
      ... 
```

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -93,6 +93,9 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
             *args: The args to pass to the Pydantic init method.
             parent_state: The parent state.
             **kwargs: The kwargs to pass to the Pydantic init method.
+
+        Raises:
+            ValueError: If a substate class shadows another.
         """
         kwargs["parent_state"] = parent_state
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


picklelo (approved):
nice fix!

PlutarcoII:
> This PR aims to prevent shadowing children substate which inherently leads to unexpected behaviors since they're not exactly shadowed. With this PR the following example will throw an error:
> 
> ```python
> import reflex as rx
> 
> class State(rx.State):
>        ...
> 
> class ChildState(State):
>        ...
> 
> class ChildState(State):
>       ... 
> ```

In reflex docs substate is an option when your app is growing up.. https://reflex.dev/docs/state/substates/
Could you please provide an example how-to do this with this new update?

masenf:
@PlutarcoII the examples in the docs _should_ still work. This PR just prevents substates from having the exact same name due to how the state tree is implemented internally. It's possible that in the future this restriction may be lifted, but for now, just use different class names for each substate and it will work.

PlutarcoII:
@masenf I arrive here due to an error in my code when I upgraded to reflex 0.2.8, but I didn't realize that I had two classes with the same name in different files. Thanks for your clarification.

PR #1828, opened 2023-09-18
rx.table __bool__ regression fix

### LINKED ISSUE #1826: Rows in component Table cannot be converted to bool for user with 'if'
**Describe the bug**
In the component Table code, rows passed to the table to be displayed is used in a if statement. For this reason, when execute reflex run, the next message appears:

```
 File "C:\Users\******\Documents\Herramientas061\herramientas061\pages\buscartrabajadores.py", line 148, in buscartrabajadores
    rx.table(
  File "C:\Python311\Lib\site-packages\reflex\components\datadisplay\table.py", line 53, in create
    if rows:
  File "C:\Python311\Lib\site-packages\reflex\vars.py", line 214, in __bool__
    raise TypeError(
TypeError: Cannot convert Var 'principal_state.buscar_trabajadores_state.resultado_busqueda' to bool for use with `if`, `and`, `or`, and `not`. Instead use `rx.cond` and bitwise operators `&` (and), `|` (or), `~` (invert).
(venv)
```

The error points to this if statement in table.py component:

```python

        if len(children) == 0:
            children = []

            if caption:
                children.append(TableCaption.create(caption))

            if headers:
                children.append(Thead.create(headers=headers))

            if rows:  # <--- ERROR HERE
                children.append(Tbody.create(rows=rows))

            if footers:
                children.append(Tfoot.create(footers=footers))
        return super().create(*children, **props)

```

Running 0.2.7 version. In 0.2.6 version worked for me. The error appeared updating reflex to 0.2.7 and perform the reflex init

Some code here:

```python

class BuscarTrabajadoresState (PrincipalState):
    text: str = ""
    unidad_funcional: str = ""
    
    @rx.var
    def resultado_busqueda(self) -> List[Tuple[str,str,str,str]] :
        if len(self.text)>0 or len(self.unidad_funcional)>0:
            return buscar_en_lineas(self.text, self.unidad_funcional)
        else:
            return []
```

And in layout:

```python

rx.table(
    headers=["DNI", "Nombre", "Categoría/Unidad Funcional", "Telefono"],
    rows=BuscarTrabajadoresState.resultado_busqueda,
    variant="striped",
    color_scheme="blue",
    width="100%"
),

```

 - Python Version: 3.11.4
 - Reflex Version: 0.2.7
 - OS: Windows 10
 - Browser (Optional): Chrome

ElijahAhianyo:
This PR fixes a regression in the table component. It updates the If statement/condition to check for `None` rather than the truthiness of a Var which is currently disallowed.
It also adds type-checking to the props of `Tfoot`, `Thead` and `Tbody`.
In addition, integration tests have been added for the component as well unit tests.

fixes #1826

### DIFF
--- a/reflex/components/datadisplay/table.py
+++ b/reflex/components/datadisplay/table.py
@@ -129,7 +129,7 @@ class Tbody(ChakraComponent):
             Component: The table body component
         """
         if len(children) == 0:
-            cls.validate_rows(rows)
+            cls.validate_rows(rows) if rows is not None else None
 
             if isinstance(rows, Var):
                 children = [
--- a/reflex/components/datadisplay/table.py
+++ b/reflex/components/datadisplay/table.py
@@ -97,13 +97,14 @@ class Thead(ChakraComponent):
             TypeError: If headers are not of type list or type tuple.
 
         """
+        allowed_types = (list, tuple)
         if (
             (
                 isinstance(headers, Var)
-                and types.get_base_class(headers.type_) not in (list, tuple)
+                and not types.check_type_in_allowed_types(headers.type_, allowed_types)
             )
             or not isinstance(headers, Var)
-            and types.get_base_class(type(headers)) not in (list, tuple)
+            and not types.check_type_in_allowed_types(type(headers), allowed_types)
         ):
             raise TypeError("table headers should be a list or tuple")
 
@@ -153,6 +154,7 @@ class Tbody(ChakraComponent):
         Raises:
             TypeError: If rows are not lists or tuples containing inner lists or tuples.
         """
+        allowed_subclasses = (List, Tuple)
         if isinstance(rows, Var):
             outer_type = rows.type_
             inner_type = (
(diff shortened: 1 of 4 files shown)


reflex/components/datadisplay/table.py
```diff
@@ -77,8 +78,20 @@ def create(cls, *children, headers=None, **props) -> Component:
...
         if len(children) == 0:
+            if (
+                (
+                    isinstance(headers, Var)
+                    and types.get_base_class(headers.type_) not in (list, tuple)
+                )
+                or not isinstance(headers, Var)
+                and types.get_base_class(type(headers)) not in (list, tuple)
+            ):
+                raise TypeError("table headers should be a list or tuple")
```
masenf, line 94: i think it would be nice to extract this logic out into a helper function for readability

masenf (approved):
one suggestion, but overall i'm good with this.

picklelo (changes requested):
Agreed with @masenf , can we refactor the repeated type checking logic into a common function to clean things up?

reflex/components/datadisplay/table.py
ElijahAhianyo: got it

picklelo (approved):

PR #1831, opened 2023-09-18
Core Graphing Rework 

Alek99:
## Note

I am still getting the error: 
```
Unhandled Runtime Error
Error: Hydration failed because the initial UI does not match what was rendered on the server.

See more info here: https://nextjs.org/docs/messages/react-hydration-error
```

##  Example Code To Test
```python
import reflex as rx


class State(rx.State):
    data = [
  {
    "name": "A",
    "uv": 4000,
    "pv": 2400,
    "amt": 2400
  },
  {
    "name": "B",
    "uv": 3000,
    "pv": 1398,
    "amt": 2210
  },
  {
    "name": "C",
    "uv": 2000,
    "pv": 9800,
    "amt": 2290
  },
  {
    "name": "D",
    "uv": 2780,
    "pv": 3908,
    "amt": 2000
  },
  {
    "name": "E",
    "uv": 1890,
    "pv": 4800,
    "amt": 2181
  },
  {
    "name": "F",
    "uv": 2390,
    "pv": 3800,
    "amt": 2500
  },
  {
    "name": "G",
    "uv": 3490,
    "pv": 4300,
    "amt": 2100
  }
]

 
def area_example():
    return rx.area_chart(
            rx.area(
                data_key="uv",
                stroke="#8884d8",
                fill="#8884d8",
            ),
            rx.area(
                data_key="pv",
                stroke="#82ca9d", 
                fill="#82ca9d",
            ),
            rx.x_axis(
                data_key="name",
            ),
            rx.y_axis(), 
            data=State.data,
            width=1000,
            height=400,
        )

def bar_example():
    return rx.bar_chart(
            rx.bar(
                data_key="uv",
                fill="#8884d8",
            ),
            rx.bar(
                data_key="pv",
                fill="#82ca9d",
            ),
            rx.x_axis(
                data_key="name",
            ),
            rx.y_axis(),
            data=State.data,
            width=500,
            height=400,
    )

def line_example():
    return rx.line_chart(
            rx.line(
                data_key="uv",
                stroke="#8884d8",
                fill="#8884d8",
            ),
            rx.line(
                data_key="pv",
                stroke="#82ca9d",
                fill="#82ca9d",
            ),
            rx.x_axis(
                data_key="name",
            ),
            rx.y_axis(),
            data=State.data,
            width=500,
            height=400,
        )


def composed():
    return 


def index():
    return rx.vstack(
        area_example(),
        rx.hstack(
            bar_example(),
            line_example(),
        ),
        width = "100%",
        padding_top = "5em",
    )


app = rx.App()
app.add_page(index)
app.compile()
```


Alek99:
The integration test are failing here because reflex web is using victory charts and I have removed them, instead of removing should we consider deprecating. However victory charts are broken because they can't work with state so totally removing also makes sense here

reflex/components/graphing/recharts/charts.py
```diff
@@ -0,0 +1,191 @@
...
+
+from typing import Any, Dict, List, Union
+
+from reflex.vars import Var
+
+from .recharts import Recharts
+
+
+class ChartBase(Recharts):
+    """A component that wraps a victory lib."""
```
masenf, line 11:
```suggestion
    """A base class for cartesian charts in Recharts."""
```

masenf (commented):
For whatever reason, I'm getting a hydration error when running the sample code... 🤔

reflex/components/graphing/recharts/charts.py
Alek99: Yeah that's expected getting the same error as well

Alek99: I'm pretty sure I just need to disable SSR but when doing this the graphs seems to go blank, investigating

Alek99: Nikhil found this which fixed https://stackoverflow.com/questions/72311188/hydration-failed-error-using-recharts-with-nextjs

PR #1832, opened 2023-09-18
Graphing Support for Matplotlib

Alek99:
Code for testing:
```python 
import reflex as rx
import altair as alt
import numpy as np
import pandas as pd
from bokeh.plotting import figure
import matplotlib.pyplot as plt

#Altair
# Compute x^2 + y^2 across a 2D grid
x, y = np.meshgrid(range(-5, 5), range(-5, 5))
z = x ** 2 + y ** 2

# Convert this grid to columnar data expected by Altair
chart_data = pd.DataFrame(
    np.random.randn(20, 3),
    columns=['a', 'b', 'c'])

chart = alt.Chart(chart_data).mark_circle().encode(
    x='a', y='b', size='c', color='c', tooltip=['a', 'b', 'c'])


#Bokeh
# prepare some data
x = [1, 2, 3, 4, 5]
y = [4, 5, 5, 7, 2]
p = figure(width=350, height=250)
circle = p.circle(x, y, fill_color="red", size=15)


# Matplotlib
np.random.seed(19680801)
fig, ax = plt.subplots()
for color in ['tab:blue', 'tab:orange', 'tab:green']:
    n = 750
    x, y = np.random.rand(2, n)
    scale = 200.0 * np.random.rand(n)
    ax.scatter(x, y, c=color, s=scale, label=color,
               alpha=0.3, edgecolors='none')
ax.legend()
ax.grid(True)

                      
class State(rx.State):
    pass


def index():
    return rx.vstack(
        rx.altair( 
            fig=chart 
        ),
        rx.bokeh(
            fig=p
        ),
        rx.pyplot(
            fig=fig
        ) 
    )
 
 
app = rx.App()
app.add_page(index)
app.compile()     
```


Alek99:
Resolving some dependency conflicts, in the pyproject.toml, but code is ready for review

Alek99:
Another thing to consider with this pr is I introduced another dependency to make the matplotlib graphs interactive mpld3. 

It is possible to support matplotlib without this but the charts would serialize to images which wouldn't be interactive to the end user. I'm open to going with this approach as well but I found the interactive charts to be a nice touch.

reflex/components/graphing/altair.py
```diff
@@ -0,0 +1,54 @@
...
+class Altair(Box):
+    """Display an Altair chart.
+
+    This component takes an Altair chart as input and renders it within a Box layout component.
+    """
+
+    # The figure to display.
+    fig: Var[Chart]
+
+    # The HTML to render.
```
picklelo, line 25: We shouldn't expose this prop to the user

reflex/components/graphing/altair.py
```diff
@@ -0,0 +1,54 @@
...
+    dangerouslySetInnerHTML: Any
+
+    def _render(self) -> Tag:
+        """Render the Altair chart as HTML and set it as the inner HTML of the Box component.
+
+        Returns:
+            The rendered component.
+        """
+        self.dangerouslySetInnerHTML = {"__html": self.fig}
+        return super()._render().remove_props("fig")
```
picklelo, line 35: Instead, here we can do `super()._render().add_props(dangerously_set_inner_HTML={"__html": self.fig}).remove_props("fig")`

picklelo (commented):
Code works for me!
* Let's not expose the HTML prop for these components, and just set them internally. That way they won't show up in the docs, etc.
* Can we add some basic tests to make sure the serialization is working

Alek99:
This may be a result of flaky ci something with tailwind test failing

reflex/components/__init__.py
```diff
@@ -15,17 +15,6 @@
 from .overlay import *
 from .typography import *
 
-# Add the convenience methods for all the components.
```
picklelo, line 18: I think we can actually bring this back now and remove the manually listing of every component since we have the `.pyi` files. But we can bring that in another PR, can we add this comment back for now?

picklelo (commented):
Can we add some unit tests for the matplotlib?

reflex/components/__init__.py
Alek99: Ticket made will uncomment it https://github.com/reflex-dev/reflex/issues/1865#issue-1910493391

Alek99:
> Hi there,
> 
> Thank you for all your work and contributions to Reflex. I am very excited to have Matplotlib support in a future release.
> 
> One question, ¿Does events like **'button_press_event'** or **'motion_notify_event'** will be supported?

Hello! Are these matplot lib specific things? I am not too familiar with it

LuisOlivaresJ:
HI,

[Here](https://matplotlib.org/stable/users/explain/figure/event_handling.html) you can see a list of available events.

PR #1839, opened 2023-09-19
pyproject.toml: add `packaging` to dep list

masenf:
This is used in prerequisites to compare node and bun versions.

Previously we were picking up this dep via `plotly`, which is no longer installed by default.


picklelo (approved):
nice catch

PR #1842, opened 2023-09-20
add option to disable Tailwind

### LINKED ISSUE #1829: Don't have Tailwind defaulty included in reflex app
Currently Tailwind is added to each fresh initialization of Reflex app. This may interfere with some other css libraries.

The `tailwindcss` module is installed defaulty. Also the is a default style style file in `.web\styles\tailwind.css`.

It would be nice, if it was included only optionally when specified in packages or installed manually.

ElijahAhianyo:
This PR provides an option to disable the out-of-the-box supported Tailwind. To do that, simply set the `tailwind` config to `None` :

`rxconfig.py`
---------------------------------
```python
import reflex as rx

config = ExampleConfig(
    app_name="example",
    tailwind= None
)
```

fixes #1829 

Docs: https://github.com/reflex-dev/reflex-web/pull/205

### DIFF
--- a/reflex/utils/prerequisites.py
+++ b/reflex/utils/prerequisites.py
@@ -434,18 +434,13 @@ def install_frontend_packages(packages: set[str]):
     config = get_config()
     if config.tailwind is not None:
         # install tailwind and tailwind plugins as dev dependencies.
-        plugins = [
-            plugin
-            for plugin in config.tailwind["plugins"]
-            if config.tailwind and "plugins" in config.tailwind
-        ]
         process = processes.new_process(
             [
                 get_install_package_manager(),
                 "add",
                 "-d",
                 constants.TAILWIND_VERSION,
-                *plugins,
+                *((config.tailwind or {}).get("plugins", [])),
             ],
             cwd=constants.WEB_DIR,
             shell=constants.IS_WINDOWS,
--- a/reflex/utils/prerequisites.py
+++ b/reflex/utils/prerequisites.py
@@ -433,22 +433,19 @@ def install_frontend_packages(packages: set[str]):
 
     config = get_config()
     if config.tailwind is not None:
-        # install tailwind
+        # install tailwind and tailwind plugins as dev dependencies.
         process = processes.new_process(
-            [get_install_package_manager(), "add", constants.TAILWIND_VERSION, "-d"],
+            [
+                get_install_package_manager(),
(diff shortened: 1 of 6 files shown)


Lendemor (commented):
Shouldn't `tailwind=None` be enough to disable tailwind rather than add a disable field?

Also I was thinking we should remove the base tailwind from package.json and only install it if tailwind is not None.

Lendemor (dismissed):
Nice!

reflex/utils/prerequisites.py
```diff
@@ -432,13 +432,23 @@ def install_frontend_packages(packages: set[str]):
     processes.show_status("Installing base frontend packages", process)
 
     config = get_config()
-    if config.tailwind is not None and "plugins" in config.tailwind:
+    if config.tailwind is not None:
+        # install tailwind
         process = processes.new_process(
-            [get_install_package_manager(), "add", *config.tailwind["plugins"]],
+            [get_install_package_manager(), "add", constants.TAILWIND_VERSION, "-d"],
```
masenf, line 438:
```suggestion
            [get_install_package_manager(), "add", "-d", constants.TAILWIND_VERSION, *config.tailwind.get("plugins", [])],
```

does this work? seems a little more compact.

masenf (dismissed):

reflex/utils/prerequisites.py
ElijahAhianyo: Yeah this works :)

reflex/utils/prerequisites.py
```diff
@@ -432,13 +432,20 @@ def install_frontend_packages(packages: set[str]):
...
+    if config.tailwind is not None:
+        # install tailwind and tailwind plugins as dev dependencies.
         process = processes.new_process(
-            [get_install_package_manager(), "add", *config.tailwind["plugins"]],
+            [
+                get_install_package_manager(),
+                "add",
+                "-d",
+                constants.TAILWIND_VERSION,
+                *config.tailwind["plugins"],
```
masenf, line 443:
```suggestion
                *config.tailwind.get("plugins", []),
```

in case no plugins are specified

ElijahAhianyo: This will fail if `tailwind` is set to `None`. I made a fix that checks that as well

masenf (approved):

PR #1843, opened 2023-09-20
allow for non-installable imports

Lendemor:
When wrapping some components, we need some more flexibility over which import tag should trigger an auto-install.

Added the `install` tag in `ImportVar`

Also used this opportunity to pack the logic of `NoRenderImportVar` inside `ImportVar` as the `render` tag.

```python
ImportVar(tag="myTag", install=False)
ImportVar(tag=None, render=False) # equivalent to NoRenderImportVar used previously
```

```python
def _get_imports(self):
    return {
        "my_lib": ImportVar(tag="myTag", install=False),
        "no_render_lib": ImportVar(tag=None, render=False),
    }
```


reflex/vars.py
```diff
@@ -1347,12 +1353,14 @@ def __hash__(self) -> int:
         Returns:
             The hash of the var.
         """
-        return hash((self.tag, self.is_default, self.alias))
+        return hash((self.tag, self.is_default, self.alias, self.install, self.render))
 
 
 class NoRenderImportVar(ImportVar):
```
masenf, line 1359: Are we deprecating this? Or leaving it as-is?

reflex/app.py
```diff
@@ -519,13 +520,14 @@ def get_frontend_packages(self, imports: Dict[str, str]):
         """
         page_imports = {
             i
-            for i in imports
+            for i, tags in imports.items()
             if i not in compiler.DEFAULT_IMPORTS.keys()
             and i != "focus-visible/dist/focus-visible"
             and "next" not in i
```
masenf, line 526: maybe out of scope for this PR, but I think these lines could be cleaned up by marking `focus-visible/dist/focus-visible` over in DEFAULT_IMPORTS with `install=False`, and changing `"next" not in i` to `"not i.startswith("next/")`.

i haven't tried this, but ultimately i _should_ be ignoring any dep that appears in the templated `package.json`, as we don't want to overwrite those pins.

masenf (approved):
overall looks good; i like the additional fields on ImportVar, I think that keeps the API more consistent.

reflex/vars.py
masenf: oh, maybe the line moved. I was referring to `NoRenderImportVar`. Are we keeping _that_?

reflex/app.py
Lendemor: Yeah it's probably a good idea to clean up this part and avoid "exception" as much as possible in how we select "to-be-installed" packages

reflex/vars.py
Lendemor: Ah yeah, I just kept it in case anyone was using it, but ultimately it will be deprecated/removed yes.
(already removed all reference to it in the codebase)

reflex/compiler/utils.py
```diff
@@ -91,8 +91,9 @@ def compile_imports(imports: imports.ImportDict) -> List[dict]:
     import_dicts = []
     for lib, fields in imports.items():
         default, rest = compile_import_statement(fields)
-        # prevent lib from being rendered on the page if all imports are NoRenderImportVar
-        if all({isinstance(f, NoRenderImportVar) for f in fields}):  # type: ignore
+
+        # prevent lib from being rendered on the page if all imports are non rendered kind
+        if all({not f.render for f in fields}):  # type: ignore
```
masenf, line 96: probably splitting hairs at this point, but i think `not any` would read easier

```suggestion
        if not any({f.render for f in fields}):  # type: ignore
```

masenf (approved):

reflex/compiler/utils.py
Lendemor: I'll make the change in a future PR since there is still some work to do around the imports :+1:

PR #1846, opened 2023-09-20
exec: print the URL, not the address the server binds to

masenf:
Fix: REF-641

### DIFF
--- a/reflex/utils/exec.py
+++ b/reflex/utils/exec.py
@@ -85,10 +85,10 @@ def run_process_and_launch_url(run_command: list[str]):
             )
         if process.stdout:
             for line in processes.stream_logs("Starting frontend", process):
-                match = re.search("ready started server on ([0-9.:]+)", line)
+                match = re.search("ready started server on ([0-9.:]+), url: (.*)", line)
                 if match:
                     if first_run:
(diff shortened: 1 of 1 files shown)


tgberkeley:
Looks great. Works for me again.

tgberkeley (approved):

nevdelap:
Doesn't that make it print that is running on localhost when it is bound to 0.0.0.0 again?

masenf:
@nevdelap Neither is technically "correct". But formatting the bind address, `0.0.0.0`, as a URL and displaying it results in a link that doesn't work when clicked.

With this change, the user can sometimes click on the `http://localhost:8000/whatever` link and have it actually take them to their app (if they are opening a browser on the same machine as the server).

Not sure what the right answer here is and happy to hear suggestions... but we had complaints from users that the link displayed did not work, so I would consider that a bug.

Users can pass `--loglevel debug` if they want to see the bind address or it could be printed separately (maybe just print that entire line from the next server instead of regexing parts out?), but it can't replace the (sometimes) working URL to the app.

nevdelap:
I wonder if it should just display the whole line rather than picking out one or the other piece.

PR #1847, opened 2023-09-20
[REF-201] Separate on_load handler from initial hydration

masenf:
Allow page navigation to trigger on_load events without having to rehydrate the entire state (for no reason)

https://linear.app/reflex-dev/issue/REF-201/move-on-load-handling-out-of-hydrate-middleware-and-into-page-initial


masenf:
I'm actually thinking a much more effective way to implement this is to rename `State` to `BaseState` and then basically bring back `DefaultState` as `rx.State`, which includes the `is_hydrated` var and `on_load_internal` event handlers. Then we can put whatever other default behavior we need in the new State class, which would always be the root of the state tree.

This solves a few other problems, allowing users to finally have multiple classes inherit from `rx.State`

masenf:
Moving back to draft for rework.

reflex/state.py
```diff
@@ -1218,6 +1220,52 @@ def dict(self, include_computed: bool = True, **kwargs) -> dict[str, Any]:
         variables = {**base_vars, **computed_vars, **substate_vars}
         return {k: variables[k] for k in sorted(variables)}
 
+    @classmethod
+    def _builtin_event_handlers(cls) -> dict[str, Callable]:
```
masenf, line 1224: i think this can go away now and we can just define `on_load_internal` directly in `rx.State`, now that it's a subclass of `rx.BaseState`

masenf (commented):
state.js:568 `const change_complete = () => addEvents(initial_events())`

needs to change to `const change_complete = () => addEvents([Event('{{state_name}}.{{const.on_load_internal}}'])`

otherwise we're still hydrating on every page load

reflex/.templates/jinja/web/utils/context.js.jinja2
```diff
@@ -23,8 +23,11 @@ export const clientStorage = {}
 {% endif %}
 
 {% if state_name %}
+export const onLoadInternalEvent = () => [Event('{{state_name}}.{{const.on_load_internal}}')]
```
masenf, line 26: also need `export const onLoadInternalEvent = () => []` in the `else` clause for stateless apps

masenf (commented):
Technically this is _my_ PR, so I cannot approve it... but LGTM! 🪨 🚀 ✨

picklelo (approved):

PR #1849, opened 2023-09-21
use jinja2 to render package.json

Lendemor:
By using jinja2 to render the package.json, it allow us to have all the listed packages from it accessible in `constants.py`.

It should make it easier when we need to manipulate that list of packages, or to run check against them.


masenf (approved):
This is nice; great addition

PR #1850, opened 2023-09-21
[REF-650] Add HtmlDangerous component and use it to render Pyplot component.

masenf:
Based on code from https://github.com/christo-pr/dangerously-set-html-content with modifications


Alek99 (approved):
Tested and works well for me will rebase my other graphing pr on this! 📊

PR #1852, opened 2023-09-21
Improvements to custom styles in rx.markdown

picklelo:
Added option to set custom styles using a component map:

```python
component_map = {
    "h1": lambda text: rx.box(rx.heading(text), rx.divider())
    "p": rx.text
}
```

Then pass it in

```python
source = """
# Hello!

Welcome to my markdown file.
"""

rx.markdown(
    source,
    component_map=component_map
)
```


masenf (dismissed):
would be cool to add e2e tests for the new var operations in `integration/test_var_operations.py`.

is this enough to fix #1283 too?

reflex/components/typography/markdown.py
```diff
@@ -38,43 +52,18 @@ class Markdown(Component):
 
     is_default = True
 
-    # Custom defined styles for the markdown elements.
-    custom_styles: Dict[str, Style] = {
```
masenf, line 42: we probably should deprecate this first, unless you think no one is using it or it's already broken

picklelo: Yes agreed, I will add this for the following release I think I can make the old version convert cleanly to the new one.

reflex/testing.py
```diff
@@ -398,7 +398,7 @@ def frontend(self, driver_clz: Optional[Type["WebDriver"]] = None) -> "WebDriver
             )
         if self.frontend_url is None:
             raise RuntimeError("Frontend is not running.")
-        driver = driver_clz() if driver_clz is not None else webdriver.Chrome()
+        driver = driver_clz() if driver_clz is not None else webdriver.Firefox()
```
masenf, line 401: not sure if we want to commit this...

masenf (approved):

reflex/testing.py
picklelo: Fixed this, just had it locally for testing

masenf (approved):

PR #1853, opened 2023-09-22
StateProxy rebinds functools.partial and methods that are bound to the proxied State

masenf:
Ensure that calling other event handlers directly or regular methods defined on the state cannot bypass the immutability protections of the StateProxy.

Add test cases to prevent regressions in this area.


picklelo (approved):
nice catch

Issue #1791, opened 2023-09-09
multi-select does not work


signup2k:
Multi-select does not work, neither with the example listed on  document's `selector` section, nor with the example code I copy and run on local machine.   

update::

I find the multselect.py and select.py both in the reflex component. 
The example given in the document, use `rx.select` only, even with multiselect example.

When I check the select.py file, the props in Select Class does not have `is_multi`, `close_menu_on_select`, etc. Which can be found in multiselect.py file.

But when I use rx.multi_select(), the Error message will be 

11, in <lambda>
    lambda e0: [Var.create_safe(f"{e0}.map(e => e.value)", is_local=True)]
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/joeyjiang/opt/anaconda3/envs/py311/lib/python3.11/site-packages/reflex/vars.py", line 238, in __bool__
    raise TypeError(
TypeError: Cannot convert Var 'true' to bool for use with `if`, `and`, `or`, and `not`. Instead use `rx.cond` and bitwise operators `&` (and), `|` (or), `~` (invert).

And I guess, the reflex mixed the select.py and multiselect.py when create Select instance (because in both select.py and multiselect.py file, the class name are the same `Select`).

I dont know how to fix this.

signup2k:
I find the multselect.py and select.py both in the reflex component. 
The example given in the document, use `rx.select` only, even with multiselect example.

When I check the select.py file, the props in Select Class does not have `is_multi`, `close_menu_on_select`, etc. Which can be found in multiselect.py file.

But when I use rx.multi_select(), the Error message will be 

11, in <lambda>
    lambda e0: [Var.create_safe(f"{e0}.map(e => e.value)", is_local=True)]
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/joeyjiang/opt/anaconda3/envs/py311/lib/python3.11/site-packages/reflex/vars.py", line 238, in __bool__
    raise TypeError(
TypeError: Cannot convert Var 'true' to bool for use with `if`, `and`, `or`, and `not`. Instead use `rx.cond` and bitwise operators `&` (and), `|` (or), `~` (invert).

[...]

PR #1861, opened 2023-09-24
multi_select somewhat usable

masenf:
Try to address #1791, but this component is _cursed_.

Because of how the component expects to get certain props (`options` is a list of dict rather than child components, so `rx.foreach` cannot be used). It's not a great candidate for wrapping in reflex because Var operations have to compile down to frontend javascript and the transformations necessary to make the component ergonomic to use are not currently implemented.

And it looks like the component was contributed without tests or documentation, which makes it even harder to use, understand and catch behavioral regressions.

I have some small tweaks in this PR that make it actually possible to work with it, although it's not the nicest component and I'm not sure if/how it works with `rx.form`... Basically it needs some serious love and attention from someone who deeply understands reflex internals (and might even require some refactoring of `rx.foreach` to emit `Var` in addition to `Component`.)

Here is some sample code that can run with this PR

```sh
pip install git+https://github.com/reflex-dev/reflex@masenf/multi-select-somewhat-usable
```

```python
import reflex as rx


class State(rx.State):
    allowed_options: list[str] = ["Foo", "bar", "baz", "qux", "quux", "corge", "grault"]
    
    @rx.var
    def multi_select_options(self) -> list[dict[str, str]]:
        return [
            {"label": o, "value": o}
            for o in self.allowed_options
        ]

    def handle_multi_select(self, values: list[str]):
        print(f"Selected: {values}")


def index() -> rx.Component:
    return rx.fragment(
        rx.vstack(
            rx.heading("Welcome to Reflex!", font_size="2em"),
            rx.multi_select(
                #options=[
                #    rx.multi_select_option(label=o, value=o)
                #    for o in ["Foo", "bar", "baz", "qux", "quux", "corge", "grault"]
                #],
                options=State.multi_select_options,
                is_multi=True,
                on_change=State.handle_multi_select,
            ),
        ),
    )


app = rx.App()
app.add_page(index)
app.compile()
```


masenf:
also needs

```diff
diff --git a/reflex/components/forms/multiselect.py b/reflex/components/forms/multiselect.py
index 9625e15..430418b 100644
--- a/reflex/components/forms/multiselect.py
+++ b/reflex/components/forms/multiselect.py
@@ -48,6 +48,7 @@ class Select(Component):
 
     library = "chakra-react-select"
     tag = "Select"
+    alias = "MultiSelect"
 
     # Focus the control when it is mounted
     auto_focus: Var[bool]
```

picklelo (approved):
works for me

cyrus0880:
i use same code 
but on ui click the select box is not any response
reflex
not show options

topkek-issou:
Hi @masenf 
Thanks for the fix, really useful!

Is there a way to populate it on the UI with a few selected options from the backend? Like with other components, we can use the `value` attribute to fill it with state vars.

So for example I have a select with Monday, Tuesday, ..., Sunday options.
And I'd like to use a state var to prefill the component with, say, State.defaults, which would be Tuesday and Thursday.
image

Thx!

PR #1878, opened 2023-09-27
Core Graphing -> Recharts

Alek99:
# Rework of Reflex's Core graphing lib.

Test with this PR on reflex web https://github.com/reflex-dev/reflex-web/pull/234


Alek99:
Looking into why the unit tests don't pass they work for me locally

reflex/components/graphing/recharts/general.py
```diff
@@ -0,0 +1,161 @@
...
+    # The width of chart container, usually calculated internally.
+    chart_width: Var[int]
+
+    # The height of chart container, usually calculated internally.
+    chart_height: Var[int]
+
+    # The margin of chart container, usually calculated internally.
+    margin: Var[Dict[str, Any]]
+
+    def get_event_triggers(self) -> dict[str, Union[Var, Any]]:
```
picklelo, line 83: Need to use `Dict` for Python 3.8

picklelo (commented):
Looks good once the build is passing

reflex/components/graphing/recharts/charts.py
```diff
@@ -0,0 +1,457 @@
...
+from reflex.components.component import Component
+from reflex.components.graphing.recharts.general import ResponsiveContainer
+from reflex.constants import EventTriggers
+from reflex.vars import Var
+
+from .recharts import RechartsCharts
+
+
+class ChartBase(RechartsCharts):
+    """A component that wraps a victory lib."""
```
masenf, line 14:
```suggestion
    """Base component for a Recharts chart."""
```

reflex/components/graphing/recharts/charts.py
```diff
@@ -0,0 +1,457 @@
...
+    ]
+
+
+class BarChart(ChartBase):
+    """A Bar chart component in Recharts."""
+
+    tag = "BarChart"
+
+    # The gap between two bar categories, which can be a percent value or a fixed value. Percentage | Number
+    bar_category_gap: Var[str]
```
masenf, line 103: can these be `Union[str, int]`? or is that not workable?

what is the unit for number values? pixels?

reflex/components/graphing/recharts/charts.py
```diff
@@ -0,0 +1,457 @@
...
+    def get_event_triggers(self) -> dict[str, Union[Var, Any]]:
+        """Get the event triggers that pass the component's value to the handler.
+
+        Returns:
+            A dict mapping the event trigger to the var that is passed to the handler.
+        """
+        return {
+            EventTriggers.ON_CLICK: lambda: [],
+            EventTriggers.ON_MOUSE_UP: lambda: [],
+            EventTriggers.ON_MOUSE_DOWN: lambda: [],
```
masenf, line 352: not sure how these are different than `on_click` if the event handler doesn't take any args, like _where_ or _what_ the user mouse_down'd on

reflex/components/graphing/recharts/recharts.py
```diff
@@ -0,0 +1,21 @@
...
+
+from reflex.components.component import Component, NoSSRComponent
+
+
+class Recharts(Component):
+    """A component that wraps a victory lib."""
+
+    library = "recharts"
+
+    lib_dependencies: List[str] = ["recharts@^2.8.0"]
```
masenf, line 13:
```suggestion
    library = "recharts@^2.8.0"
```

does this work? if so i think that's preferred

masenf (dismissed):
* need to generate pyi files at some point (but that would make review more challenging)
* is there anything victory had that recharts does not have?
* are there any important breaking changes that victory users need to apply when upgrading?

reflex/components/graphing/recharts/charts.py
Alek99: Good idea I'll go through one more time and fix the spots that can also be percentages

reflex/components/graphing/recharts/charts.py
Alek99: I agree these function the same I initially added because they were used in recharges, but removing would be the best choice for simplicity.

Alek99:
> * need to generate pyi files at some point (but that would make review more challenging)
> * is there anything victory had that recharts does not have?
> * are there any important breaking changes that victory users need to apply when upgrading?

Ok I'll need to answer these in the docs and post about recharts.

from what I can tell recharts has all the functionality Victory has minus a few charts Candlestick, Voronoi

Based on the current state of the pr the user will have to completely change to Recharts we would no longer support victory. Most of the changes would be not using the rx.data functionality, as I did not add this in recharts

Alek99:
Ok I'll make these update this was not necessarily the final docs for these just a first pass, I din't want the pr to get too big

Alek99:
Integration test failing because we will need to merge in my pr first on reflex-web

masenf (approved):

picklelo (approved):

PR #1891, opened 2023-09-29
Expose DOM event actions on EventHandler, EventSpec, and EventChain (stopPropagation)

### LINKED ISSUE #1621: [REF-675] Stop click propagation
Is there a way of stopping the propagation of "onClick" events?

I have a button that sits on top of a clickable hstack, when I click  the button, the hstack also gets clicked.

My current workaround is to have a variable to prevent the double click but this obviously makes the code messier than it has to be.

Any way of preventing the double click?

[https://github.com/reflex-dev/reflex/pull/1891](https://github.com/reflex-dev/reflex/pull/1891)

<sub>From [SyncLinear.com](https://synclinear.com) | [REF-675](https://linear.app/reflex-dev/issue/REF-675/stop-click-propagation)</sub>

masenf:
This change adds properties to `EventHandler`, `EventSpec` and `EventChain`:

* `stop_propagation`: instruct the frontend to call `_e.stopPropagation` when processing the DOM event
* `prevent_default`: instruct the frontend to call `_e.preventDefault` when processing the DOM event

These are properties that return the object they are accessed through and can be chained.

```python
rx.box(
    rx.link(
        "Click only me",
        href="/will-never-nav-here",
        on_click=State.do_thing.stop_propagation.prevent_default,
    ),
    on_click=State.will_not_receive_link_click_events,
)
```

In a list-based event chain (deprecated), any EventHandler/EventSpec that defines one of these actions, will have that apply to the entire chain, since that's the only way this can work.

-----------------

This allows nested trees of elements to have more control over which layer receives events. It also allows links and other elements with default actions have event handlers that explicitly prevent the default (like not navigating when a link is clicked).

There are also two new top-level attributes: `rx.stop_propagation` and `rx.prevent_default` which can be used for an event trigger to block propagation or defaults for specific events without actually triggering their own event. This might be useful on a form or editable that is sitting on top of a clickable element to prevent clicks from propagating to the parent, but otherwise no action should be taken for the app (because it's implicit via `on_submit` or in the component internals).

Fix #1621 

This change also removes the special case `preventDefault` for submit events from `state.js` and moves that logic into the definition of the `rx.Form` component itself, which uses this new mechanism.

-----------------

I have included an AppHarness integration test for this behavior as well as unit tests. Here is some additional sample code that I was testing with:

```python
import reflex as rx


class State(rx.State):
    editable_val: str = "edit me"
    def on_click(self, val: str):
        print(f"clicked {val}")

    def on_click_outer(self):
        print("clicked outer")

    def on_submit(self, ev):
        print(f"editable submitted {ev}")

    def on_submit_form(self, form_data):
        print(f"form submitted {form_data}")


def index() -> rx.Component:
    return rx.vstack(
        rx.button("Click me", on_click=State.on_click("inner").stop_propagation),
        rx.editable(
            rx.editable_preview(),
            rx.editable_input(),
            value=State.editable_val,
            on_change=State.set_editable_val,
            on_submit=State.on_submit,
            on_click=rx.stop_propagation,
        ),
        rx.link("Click the Link", href="/foobar", on_click=State.on_click("link").stop_propagation.prevent_default),
        rx.form(
            rx.input(type="text", id="foo"),
            rx.button("Submit", type_="submit"),
            on_submit=State.on_submit_form,
            on_click=rx.stop_propagation,
        ),
        on_click=State.on_click_outer,
        padding_top="10%",
    )


app = rx.App()
app.add_page(index)
app.compile()
```


Lendemor (approved):
Tested it, seems good for me, nice work :+1:

PR #1894, opened 2023-09-30
Added conditional rendering func to compile workflow

### LINKED ISSUE #1830: Data Table behavior is inconsistent when empty
**Describe the bug**
The data table element will sometimes return a text saying "No matching records found" and other times just empty rows when you have an empty list, or a state with the value being an empty list.

**To Reproduce**
Steps to reproduce the behavior: create a data table with data=[] and reload the page a few times.
 - Code/Link to Repo:

`rx.data_table(data=[], columns=["A", "B"])`

**Expected behavior**
Everytime the data_table has no data it should return the text "No matching records found", no matter how many times you reload

**Screenshots**
image
image

 - Python Version: 3.8
 - Reflex Version: 0.2.7
 - OS: Docker container (Linux)
 - Browser (Optional): Chrome
 
**Additional context**

RyleyGG:
### All Submissions:

- [ x ] Have you followed the guidelines stated in [CONTRIBUTING.md](https://github.com/reflex-dev/reflex/blob/main/CONTRIBUTING.md) file?
- [ x ] Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/reflex-dev/reflex/pulls ) for the desired changed?

### Type of change

- [ x ] Bug fix (non-breaking change which fixes an issue)

### New Feature Submission:

- [ x ] Does your submission pass the tests? 
- [ x ] Have you linted your code locally prior to submission?

### Change Description
Per #1830, there is an issue with empty data table behavior wherein the table is sometimes rendered without the expected "No matching records found" message. From my research, this appears to be due to multiple re-renders of the component. As of 4f6b3c0, these re-renders can clearly be seen on page load, even if the no data message is ultimately displayed. These re-renders are due to the contexts associated with the component state that are included in the compiled pages. Namely:
```
const state = useContext(StateContext)
/* some other code... */
const [addEvents, connectError] = useContext(EventLoopContext)
```

Specifically, the re-renders are due to the _is\_hydrated_ property in the state being updated from _false_ to _true_. As a first-time contributor to this repo, I do not feel comfortable manipulating the state without understanding fully how its used. Therefore, this PR includes a new method used in the compile workflow to solve #1830. The method uses regex to check the compilation template after its first generated and adds conditional rendering logic to elements as required. Currently, the conditional rendering is only added to DataTableGrid elements if they have no data.

closes #1830

### DIFF
--- a/reflex/compiler/compiler.py
+++ b/reflex/compiler/compiler.py
@@ -133,7 +133,9 @@ def _compile_conditional_rendering(template_obj: str) -> str:
     Returns:
         The template after relevant modifications
     """
-    pattern = re.compile(r'(<DataTableGrid[^>]*\/>)')
+    pattern = re.compile(
+        r'(<DataTableGrid[^>]*\bdata\s*=\s*\{\s*\[\s*\]\s*\}[^>]*\/>)'
+    )
     replacement = r'{ state.is_hydrated ? \1 : null }'
 
     template_obj = re.sub(pattern, replacement, template_obj)
--- a/reflex/compiler/compiler.py
+++ b/reflex/compiler/compiler.py
@@ -123,7 +123,16 @@ def _compile_page(
     template_obj = _compile_conditional_rendering(template_obj)
     return template_obj
 
-def _compile_conditional_rendering(template_obj: str):
+def _compile_conditional_rendering(template_obj: str) -> str:
+    """Update generated template to add required conditional rendering
+       for elements in the template
+
+    Args:
+        template_obj: Template to be modified
+
+    Returns:
+        The template after relevant modifications
+    """
     pattern = re.compile(r'(<DataTableGrid[^>]*\/>)')
     replacement = r'{ state.is_hydrated ? \1 : null }'
 
--- a/reflex/compiler/compiler.py
+++ b/reflex/compiler/compiler.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import os
+import re
 from pathlib import Path
 from typing import Type
 
@@ -111,7 +112,7 @@ def _compile_page(
     imports = utils.compile_imports(imports)
 
     # Compile the code to render the component.
-    return templates.PAGE.render(
+    template_obj = templates.PAGE.render(
         imports=imports,
         dynamic_imports=component.get_dynamic_imports(),
         custom_codes=component.get_custom_code(),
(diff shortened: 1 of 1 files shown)


reflex/compiler/compiler.py
```diff
@@ -111,14 +112,34 @@ def _compile_page(
...
+    Args:
+        template_obj: Template to be modified
+
+    Returns:
+        The template after relevant modifications
+    """
+    pattern = re.compile(
+        r'(<DataTableGrid[^>]*\bdata\s*=\s*\{\s*\[\s*\]\s*\}[^>]*\/>)'
+    )
+    replacement = r'{ state.is_hydrated ? \1 : null }'
```
masenf, line 139: the state var in the page is not always called `state`, the name is based on the class name that inherits from `rx.State`

masenf (changes requested):
Thanks for jumping in with a contribution 💯.

I think the idea is sound, but the approach here is not sustainable.

In my experience the proper place to implement this kind of logic is inside the component itself:

https://github.com/reflex-dev/reflex/blob/4f6b3c049be876def1277a72b5880600d5b572d4/reflex/components/datadisplay/datatable.py#L99-L103

Instead of returning the component here, the code should return a cond if the data source comes from a var (if it does NOT come from a var, then there's no reason to wait for state hydration)

```python
data_table_component = super().create(
        *children,
        **props,
    )

state = None
if isinstance(data, Var):
    state = data.state
elif isinstance(columns, Var):
    state = columns.state
if state is not None:
    return Cond(
        getattr(state, constants.CompileVars.IS_HYDRATED),
        data_table_component,
    )

# table does not depend on state, so no cond needed
return data_table_component
```

reflex/compiler/compiler.py
RyleyGG: I agree that it would be much better for this logic to be encapsulated. Reviewing your code and I'm having difficulty understanding where or how the incoming data prop is set to type Var. On my end I'm only seeing it as type list when setting up the test application to match the conditions from #1830

masenf: `data` would be a `Var` if it was coming from the state

```python
class State(rx.State):
    data: list[list[int]] = []

...

rx.data_table(data=State.data, columns=["A", "B"])
```

In this case, the `data` prop would be a `Var`. You can see this if you put a breakpoint in the `DataTable.create` method.

That said, I haven't attempted to repro this locally, so it's possible I might be misunderstanding the root cause of #1830

masenf:
Closing stale PR without updates.

Issue #1865, opened 2023-09-25
Generate static convenience methods for all components


Alek99:
We're currently generating .pyi files for type hinting in IDEs, and it would also be nice to generate the `__init__.py` files inside `reflex/components` that expose the `ComponentX.create` functions as `component_x`.

We have this bit of code commented out:

https://github.com/reflex-dev/reflex/blob/2f284a08b6afabd1d3141830d846b7c934fccf1f/reflex/components/__init__.py#L18-L25

But this approach actually breaks our IDE integration, because we need to statically link the `snake_case` component name to the `Component.create` function for the IDE to find the pyi type hints.

To solve this issue the `pyi_generator` (or a new generator) needs to exist that scans through component modules, and creates a hierarchy of `__init__.py` files similar to the ones we already have in https://github.com/reflex-dev/reflex/blob/main/reflex/components/__init__.py

adkbbx:
Hey, I am intersted in contributing to this issue do let me know how I can help.

picklelo:
@adkbbx Pretty much we want to remove all the manual redefinitions within that file. For example the lines that say `box = Box.create`. And replace it with the commented code. We had to do the manual work because the definitions weren't getting picked up by IDEs before, but now we generate `.pyi` files that define the signatures, so we don't need to do this anymore.

y9rabbito:
@Alek99 @adkbbx and others you could review that and see if it's alright or whether any changes are needed.

masenf:
I think the pyi_generator also needs be updated to generate the stub functions, otherwise the dynamically created names aren't visible in the IDE.

Billa05:
is this issue resolved?

danik292:
@measenf resolved

masenf:
This issue is still unresolved because we have to manually create entries in `reflex/components/__init__.py`. I updated the description to make it more clear what is needed here. A simple dynamic loop is not sufficient, we need code generated in real python files in order to retain good IDE integration and type checking.

danik292:
So how to do it ? @masenf

picklelo:
This is not needed anymore.

PR #1895, opened 2023-10-01
Added dynamic convenience methods and remove manual redifinitions in __init.py__

y9rabbito:
This PR is according to the issue #1865, autogenerate convenience methods for all the components.


masenf (commented):
I pushed a change to get the tests passing, but I'm not sure this is actually what we want, as it seems to break auto-complete in the IDE. Typing `rx.button(` for example no longer shows the correct function/type signatures with this change.

y9rabbito (commented):
Imports the to_snake_case(name), defined in reflex_utils_format, and also delete import of utils

Billa05:
> I pushed a change to get the tests passing, but I'm not sure this is actually what we want, as it seems to break auto-complete in the IDE. Typing `rx.button(` for example no longer shows the correct function/type signatures with this change.

 maybe it is better to keep manual redefinitions to maintain a good developer experience

masenf:
We'll keep the static definitions for now for better type checking and IDE integration. In the future we will look at generating the component `__init__.py` helpers/wrappers similarly to the `.pyi` files that are already generated.

PR #1898, opened 2023-10-01
state: subclass of MutableState must return _mark_dirty return value

masenf:
Some of the methods actually return useful values, like `pop`, so the return value must be passed through after marking dirty.

Previous fix https://github.com/reflex-dev/reflex/pull/1876 corrected this issue for normal event handlers using MutableProxy. This fix is needed for background tasks that use ImmutableMutableProxy.

### DIFF
--- a/tests/test_state.py
+++ b/tests/test_state.py
@@ -1624,7 +1624,7 @@ class BackgroundTaskState(State):
     """A state with a background task."""
 
     order: List[str] = []
-    dict_list: Dict[str, List[int]] = {"foo": []}
+    dict_list: Dict[str, List[int]] = {"foo": [1, 2, 3]}
 
     @rx.background
     async def background_task(self):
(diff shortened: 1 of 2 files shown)


picklelo (approved):

tests/test_state.py
```diff
@@ -1656,6 +1656,9 @@ async def background_task(self):
                 pass  # update proxy instance
 
         async with self:
+            # Methods on ImmutableMutableProxy should return their wrapped return value.
+            assert self.dict_list.pop("foo") is not None
```
Lendemor, line 1660: Should we check the actual value of `foo` rather than `not None` ?

Lendemor (approved):

tests/test_state.py
masenf: might not hurt. i'll add it to a subsequent PR

PR #1905, opened 2023-10-02
fix menu items= API

### LINKED ISSUE #1903: New rx.menu() API does not work correctly
**Describe the bug**
When using `rx.menu()`, I have to provide at least the arguments `button` and `items`, which are both type-hinted as `Unknown`. When I provide an avatar as the button and a list of strings as the items, the resulting menu only contains one item (named after all the strings combined) which is also unusable. 

**To Reproduce**
 ```python
def avatar_menu():
    return rx.menu(button=rx.avatar(name="Max Hustemann"), items=["A", "B", "C"])
```

**Expected behavior**
A menu that contains the selectable items A, B and C.

 - Python Version: 3.11.5
 - Reflex Version: 0.2.8
 - OS: Fedora 
 - Browser (Optional): Brave
 
**Additional context**
I suspect this happens because no `MenuItem` is ever created for each item and a `MenuList` is instead used directly. Also, I'd suggest to type hint the items as a `List`.

### PR DESCRIPTION
**Pull request by Lendemor**
Fix #1903

### DIFF
--- a/reflex/components/overlay/menu.py
+++ b/reflex/components/overlay/menu.py
@@ -1,7 +1,7 @@
 """Menu components."""
 from __future__ import annotations
 
-from typing import Any, List, Union
+from typing import Any, List, Optional, Union
 
 from reflex.components.component import Component
 from reflex.components.libs.chakra import ChakraComponent
@@ -74,7 +74,13 @@ class Menu(ChakraComponent):
         }
 
     @classmethod
-    def create(cls, *children, button=None, items=None, **props) -> Component:
+    def create(
+        cls,
+        *children,
+        button: Optional[Component] = None,
+        items: Optional[List] = None,
+        **props,
+    ) -> Component:
         """Create a menu component.
 
         Args:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: reflex/components/overlay/menu.py [hunk 1]
```diff
@@ -74,7 +74,13 @@ def get_event_triggers(self) -> dict[str, Union[Var, Any]]:
         }
 
     @classmethod
-    def create(cls, *children, button=None, items=None, **props) -> Component:
+    def create(
+        cls,
+        *children,
+        button: Optional[Component] = None,
+        items: Optional[List] = None,
```
masenf, line 81: is it possible for `items` to be a `Var`? doesn't look like that case is handled

Lendemor: It wasn't supported initially, should I add this in this PR, or I can add it in a separate one.

masenf: i'm fine either way, i just anticipate that coming as the next question 😂

picklelo: We can make a new PR for this

picklelo (approved):

PR #1619, opened 2023-08-18
Added fix for serializing PIL images

wassafshahzad:
### All Submissions:

- [x] Have you followed the guidelines stated in [CONTRIBUTING.md](https://github.com/pynecone-io/pynecone/blob/main/CONTRIBUTING.md) file?
- [x] Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/pynecone-io/pynecone/pulls ) for the desired changed?

### Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

#### Description
Updates states format_state.py to handle list, set and tuple

#### Linked Issue
Closes #1578
Closes #1522

PR #1915, opened 2023-10-04
fix rx.image src not working with state

### LINKED ISSUE #1914: rx.image src can not load from State variable
## Description
When setting a path to an asset or an image url in the state rx.image is not able to load this.

```py
from rxconfig import config

from PIL import Image
import requests

import reflex as rx



class State(rx.State):
    url = "https://picsum.photos/id/1/200/300"
    image_0 = Image.open(requests.get(url, stream=True).raw)
    image_1 = "/placeholder.jpeg"


def index() -> rx.Component:
    return rx.vstack(
            rx.image(src=State.image_0),
            rx.image(src=State.image_1),
            rx.image(src=State.url),
            )



app = rx.App()
app.add_page(index)
app.compile()
```

**Expected behavior**
loads the image specified in the states attributes

**Screenshots**

 - Python Version: 3.10.13
 - Reflex Version: 0.2.8
 - OS: ubuntu 23 and macOS Ventura 13.4.1
 - Browser (Optional): Firefox, Chrome, Chromium, Arc browser(also Chromium)

Lendemor:
Serialization update may have broken the `src` parameter of `rx.image` working with State var. (haven't checked if it worked before)

Now work with State var (PIL, absolute link or relative link) or raw values

Closes #1914

### DIFF
--- a/reflex/components/media/image.py
+++ b/reflex/components/media/image.py
@@ -76,7 +76,7 @@ class Image(ChakraComponent):
             The Image component.
         """
         src = props.get("src", None)
-        if src is not None and isinstance(src, str):
+        if src is not None and not isinstance(src, (Var)):
             props["src"] = Var.create(value=src, is_string=True)
         return super().create(*children, **props)
 
--- a/reflex/components/media/image.py
+++ b/reflex/components/media/image.py
@@ -7,7 +7,6 @@ from typing import Any, Optional, Union
 
 from reflex.components.component import Component
 from reflex.components.libs.chakra import ChakraComponent
-from reflex.components.tags import Tag
 from reflex.utils.serializers import serializer
 from reflex.vars import Var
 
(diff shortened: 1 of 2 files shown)


masenf:
I had these examples working a bit ago, do we know what regressed this functionality? https://github.com/masenf/example-pangea

https://github.com/reflex-dev/reflex/pull/1619#issuecomment-1683262379

Lendemor:
> I had these examples working a bit ago, do we know what regressed this functionality? https://github.com/masenf/example-pangea
> 
> [#1619 (comment)](https://github.com/reflex-dev/reflex/pull/1619#issuecomment-1683262379)

I'm not sure, I didn't dig to find the cause of the regression. 
I'm assuming some changes during the serialization changes broke something, need further investigation if we want the exact cause.

masenf:
It broke in 0.2.8, was working in 0.2.7; Mainly i just want to know if we have a class of bugs that was introduced in that release or if this is a one off problem.

masenf:
Here is the change that introduced this issue: https://github.com/reflex-dev/reflex/commit/1938a6cc588423beef4c91789240ee0d99166cd5#diff-4652c902f558b2570fb46d70122cd9bcf9045c7235cebb2c49bd10a8054ace1fL63-R65

Looks like our testing is better now, but we're still missing the case where the src comes from a state Var

masenf (approved):

PR #1917, opened 2023-10-04
Deprecate set_cookie and set_local_storage

ElijahAhianyo:


masenf (commented):
can we dep `get_local_storage` as well

masenf (approved):

PR #1921, opened 2023-10-05
Props as Literals

### LINKED ISSUE #1864: Allow Props to be Literals
I came across this when implementing my recharts switch. 

https://recharts.org/en-US/api/Area

There was a prop `type` that needs to accept a certain sect of values `'basis' | 'basisClosed' | 'basisOpen' | 'bumpX' | 'bumpY' | 'bump' | 'linear' | 'linearClosed' | 'natural' | 'monotoneX' | 'monotoneY' | 'monotone' | 'step' | 'stepBefore' | 'stepAfter' `

I initially that an enum would work but then the user would have to remember to use etc...

Allowing  something like 

```python
class Area(Cartesian):
    """An Area component in Recharts."""

    tag = "Area"

    stroke: Var[str]

    fill: Var[str]

    type_: Var[Literal['basis', 'basisClosed', 'basisOpen','bumpX','bumpY','bump','linear','linearClosed','natural','monotoneX','monotoneY','monotone','step','stepBefore','stepAfter']]
```

This seems like a cleaner api than using an enum here example below

```python 
def area_example():
    return rx.area_chart(
            rx.area(
                data_key="uv",
                stroke="#8884d8",
                fill="#8884d8",
                type_="natural",
            ),
            rx.y_axis(), 
            data=State.data,
            width=1000,
            height=400,
        )
```

ElijahAhianyo:
This PR adds support for prop Literals. 

Usage:

```python

class ComponentName(Component):
    ...
    size: Var[Literal["sm", "md", "lg", "xs"]]   
    base_value: Var[Union[Literal["dataMin", "dataMax"], int],  # use with union for multiple types


```
fixes #1864


reflex/components/forms/button.py
```diff
@@ -29,21 +29,41 @@ class Button(ChakraComponent):
...
+            "pink",
+            "whiteAlpha",
+            "blackAlpha",
+            "linkedin",
+            "facebook",
+            "messenger",
+            "whatsapp",
+            "twitter",
+            "telegram",
+        ]
```
masenf, line 60: i think we should move this to reflex/components/libs/chakra.py and import it where it's used.

reflex/components/layout/stack.py
```diff
@@ -13,7 +15,7 @@ class Stack(ChakraComponent):
     align_items: Var[str]
 
     # The direction to stack the items.
-    direction: Var[str]
+    direction: Var[Union[Literal["row", "column"], List[str]]]
```
masenf, line 18: what is the `List[str]` for?

reflex/components/forms/button.py
ElijahAhianyo: Do you mean refactoring this so to unpack (`Literal[*color_scheme])? If so, then it won't work as its only supported in python 3.11. I had this implemented initially but to revert as a  result. Alternatively, I figured using a Mixin class would work,

reflex/components/disclosure/tabs.py
```diff
@@ -13,7 +18,7 @@ class Tabs(ChakraComponent):
     tag = "Tabs"
 
     # The alignment of the tabs ("center" | "end" | "start").
-    align: Var[str]
+    align: Var[LiteralTagAlign]
```
masenf, line 21:
```suggestion
    align: Var[LiteralTabsAlign]
```

This is just for the tabs yeah?

masenf (dismissed):
This looks great!

I found one spot that I wasn't sure about.

masenf:
I love that this code passes pyright now, even if there's a bit of lying going on under the covers!!

```python
import reflex as rx
from reflex.components.libs.chakra import LiteralStatus


class State(rx.State):
    alert_status: LiteralStatus = "info"

    def set_alert_status(self, status: LiteralStatus):
        self.alert_status = status


def index() -> rx.Component:
    return rx.vstack(
        rx.select(
            ["success", "info", "warning", "error"],
            value=State.alert_status,
            on_change=State.set_alert_status,
        ),
        rx.alert(
            "This is an alert!",
            icon=True,
            title="My Alert",
            desc="This is the desc prop",
            variant="top-accent",
            status=State.alert_status,
            margin="1em",
            padding="1em",
        ),
    )


app = rx.App()
app.add_page(index)
app.compile()
```

picklelo (approved):

PR #1925, opened 2023-10-05
Lendemor/fix backward compat of menu api

Lendemor:
follow up to #1905, some issue in MenuList cause hydration error in `reflex-web`

### DIFF
--- a/reflex/components/overlay/menu.py
+++ b/reflex/components/overlay/menu.py
@@ -142,7 +142,7 @@ class MenuList(ChakraComponent):
             The MenuList component.
         """
         if len(children) == 0:
-            if not items:
+            if items is None:
                 items = []
             children = [
                 child if isinstance(child, MenuItem) else MenuItem.create(child)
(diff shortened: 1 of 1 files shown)


reflex/components/overlay/menu.py
```diff
@@ -96,10 +97,13 @@ def create(
...
             if button:
-                children.append(MenuButton.create(button))
+                if not isinstance(button, (MenuButton, Button)):
+                    children.append(MenuButton.create(button))
+                else:
+                    children.append(button)
             if not items:
                 items = []
-            children.append(MenuList.create(*items))
+            children.append(MenuList.create(items=items))
```
masenf, line 106: what is the advantage of adding another arg, `items`, here?

Lendemor: It's to avoid conflict with the "manual" API where users defined every component by themselves.

(Like in `reflex-web` where the menu items in the navbar are defined manually.)

masenf (approved):

PR #1927, opened 2023-10-05
Update Deprecations

ElijahAhianyo:
update Deprecation versions and pushed the deprecation versions of deprecated features that require extra testing back

### DIFF
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -241,7 +241,7 @@ class Component(Base, ABC):
                     feature_name="EventChain",
                     reason="to avoid confusion, only use yield API",
                     deprecation_version="0.2.8",
-                    removal_version="0.2.9",
+                    removal_version="0.3.1",
                 )
             events = []
             for v in value:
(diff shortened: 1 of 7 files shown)


masenf (dismissed):

reflex/components/component.py
```diff
@@ -241,7 +241,7 @@ def _create_event_chain(
                     feature_name="EventChain",
                     reason="to avoid confusion, only use yield API",
                     deprecation_version="0.2.8",
-                    removal_version="0.2.9",
+                    removal_version="0.3.1",
```
ElijahAhianyo, line 244: pushing the deprecation dates back as it will require additional testing

reflex/event.py
```diff
@@ -537,7 +537,7 @@ def call_event_handler(
             feature_name="EVENT_ARG API for triggers",
             reason="Replaced by new API using lambda allow arbitrary number of args",
             deprecation_version="0.2.8",
-            removal_version="0.2.9",
+            removal_version="0.3.1",
```
ElijahAhianyo, line 540: pushing this back for the same reason

Lendemor (approved):
Seems good :+1:

PR #1943, opened 2023-10-10
Var field cleanup

masenf:
Rename fields on `Var` class to avoid shadowing names on classes that the var is referencing
  * `name` -> `_var_name`
  * `type_` -> `_var_type`
  * `state` -> `_var_state`
  * `is_local` -> `_var_is_local`
  * `is_string` -> `_var_is_string`
  * `full_name` -> `_var_full_name`
  * `set_state` -> `_var_set_state`

Move from pydantic BaseModel class to regular python `dataclass` because pydantic treats underscore prefix in a weird way. This will require dropping py3.7.

AppHarness now reads output from the frontend process to avoid deadlocking it when errors get dumped to the console (found while debugging).


picklelo (approved):
Looks good to me. I don't think there will be many breaking changes because even within `reflex-web` we never accessed these fields, so we should be good

picklelo (approved):

PR #1945, opened 2023-10-11
Remove victory charts

ElijahAhianyo:
remove deprecated victory charts


masenf (approved):
approve, but lets hold off on merging this until we're really ready to ship... that was if some show stopper bug from 0.2.9 pops up, we can still fix it as 0.2.10;

once this goes to `main`, we basically _have_ to ship the breaker 0.3.0

PR #1950, opened 2023-10-12
No state No Websocket

ElijahAhianyo:
This PR removes the need for Websockets when there's no state.

### DIFF
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -489,7 +489,7 @@ export const useEventLoop = (
       return;
     }
     // only use websockets if state is present
-    if (Object.keys(initial_state).length > 0) {
+    if (Object.keys(state).length > 0) {
       // Initialize the websocket connection.
       if (!socket.current) {
         connect(socket, dispatch, ['websocket', 'polling'], setConnectError, client_storage)
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -488,17 +488,19 @@ export const useEventLoop = (
     if (!router.isReady) {
       return;
     }
-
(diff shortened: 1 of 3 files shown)


reflex/.templates/web/utils/state.js
```diff
@@ -488,17 +488,19 @@ export const useEventLoop = (
...
-
-    // Initialize the websocket connection.
-    if (!socket.current) {
-      connect(socket, dispatch, ['websocket', 'polling'], setConnectError, client_storage)
-    }
-    (async () => {
-      // Process all outstanding events.
-      while (event_queue.length > 0 && !event_processing) {
-        await processEvent(socket.current)
+    // only use websockets if state is present
```
picklelo, line 491: Maybe refactor this to be 

```js
if (Object.keys(state).length > 0) {
    return
}
```

to avoid the nestedness

reflex/app.py
```diff
@@ -175,32 +175,33 @@ def __init__(self, *args, **kwargs):
...
-            cors_allowed_origins="*"
-            if config.cors_allowed_origins == ["*"]
-            else config.cors_allowed_origins,
-            cors_credentials=True,
-            max_http_buffer_size=constants.POLLING_MAX_HTTP_BUFFER_SIZE,
-            ping_interval=constants.Ping.INTERVAL,
-            ping_timeout=constants.Ping.TIMEOUT,
-        )
+        if self.state is not DefaultState:
+            # Set up the Socket.IO AsyncServer.
```
picklelo, line 179: We can move everything in this `if` statement into a function `setup_backend` or something to clean this up as this function is getting quite large now

picklelo (commented):
Great work, code works well for me!

Just a couple refactoring comments

masenf (approved):

PR #1954, opened 2023-10-12
[REF-843] Automatically update api_url and deploy_url

masenf:
When backend_port or frontend_port are provided, reformat the defaults for api_url and deploy_url to match.

If either api_url or deploy_url are provided, then they are not implicitly updated (respect the user decision)

Fix REF-843

------------------

After this change, just setting `reflex run --backend-port` will just work without additional configuration. Also, testing apps on other devices is easier, because the API_URL does not need to be set.


reflex/.templates/web/utils/state.js
```diff
@@ -67,6 +67,24 @@ export const getToken = () => {
   return token;
 };
 
+/**
+ * Get the URL for the websocket connection
+ * @returns The websocket URL object.
+ */
+export const getEventURL = () => {
```
picklelo, line 74: Since we're modifying it here, the `api_url` printed in the debug logs will show the wrong one still. Is it possible to modify the `api_url` on the Python side instead? Or maybe not since this case handles it wherever they end up deploying it.

reflex/config.py
```diff
@@ -274,6 +291,34 @@ def get_event_namespace(self) -> str | None:
...
+            **kwargs: The kwargs passed to the config or from the env.
+        """
+        if "api_url" not in self._non_default_attributes and "backend_port" in kwargs:
+            self.api_url = f"http://localhost:{kwargs['backend_port']}"
+
+        if (
+            "deploy_url" not in self._non_default_attributes
+            and "frontend_port" in kwargs
+        ):
+            self.deploy_url = f"http://localhost:{kwargs['frontend_port']}"
```
picklelo, line 307: I don't think the `deploy_url` should ever be localhost. We only use this currently for the NextJS sitemap for indexing websites

picklelo (commented):
This is great, tested and works well for me. Not sure if we need to set the `deploy_url` though, as it doesn't make sense for it to be localhost

reflex/.templates/web/utils/state.js
masenf: The api_url in the logs will show whatever it is set to, there's no real way to munge that correctly on the python side.

I suppose a possible alternative would be to leave "localhost" alone and instead make the magic `0.0.0.0` mean "wherever i end up"?

reflex/config.py
masenf: The default `deploy_url` is already "localhost" in the `Config`... what should it be?

picklelo: I guess `localhost` is accurate in dev mode, it's just currently only used for the sitemap so it's a bit meaningless.

picklelo (approved):

PR #1961, opened 2023-10-13
Deprecate rx.get_cookies

ElijahAhianyo:


masenf (approved):
I'm less sure about removing `get_cookies` as it is part of the websocket request... but the likelihood of people using this API and getting stale data and not knowing why is pretty high.

picklelo (commented):
Should we also link to the relative docs for deprecation warnings? Maybe can be another field we add to `console.deprecate`

PR #1963, opened 2023-10-13
AppHarness: support headless mode and driver selection

masenf:
Using environment variables `APP_HARNESS_HEADLESS` and `APP_HARNESS_DRIVER`, support switching out the browser engine used by default in the AppHarness.

### DIFF
--- a/reflex/testing.py
+++ b/reflex/testing.py
@@ -397,7 +397,23 @@ class AppHarness:
             )
         if self.frontend_url is None:
             raise RuntimeError("Frontend is not running.")
-        driver = driver_clz() if driver_clz is not None else webdriver.Chrome()
+        want_headless = False
+        options = None
+        if os.environ.get("APP_HARNESS_HEADLESS"):
+            want_headless = True
+        if driver_clz is None:
+            requested_driver = os.environ.get("APP_HARNESS_DRIVER", "Chrome")
+            driver_clz = getattr(webdriver, requested_driver)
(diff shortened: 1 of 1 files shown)


reflex/testing.py
```diff
@@ -397,7 +397,23 @@ def frontend(self, driver_clz: Optional[Type["WebDriver"]] = None) -> "WebDriver
             )
         if self.frontend_url is None:
             raise RuntimeError("Frontend is not running.")
-        driver = driver_clz() if driver_clz is not None else webdriver.Chrome()
+        want_headless = False
+        options = None
+        if os.environ.get("APP_HARNESS_HEADLESS"):
```
Lendemor, line 402: Do we start adding the `REFLEX_` prefix now for the env vars, or do we use a different PR to change all of them?

reflex/testing.py
```diff
@@ -397,7 +397,23 @@ def frontend(self, driver_clz: Optional[Type["WebDriver"]] = None) -> "WebDriver
...
+            want_headless = True
+        if driver_clz is None:
+            requested_driver = os.environ.get("APP_HARNESS_DRIVER", "Chrome")
```
Lendemor, line 405: same comment as above.

Lendemor (approved):

reflex/testing.py
picklelo: Let's have a separate PR where we do this for all the env vars

PR #1967, opened 2023-10-13
deprecate get_ methods for router_data, use BaseVars instead

Lendemor:
New API for accessing router_data.

Deprecate all the previous methods: `get_token`, `get_headers`, ....

Introduced 3 new `BaseVar` in State: `State.headers`, `State.session`, `State.page`

Will fix the units tests soon.

```python
def index() -> rx.Component:
    return rx.vstack(
        rx.text(f"State.page.path: {State.page.path}"),
        rx.text(f"State.page.raw_path: {State.page.raw_path}"),
        rx.text(f"State.page.full_path: {State.page.full_path}"),
        rx.text(f"State.page.full_raw_path: {State.page.full_raw_path}"),
        rx.text(f"State.page.host: {State.page.host}"),
        rx.text(f"State.headers.origin: {State.headers.origin}"),
        rx.text(f"State.session.client_token: {State.session.client_token}"),
        rx.text(f"State.session.client_ip: {State.session.client_ip}"),
        rx.text(f"State.session.session_id: {State.session.session_id}"),
        rx.foreach(State.page.params, lambda k, i: rx.text(f"{k[0]}: {k[1]}")),
    )
```


masenf:
My initial gut reaction is, "what if i want my own state vars called `session` or `page`?"

To provide any API on the `State` class naturally precludes users from choosing the names they would prefer, but maybe we could make these like second-order names, like `state.router.page` and `state.router.session`? At least then we're only blocking a single name ("router"). Or maybe `reflex` would be a better entry point? `state.reflex.page`? Then we could put all other public API entry points hanging off of `.reflex` and that would be the only name users couldn't use.

Thoughts?

Lendemor:
> My initial gut reaction is, "what if i want my own state vars called `session` or `page`?"
> 
> To provide any API on the `State` class naturally precludes users from choosing the names they would prefer, but maybe we could make these like second-order names, like `state.router.page` and `state.router.session`? At least then we're only blocking a single name ("router"). Or maybe `reflex` would be a better entry point? `state.reflex.page`? Then we could put all other public API entry points hanging off of `.reflex` and that would be the only name users couldn't use.
> 
> Thoughts?

I think it's a good point, `router` has much less chance of being used yeah.

We could make a
```
class RouterData(Base):
    ...
```
that contains everything.

masenf (approved):
I'm realizing that a consequence of this approach is that we're now sending back all of the router data vars over the websocket with any navigation event. Maybe that's okay, it's certainly more convenient if we need to render these values.

Lendemor:
> I'm realizing that a consequence of this approach is that we're now sending back all of the router data vars over the websocket with any navigation event. Maybe that's okay, it's certainly more convenient if we need to render these values.

I think it's okay yeah.
If there is an issue with it later we can figure out a solution.

PR #1978, opened 2023-10-16
Remove Default state

ElijahAhianyo:
This PR removes the need for the default placeholder state as a follow up to #1950

### DIFF
--- a/reflex/.templates/jinja/web/utils/context.js.jinja2
+++ b/reflex/.templates/jinja/web/utils/context.js.jinja2
@@ -1,14 +1,29 @@
 import { createContext, useState } from "react"
 import { Event, hydrateClientStorage, useEventLoop } from "/utils/state.js"
 
+{% if initial_state %}
 export const initialState = {{ initial_state|json_dumps }}
+{% else %}
+export const initialState = {}
+{% endif %}
+
 export const ColorModeContext = createContext(null);
 export const StateContext = createContext(null);
 export const EventLoopContext = createContext(null);
+{% if clientStorage %}
 export const clientStorage = {{ client_storage|json_dumps }}
+{% else %}
+export const clientStorage = {}
+{% endif %}
+
+{% if state_name %}
 export const initialEvents = [
     Event('{{state_name}}.{{const.hydrate}}', hydrateClientStorage(clientStorage)),
 ]
+{% else %}
+export const initialEvents = []
+{% endif %}
+
 export const isDevMode = {{ is_dev_mode|json_dumps }}
 
 export function EventLoopProvider({ children }) {
--- a/reflex/.templates/jinja/web/utils/context.js.jinja2
+++ b/reflex/.templates/jinja/web/utils/context.js.jinja2
@@ -10,7 +10,7 @@ export const initialState = {}
 export const ColorModeContext = createContext(null);
 export const StateContext = createContext(null);
 export const EventLoopContext = createContext(null);
-{% if clientStorage %}
+{% if client_storage %}
 export const clientStorage = {{ client_storage|json_dumps }}
 {% else %}
 export const clientStorage = {}
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -210,7 +210,7 @@ class App(Base):
         self.setup_admin_dash()
 
         # If a State is not used and no overlay_component is specified, do not render the connection modal
-        if not self.state and self.overlay_component is default_overlay_component:
+        if self.state is None and self.overlay_component is default_overlay_component:
             self.overlay_component = None
 
     def __repr__(self) -> str:
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -97,7 +97,7 @@ class App(Base):
     state: Optional[Type[State]] = None
 
     # Class to manage many client states.
-    state_manager: Optional[StateManager] = None
+    _state_manager: Optional[StateManager] = None
 
     # The styling to apply to each component.
     style: ComponentStyle = {}
@@ -177,7 +177,7 @@ class App(Base):
 
         if self.state:
             # Set up the state manager.
-            self.state_manager = StateManager.create(state=self.state)
+            self._state_manager = StateManager.create(state=self.state)
 
             # Set up the Socket.IO AsyncServer.
             self.sio = AsyncServer(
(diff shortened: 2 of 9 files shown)


reflex/app.py
```diff
@@ -211,10 +209,7 @@ def __init__(self, *args, **kwargs):
         self.setup_admin_dash()
 
         # If a State is not used and no overlay_component is specified, do not render the connection modal
-        if (
-            self.state is DefaultState
-            and self.overlay_component is default_overlay_component
-        ):
+        if not self.state and self.overlay_component is default_overlay_component:
```
masenf, line 212:
```suggestion
        if self.state is None and self.overlay_component is default_overlay_component:
```

I'd prefer to see the explicit `None` check here

reflex/app.py
```diff
@@ -714,8 +719,10 @@ async def modify_state(self, token: str) -> AsyncIterator[State]:
         """
         if self.event_namespace is None:
             raise RuntimeError("App has not been initialized yet.")
+
+        self.validate_state_manager()
         # Get exclusive access to the state.
-        async with self.state_manager.modify_state(token) as state:
+        async with self.state_manager.modify_state(token) as state:  # type: ignore
```
masenf, line 725: is this `type: ignore` because `state_manager` could possibly be None? 

It would be nice if we could find a workaround for that... like maybe change `state_manager` to `_state_manager` (and make it a `pydantic.PrivateAttr`), and then have `state_manager` be a `property` that always returns a `StateManager` instance or raises an exception?

With this approach, we could get rid of the explicit `validate_state_manager` calls and all the other bits of test code that have to `assert app.state_manager is not None` to avoid typing errors.

reflex/app.py
```diff
@@ -860,6 +868,10 @@ async def upload_file(files: List[UploadFile]):
         for file in files:
             assert file.filename is not None
             file.filename = file.filename.split(":")[-1]
+
+        if app.state_manager is None:
+            raise ValueError("The state manager has not been initialized.")
```
masenf, line 873: isn't this what `validate_state_manager` is for? at any rate, see comment above.

reflex/compiler/compiler.py
```diff
@@ -98,11 +98,15 @@ def _compile_contexts(state: Type[State]) -> str:
...
-        is_dev_mode=os.environ.get("REFLEX_ENV_MODE", "dev") == "dev",
+    return (
+        templates.CONTEXT.render(
+            initial_state=utils.compile_state(state),
+            state_name=state.get_name(),
+            client_storage=utils.compile_client_storage(state),
+            is_dev_mode=os.environ.get("REFLEX_ENV_MODE", "dev") == "dev",
+        )
+        if state
+        else templates.CONTEXT.render()
```
masenf, line 109: `is_dev_mode` can probably still be rendered yeah?

reflex/state.py
```diff
@@ -1234,12 +1234,6 @@ def __setattr__(self, name: str, value: Any) -> None:
         super().__setattr__(name, value)
 
 
-class DefaultState(State):
```
masenf, line 1237: 🙌  heck yes! I always thought `DefaultState` was weird and useless 😂

masenf (changes requested):

masenf:
Fixed the conflict from my PR that just merged.

reflex/compiler/compiler.py
```diff
@@ -125,14 +129,17 @@ def _compile_page(
...
+        "imports": imports,
+        "dynamic_imports": component.get_dynamic_imports(),
+        "custom_codes": component.get_custom_code(),
+        "hooks": component.get_hooks(),
+        "render": component.render(),
+    }
+    if state:
+        kwargs["state_name"] = state.get_name()
+
+    return templates.PAGE.render(**kwargs)
```
masenf, line 142:
```suggestion
    kwargs = {"state_name": state.get_name()} if state else {}

    return templates.PAGE.render(
        imports=imports,
        dynamic_imports=component.get_dynamic_imports(),
        custom_codes=component.get_custom_code(),
        hooks=component.get_hooks(),
        render=component.render(),
        **kwargs,
    )
```

Better type checking/hinting if we only pass dynamic kwargs for what is actually dynamic.

reflex/.templates/jinja/web/utils/context.js.jinja2
```diff
@@ -1,14 +1,29 @@
...
+{% if initial_state %}
 export const initialState = {{ initial_state|json_dumps }}
+{% else %}
+export const initialState = {}
+{% endif %}
+
 export const ColorModeContext = createContext(null);
 export const StateContext = createContext(null);
 export const EventLoopContext = createContext(null);
+{% if clientStorage %}
```
masenf, line 13:
```suggestion
{% if client_storage %}
```

This is why app harness tests were failing ‼️

masenf (changes requested):

reflex/compiler/compiler.py
ElijahAhianyo: This renders as `null` if there's no state, Im guessing `is_dev_mode`  is required regardless of whether state is used or not

reflex/app.py
ElijahAhianyo: This is a better approach, got it.

reflex/compiler/compiler.py
masenf: `is_dev_mode` isn't _really_ required at all, it's just an extra var that can be imported to determine if reflex is running in dev or prod mode. The value can be determined regardless of state, as it only depends on checking an env var at compile time.

reflex/compiler/compiler.py
```diff
@@ -98,11 +98,16 @@ def _compile_contexts(state: Type[State]) -> str:
     Returns:
         The compiled context file.
     """
-    return templates.CONTEXT.render(
-        initial_state=utils.compile_state(state),
-        state_name=state.get_name(),
-        client_storage=utils.compile_client_storage(state),
-        is_dev_mode=os.environ.get("REFLEX_ENV_MODE", "dev") == "dev",
+    is_dev_mode = (os.environ.get("REFLEX_ENV_MODE", "dev") == "dev",)
```
masenf, line 101:
```suggestion
    is_dev_mode = os.environ.get("REFLEX_ENV_MODE", "dev") == "dev"
```

is_dev_mode should just be a bool

masenf (approved):

masenf (approved):
With the latest fix in `reflex/testing.py` to accommodate the change to `_state_manager`, this is passing all integration tests. Looks good from my side!

PR #1982, opened 2023-10-16
convert initialEvents to a function

### LINKED ISSUE #1979: LocalStorage state variable not being read
**Describe the bug**
We're creating a PoC to better understand if this framework can be a good fit for us.
We've implemented a sample login system and we're saving user data into local storage via a state variable. Based on this variable we've created a bunch of @rx.var in order to check if the user is authenticated.
When we do a successful login, after we redirect the user to the main page, we see for a fraction of second the expected layout for a logged in user, then suddenly we get the layout we would expected for a not logged in user. This only happens if we land directly on the login page or if we refresh it, instead if we navigate to the login page passing through the main page everything works fine.

**To Reproduce**
In order to reproduce this bug, you should perform a successful login after refreshing the login page or navigating directly to it without passing through the main page.
Here's the link to the code: https://pastebin.com/qS7ycZNa

**Expected behavior**
We would just expect to see the main page content for logged in users.

**Screenshots**
We recorded a video here: https://streamable.com/j4e0qc 

 - Python Version: 3.9.6
 - Reflex Version: 0.2.9
 - OS: Mac OS
 - Browser (Optional): Chrome

masenf:
Convert the `initialEvents` list into a function that returns the list of initial events for two advantages:

1. Get new event objects for every call, avoids the weird `.map((e) => ...e)` to copy the list
2. Call `hydrateClientStorage` each time to ensure that the local storage and cookie hydrate dict contains the most up to date value.

Fix #1979
 
-------

Add a new app harness test simulating the login flow that reproduced this issue initial. Since some additional refactoring is coming around on_load and hydrate, I wanted to be sure this flow does not regress after it is fixed.

### DIFF
--- a/integration/test_login_flow.py
+++ b/integration/test_login_flow.py
@@ -1,0 +1,144 @@
+"""Integration tests for client side storage."""
+from __future__ import annotations
+
+from typing import Generator
+
+import pytest
+from selenium.common.exceptions import NoSuchElementException
+from selenium.webdriver.common.by import By
+from selenium.webdriver.remote.webdriver import WebDriver
+
+from reflex.testing import AppHarness
+
+from . import utils
+
+
+def LoginSample():
+    """Sample app for testing login/logout with LocalStorage var."""
+    import reflex as rx
+
+    class State(rx.State):
+        auth_token: str = rx.LocalStorage("")
+
+        def logout(self):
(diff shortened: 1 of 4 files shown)


picklelo (approved):
nice

PR #1977, opened 2023-10-16
Default Json serializer for rx.Base

ElijahAhianyo:
fixes #1976

PR #1984, opened 2023-10-17
fix serialization as a whole for list/dict/Base containing custom items to serialize

### LINKED ISSUE #1976: Need for custom json serializer
During compile time, any non-serializable object encountered during the JSON serialization process is treated as a list, This means the code below:

```python
import reflex as rx
from datetime import datetime
from sqlmodel import Field

class User(rx.Model, table=True):
      first_name: str = Field()
      last_name: str = Field()
      created_at: datetime = Field(default=datetime.now())
      updated_at: datetime = Field(default=datetime.now())


class State(rx.State):
    @rx.var
     def user(self) -> User:
         with rx.session as session:
             user = session.exec(User.select.where(User.id == 1)).first()
             return user


```
Fails with error:

```
  File "/Users/eli/Documents/work/Pynecone-io/pynecone-forks/main/reflex/reflex/app.py", line 951, in on_event
    await self.emit_update(update=update, sid=sid)
  File "/Users/eli/Documents/work/Pynecone-io/pynecone-forks/main/reflex/reflex/app.py", line 921, in emit_update
    self.emit(str(constants.SocketEvent.EVENT), update.json(), to=sid)
                                                ^^^^^^^^^^^^^
  File "/Users/eli/Documents/work/Pynecone-io/pynecone-forks/main/reflex/reflex/base.py", line 33, in json
    return self.__config__.json_dumps(self.dict(), default=list)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
          ^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/json/encoder.py", line 200, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/json/encoder.py", line 258, in iterencode
    return _iterencode(o, 0)
           ^^^^^^^^^^^^^^^^^
TypeError: 'datetime.datetime' object is not iterable

```
Ideally, we should have a custom serializer to handle this

Lendemor:
Improve the serialization logic to handle the case where we want to serialize a list of custom objects.

Before PR, a datetime can be serialized,  but not a list of datetime or a dict containing datetime as value.
After PR, they can all be serialized.

Include the changes suggested by @ElijahAhianyo in #1977.

Closes #1976

### DIFF
--- a/tests/utils/test_serializers.py
+++ b/tests/utils/test_serializers.py
@@ -1,4 +1,5 @@
 import datetime
+from enum import Enum
 from typing import Any, Dict, List, Type
 
 import pytest
@@ -93,6 +94,25 @@ def test_add_serializer():
     assert not serializers.has_serializer(Foo)
 
 
+class TestEnum(str, Enum):
+    """TestEnum."""
+
+    FOO = "foo"
+    BAR = "bar"
+
+
+class TestEnum2(str, Enum):
+    """TestEnum2."""
+
+    FOO = "foo"
+    BAR = "bar"
+
+
+@serializers.serializer
+def serialize_TestEnum2(enum: TestEnum2) -> str:
+    return "prefix_" + enum.value
+
+
 @pytest.mark.parametrize(
     "value,expected",
     [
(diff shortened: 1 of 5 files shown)


ElijahAhianyo (dismissed):

tests/utils/test_serializers.py
```diff
@@ -93,6 +95,31 @@ def serialize_foo(value: Foo) -> str:
...
 
 
+class TestEnum(str, Enum):
+    """TestEnum."""
+
+    FOO = "foo"
+    BAR = "bar"
+
+
+class TestEnum2(Enum):
```
masenf, line 105:
```suggestion
class FooEnumWithPrefix(Enum):
```

2 suggestions here

* remove the `Test` prefix from the name, so pytest doesn't try to think this is a test class (also do this for `TestEnum` -> `FooEnum`)
* Add `WithPrefix` to the end to help visually distinguish why these otherwise identical classes are different (this would particularly help in the parametrizations, which I had to re-read like 3 times before I noticed the `2`)

tests/utils/test_serializers.py
```diff
@@ -93,6 +95,31 @@ def serialize_foo(value: Foo) -> str:
...
+    FOO = "foo"
+    BAR = "bar"
+
+
+@serializers.serializer
+def serialize_TestEnum2(enum: TestEnum2) -> str:
+    return "prefix_" + enum.value
+
+
+class TestBase(Base):
```
masenf, line 117:
```suggestion
class BaseSubclass(Base):
```

Again, avoiding the `Test` prefix to not get picked up in pytest collection routine (and avoid warning)

masenf: image

masenf (changes requested):
This change is so good. It lets us avoid having to custom-define `dict` in an `rx.Model` to serialize datetime and other complex types.

Just a few suggestions:

tests/utils/test_serializers.py
Lendemor: So THAT was the reason for this warnings, thanks!

masenf (approved):

PR #1990, opened 2023-10-17
Stop double compiles in dev mode

picklelo:
Create a `.web/.nocompile` function that stops the first initial compile on the backend process in dev mode. After the first skip, the file is deleted, enabling hot reload to keep working.

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -594,9 +594,9 @@ class App(Base):
             return False
 
         # Check the nocompile file.
-        if os.path.exists(constants.compiler.NOCOMPILE_FILE):
+        if os.path.exists(constants.NOCOMPILE_FILE):
             # Delete the nocompile file
-            os.remove(constants.compiler.NOCOMPILE_FILE)
+            os.remove(constants.NOCOMPILE_FILE)
             return False
 
         # By default, compile the app.
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -583,10 +583,30 @@ class App(Base):
             parent = child
         return root
 
+    def _should_compile(self) -> bool:
+        """Check if the app should be compiled.
+
+        Returns:
+            Whether the app should be compiled.
+        """
+        # Check the environment variable.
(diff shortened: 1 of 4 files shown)


reflex/utils/exec.py
```diff
@@ -154,6 +154,13 @@ def run_backend(
     """
     config = get_config()
     app_module = f"{config.app_name}.{config.app_name}:{constants.CompileVars.APP}"
+
+    # Create a .nocompile file to skip compile for backend.
+    if os.path.exists(constants.base.Dirs.WEB):
+        with open(constants.compiler.NOCOMPILE_FILE, "w"):
```
Lendemor, line 160: We could add `NOCOMPILE_FILE` to `constants/__init__.py`

reflex/utils/exec.py
```diff
@@ -154,6 +154,13 @@ def run_backend(
     """
     config = get_config()
     app_module = f"{config.app_name}.{config.app_name}:{constants.CompileVars.APP}"
+
+    # Create a .nocompile file to skip compile for backend.
+    if os.path.exists(constants.base.Dirs.WEB):
```
Lendemor, line 159: We can use `constants.Dirs.WEB`

reflex/app.py
```diff
@@ -583,10 +583,30 @@ def _app_root(self, app_wrappers):
...
 
+    def _should_compile(self) -> bool:
+        """Check if the app should be compiled.
+
+        Returns:
+            Whether the app should be compiled.
+        """
+        # Check the environment variable.
+        if os.environ.get(constants.SKIP_COMPILE_ENV_VAR) == "yes":
+            return False
```
masenf, line 594: can we just get rid of this? i don't think it was being used anywhere else

picklelo: I think this is used in other places in `reflex.py`. For example we call `_skip_compile()` in `db_init` and `migrate`. Agreed we should have only one way to do this though as this could be a bit messy to have two methods.

masenf (approved):

Lendemor (approved):

PR #1994, opened 2023-10-18
Wrap Moment Component

Lendemor:
Wrap Moment, a react component allowing to easily display relative dates, times and durations.

The component also provide a callback that trigger when the timer update (at every `interval`)

How to use : 
```python

class State(rx.State):
    """The app state."""
    ts: datetime.datetime = datetime.datetime.fromisoformat("2023-10-17T15:35:26")
```

```python
def index():
    return rx.vstack(
        rx.moment(State.ts, duration_from_now=True),
        rx.moment(State.ts, from_now=True),
        rx.moment(State.ts, duration_from_now=True, interval=1000),
        rx.moment(State.ts, from_now=True, interval=1000),
        rx.moment(State.ts, format="YYYY"),
        rx.moment(State.ts, format="YYYY/MM"),
        rx.moment(State.ts, format="YYYY/MM/DD"),
    )
```


masenf (changes requested):
needs comments for each prop so they are nicely rendered in the docs. otherwise, looks great and will come in handy.

i particularly like the `on_change` and `interval`, which opens up options for client-side polling without blocking chain calls or background tasks.

masenf:
`tz` prop requires an additional dep. i was playing around with this

```diff
diff --git a/reflex/components/datadisplay/moment.py b/reflex/components/datadisplay/moment.py
index 4598980..6d4fe2e 100644
--- a/reflex/components/datadisplay/moment.py
+++ b/reflex/components/datadisplay/moment.py
@@ -2,8 +2,8 @@
 
 from typing import Any, Dict, List
 
-from reflex.components.component import NoSSRComponent
-from reflex.vars import Var
+from reflex.components.component import Component, NoSSRComponent
+from reflex.vars import ImportVar, Var
 
 
 class Moment(NoSSRComponent):
@@ -14,6 +14,19 @@ class Moment(NoSSRComponent):
     library: str = "react-moment"
     lib_dependencies: List[str] = ["moment"]
 
+    @classmethod
+    def create(cls, *children, **props) -> Component:
+        comp = super().create(*children, **props)
+        if "tz" in props:
+            comp.lib_dependencies.append("moment-timezone")
+        return comp
+
+    def _get_imports(self) -> Dict[str, Any]:
+        imports = super()._get_imports()
+        if self.tz is not None:
+            imports["moment-timezone"] = {ImportVar(tag="")}
+        return imports
+
     # interval (how often time update / 0 to disable)
     interval: Var[int]
```

masenf (approved):

PR #2003, opened 2023-10-19
Drop python3.7 support

masenf:
The recent Literal props PR #1921 requires the use of typing.Literal, which is not present in 3.7

Since 3.7 is end of life, drop support.


Lendemor (approved):
Finally :laughing:

PR #2006, opened 2023-10-19
Support Python 3.12

masenf:
Fix outstanding py3.12 issues (there might be other issues we are not yet aware of)

* Poetry install conflict due to old version of pandas (for unit tests)
* Computed var dependency tracking change (because py312 remove LOAD_METHOD opcode from bytecode)
* MutableProxy refactored to use `__getattr__` due to changes in how py312 handles `__getattribute__` that do not play well with `wrapt.ObjectProxy`
* ~**[HACK]** Patch `uvicorn.server` module to avoid shutdown hang, because the uvicorn project has yet to acknowledge the bug and our users are starting to hit it. This only applies to py3.12 interpreter version.~ Removed now that uvicorn 0.24.0 with py3.12 support has been released.

Add 3.12.0 to github actions matrix for unit tests and integration tests.


picklelo (approved):
awesome!

PR #2009, opened 2023-10-20
[reflex export] backend.zip excludes dirs that look like venv dirs

jackie-pc:
We leave `frontend.zip` alone, because by design we only zip up generated `.web/`. Chance of false positive vs. false negative would be higher there (for excluding something that shouldn't).

For `backend.zip`, we exclude venv looking files (detected by presence of `pyvenv.cfg`).

As a drive-by we also exclude `*.db` from `backend.zip`.

### DIFF
--- a/reflex/utils/build.py
+++ b/reflex/utils/build.py
@@ -59,6 +59,8 @@ def _zip(
     component_name: constants.ComponentName,
     target: str,
     root_dir: str,
+    exclude_venv_dirs: bool,
+    exclude_sqlite_db_files: bool,
     dirs_to_exclude: set[str] | None = None,
     files_to_exclude: set[str] | None = None,
 ) -> None:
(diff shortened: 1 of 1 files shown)


masenf (approved):
looks fine from here

reflex/utils/build.py
```diff
@@ -59,6 +59,8 @@ def _zip(
     component_name: constants.ComponentName,
     target: str,
     root_dir: str,
+    exclude_venv_dirs: bool,
```
picklelo, line 62: should we add default values for these args?

picklelo (approved):

reflex/utils/build.py
jackie-pc: I don't think so.  Default values are usually for convenience:
* We cannot (or don't want to) update existing call sites.
* We think it's too burdensome to ask new call sites to be explicit.
But the drawback is we lose (IMO) a lot of readability.

In this case the trade-off seems quite skewed toward not doing defaults.  It's a private function that has two existing call sites.

PR #2010, opened 2023-10-20
Improve Var type handling for better rx.Model attribute access

### LINKED ISSUE #1814: rx.foreach Union support
In my app I'm trying to use an union to hold slightly different classes (common base members but extra members) and using rx.foreach to try to display those entities. If I specify the union, the foreach component won't run, citing :

```AttributeError: The State var `_` has no attribute 'MY_ATTRIBUTE' or may have been annotated wrongly.```

I've tried to simply make these classes inherit from a single class but I of course cannot access the other members not from the class.

Could I request support for unions for rx.foreach? Would it be hard to implement?
Edit: One option would be to provide a way of specifying a render_fn per type and a default one?

My example setup:

```python
from pydantic import Field
import reflex as rx

class Reaction1(rx.Model):
    reaction: str = Field("instant", const=True)

class Reaction2(rx.Model):
    reaction: str = Field("delay", const=True)
    delay: int = Field(default=20)

ReactionTypes = Reaction1 | Reaction2

reaction_list: list[ReactionTypes] = [
    Reaction1(**{"reaction": "instant"}),
    Reaction2(**{"reaction": "delay", "delay": 40}),
]

class State(rx.State):

    @rx.var
    def reaction_list(self) -> list[ReactionTypes]:
        return reaction_list
    
def show_entry(entry: ReactionTypes) -> rx.Component:
    if isinstance(entry, Reaction2):
        return rx.hstack(
            rx.text(entry.reaction),
            rx.text(f"delay: {entry.delay}"),
        )
    else:
        return rx.text(entry.reaction)
        
def index() -> rx.Component:
    return rx.vstack(
        rx.foreach(
            State.reaction_list,
            show_entry,
        ),
    )


app = rx.App()
app.add_page(index)
app.compile()

```

masenf:
Support accessing relationship attributes on Models

Support accessing all possible attributes when Var is annotated as a Union (fix #1814)

Use null coalescing when the var is Optional or has a default value of None

`reflex.utils.types.is_union` supports new style `UnionType` on py3.10


tests/test_state.py
```diff
@@ -2182,3 +2182,52 @@ class MutableResetState(State):
...
+
+    c2: Optional[Custom2] = None
+    c2r: Custom2
+
+
+def test_state_union_optional():
+    """Test that state can be defined with Union and Optional vars."""
+
+    class UnionState(State):
+        int_float: Union[int, float] = 0
```
masenf, line 2211: I don't think other Var operations are prepared to handle bare Union types...

reflex/utils/types.py
```diff
@@ -85,7 +87,58 @@ def is_literal(cls: GenericType) -> bool:
...
+    Args:
+        cls: The class to check.
+
+    Returns:
+        Whether the class is an Optional.
+    """
+    return is_union(cls) and type(None) in get_args(cls)
+
+
+def can_access_attribute(cls: GenericType, name: str) -> GenericType | None:
```
picklelo, line 105: Rename function since it doesn't return a bool

reflex/utils/types.py
```diff
@@ -85,7 +87,58 @@ def is_literal(cls: GenericType) -> bool:
...
+    Supports pydantic models, unions, and annotated attributes on rx.Model.
+
+    Args:
+        cls: The class to check.
+        name: The name of the attribute to check.
+
+    Returns:
+        The type of the attribute, if accessible, or None
+    """
+    from ..model import Model
```
picklelo, line 117: `from reflex.model import Model`

reflex/utils/types.py
```diff
@@ -85,7 +87,58 @@ def is_literal(cls: GenericType) -> bool:
...
+            type_ = hints[name]
+            if isinstance(type_, ModelField):
+                return type_.type_
+            return type_
+    elif is_union(cls):
+        # check in each arg of the annotation
+        for arg in get_args(cls):
+            type_ = can_access_attribute(arg, name)
+            if type_ is not None:
+                return type_
```
picklelo, line 141: Add `return None`

picklelo (approved):

PR #2012, opened 2023-10-21
Support Form controls via name attribute (no ID or ref)

### LINKED ISSUE #2074: Short-hand for clearing a form after submitting
Please create an alternative method that can set all the elements inside a form to "" after the on_submit takes place.

We used reflex in CalHacks 10.0, and this is one of the enhancements we think would be great!

masenf:
Dynamic refs / ids don't really work... but dynamic `name` is not a problem!

```python
import reflex as rx


class State(rx.State):
    n_fields: int = 0
    grid: int = 1

    def handle_submit(self, form_data: dict):
        print(form_data)

    def add_field(self):
        self.n_fields += 1

    def add_grid(self):
        self.grid += 1


def index() -> rx.Component:
    return rx.vstack(
        rx.form(
            rx.input(id="text", placeholder="text"),
            #rx.input(name="text", placeholder="text"),
            rx.pin_input(id="pin", length=4),
            rx.input(name="foo", placeholder="foo"),
            rx.button("+", on_click=State.add_field),
            rx.box(
                rx.foreach(
                    rx.Var.range(State.n_fields),
                    lambda i: rx.input(name=f"field_{i}", placeholder=f"field {i}", key=i)
                ),
                width="70vw",
            ),
            rx.button("+", on_click=State.add_grid),
            rx.responsive_grid(
                rx.foreach(
                    rx.Var.range(State.grid),
                    lambda i: rx.foreach(
                        rx.Var.range(State.grid),
                        lambda j: rx.input(
                            name=f"grid_{i}_{j}", placeholder=f"grid {i} {j}", key=f"{i}_{j}", width="4em",
                        ),
                    ),
                ),
                columns=[State.grid],
            ),
            rx.button("Submit", type_="submit"),
            reset_on_submit=True,
        ),
    )


app = rx.App()
app.add_page(index)
app.compile()
```

This PR adds a custom handleSubmit function that all forms will use, which pre-processes the fields present in the submit event via `FormData` API. It combines the form data with the ref data, allowing both methods to interoperate without breaking any existing code.

New form-based code should prefer to use the "name" attribute to avoid the overhead of using refs for no specific gain (unless focus or value setting is required).

Closes #2074


masenf:
The new `FormSubmitName` app compiles this component

```javascript
  const form_state = useContext(StateContext);
  const router = useRouter();
  const [colorMode, toggleColorMode] = useContext(ColorModeContext);
  const focusRef = useRef();

  // Main event loop.
  const [addEvents, connectError] = useContext(EventLoopContext);

  // Set focus to the specified element.
  useEffect(() => {
    if (focusRef.current) {
      focusRef.current.focus();
    }
  });

  // Route after the initial page hydration.
  useEffect(() => {
    const change_complete = () => addEvents(initialEvents());
    router.events.on("routeChangeComplete", change_complete);
    return () => {
      router.events.off("routeChangeComplete", change_complete);
    };
  }, [router]);

  const handleSubmitrxvpfysf = useCallback((ev) => {
    const $form = ev.target;
    ev.preventDefault();
    const form_data = {
      ...Object.fromEntries(new FormData($form).entries()),
      ...{},
    };

    addEvents([Event("form_state.form_submit", { form_data: form_data })]);

    if (false) {
      $form.reset();
    }
  });

  const ref_token = useRef(null);
  refs["ref_token"] = ref_token;

  return (
    <Fragment>
      <Fragment>
        {isTrue(connectError !== null) ? (
          <Fragment>
            <Modal isOpen={connectError !== null}>
              <ModalOverlay>
                <ModalContent>
                  <ModalHeader>{`Connection Error`}</ModalHeader>
                  <ModalBody>
                    <Text>
                      {`Cannot connect to server: `}
                      {connectError !== null ? connectError.message : ""}
                      {`. Check if server is reachable at `}
                      {getEventURL().href}
                    </Text>
                  </ModalBody>
                </ModalContent>
              </ModalOverlay>
            </Modal>
          </Fragment>
        ) : (
          <Fragment />
        )}
      </Fragment>
      <VStack sx={{ height: "100vh" }}>
        <Input
          id={`token`}
          isReadOnly={true}
          ref={ref_token}
          type={`text`}
          value={form_state.router.session.client_token}
        />
        <Box action={`/invalid`} as={`form`} onSubmit={handleSubmitrxvpfysf}>
          <VStack>
            <Input name={`name_input`} type={`text`} />
            <HStack>
              <PinInput length={4} name={`pin_input`}>
                {Array(4)
                  .fill(0)
                  .map((_, eavcvpyd) => (
                    <PinInputField
                      index={eavcvpyd}
                      key={eavcvpyd}
                      name={`pin_input-${eavcvpyd}`}
                    />
                  ))}
              </PinInput>
            </HStack>
            <NumberInput name={`number_input`}>
              <NumberInputField />
              <NumberInputStepper>
                <NumberIncrementStepper />
                <NumberDecrementStepper />
              </NumberInputStepper>
            </NumberInput>
            <Checkbox name={`bool_input`} value={true} />
            <Switch name={`bool_input2`} value={true} />
            <Checkbox name={`bool_input3`} value={true} />
            <Switch name={`bool_input4`} value={true} />
            <Slider name={`slider_input`}>
              <SliderTrack>
                <SliderFilledTrack />
              </SliderTrack>
              <SliderThumb />
            </Slider>
            <RangeSlider name={`range_input`}>
              <RangeSliderTrack>
                <RangeSliderFilledTrack />
              </RangeSliderTrack>
              <RangeSliderThumb index={0} />
              <RangeSliderThumb index={1} />
            </RangeSlider>
            <RadioGroup name={`radio_input`}>
              <Radio value={`option1`}>{`option1`}</Radio>
              <Radio value={`option2`}>{`option2`}</Radio>
            </RadioGroup>
            <Select name={`select_input`}>
              <option value={`option1`}>{`option1`}</option>
              <option value={`option2`}>{`option2`}</option>
            </Select>
            <Textarea name={`text_area_input`} />
            <DebounceInput
              debounceTimeout={0}
              element={Input}
              name={`debounce_input`}
              onChange={(_e0) =>
                addEvents(
                  [Event("_console", { message: _e0.target.value })],
                  _e0,
                )
              }
              type={`text`}
            />
            <Button type={`submit`}>{`Submit`}</Button>
          </VStack>
        </Box>
        <Spacer />
      </VStack>
      <NextHead>
        <title>{`Reflex App`}</title>
        <meta content={`A Reflex app.`} name={`description`} />
        <meta content={`favicon.ico`} property={`og:image`} />
      </NextHead>
    </Fragment>
  );
}
```

reflex/utils/format.py
```diff
@@ -625,3 +625,25 @@ def unescape_double_quotes_in_var(m: re.Match) -> str:
         string=value,
         flags=re.VERBOSE,
     )
+
+
+def collect_form_dict_names(form_dict: dict[str, Any]) -> dict[str, Any]:
+    """Collapse keys with consecutive suffixes into a single list value.
+
+    Seprators dash and underscore are removed, unless this would overwrite an existing key.
```
picklelo, line 633: typo: Separators

reflex/vars.py
```diff
@@ -1151,6 +1151,59 @@ def foreach(self, fn: Callable) -> Var:
...
+                f"Cannot get start of range on non-int var {start._var_full_name}."
+            )
+
+        return BaseVar(
+            _var_name=f"Array({self._var_full_name}).fill(0).map((_, i) => i + {start})",
+            _var_type=list[self._var_type],
+            _var_is_local=False,
+        )
+
+    def forrange(self, fn: Callable) -> Var:
```
picklelo, line 1181: I wonder if it makes more sense to introduce a new var if they do `rx.range(i)` which can match the API of the normal Python `range` operator. Then people can do something like

```python
rx.foreach(rx.range(5), my_function)
```

We can even then support start/stop indices

```python
rx.foreach(rx.range(1, 10), my_function)
```

Just because I think we should try to match the Python API as much as possible and not introduce new function names.

reflex/vars.py
```diff
@@ -1151,6 +1151,59 @@ def foreach(self, fn: Callable) -> Var:
             _var_is_local=self._var_is_local,
         )
 
+    def range(self, start: Var | int = 0) -> Var:
```
picklelo, line 1154: Maybe instead of this we make a special function `rx.range` to match the Python API. Then we can have more flexibility

reflex/utils/format.py
masenf, line 633:
```suggestion
    Separators dash and underscore are removed, unless this would overwrite an existing key.
```

reflex/components/forms/checkbox.py
```diff
@@ -50,6 +50,9 @@ class Checkbox(ChakraComponent):
     # The name of the input field in a checkbox (Useful for form submission).
     name: Var[str]
 
+    # The value of the input field when checked (use is_checked prop for a bool)
```
picklelo, line 53: Should we call this something like `checked_value`? It may be a bit confusing because for other components `value` is used as the name for the controlled input

picklelo (approved):
Awesome - love the ability to clear forms now

PR #2016, opened 2023-10-22
CLI switch to prod server

martinxu9:
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

- CLI commands switch to prod server for hosting service.
- Refactor hosting constants, moving those only internal to hosting service to the same file.
- Streamline the login and check authentication in all the hosting service commands. If current config is invalid, run the workflow to let user authenticate (on web) again.
- Add a command to show the supported regions in hosting service: `reflex deployments regions` a call to public endpoint
- Add an option to include sqlite db file when `backend_exclude_sqlite_db_files == True` in deploy command.


reflex/config.py
```diff
@@ -190,6 +190,11 @@ class Config:
     # The rxdeploy url.
     rxdeploy_url: Optional[str] = None
 
+    # The hosting service backend URL.
+    cp_backend_url: str = constants.Hosting.CP_BACKEND_URL
+    # The hosting service frontend URL.
+    cp_web_url: str = constants.Hosting.CP_WEB_URL
```
masenf, line 196: what users are expected to configure these knobs?

reflex/utils/hosting.py
```diff
@@ -790,14 +814,17 @@ async def get_logs(
                         row_json["timestamp"] = convert_to_local_time(
                             row_json["timestamp"]
                         )
+                    for k, v in row_json.items():
+                        if v is None:
+                            row_json[k] = str(v)
```
masenf, line 819: generally should avoid modifying the dict during iteration, prefer a comprehension here

```python
row_json_w_str = {
    k: v
    for k, v in row_json.items()
    if v is not None
}
```

reflex/utils/hosting.py
```diff
@@ -19,32 +19,55 @@
 
 from reflex import constants
 from reflex.base import Base
+from reflex.config import get_config
 from reflex.utils import console
 
+config = get_config()
+# Endpoint to create or update a deployment
+POST_DEPLOYMENTS_ENDPOINT = f"{config.cp_backend_url}/deployments"
```
masenf, line 27: this sort of thing makes me a little nervous because `get_config()` can return a different object during the lifetime of the process, however once this module is imported, the constants are stuck with whatever values were provided in the config at import time.

i don't think it's a problem now, but if we wanted to expose `cp_backend_url` as a command line flag (for example), then we'd have some more complex refactoring in this module to deal with the fact that the config when imported isn't actually the config we wanted to use.

masenf (approved):
Good clean up and consolidation of the auth token code. No serious complaints here, but a few comments.

picklelo (approved):

reflex/config.py
martinxu9: not really. Nikhil just mentioned this to me in person too. Since our docs are auto generated, these will show up there. Having them here is mainly for internally we can easily change the endpoint to staging and dev for testing instead of having a separate branch with code changes. I guess we could leave them or add comments there are internal use only. Users do not need to set them?

reflex/utils/hosting.py
martinxu9: ah, how about I just create a new one then. I still need the None to be printed. For some reason, the tabulate package is dumb enough to trip over a value being None. Skipping likely causes problem, because it's a table, column numbers need to match.

martinxu9: fixed in https://github.com/reflex-dev/reflex/pull/2016/commits/02d95ffbbb4b0544cd7c315204041d151e1ca458

masenf (approved):

PR #2017, opened 2023-10-22
[Experimental] component as props for IconButton

Lendemor:
Only accept static component because we still need to define the imports during compilation time.

```python
rx.icon_button(icon=rx.icon(tag="add"))
```

### DIFF
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -20,6 +20,7 @@ from reflex.event import (
 )
 from reflex.style import Style
 from reflex.utils import console, format, imports, types
+from reflex.utils.serializers import serializer
 from reflex.vars import BaseVar, ImportVar, Var
 
 
@@ -966,3 +967,17 @@ class NoSSRComponent(Component):
             else ""
         )
         return "".join((library_import, mod_import, opts_fragment))
+
+
+# We need to define it here instead of inside serializers.py to avoid circular import.
+@serializer
+def serialize_component(comp: Component):
+    """Serialize a component.
+
+    Args:
+        comp: The component to serialize.
+
+    Returns:
+        The serialized component.
+    """
+    return str(comp)
--- a/reflex/components/forms/iconbutton.py
+++ b/reflex/components/forms/iconbutton.py
@@ -1,6 +1,8 @@
 """An icon button component."""
 
+from reflex.components.component import Component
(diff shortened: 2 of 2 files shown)


masenf (commented):
This is cool! Would it be possible to set a component var in State now?

The private attr in Component makes me a bit nervous, we should add a test to verify that is actually a per-instance value with no bleed over between instances.

Lendemor:
I did test setting a component as var in the State, sadly it doesn't work, for two reason: 

1. Since a var State value will be determined at runtime, we can't find the import that will be needed (they must be written during compilation time.
2. Even when the import is present, the Component set inside the state var is not evaluated correctly.

I don't know if it's definitely impossible or if I'm missing some React knowledge to make it work though.

ElijahAhianyo:
> I did test setting a component as var in the State, sadly it doesn't work, for two reason:
> 
> 1. Since a var State value will be determined at runtime, we can't find the import that will be needed (they must be written during compilation time.
> 2. Even when the import is present, the Component set inside the state var is not evaluated correctly.
> 
> I don't know if it's definitely impossible or if I'm missing some React knowledge to make it work though.

I havent thought this well through but perhaps we can construct the imports dynamically at runtime using [dynamic imports](https://www.w3docs.com/learn-javascript/dynamic-imports.html) and since we know the component we should be able to extract the import paths. There could be some pitfalls I haven't thought through though.

Lendemor:
> > I did test setting a component as var in the State, sadly it doesn't work, for two reason:
> > 
> > 1. Since a var State value will be determined at runtime, we can't find the import that will be needed (they must be written during compilation time.
> > 2. Even when the import is present, the Component set inside the state var is not evaluated correctly.
> > 
> > I don't know if it's definitely impossible or if I'm missing some React knowledge to make it work though.
> 
> I havent thought this well through but perhaps we can construct the imports dynamically at runtime using [dynamic imports](https://www.w3docs.com/learn-javascript/dynamic-imports.html) and since we know the component we should be able to extract the import paths. A problem I see may be constructing the JSX

This might be possible, though I see this as a separate PR later on that we can add.
Constructing dynamic import should not be too hard, but constructing the React component from a state value might be a tad more complicated.

I'm closing this PR since #2124 has the proper implementation of this feature.

PR #2019, opened 2023-10-23
Improve Event handler Error message

ElijahAhianyo:
When the number of arguments of an event handler is not the same as that of when its been invoked, we throw an error which doesn't easily tell where the issue is from. We can use `__qualname__` to get a more accurate path and propagate the error to the component class level to tell what component has triggered the said error

### DIFF
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -268,8 +268,8 @@ class Component(Base, ABC):
                         event = call_event_handler(v, arg_spec)  # type: ignore
                     except ValueError as err:
                         raise ValueError(
-                            str(err) + f" defined in the `{type(self).__name__}` component"
-                        )
+                            f" {str(err)} defined in the `{type(self).__name__}` component"
+                        ) from err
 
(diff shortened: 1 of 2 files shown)


reflex/components/component.py
```diff
@@ -264,7 +264,12 @@ def _create_event_chain(
             for v in value:
                 if isinstance(v, EventHandler):
                     # Call the event handler to get the event.
-                    event = call_event_handler(v, arg_spec)  # type: ignore
+                    try:
+                        event = call_event_handler(v, arg_spec)  # type: ignore
+                    except ValueError as err:
+                        raise ValueError(
+                            f" {str(err)} defined in the `{type(self).__name__}` component"
```
masenf, line 271:
```suggestion
                            f" {err} defined in the `{type(self).__name__}` component"
```

the `str` is not necessary

masenf (dismissed):

masenf (approved):

PR #2023, opened 2023-10-24
Generate state delta from processed state instance

masenf:
When using redis with background tasks, the background task can generate state updates that are persisted in redis, but are never updated back in the instance of the state that was used to initiate the background task (`async with self` only updates the state instance present in the wrapper).

Since `_process_event` is called against this instance of the state that isn't used for processing, it will likely be out of date, if the background task changed anything.

With this fix, `_process_event` uses the instance of the `State` (actually a `StateProxy`) that ran the background task to determine the delta to send back from the client, instead of `self` (the not-updated instance of the state retrieved from redis as the event was started).

`StateProxy.__setattr__` logic that determines whether an attribute can be set on the state was updated for readability and to explicitly ignore `get_skip_vars`, which are not persisted to redis. Since these values are not in redis, they are really only per-instance attributes and do not need locking/sync since they only matter in the context of the instance and anyone holding a reference to a specific instance will see the changes immediately without the possibility of data race.

Fix REF-989

Added a test case which reproduces the issue seen in lorem-stream when running with redis.

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -963,14 +963,19 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
         Returns:
             The valid StateUpdate containing the events and final flag.
         """
+        # get the delta from the root of the state tree
+        state = self
+        while state.parent_state is not None:
+            state = state.parent_state
+
         token = self.router.session.client_token
 
         # Convert valid EventHandler and EventSpec into Event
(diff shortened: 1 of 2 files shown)


picklelo (approved):
Nice catch - will try to redeploy the lorem app today with @tgberkeley

PR #2025, opened 2023-10-24
Radix Themes style notation fix

ElijahAhianyo:
Fix for a bug where `sx`  is used for radix style props instead of `style`

### DIFF
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -456,6 +456,14 @@ class Component(Base, ABC):
             child.add_style(style)
         return self
 
+    def _get_style(self) -> dict:
+        """Get the style for the component.
+
+        Returns:
+            The dictionary of the component style as value and the style notation as key.
+        """
+        return {"sx": self.style}
+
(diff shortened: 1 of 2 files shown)


reflex/components/component.py
```diff
@@ -467,10 +475,10 @@ def render(self) -> Dict:
             tag.add_props(
                 **self.event_triggers,
                 key=self.key,
-                sx=self.style,
                 id=self.id,
                 class_name=self.class_name,
                 **self.custom_attrs,
+                **self._get_style(),
```
masenf, line 481:
```suggestion
                **self._get_style(),
                **self.custom_attrs,
```

lets flip these around, so that `custom_attrs` always takes precedence

masenf (changes requested):
much simpler approach, very nice

reflex/components/component.py
ElijahAhianyo: got it

masenf (approved):

PR #2029, opened 2023-10-25
check_generated_pyi: fix diff checking

masenf:
`git diff` doesn't return non-zero if there is a diff.

Expected to fail, because we are currently out of date since this check hasn't been working correctly.


Lendemor (approved):
I'm not too sure how/why `git status --porcelain` fit in this check, but seems good otherwise.

picklelo (approved):

PR #2030, opened 2023-10-25
add region check upfront when user deploys interactively

martinxu9:
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

- Fix a bug that does not print the list of enabled regions in `reflex deployments regions` command.
- Fetch enabled regions when user deploys interactively. In this case, there is a prepare_deploy call to the hosting server. If we have a list of enabled regions, we check user input against it.
- Example of it running: image
- Change the prompt for deployment name to 

```
Choose a name for your deployed app (quiz-lime-moon):
```

- Add a prompt to let user confirm if they have all deps in requirements.txt if only reflex is found in there: image


reflex/reflex.py
```diff
@@ -544,9 +552,19 @@ def deploy(
...
-        )
+        console.debug(f"{enabled_regions=}")
+        while True:
+            region_input = console.ask(
+                "Region to deploy to", default=regions[0] if regions else "sjc"
+            )
+
+            if enabled_regions is None or region_input in enabled_regions:
+                break
+            else:
```
picklelo, line 563: nit: don't need the `else` case since the `if` case breaks

picklelo (approved):

PR #2032, opened 2023-10-25
Improve pyi script with stubgen and AST.

Lendemor:


Lendemor:
@masenf included all this changes in #2034

PR #2033, opened 2023-10-25
Autogenerating Dependencies During Deployment

Alek99:
## Current Approach

Generate a requirements.txt with just reflex. When the user goes to deploy if they do not add any packages to requirements.txt warn them about this and proceed with their deployment. 

If they are missing any the deployment will fail when running the backend.

## Proposed Approach 
If a requirements.txt is not present in the user's project when they go to deploy generate a requirements.txt using only their top-level dependencies.

If the requirements.txt is present skip.

```
Could not find requirements.txt. Do you want to generate it? [y/n]: y
Name of deployment (test-teal-apple): 
...
```

Generates a requirement.txt in the root of their project

### Pros 
- Saves the user time during deployment, when generating top-level dependencies
- Ensures they don't miss any requirements in their project which would cause the deployment to fail
- Works with venv, poetry and also without using a venv

### Cons
- Catches all top-level dependencies so may get extra if the user has added a dependency and then taken it out of their project without updating their environment
- Introduces another package (It is pretty small though)


reflex/utils/dependency.py
```diff
@@ -0,0 +1,46 @@
...
+            f.write(line + "\n")
+
+
+def check_requirements():
+    """Check if the requirements are installed."""
+    if not os.path.exists(constants.RequirementsTxt.FILE):
+        console.warn("It seems like there's no requirements.txt in your project.")
+        response = console.ask(
+            "Would you like us to auto-generate one based on your current environment?",
+            choices=["y", "n"],
```
martinxu9, line 37: Nice. Do we want to give them Y/n, enter is yes

reflex/utils/dependency.py
```diff
@@ -0,0 +1,47 @@
...
+
+from reflex import constants
+from reflex.utils import console
+
+
+def generate_requirements():
+    """Generate a requirements.txt file based on the current environment."""
+    # Run the command and get the output
+    result = subprocess.run(
+        ["pipdeptree", "--warn", "silence"],
```
masenf, line 17:
```suggestion
        [sys.executable, "-m", "pipdeptree", "--warn", "silence"],
```

more reliable invocation for venvs

reflex/utils/dependency.py
Alek99: I would rather not here just because I want this command to be very explicit

masenf (approved):

masenf (approved):

PR #2034, opened 2023-10-26
Refactor pyi_generator to use AST directly

masenf:
* Dynamically import modules to `eval` type hints
* Use a `NodeTransformer` to walk the abstract syntax tree of the original module
* Using AST data and dynamic runtime data, construct a new `.create` method for `Component` classes that includes all props declared as pydantic fields.
* Remove annotated assignment targets
* Remove assignments entirely (unless it's assigning `typing.Any` as an ImportError fallback)
* Add docstrings and type annotations throughout the script


masenf:
Docstrings for Accordion.create and ColorModeButton.create are getting generated weird because some of the props in `create` overlap with props defined in the class.

scripts/pyi_generator.py
```diff
@@ -177,61 +46,457 @@
...
+        defaults=[],
+    )
+    definition = ast.FunctionDef(
+        name="create",
+        args=create_args,
+        body=[
+            ast.Expr(
+                value=ast.Constant(value=_generate_docstrings(all_classes, all_props))
+            ),
+        ],
```
Lendemor, line 323: Should probably include an `ast.Ellipsis` after the docstring for cleanliness ?

scripts/pyi_generator.py
```diff
@@ -244,215 +509,57 @@ class PyiGenerator:
...
+                pyi_content.append(formatted_line)
 
-    def _scan_file(self, file):
-        self.current_module_path = os.path.splitext(file)[0]
-        module_import = os.path.splitext(os.path.join(self.root, file))[0].replace(
-            "/", "."
-        )
+        pyi_path = module_path.with_suffix(".pyi")
+        pyi_path.write_text("\n".join(pyi_content))
+        logger.info("Wrote %s", pyi_path)
```
Lendemor, line 534: Any reason that prevent using f-string here?

Lendemor (changes requested):
Some small additions/changes, but I love it, should make it much easier to extend later if needed.

scripts/pyi_generator.py
masenf: no, i can fix that

Lendemor (approved):

PR #2037, opened 2023-10-26
Reduce Syntax highlighter footprint

### LINKED ISSUE #1997: Reducing React Syntax Highlighter Footprint
Right now we are using the full build of syntax highlighter I believe we can use a lighter build which should sugnificantly reduce the footprint

https://github.com/react-syntax-highlighter/react-syntax-highlighter#prism

```javascript
import { Light as SyntaxHighlighter } from 'react-syntax-highlighter';
import js from 'react-syntax-highlighter/dist/esm/languages/hljs/javascript';
import docco from 'react-syntax-highlighter/dist/esm/styles/hljs/docco';

SyntaxHighlighter.registerLanguage('javascript', js);
```

Would be this instead of using prism directly. The trick with this would be to dynamically import the specific languages based on what they used in there code

For reference syntax highlighter is often the takes the largest to render of any component by a large margin. So this would significantly speed up rendering times where codeblock is used.
image

ElijahAhianyo:
Running without PR(before):
Screenshot 2023-10-26 at 2 36 57 PM

Running with PR(after):
Screenshot 2023-10-26 at 2 35 27 PM

fixes #1997


reflex/components/libs/chakra.py
```diff
@@ -121,6 +121,55 @@ class ChakraColorModeProvider(Component):
 LiteralVariant = Literal["solid", "subtle", "outline"]
 LiteralDividerVariant = Literal["solid", "dashed"]
 LiteralTheme = Literal["light", "dark"]
+LiteralCodeBlockTheme = Literal[
```
masenf, line 124: can these be defined in `code.py`, since they're not really chakra-related and will be used even with radix.

masenf (changes requested):

Alek99:
Is this ready for review?

ElijahAhianyo:
> Is this ready for review?

yes it is

masenf (dismissed):

reflex/components/datadisplay/code.py
```diff
@@ -1,33 +1,365 @@
...
     library = "react-syntax-highlighter@15.5.0"
 
-    tag = "Prism"
+    tag = "PrismAsyncLight"
+
+    alias = "SyntaxHighlighter"
 
     # The theme to use ("light" or "dark").
-    theme: Var[LiteralTheme]
+    theme: Var[LiteralCodeBlockTheme] = "prism"  # type: ignore
```
picklelo, line 359: Should we update the comment for this now?

reflex/components/datadisplay/code.py
```diff
@@ -75,6 +437,11 @@ def create(
         # This component handles style in a special prop.
         custom_style = props.pop("custom_style", {})
 
+        # react-syntax-highlighter doesnt have an explicit "light" theme so we use the default
+        # prism theme to ensure code compatibility
+        if "theme" in props and props["theme"] == "light":
+            props["theme"] = "prism"
```
picklelo, line 443: This should be `one-light` instead of `prism` to keep our main website consistent. I think the dark we should map to `one-dark` also

picklelo (commented):
Awesome! Just need to fix the default theme and this looks good

picklelo (changes requested):
Need to fix the default style

reflex/components/datadisplay/code.py
ElijahAhianyo: We could do that, however, I expected this issue https://linear.app/reflex-dev/issue/REF-1078/in-props-section-add-a-section-for-valid-literals to handle displaying the valid prop values from the literals.

picklelo (approved):
looks good to me!

PR #2041, opened 2023-10-26
Apply new pyi script

Lendemor:
For all components ~except~ including plotly, need to resolve resolution conflict.


Alek99:
I'm not so comfortable changing all the pyi files this close to calhacks, lets just generate this for the new data editor

Alek99 (changes requested):
I'm not comfortable with this large of a pr so close to the release and calhacks, lets just add the new pyi for the data editor and keep the old pyi for the other files

reflex/components/base/body.pyi
```diff
@@ -1,21 +1,85 @@
...
-from reflex.components.component import Component
+from typing import Any, Literal, overload, Dict, Optional, Union, List
 from reflex.vars import Var, BaseVar, ComputedVar
-from reflex.event import EventHandler, EventChain, EventSpec
+from reflex.event import EventChain, EventHandler, EventSpec
+from reflex.style import Style
+from reflex.components.component import Component
 
 class Body(Component):
+    ...
```
masenf, line 13: oops, these shouldn't be here if the class is not empty...we should move these lines (https://github.com/reflex-dev/reflex/blob/9e5a47b52d51b7a60608355515de82a57ee2ba69/scripts/pyi_generator.py#L449-L451) to the end of that `visit_ClassDef` function to ensure we dont add the ellipsis when we will generate a function.

masenf (changes requested):
Looks good, but looks big. I can understand the hesitation to bring this in this close to calhacks.

One other fixup is needed from my previous PR:

Lendemor:
> I'm not comfortable with this large of a pr so close to the release and calhacks, lets just add the new pyi for the data editor and keep the old pyi for the other files

Then I'll add it on the datatable PR and we can merge this one after the release.

masenf (approved):
nice

PR #2049, opened 2023-10-27
fix onload method not working in prod mode (and hosting)

Lendemor:


picklelo (approved):
nice catch!

masenf:
Wow, very good fix... we miss this in testing because `test_dynamic_routes` doesn't use the page decorator.

PR #2050, opened 2023-10-27
Update base template styling

picklelo:
Don't use utf-8 encoding.

### DIFF
--- a/reflex/.templates/apps/demo/code/sidebar.py
+++ b/reflex/.templates/apps/demo/code/sidebar.py
@@ -169,7 +169,7 @@ def sidebar() -> rx.Component:
             sidebar_footer(),
             height="100dvh",
         ),
-        display=["none", "none", "flex"],
+        display=["none", "none", "block"],
         min_width=sidebar_width,
         height="100%",
         position="sticky",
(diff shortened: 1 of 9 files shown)


masenf:
why is the encoding fix necessary? shouldn't we be reading these files as text?

picklelo:
Undid the utf-8 encoding. This PR is now for just updating the base template styling and adding the page kwargs to the template.

masenf (commented):
sidebar doesnt show up at all on mobile now, not clear how to display it, am i missing something?

picklelo:
@masenf That is the intention - I moved the links to the top menu bar for mobile.

masenf:
Oh alright i see that now.. second question, why does the content always scroll off the page?

reflex/.templates/apps/demo/code/sidebar.py
```diff
@@ -169,6 +169,7 @@ def sidebar() -> rx.Component:
             sidebar_footer(),
             height="100dvh",
         ),
+        display=["none", "none", "flex"],
```
masenf, line 172:
```suggestion
        display=["none", "none", "block"],
```

masenf (changes requested):

masenf (dismissed):

masenf (approved):

PR #2055, opened 2023-10-27
Add none to color scheme literal

ElijahAhianyo:


masenf (approved):
this sort of seems like the wrong approach, since `"none"` isn't a valid color scheme according to chakra... but i get it, ship it.

picklelo:
Yeah just to make sure we don't break too many people's existing apps.. But we will think of a cleaner way later.

PR #2056, opened 2023-10-27
Remove .pyc and __pycache__ from template dir

masenf:
Since the .py files are having imports rewritten anyway, these bytecode cache files are not helpful and will have to be regenerated. Explicitly deleting them avoids hitting an exception attempting to read them as utf-8 text.


picklelo (dismissed):

picklelo (approved):
nice fix

PR #2062, opened 2023-10-27
Bump to v0.3.0

picklelo:


masenf:
Reflex-web is failing because it's strict pinned to 0.2.9 and needs to be changed to `==0.3.1` then `reflex-ci` branch updated.

masenf (approved):

PR #2065, opened 2023-10-27
Support 3.9 on templates

picklelo:


Alek99 (approved):

masenf (commented):
should we put this in all files?

picklelo:
@masenf yes we should probably. The unit tests usually catch this but didn't for the template. I'm going to cut another release right now - 0.3.1

PR #2073, opened 2023-10-28
[REF-144] Automatically memoize stateful components

masenf:
If a component depends on state or emits events, render it as its own component function outside of the page component.

The page itself no longer depends on state; this means any non-stateful components will not re-render when the state changes.

This results in a 10x re-render speedup on complex pages of reflex-web, like `/docs/state/events/`. With current `main` branch a react re-render consumes approximately 110ms, and with this branch, its down to about 13ms! ⚡️

Some parts are still a bit ugly and need refinement, but this can now be tested against `reflex-web` repo.


masenf:
Current implementation is not working with style props that depend on state or flexdown evaluated components that depend on state.

masenf:
close in favor of #2198

PR #2085, opened 2023-10-30
Hosting CLI: use http endpoint to return deploy milestones

martinxu9:
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

- For users with less ideal internet conditions, it's possible the websocket streaming deploy milestones get disconnected. There was no retry mechanism. Instead, hosting service adds an equivalent endpoint to query these events once. CLI can query this endpoint peridiocally.

### DIFF
--- a/reflex/utils/hosting.py
+++ b/reflex/utils/hosting.py
@@ -9,7 +9,7 @@ import re
 import time
 import uuid
 import webbrowser
-from datetime import datetime
+from datetime import datetime, timedelta
 from http import HTTPStatus
 from typing import List, Optional
 
@@ -41,6 +41,8 @@ GET_DEPLOYMENT_STATUS_ENDPOINT = f"{config.cp_backend_url}/deployments"
 GET_REGIONS_ENDPOINT = f"{config.cp_backend_url}/deployments/regions"
 # Websocket endpoint to stream logs of a deployment
 DEPLOYMENT_LOGS_ENDPOINT = f'{config.cp_backend_url.replace("http", "ws")}/deployments'
+# The HTTP endpoint to fetch logs of a deployment
+POST_DEPLOYMENT_LOGS_ENDPOINT = f"{config.cp_backend_url}/deployments/logs"
 # Expected server response time to new deployment request. In seconds.
 DEPLOYMENT_PICKUP_DELAY = 30
 # End of deployment workflow message. Used to determine if it is the last message from server.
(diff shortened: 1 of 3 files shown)


martinxu9:
@ElijahAhianyo could you help me try this PR for deploying an app? I think you ran into some issues https://discord.com/channels/1029853095527727165/1168071548817645599/1168090802740809778 I just switched out the websocket and replaced with http and retries. Let me know if it helps with this particular situation. There'll be a different PR for the overall slowness.

reflex/utils/hosting.py
```diff
@@ -726,7 +728,22 @@ def get_deployment_status(key: str) -> DeploymentStatusResponse:
...
+def convert_to_local_time_with_tz(iso_timestamp: str) -> datetime | None:
+    """Helper function to convert the iso timestamp to local time.
+
+    Args:
+        iso_timestamp: The iso timestamp to convert.
+
+    Returns:
+        The converted timestamp with timezone.
+    """
+    try:
```
picklelo, line 740: What causes this to raise an exception? Generally shouldn't catch bare Exceptions

reflex/utils/hosting.py
```diff
@@ -736,7 +753,9 @@ def convert_to_local_time(iso_timestamp: str) -> str:
         The converted timestamp string.
     """
     try:
-        local_dt = datetime.fromisoformat(iso_timestamp).astimezone()
+        local_dt = convert_to_local_time_with_tz(iso_timestamp)
+        if local_dt is None:
+            return iso_timestamp
         return local_dt.strftime("%Y-%m-%d %H:%M:%S.%f %Z")
     except Exception as ex:
```
picklelo, line 760: Same here, can we catch a more specific error / do we know why we need the catch?

reflex/utils/hosting.py
```diff
@@ -1006,6 +1025,79 @@ def log_out_on_browser():
...
+                    maybe_timestamp := convert_to_local_time_with_tz(row["timestamp"])
+                ) is not None:
+                    console.debug(
+                        f"Updating from {from_iso_timestamp} to {maybe_timestamp}"
+                    )
+                    # Add a small delta so does not poll the same logs
+                    from_iso_timestamp = maybe_timestamp + timedelta(microseconds=1e5)
+                else:
+                    console.error(f"Unable to parse timestamp {row['timestamp']}")
+                    raise ValueError
```
picklelo, line 1081: Should we specify an error in the ValueError? Also, do we want to break out of the loop here or just continue?

reflex/utils/hosting.py
martinxu9: the conversion can raise TypeError, ValueError, let me change to the more specific exception types.

reflex/utils/hosting.py
martinxu9: Actually we don't need to catch here, it's already caught in the call to convert.

reflex/utils/hosting.py
martinxu9: Good catch, we should continue here instead.

picklelo (approved):

PR #2087, opened 2023-10-31
rx.background: make yield after async with self optional

masenf:
Previously, the StateProxy was responsible for emitting state updates after exiting the `async with self` block, using `app.modify_state` helper to actually do the heavy lifting. This created a redundant path where `async with self` updates were being handled in a different spot than normal `yield` from a background task.

With this change, the `@rx.background` decorator gets a new `automatic_yield_after_modifications` kwarg (default True), which controls this behavior. The StateProxy no longer uses `app.modify_state`, but instead `app.state_manager.modify_state` (which does NOT send the update automatically).

A new module, `reflex.utils.rewrite` has been added, which facilitates rewriting background event handlers to include a `yield` after `async with self` blocks. Since the yield is part of the function, the updates can be handled via the normal `_process_event` path and emitted from the background task coroutine. This arrangement also allows users to explicitly opt-out of automatic updates and function rewriting if they so choose.

Additionally, some event handlers, like upload, cannot rely on the websocket to send their deltas, so removing the out-of-band update path which relied on the websocket allows upload handlers to be implemented as background task, for example when processing large uploads.

In the future, I can see this mechanism being used to smooth out some syntactical difficulties around native `for` and `if`, by rewriting the AST of the render functions for the frontend?

TODO:
- [ ] use `astunparse` for py3.8 compatibility
- [ ] figure out what to do when source is not available (reflex-web)

Is this just too complicated? Is there a better way?

To support the upload case, we could just say you have to explicitly `yield` to send state updates... but `rx.background` is already weird enough without extra rules, hence trying to make it easier for users by retaining the automatic delta behavior.

### DIFF
--- a/reflex/event.py
+++ b/reflex/event.py
@@ -17,7 +17,7 @@ from typing import (
 
 from reflex import constants
 from reflex.base import Base
-from reflex.utils import console, format
+from reflex.utils import console, format, rewrites
 from reflex.utils.types import ArgsSpec
 from reflex.vars import BaseVar, Var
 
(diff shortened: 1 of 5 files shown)


masenf:
This is crazy and no one should be doing this 😄

PR #2099, opened 2023-10-31
add column resize for data editor

Lendemor:
Allow resizing columns with onColumnResize.

How it works: 
```python
data_ = {
    "Column1": [1, 2, 3, 4, 5],
    "Column2": ["A", "B", "C", "D", "E"],
    "Column3": [0.1, 0.2, 0.3, 0.4, 0.5],
}

class State(rx.State):
    columns: list = [
        {"title": "Column1", "id": "column1", "type": "int"},
        {"title": "Column2", "id": "column2", "type": "str"},
        {"title": "Column3", "id": "column3", "type": "float"},
    ]
    data: list = [
        dict((key.lower(), value) for key, value in zip(data_.keys(), values))
        for values in zip(*data_.values())
    ]

    def col_resized(self, column, width):
        print(column, width)
        for col in self.columns:
            if col.get("id") == column.get("id"):
                col["width"] = width


def index():
    return rx.hstack(
        # rx.foreach(["B1", "B2", "B3"], button_stop),
        # on_click=State.outer_click,
        rx.data_editor(
            columns=State.columns,
            data=State.data,
            rows=5,
            on_column_resize=State.col_resized,
            overscroll_x=100,
        )
    )
```


Lendemor:
Need https://github.com/reflex-dev/reflex-web/pull/292

picklelo (approved):

PR #2106, opened 2023-11-01
REF-1050/Investigate Windows Hot reload

ElijahAhianyo:
This PR fixes a peer dependency resolution error by the postcss lib causing a relock of the  `package-lock.json` file.

### DIFF
--- a/reflex/config.py
+++ b/reflex/config.py
@@ -201,6 +201,9 @@ class Config(Base):
     # The worker class used in production mode
     gunicorn_worker_class: str = "uvicorn.workers.UvicornH11Worker"
 
+    # npm prefer-offline flag to prefer local cache.
+    npm_prefer_offline: bool = False
+
     # Attributes that were explicitly set by the user.
     _non_default_attributes: Set[str] = pydantic.PrivateAttr(set())
 
--- a/reflex/config.py
+++ b/reflex/config.py
@@ -201,9 +201,6 @@ class Config(Base):
     # The worker class used in production mode
     gunicorn_worker_class: str = "uvicorn.workers.UvicornH11Worker"
 
-    # npm prefer-offline flag to prefer local cache.
-    npm_prefer_offline: bool = False
-
     # Attributes that were explicitly set by the user.
     _non_default_attributes: Set[str] = pydantic.PrivateAttr(set())
 
--- a/reflex/config.pyi
+++ b/reflex/config.pyi
@@ -95,7 +95,6 @@ class Config(Base):
         rxdeploy_url: Optional[str] = None,
         cp_backend_url: Optional[str] = None,
         cp_web_url: Optional[str] = None,
-        npm_prefer_offline: bool = False,
         username: Optional[str] = None,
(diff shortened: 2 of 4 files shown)


masenf (approved):
cool!

masenf:
It doesn't seem significantly faster on my windows VM. Still installs tailwind every time...

ElijahAhianyo:
> It doesn't seem significantly faster on my windows VM. Still installs tailwind every time...

I see, when you look at the npm logs for tailwind, does the cache get revalidated every time?

masenf:
> does the cache get revalidated every time

not sure how to tell...

ElijahAhianyo:
> also seems like this turns up the log level of npm by default? my terminal is displaying a lot more output now

I added this as a result of this issue(https://github.com/reflex-dev/reflex/issues/2115) where the installation fails but theres not much details on why. Probably there's a better way to handle this, Maybe move that a log level up(Error instead of debug)?

ElijahAhianyo:
> > does the cache get revalidated every time
> 
> not sure how to tell...

usually you'd see this in the logs when its revalidated:
```
Debug: Running command: ['/opt/homebrew/bin/npm', 'add', '-d', 'tailwindcss@3.3.2', '@tailwindcss/typography']
Debug: Installing tailwind
⠙ Installing tailwindnpm info using npm@9.8.0
npm info using node@v18.17.0
npm http fetch GET 200 https://registry.npmjs.org/@tailwindcss%2ftypography 2824ms (cache revalidated)
npm http fetch GET 200 https://registry.npmjs.org/tailwindcss 1195ms (cache revalidated)
npm http fetch POST 200 https://registry.npmjs.org/-/npm/v1/security/advisories/bulk 489ms
```

Also just to get a sense of how little of a difference it makes for you, what are the time differences(or numbers) when you run with the `npm_prefer_offline` vs when you dont?

masenf:
When i run with npm_prefer_offline, it takes about 12 seconds from when i hit save to when the npm output stops scrolling.

When i run without npm_prefer_offline, it's about the same, 13 seconds from when i hit save.

As for the tailwind thing, i think my output is getting screwed up on windows, because i see this when it mentions tailwind

```console
npm info ok
⠋ Installing tailwindnpm info using npm@9.6.7
npm info using node@v18.17.0
npm http fetch GET 200 https://registry.npmjs.org/tailwindcss 673ms (cache updated)
⠋ Installing tailwind[##################] / reify:@next/swc-darwin-arm64: timing reifyNode:node_modules/@next/swc-darwinnpm http fetch POST 200 https://registry.npmjs.org/-/npm/v1/security/advisories/bulk 130ms
⠙ Installing tailwind[##################] \ reify:@next/swc-darwin-arm64: http fetch POST 200 https://registry.npmjs.org
```

Alek99:
I see this is approved, but may have some issues. What is the best way to go about testing this

ElijahAhianyo:
Moving the performance tests to another PR as I need to do a bit more investigation on why this behaves differently on other windows

picklelo (approved):

PR #2109, opened 2023-11-01
Use blank template as default

picklelo:
Switch back to the original blank template. The sidebar template can be accessed through `reflex init --template sidebar`

TODO: I still have to switch the file names from "base" -> "sidebar" before we can merge this.


masenf (approved):
🚢

PR #2117, opened 2023-11-02
Compatibility with older typer versions

masenf:
Allow typer versions that do not understand pretty_exceptions_enable kwarg


picklelo (approved):
Looks good - we can also just raise our minimum typer requirement at this point too. The only reason we had it lower was to disable these pretty prints

masenf:
> we can also just raise our minimum typer requirement

yeah i was going back and forth about this... if we want to maximize our compatibility with other libs, then we should be generous about our supported dep ranges, but then that increases our testing burden.

PR #2119, opened 2023-11-02
[WIP] add search support to data_editor

Lendemor:
Implement search props and event triggers for data_editor.

State and data:
```python
data_ = {
    "Column1": [1, 2, 3, 4, 5],
    "Column2": ["A", "B", "C", "D", "E"],
    "Column3": [0.1, 0.2, 0.3, 0.4, 0.5],
}

class State(rx.State):
    search_in_grid: bool = False
    search_value: str = ""
    search_results: list = [[0, 0], [2, 0], [2, 2]]
    columns: list = [
        {"title": "Column1", "id": "column1", "type": "int"},
        {"title": "Column2", "id": "column2", "type": "str"},
        {"title": "Column3", "id": "column3", "type": "float"},
    ]
    data: list = [
        dict((key.lower(), value) for key, value in zip(data_.keys(), values))
        for values in zip(*data_.values())
    ]
    
    def toggle_search(self):
        self.search_in_grid = not self.search_in_grid

    def search_for(self, val):
        self.search_value = val
        if val:
            self.search_results = [[0, 0], [1, 1], [2, 2]]
```

Render:
```python
    rx.data_editor(
        columns=State.columns,
        data=State.data,
        rows=5,
        show_search=State.search_in_grid,
        search_value=State.search_value,
        on_search_value_change=State.search_for,
        on_search_close=State.toggle_search,
         search_results=State.search_results,
    )
```

### DIFF
--- a/reflex/components/datadisplay/dataeditor.py
+++ b/reflex/components/datadisplay/dataeditor.py
@@ -197,6 +197,15 @@ class DataEditor(NoSSRComponent):
     # Initial scroll offset on the vertical axis.
     scroll_offset_y: Var[int]
 
+    # Enable the built-in search of the data editor.
+    show_search: Var[bool]
+
+    # The search value.
+    search_value: Var[str]
+
+    # The results of the search as a list of cells positions to highlight.
+    search_results: Var[list[list[int]]]
+
     # global theme
     theme: Var[Union[DataEditorTheme, Dict]]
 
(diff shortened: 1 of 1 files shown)


Lendemor:
### Blocking: 
When search_value is fully controlled (`value`  and `on_search_value_change` are set) thousand of events with empty string are generated, making the input unusable.
Need to investigate before it can be added.

Bonus: 
Need to apply some debounce to the search input.

PR #2122, opened 2023-11-02
Prompt for template on reflex init

picklelo:
Ask the user to specify a template interactively on the first `reflex init`.

### DIFF
--- a/docker-example/Dockerfile
+++ b/docker-example/Dockerfile
@@ -17,7 +17,7 @@ RUN python3.11 -m venv $VIRTUAL_ENV
 RUN pip install -r requirements.txt
 
 # Deploy templates and prepare app
-RUN reflex init --template blank
+RUN reflex init
 
 # Export static copy of frontend to /app/.web/_static
 RUN reflex export --frontend-only --no-zip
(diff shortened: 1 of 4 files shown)


picklelo:
One thing I haven't included yet - if the user specifies a template up front we should skip this check. Currently if they do `reflex init --template blank` it still goes through the prompt. Maybe we can remove the default argument in the init command to catch this.

masenf (commented):
yeah should be easy enough to kill the default template on line 125, technically there is no default template now

masenf (commented):
need to add `--template blank` to the docker integration test

docker-example/Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN python3.11 -m venv $VIRTUAL_ENV
 RUN pip install -r requirements.txt
 
 # Deploy templates and prepare app
-RUN reflex init
+RUN reflex init --template blank
```
masenf, line 20: we actually probably don't need this in the dockerfile, because the app should already exist when the container is being built... not sure if it hurts anything though.

masenf (approved):

docker-example/Dockerfile
picklelo: Good point - I'll remove this

masenf (approved):

PR #2125, opened 2023-11-03
Github Codespaces Support 🪐

### LINKED ISSUE #698: Websockets in Github Codespaces
I've tried using Github Codespaces to run example code. Codespaces lets me install pynecone and run it. Port 3000/8000 are opened and a link is provided to the view the webpage. Keep in mind, the hostname is a custom hostname generated by Github, it obviously can't use localhost or 192.168.1.1.

However, the site is non-responsive. To make sure my code wasn't the issue, I've cloned the _todo_ project. The website renders. If I _add an item_ it is never added to the list. Marking off an item also has no effect.

Sure enough if I look at the Developer Tools, websockets is pointing to 'localhost' still. `WebSocket connection to 'ws://localhost:8000/event/?EIO=4&transport=websocket' failed: Error in connection establishment: net::ERR_CONNECTION_REFUSED`

Is there a config so websockets knows to use the host given by GitHub and not localhost?

### LINKED ISSUE #1260: Github Code Spaces Template
If anyone if familiar with GitHub code spaces it would be awesome if we have a template with Pynecone preloaded so new users can experiment as fast as possible. We can also put a link to the template in the readme.

masenf:
Make reflex work on github codespace.

Fix #1260
Fix #698 

https://github.com/reflex-dev/reflex/assets/1524005/d2faa8eb-dd86-4332-a41a-7027dbc6fa8d

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,7 +1,14 @@
 {
 	"image": "mcr.microsoft.com/devcontainers/python:1-3.11-bookworm",
-	"features": {
-		"ghcr.io/devcontainers/features/node:1": {}
-	},
-	"postCreateCommand": "/bin/bash -c '/usr/local/bin/pip install poetry'"
+	"postCreateCommand": "/bin/bash -c 'python -m pip install poetry && python -m poetry install & git clone https://github.com/reflex-dev/reflex-examples; wait'",
+	"forwardPorts": [3000, 8000],
+	"portsAttributes": {
+		"3000": {
+			"label": "Frontend",
+			"onAutoForward": "notify"
+		},
+		"8000": {
+			"label": "Backend"
+		}
+	}
 }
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,14 @@
...
+	"image": "mcr.microsoft.com/devcontainers/python:1-3.11-bookworm",
+	"postCreateCommand": "/bin/bash -c 'python -m pip install poetry && python -m poetry install & git clone https://github.com/reflex-dev/reflex-examples; wait'",
+	"forwardPorts": [3000, 8000],
+	"portsAttributes": {
+		"3000": {
+			"label": "Frontend",
+			"onAutoForward": "notify"
+		},
+		"8000": {
+			"label": "Backend"
```
picklelo, line 11: Should we auto-forward the backend port?

picklelo (approved):

.devcontainer/devcontainer.json
masenf: As far as i know, we have to forward it, and explicitly make it public (since you can't do that in the conf here). It's because each port actually runs on different subdomains through github's proxying infrastructure.

PR #2126, opened 2023-11-03
Set unique index vars in rx.foreach

picklelo:
Previously we always set the same index var name, so nested foreach loops would overshadow the index var. This PR introduces a unique index name for every foreach component.

Test code:

```python
from rxconfig import config

import reflex as rx
from typing import Tuple, List

class State(rx.State):

    my_tuple: Tuple[str, int, int] = []
    nested_tuple: List[List[Tuple[str, int, int]]] = []

    def my_function(self):
        first_string = 'hello'
        one_string = '1'
        twostring = '2'


        for _ in range(0, 5):
            mylist = []
            for _ in range(0, 5):
                a_tuple = (first_string, int(one_string), int(twostring))
                mylist.append(a_tuple)
            self.nested_tuple.append(mylist)

def display_tuple(a_list, i):
    return rx.foreach(
        a_list,
        lambda thetuple, j: rx.cond(
            # TypeError: Unsupported Operand type(s) for >: `.at(1) of type str and 10` of type int
            True,
            rx.box(i, j),
        ),
    )

def index() -> rx.Component:
    return rx.fragment(
        rx.foreach(
            State.nested_tuple,
            display_tuple,
        ),
    )

app = rx.App()
app.add_page(index, on_load=State.my_function)
app.compile() 
```

### DIFF
--- a/reflex/components/layout/foreach.py
+++ b/reflex/components/layout/foreach.py
@@ -47,15 +47,20 @@ class Foreach(Component):
                 f"Could not foreach over var of type Any. (If you are trying to foreach over a state var, add a type annotation to the var.)"
             )
         arg = BaseVar(_var_name="_", _var_type=type_, _var_is_local=True)
+        comp = IterTag(iterable=iterable, render_fn=render_fn).render_component(arg)
         return cls(
             iterable=iterable,
             render_fn=render_fn,
-            children=[IterTag.render_component(render_fn, arg=arg)],
+            children=[comp],
             **props,
         )
 
     def _render(self) -> IterTag:
-        return IterTag(iterable=self.iterable, render_fn=self.render_fn)
+        return IterTag(
+            iterable=self.iterable,
+            render_fn=self.render_fn,
+            index_var_name=get_unique_variable_name(),
+        )
 
     def render(self):
         """Render the component.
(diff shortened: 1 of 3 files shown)


reflex/components/layout/foreach.py
```diff
@@ -66,9 +71,9 @@ def render(self):
         tag = self._render()
         try:
             type_ = (
-                self.iterable._var_type
-                if self.iterable._var_type.mro()[0] == dict
-                else self.iterable._var_type.__args__[0]
+                tag.iterable._var_type
+                if tag.iterable._var_type.mro()[0] == dict
+                else tag.iterable._var_type.__args__[0]
```
masenf, line 76:
```suggestion
                else typing.get_args(tag.iterable._var_type)[0]
```
better to use the new-style typing functions ala REF-971

reflex/components/tags/iter_tag.py
```diff
@@ -23,37 +20,36 @@ class IterTag(Tag):
...
         return BaseVar(
-            _var_name=INDEX_VAR,
+            _var_name=self.index_var_name,
             _var_type=int,
         )
 
-    @staticmethod
-    def get_index_var_arg() -> Var:
+    def get_index_var_arg(self) -> Var:
         """Get the index var for the tag.
```
masenf, line 38: can this docstring help untangle why we have `get_index_var` and `get_index_var_arg` because i think i get it, but it's still not super clear.

masenf (approved):

masenf (approved):

PR #2129, opened 2023-11-03
Assume wss:// with no port when frontend is HTTPS

masenf:
If the frontend is being accessed via HTTPS and the API_URL is either localhost, 0.0.0.0 or :: and uses a non-secure protocol, assume secure protocol (wss://) and no port. If secure protocol with a specific port is actually needed, it must be specified with `API_URL=https://0.0.0.0:7878`

Fix REF-890


Lendemor (approved):

masenf:
> should we remove that now?

No we still need that. Because we ask the user to specify the API_URL with an http(s) prefix, then we have to convert that to `ws` or `wss` for the eventurl specifically.

PR #2142, opened 2023-11-07
Lendemor/next 14

Lendemor:
upgrade nextjs version to 14.0.1


reflex/constants/installer.py
```diff
@@ -95,8 +95,8 @@ class Commands(SimpleNamespace):
         """The commands to define in package.json."""
 
         DEV = "next dev"
-        EXPORT = "next build && next export -o _static"
-        EXPORT_SITEMAP = "next build && next-sitemap && next export -o _static"
+        EXPORT = "next build"
+        EXPORT_SITEMAP = "next build && next-sitemap"
```
picklelo, line 99: The next-sitemap may have to come first to make sure it gets included in the export

reflex/constants/installer.py
```diff
@@ -95,8 +95,8 @@ class Commands(SimpleNamespace):
         """The commands to define in package.json."""
 
         DEV = "next dev"
-        EXPORT = "next build && next export -o _static"
-        EXPORT_SITEMAP = "next build && next-sitemap && next export -o _static"
+        EXPORT = "next build"
+        EXPORT_SITEMAP = "next-sitemap && next build"
```
picklelo, line 99: Seems like this change isn't working.

On the `main` branch if you do `reflex run --env prod` on reflex-web, you can navigate to localhost:3000/sitemap.xml and see the sitemap. After this PR, that page returns a 404.

reflex/utils/prerequisites.py
```diff
@@ -357,6 +357,11 @@ def update_next_config(next_config: str, config: Config) -> str:
         f'basePath: "{config.frontend_path or ""}"',
         next_config,
     )
+
+    if export:
```
picklelo, line 361: Would it be possible to read this in as a json and then do the manipulations rather than messing with the strings directly?

picklelo (commented):
Everything works well for me except for the sitemap comment

reflex/utils/prerequisites.py
Lendemor: sure

reflex/constants/base.py
```diff
@@ -24,7 +24,7 @@ class Dirs(SimpleNamespace):
     # The name of the utils file.
     UTILS = "utils"
     # The name of the output static directory.
-    STATIC = "_static"
+    STATIC = "out"
```
masenf, line 27: is this rename necessary? personally, this will break my docker builds, github pages publishing.

deploying to hosting and using the .zip files still work tho

masenf (commented):
Deployed reflex-web to the hosting service with this version of reflex and it did work. But i agree there is something funky about the sitemap generation: https://pcweb-next14.reflex.run/sitemap-0.xml

reflex/constants/base.py
Lendemor: yes it is, next JS changed the static folder to out for exporting

reflex/utils/build.py
```diff
@@ -35,21 +35,25 @@ def set_os_env(**kwargs):
...
-            "siteUrl": deploy_url,
-            "generateRobotsTxt": True,
-        }
-    )
+    config = {
+        "siteUrl": deploy_url,
+        "generateRobotsTxt": True,
+    }
+
+    if export:
```
picklelo, line 53: We can just always keep this right? It will only apply when they're exporting

reflex/utils/console.py
```diff
@@ -45,7 +45,12 @@ def debug(msg: str, **kwargs):
         kwargs: Keyword arguments to pass to the print function.
     """
     if _LOG_LEVEL <= LogLevel.DEBUG:
-        print(f"[blue]Debug: {msg}[/blue]", **kwargs)
+        msg_ = f"[blue]Debug: {msg}[/blue]"
+        progress = kwargs.pop("progress", None)
```
picklelo, line 49: Since we only need Python 3.8 this can be made to
```python
if progress := kwargs.pop("progress", None):
    ...
```

reflex/utils/prerequisites.py
```diff
@@ -337,27 +329,37 @@ def init_reflex_json():
...
-    )
-    next_config = re.sub(
-        'basePath: ".*?"',
-        f'basePath: "{config.frontend_path or ""}"',
-        next_config,
-    )
-    return next_config
+    next_config_file = os.path.join(constants.Dirs.WEB, constants.Next.CONFIG_FILE)
+
+    with open(next_config_file, "r") as file:
```
picklelo, line 340: Why do we have to read this file? Seems like in the next line we're overriding this variable completely

reflex/utils/prerequisites.py
```diff
@@ -337,27 +329,37 @@ def init_reflex_json():
...
+
+
+def _update_next_config(config, export=False):
+    next_config = {
+        "basePath": config.frontend_path or "",
+        "compress": config.next_compression,
+        "reactStrictMode": True,
+        "trailingSlash": True,
+    }
+    if export:
```
picklelo, line 357: Like the other comment - is there any reason we can't have this all the time? Won't it only apply to exports or am I missing something

picklelo (commented):
Works great for me! Just some small comments

reflex/utils/build.py
Lendemor: No we only need it when exporting, it break prod mode otherwise.

reflex/utils/prerequisites.py
Lendemor: It's needed, both export and prod mode can't work with the same config file.

reflex/utils/prerequisites.py
Lendemor: some remains from the previous logic, will remove

PR #2144, opened 2023-11-07
Lazy import modules in reflex

picklelo:
This PR is intended to speed up the time to run `import reflex`. Before we were loading everything eagerly, and it could take several seconds simply to import the package. We now offload the importing until the actual field is called, speeding up import times.

On my MacBook Pro M1: `import reflex as rx`
Before:
  Time (mean ± σ):      2.230 s ±  0.418 s    [User: 2.416 s, System: 1.922 s]                                                                                                    
  Range (min … max):    2.061 s …  3.417 s    10 runs  

After:
  Time (mean ± σ):     395.3 ms ±  46.2 ms    [User: 272.1 ms, System: 69.7 ms]
  Range (min … max):   370.5 ms … 525.2 ms


picklelo:
This still needs to work with `.pyi` before merging. Maybe we can generate one for the base `reflex/__init__.py`

masenf (approved):
brilliant

masenf (approved):

PR #2146, opened 2023-11-08
RED-1052/rx.State as Base State

ElijahAhianyo:
This PR renames `rx.State` to `rx.BaseState` and creates a base state(`rx.State`) for all client apps. With this approach every substate will be a child(or substate) of `rx.State`. Also allows multiple inheritance from `rx.State`:

```python
import reflex as rx

class FirstState(rx.State):
  ...

class SecondState(rx.State):
   ...
```

docs:  https://github.com/reflex-dev/reflex-web/pull/306


reflex/state.py
```diff
@@ -300,6 +300,32 @@ def __init_subclass__(cls, **kwargs):
...
             cls.inherited_backend_vars = parent_state.backend_vars
 
+            if (
+                any(
+                    [
+                        c
+                        for c in cls.class_subclasses[parent_state]
+                        if c.__name__ == cls.__name__
+                    ]
+                )
```
masenf, line 310:
```suggestion
                cls.__name__ in set(c.__name__ for c in cls.class_subclasses[parent_state])
```

reflex/state.py
```diff
@@ -167,11 +169,16 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
     # The event handlers.
     event_handlers: ClassVar[Dict[str, EventHandler]] = {}
 
+    # A mapping of classes and corresponding subclassses.
+    class_subclasses: ClassVar[
+        DefaultDict[Type[BaseState], Set[Type[BaseState]]]
+    ] = defaultdict(set)
```
masenf, line 175: i'm still unclear on why this has to be a dict, shouldn't each class only need to know about its subclasses?

reflex/constants/base.py
```diff
@@ -173,3 +173,4 @@ class Ping(SimpleNamespace):
 # Testing variables.
 # Testing os env set by pytest when running a test case.
 PYTEST_CURRENT_TEST = "PYTEST_CURRENT_TEST"
+RELOAD_CONFIG = "RELOAD_CONFIG"
```
masenf, line 176:
```suggestion
RELOAD_CONFIG = "__REFLEX_RELOAD_CONFIG"
```

i think we should namespace this magic var, like `SKIP_COMPILE_ENV_VAR` so it is less likely to be accidentally set

integration/test_client_storage.py
```diff
@@ -245,68 +245,78 @@ async def test_client_side_state(
...
     set_sub_state_button.click()
+    time.sleep(0.5)
     state_var_input.send_keys("c4")
     input_value_input.send_keys("c4 value")
     set_sub_state_button.click()
+    time.sleep(0.5)
     state_var_input.send_keys("c5")
     input_value_input.send_keys("c5 value")
     set_sub_state_button.click()
+    time.sleep(0.5)
```
masenf, line 260: are all these sleeps really necessary? what changed about the test to need them?

masenf (changes requested):

integration/test_client_storage.py
ElijahAhianyo: @masenf  This should be removed now. There seem to have been some race condition when I ran with redis locally that prompted the sleeps but they pass the CI now.

reflex/app.py
```diff
@@ -149,25 +150,24 @@ def __init__(self, *args, **kwargs):
...
 
-        # Special case to allow test cases have multiple subclasses of rx.State.
+        # Special case to allow test cases have multiple subclasses of rx.BaseState.
         if not is_testing_env:
-            # Only one State class is allowed.
+            # Only one Base State class is allowed.
             if len(state_subclasses) > 1:
                 raise ValueError(
-                    "rx.State has been subclassed multiple times. Only one subclass is allowed"
+                    "rx.BaseState cannot be subclassed multiple times. use rx.State instead"
```
masenf, line 161:
```suggestion
                    "rx.BaseState cannot be subclassed multiple times. Use rx.State instead."
```

reflex/testing.py
```diff
@@ -162,6 +162,9 @@ def _initialize_app(self):
         with chdir(self.app_path):
             # ensure config and app are reloaded when testing different app
             reflex.config.get_config(reload=True)
+            # reset rx.State subclasses
+            State.class_subclasses.clear()
+            # self.app_module.app.
```
masenf, line 167:
```suggestion
            # Reset rx.State subclasses.
            State.class_subclasses.clear()
```

reflex/utils/prerequisites.py
```diff
@@ -124,11 +124,13 @@ def get_app(reload: bool = False) -> ModuleType:
     Returns:
         The app based on the default config.
     """
+    os.environ[constants.RELOAD_CONFIG] = str(reload)
     config = get_config()
     module = ".".join([config.app_name, config.app_name])
     sys.path.insert(0, os.getcwd())
     app = __import__(module, fromlist=(constants.CompileVars.APP,))
     if reload:
+
```
masenf, line 133:
```suggestion
```

masenf (approved):
a few more small things

picklelo (approved):
Looks good to me!

PR #2138, opened 2023-11-06
[REF/1116]Bun as runtime on Mac and Linux

ElijahAhianyo:

PR #2151, opened 2023-11-09
Use --bun flag for bun run

picklelo:
Follow up for #2138 as we are still using the node runtime instead of bun. Tested with `reflex init` and `reflex run`. Though there may be some issues with the `next build` for the export step... We may have to revert the Bun change if the export is not stable.


reflex/utils/prerequisites.py
```diff
@@ -722,8 +722,7 @@ def prompt_for_template() -> constants.Templates.Kind:
     Returns:
         The template the user selected.
     """
-    # Show the user the URLs of each temlate to preview.
-    console.print("\nGet started with a template:")
+    # Show the usackagetarted with a template:")
```
masenf, line 725: 🤔

picklelo: oops 😅

PR #2168, opened 2023-11-14
Set is_hydrated var explicitly on rx.State

ElijahAhianyo:
Define an explicit `is_hydrated` var on the state so it is converted properly when doing `State._init_vars` on base vars in the base state

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -194,6 +194,9 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
     # The router data for the current page
     router: RouterData = RouterData()
 
+    # The hydrated bool.
+    is_hydrated: bool = False
+
     def __init__(self, *args, parent_state: State | None = None, **kwargs):
         """Initialize the state.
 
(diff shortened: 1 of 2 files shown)


masenf:
From discord [[1]](https://discord.com/channels/1029853095527727165/1173887438142320710/1174042405700386947)

> in v0.3.2, the hydrate middleware seems to be loaded first, adds the field to the base `rx.State`, and then the app module defines a substate and hits the `__init_subclass__` logic that **iameli** described above.
>
> In `main` the hydrate middleware loads _after_ we start running the module code, which actually makes sense, because importing `rx.State` is not dependent on `rx.App` (which is what brings in the middleware import).
>
> If i modify my app code to deref `rx.App` first before defining any substates, then the problem go away. Even just `print(rx.App)` at the top of the file makes this go away.

reflex/state.py
```diff
@@ -194,6 +194,9 @@ class State(Base, ABC, extra=pydantic.Extra.allow):
     # The router data for the current page
     router: RouterData = RouterData()
 
+    # The hydrated bool.
```
picklelo, line 197: Do we have to do anything to specify that this is a `Var`? Because that logic currently lives in the `init_subclass` method

masenf: substates will pick it up as a Var, but it wouldn't be directly usable off of `rx.State` itself, as far as i can tell.

PR #2172, opened 2023-11-15
Use singleton app provider to speed up compiles

picklelo:
We were creating the AppProvider for every component - this change creates a singleton for the Chakra provider to speed up build times.


masenf (commented):
Is this actually a noticeable difference??

picklelo:
The provider added around 1-2 seconds to the compile time so it made a difference. With the changes so far I've brought it down to 4s now.

masenf (approved):

PR #2178, opened 2023-11-15
Bump to v0.3.3

picklelo:


masenf (approved):
in the future we should bring in the version bump before cutting the release i think... just so the released bits match the tag. as it is now, if i check out tag `v0.3.3`, then the reflex i get is actually marked 0.3.2, which could be confusing as it doesn't match what you get with `pip install reflex==0.3.3`

picklelo:
Yes agreed will make that part of the checklist before cutting in the future

PR #2194, opened 2023-11-17
[fixup] _valid_children and _invalid_children pyi_generator and graphing components

masenf:
While merging latest changes from `main` into #2029, i noticed a few discrepancies


picklelo (approved):
good catch

PR #2198, opened 2023-11-19
[REF-144] Add context in each component to prevent rerenders

masenf:
[REF-1129] Memoize components explicitly
[REF-1130] Move memoized components to their own module if shared

* Implement `StatefulComponent`, which is rendered out to a separate component function if any Vars depend on state
  * Memoize event handlers within components to avoid unnecessary re-renders
* More determinism, less reliance on `get_unique_variable_name`, which causes otherwise identical components to render out with different code
* If a component is used in multiple pages and renders out to identical code, then write it to `utils/stateful_components.js` and import it into all pages where it is used.


reflex/app.py
```diff
@@ -660,10 +660,27 @@ def compile(self):
             app_wrappers[(20, "Theme")] = self.theme
 
         with progress:
-            for route, component in self.pages.items():
+            for _route, component in self.pages.items():
                 # TODO: this progress does not reflect actual threaded task completion
                 progress.advance(task)
                 component.add_style(self.style)
+                # add component.get_imports() to all_imports
```
picklelo, line 667: nit: can we capitalize the start of the comment and end with periods for consistency

reflex/app.py
```diff
@@ -660,10 +660,27 @@ def compile(self):
...
+                custom_components |= component.get_custom_components()
+
+            (
+                stateful_components_path,
+                stateful_components_code,
+                page_components,
+            ) = compiler.compile_stateful_components(self.pages.values())
+            compile_results.append((stateful_components_path, stateful_components_code))
+
+            for route, component in zip(self.pages.keys(), page_components):
```
picklelo, line 683: I think you shouldn't need the `.keys()`

reflex/compiler/compiler.py
```diff
@@ -205,6 +209,49 @@ def _compile_components(components: set[CustomComponent]) -> str:
...
+    Args:
+        page_components: The Components or StatefulComponents to compile.
+
+    Returns:
+        The rendered stateful components code.
+    """
+    all_import_dicts = []
+    rendered_components = {}
+
+    def get_shared_components_recursive(component: BaseComponent):
```
picklelo, line 224: Can we add more comments on what shared components are

reflex/compiler/compiler.py
```diff
@@ -205,6 +209,49 @@ def _compile_components(components: set[CustomComponent]) -> str:
...
+        """Get the shared components for a component and its children.
+
+        Args:
+            component: The component to collect shared StatefulComponents for.
+        """
+        for child in component.children:
+            get_shared_components_recursive(child)
+        if isinstance(component, StatefulComponent) and component.references > 1:
```
picklelo, line 232: Add comment describing what `component.references > 1` check is for. I understand now it's factoring out common components, but let's just add more documentation throughout the code to make onboarding to the code easier

reflex/compiler/compiler.py
```diff
@@ -205,6 +209,49 @@ def _compile_components(components: set[CustomComponent]) -> str:
...
+            component.rendered_as_shared = False
```
picklelo, line 233: Same here, why do we have to set the flags to False/True

reflex/compiler/compiler.py
```diff
@@ -330,6 +377,25 @@ def compile_components(components: set[CustomComponent]):
     return output_path, code
 
 
+def compile_stateful_components(
+    pages: Iterable[Component],
+) -> tuple[str, str, list[BaseComponent]]:
+    """Compile the stateful components.
```
picklelo, line 383: Let's expand on what stateful components are

tests/test_event.py
```diff
@@ -185,9 +185,9 @@ def test_set_focus():
     assert spec.handler.fn.__qualname__ == "_set_focus"
     assert spec.args[0][0].equals(Var.create_safe("ref"))
     assert spec.args[0][1].equals(Var.create_safe("ref_input1"))
-    assert format.format_event(spec) == 'Event("_set_focus", {ref:ref_input1})'
+    assert format.format_event(spec) == 'Event("_set_focus", {ref:`ref_input1`})'
```
picklelo, line 188: Is this what we want it to compile out to? Won't the backticks mess it up?

tests/components/layout/test_foreach.py
```diff
@@ -67,11 +67,11 @@ def display_nested_color_with_shades_v2(color):
 
 
 def display_color_tuple(color):
-    return box(text(color))
+    return box(text(color, "tuple"))
```
picklelo, line 70: What are the type annotations for?

reflex/event.py
```diff
@@ -459,7 +466,7 @@ def set_focus(ref: str) -> EventSpec:
     return server_side(
         "_set_focus",
         get_fn_signature(set_focus),
-        ref=Var.create_safe(format.format_ref(ref)),
+        ref=Var.create_safe(format.format_ref(ref), _var_is_string=True),
```
picklelo, line 469: Why are refs strings now?

reflex/components/layout/foreach.py
```diff
@@ -34,33 +35,38 @@ def create(cls, iterable: Var[Iterable], render_fn: Callable, **props) -> Foreac
...
         )
+        # Keep a ref to a rendered component to determine correct imports.
+        component.children = [
+            component._render(props=dict(index_var_name="i")).render_component()
+        ]
+        return component
 
-    def _render(self) -> IterTag:
+    def _render(self, props: dict[str, Any] | None = None) -> IterTag:
+        props = {} if props is None else props.copy()
```
picklelo, line 55: Add comments / blocks to this function

reflex/components/layout/foreach.py
```diff
@@ -34,33 +35,38 @@ def create(cls, iterable: Var[Iterable], render_fn: Callable, **props) -> Foreac
...
+        render_sig = inspect.signature(self.render_fn)
+        params = list(render_sig.parameters.values())
+        if len(params) >= 1:
+            props.setdefault("arg_var_name", params[0].name)
```
picklelo, line 59: Why use `setdefault` instead of just setting it through an index?

reflex/components/component.py
```diff
@@ -953,6 +1029,8 @@ def get_custom_components(
         if seen is None:
             seen = set()
         for child in self.children:
+            if not isinstance(child, Component):
```
picklelo, line 1032: Same here

reflex/components/component.py
```diff
@@ -1235,3 +1310,273 @@ def serialize_component(comp: Component):
...
+    ) -> tuple[str, str]:
+        """Render the code for a stateful component.
+
+        Args:
+            component: The component to render.
+
+        Returns:
+            The tag name and the rendered code.
+        """
+        rendered_code = component.render()
```
picklelo, line 1424: Block / comment this function

reflex/components/component.py
```diff
@@ -1235,3 +1310,273 @@ def serialize_component(comp: Component):
...
+        """Memoize event handler functions with useCallback to avoid unnecessary re-renders.
+
+        Args:
+            component: The component with events to memoize.
+
+        Returns:
+            A dict of event trigger name to a tuple of the memoized event trigger Var and
+            the hook code that memoizes the component.
+        """
+        trigger_memo = {}
```
picklelo, line 1471: This function also

reflex/components/component.py
```diff
@@ -1235,3 +1310,273 @@ def serialize_component(comp: Component):
...
+        for event_trigger, event_args in component._get_vars_from_event_triggers(
+            component.event_triggers
+        ):
+            if event_trigger in {
+                EventTriggers.ON_MOUNT,
+                EventTriggers.ON_UNMOUNT,
+                EventTriggers.ON_SUBMIT,
+            }:
+                continue  # do not memoize lifecycle or submit events
```
picklelo, line 1480: Put comments above the line

reflex/components/forms/debounce.py
```diff
@@ -35,95 +34,90 @@ class DebounceInput(Component):
...
-        if isinstance(child, type(self)) or len(self.children) > 1:
+        if len(children) != 1:
             raise RuntimeError(
                 "Provide a single child for DebounceInput, such as rx.input() or "
                 "rx.text_area()",
             )
+        child = children[0]
         if "on_change" not in child.event_triggers:
             raise ValueError("DebounceInput child requires an on_change handler")
+        props_from_child = {
```
picklelo, line 72: Add new lines / comments between sections to make the code more readable

picklelo (commented):
Left some initial comments - haven't done a thorough review of the logic in component.py yet, will do that on the next pass

reflex/components/component.py
```diff
@@ -1235,3 +1310,273 @@ def serialize_component(comp: Component):
...
+
+        Args:
+            component: The component to memoize.
+
+        Returns:
+            The stateful component or None if the component should not be memoized.
+        """
+        from reflex.components.base.bare import Bare
+
+        if component.tag is None:
```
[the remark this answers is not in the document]
masenf: This would be for things like `Cond` and `Foreach`... basically any subclass of `Component` that does not explicitly set a `tag` attribute will get `None`. In these cases, we assume you're doing something fancy and do no further processing.

This function returning None indicates that the component should NOT be memoized and just render it as it is.

reflex/event.py
masenf: that's what causes them to get rendered with backticks now, see example above.

masenf:
This does seem to break `can_copy=True` of the code block, need to investigate.

reflex/app.py
```diff
@@ -659,81 +654,128 @@ def compile(self):
...
-            for route, component in self.pages.items():
-                # TODO: this progress does not reflect actual threaded task completion
+        with progress, concurrent.futures.ThreadPoolExecutor() as thread_pool:
+            fixed_pages = 7
+            compile_total = sum(
+                [
+                    len(self.pages),  # Pre-processing
+                    len(self.pages),  # Memoize stateful components
+                    len(self.pages) + fixed_pages,  # Compile pages
+                    len(self.pages) + fixed_pages,  # Write pages
```
picklelo, line 664: I don't think we need to separate out all the writes, they happen pretty instantly. We should remove them or combine them to one step

reflex/app.py
```diff
@@ -659,81 +654,128 @@ def compile(self):
...
-        for output_path, code in compile_results:
-            thread_pool.apply_async(compiler_utils.write_page, args=(output_path, code))
-        thread_pool.close()
-        thread_pool.join()
+            # Wait for all compilation tasks to complete.
+            for future in concurrent.futures.as_completed(result_futures):
+                compile_results.append(future.result())
+
+            # Write the pages at the end to trigger the NextJS hot reload only once.
+            progress.update(task, description="Write:")
```
picklelo, line 770: This after the `get_frontend_packages` cause the progress back to flicker weirdly for me, I think we should get rid of this

picklelo (commented):
The component code is looking good for me - I think we should change the compile bar back to how it was before. It's a bit intimidating with the size now, I preferred when it reflected the number of pages. We can add a step for the extra pages if needed, but we don't need so many steps

masenf:
discovered a regression in this branch w.r.t. https://github.com/reflex-dev/reflex-examples/pull/165 form designer example.

```console

 1 of 2 unhandled errors

Unhandled Runtime Error
ReferenceError: Link is not defined

Source
pages/edit/form/[form_id].js (28:3) @ Link

  26 | <Box key={index_7dea9f7e65b2d3117d8d5b1b92f54044} sx={{"border": "1px solid black", "padding": "5px", "width": "100%"}}>
  27 | <HStack>
> 28 | <Link as={NextLink} href={`/edit/form/${state.form_id}/field/${field.id}`}>
     | ^
  29 | <Fragment>
  30 | {isTrue(field) ? (
  31 | <Fragment>
Call Stack
map
pages/edit/form/[form_id].js (25:40)
```

the code in question corresponds to this component

```python
def field_edit_view(field: Field):
    return rx.box(
        rx.hstack(
            rx.link(
                field_prompt(field, show_name=True),
                href=routes.edit_field(State.form_id, field.id),
            ),
            rx.spacer(),
            rx.link("X", on_click=FormEditorState.delete_field(field.id)),
        ),
        field_input(field),
        rx.text(rx.cond(field.required, "Required", "Optional")),
        border="1px solid black",
        padding="5px",
        width="100%",
    )
```

which is used inside a simple `foreach` when building the edit form

```python
        rx.foreach(
            FormEditorState.form.fields,
            field_edit_view,
        ),
```

somehow the compiler is missing the import of `Link`...

reflex/app.py
```diff
@@ -659,81 +654,119 @@ def compile(self):
...
-                # TODO: this progress does not reflect actual threaded task completion
+        with progress, concurrent.futures.ThreadPoolExecutor() as thread_pool:
+            fixed_pages = 7
+            compile_total = sum(
+                [
+                    len(self.pages),  # Memoize stateful components
+                    len(self.pages) + fixed_pages,  # Compile pages
+                ]
+            )
+            task = progress.add_task("Memoize:", total=compile_total)
```
picklelo, line 665: I don't think we should expose "memoize" to the user, it may add confusion. We can keep it all under the same "compiling" flag.

reflex/app.py
```diff
@@ -659,81 +654,119 @@ def compile(self):
...
+
+            def mark_complete(_=None):
```
picklelo, line 667: Can we combine this whole step into just one on the progress bar. Reducing the number on the steps makes the framework seem more lightweight, plus each of these steps is pretty instant on my machine

reflex/app.py
```diff
@@ -659,81 +654,119 @@ def compile(self):
...
-                "content", constants.Tailwind.CONTENT
-            )
-            compile_results.append(compiler.compile_tailwind(config.tailwind))
+            # Empty the .web pages directory.
+            compiler.purge_web_pages_dir()
 
-        # Empty the .web pages directory
-        compiler.purge_web_pages_dir()
+            # Install frontend packages.
+            self.get_frontend_packages(all_imports)
```
picklelo, line 756: Not sure if it matters, but the compile bar flickers when run in progress with this command for me

picklelo (approved):
Awesome work with this, excited for the speedups 🚀

PR #2205, opened 2023-11-20
Fix Prop conversion to camel case

### LINKED ISSUE #2204: [REF-1204] Handle style props beginning with underscores
The `to_camel_case` method currently doesn't support props that start with underscores. For example the `_hover` prop isn't working currently.

<sub>From [SyncLinear.com](https://synclinear.com) | [REF-1204](https://linear.app/reflex-dev/issue/REF-1204/handle-style-props-beginning-with-underscores)</sub>

ElijahAhianyo:
Fix conversion of snake/kebab case to camel case

fixes #2204

### DIFF
--- a/tests/utils/test_format.py
+++ b/tests/utils/test_format.py
@@ -142,6 +142,11 @@ def test_to_snake_case(input: str, output: str):
         ("kebab-case", "kebabCase"),
         ("kebab-case-two", "kebabCaseTwo"),
         ("snake_kebab-case", "snakeKebabCase"),
+        ("_hover", "_hover"),
+        ("-starts-with-hyphen", "-startsWithHyphen"),
+        ("--starts-with-double-hyphen", "--startsWithDoubleHyphen"),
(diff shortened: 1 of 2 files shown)


tests/utils/test_format.py
```diff
@@ -142,6 +142,11 @@ def test_to_snake_case(input: str, output: str):
         ("kebab-case", "kebabCase"),
         ("kebab-case-two", "kebabCaseTwo"),
         ("snake_kebab-case", "snakeKebabCase"),
+        ("_hover", "_hover"),
+        ("-starts-with-hyphen", "-startsWithHyphen"),
```
Lendemor, line 146: Is there a use case where `-starts-with-hyphen` is a valid variable name? 
Won't it clash with the `-` minus symbol?

picklelo: Yeah I don't think it can ever start with that

picklelo: Will merge this in as is since we should never hit this case really

picklelo (approved):

PR #2206, opened 2023-11-20
REF-1202: Upgrade bun version if it differs from reflex set version

ElijahAhianyo:
This PR ensures that on systems where reflex installs bun (Mac, linux), bun is installed when the host version differs from the latest version set in reflex(`reflex.constants.Bun.version`)

### DIFF
--- a/reflex/utils/prerequisites.py
+++ b/reflex/utils/prerequisites.py
@@ -487,7 +487,9 @@ def install_bun():
         return
 
     # Skip if bun is already installed.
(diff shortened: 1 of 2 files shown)


reflex/utils/prerequisites.py
```diff
@@ -487,7 +487,9 @@ def install_bun():
         return
 
     # Skip if bun is already installed.
```
Lendemor, line 489: maybe update the comment to include the additional version check

picklelo (approved):

PR #2215, opened 2023-11-22
Markdown custom code block with can_copy

masenf:
Tweaks (hacks?) to `rx.code_block` so that the `can_copy` prop is usable when passed in an `rx.markdown` custom `component_map` dict.

Adds a test case to ensure `rx.code_block` works standalone and with `rx.markdown`, with or without `can_copy=True`.


masenf:
close in favor of #2219

PR #2223, opened 2023-11-27
add next/image as default rx.image

Lendemor:


masenf (commented):
i think we should keep `rx.image` pointing to the Chakra image for now, but we should get this out there and switch reflex-web over to use it and do some more battle testing.

reflex/components/next/video.py
```diff
@@ -0,0 +1,33 @@
...
+
+from .base import NextComponent
+
+
+class Video(NextComponent):
+    """A video component from NextJS."""
+
+    tag = "Video"
+    library = "next-video"
+    is_default = True
```
masenf, line 16: what does this component actually do? reading through the docs here https://next-video.dev/docs

it seems like it's a video uploading and hosting service with nextjs integration. i'm not 100% sure this makes sense in reflex, especially not without the associated next.config.js

```
/** @type {import('next').NextConfig} */
const { withNextVideo } = require('next-video/process');
 
const nextConfig = {}; // Your current Next Config object
 
module.exports = withNextVideo(nextConfig);
```

I think this could be part of our asset story, particularly with the hosting service which does not have a lot of disk, but it seems to be missing parts and definitely needs documentation.

masenf (approved):
looks good, the image stuff works and serialization of PIL images is still working.

i'm not sure about next-video, but it doesn't seem to be harming anything here, just not sure if it can actually be used.

PR #2230, opened 2023-11-29
[reflex-web tweaks] Do not memoize children of InputGroup

masenf:
The Chakra InputGroup component depends on its children being specific types of components, so a memoized Input_* is not "seen" as the input component.

### DIFF
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -1643,10 +1643,11 @@ class StatefulComponent(BaseComponent):
         Returns:
             The memoized component tree.
         """
+        from reflex.components.forms.input import InputGroup
         from reflex.components.layout.foreach import Foreach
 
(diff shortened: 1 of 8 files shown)


reflex/components/component.py
```diff
@@ -1643,10 +1643,11 @@ def compile_from(cls, component: BaseComponent) -> BaseComponent:
         Returns:
             The memoized component tree.
         """
+        from reflex.components.forms.input import InputGroup
         from reflex.components.layout.foreach import Foreach
 
         # Foreach must be memoized as a single component to retain index Var context.
-        if not isinstance(component, Foreach):
+        if not isinstance(component, (Foreach, InputGroup)):
```
picklelo, line 1650: We should update the comment here too. Do we need to add any other components to this tuple?

PR #2237, opened 2023-11-30
deprecate state arg in App()

Lendemor:
Due to the recent PRs allowing any number of subclass for `rx.state`, the `state` argument for `App()` is no longer needed.

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -164,7 +164,7 @@ class App(Base):
             if "state" in kwargs:
                 console.deprecate(
                     feature_name="`state` argument for App()",
-                    reason="due to multiple rx.State subclasses being allowed.",
+                    reason="due to all `rx.State` subclasses being inferred.",
                     deprecation_version="0.3.5",
                     removal_version="0.4.0",
                 )


reflex/app.py
```diff
@@ -145,6 +145,13 @@ def __init__(self, *args, **kwargs):
                         Also, if there are multiple client subclasses of rx.State(Subclasses of rx.State should consist
                         of the DefaultState and the client app state).
         """
+        if "state" in kwargs:
```
masenf, line 148: can you move it back under the `if not is_testing_env`? we don't want a bunch of our tests to hit this warning

masenf (dismissed):

reflex/app.py
```diff
@@ -161,11 +161,12 @@ def __init__(self, *args, **kwargs):
...
 
-            # verify that provided state is valid
-            if self.state and self.state is not State:
-                console.warn(
-                    f"Using substate ({self.state.__name__}) as root state in `rx.App` is currently not supported."
-                    f" Defaulting to root state: ({State.__name__})"
+            if "state" in kwargs:
+                console.deprecate(
+                    feature_name="`state` argument for App()",
+                    reason="due to multiple rx.State subclasses being allowed.",
```
picklelo, line 167: Should we more explicitly mention that since v0.2.x the substates are automatically inferred. So that people with older apps who weren't getting warnings before know why it's to required.

picklelo (approved):

PR #2240, opened 2023-12-01
[REF-1349] RechartsCharts and ResponsiveContainer must be memo leaf

masenf:
The children of a RechartsChar or ResponsiveContainer should not be memoized separately, they must be memoized as a single unit.

Implement MemoizationDisposition NEVER and ALWAYS to help control memoization in downstream classes without making special cases in the compiler.


picklelo (approved):
fixes the render issue for me

PR #2242, opened 2023-12-01
[REF-1356] Track changes applied to `Base` subclass via helper method.

### LINKED ISSUE #2239: [REF-1356] Updates Custom Var with collection does not trigger re-render
**Describe the bug**
When updating fields of a custom var via instance methods, the frontend does not receive the delta to update the state on it's side. (Almost identical to https://github.com/reflex-dev/reflex/issues/1675.)

**To Reproduce**

Using https://github.com/reflex-dev/reflex/issues/1675 as a base, but with one small change:

```python
class User(rx.Base):
    email: str = ""
    first_name: str = ""
    last_name: str = ""

    # Previously was at the State level
    def set_email(self, email):
        self.email = email

class State(rx.State):
    user: User = User(email="foo@gg.com", first_name="A", last_name="B")

    def set_user_email(self, mail):
        # Previous implementation, which does work:
        # self.user.email = mail
        self.user.set_email(mail)

@rx.page(route="/")
def idx():
    return rx.vstack(
        rx.text(f"Welcome {State.user.email}"),
        rx.input(value=State.user.email, on_change=State.set_user_email),
    )
```

**Expected behavior**
The page re-renders after calling the mutating method on the custom var

 - Python Version: 3.11
 - Reflex Version: 0.3.4
 - OS: MacOS 12.6.8
 
**Additional context**
Unclear if this is actually a "bug"; the documentation doesn't explicitly say this isn't allowed, but also doesn't show an explicit example of this.

My 2 cents: The proposed behavior would be an intuitive way to decompose logic amongst multiple models, and particularly useful as apps get more complex. Without it, the `State` class will bloat with detailed logic mutating `rx.Base` objects directly. (And for newer Reflex developers like me, leave them confused why the page isn't rendering after the state has been mutated, before they track down the issue.)

Moreover, my initial sense is that being able to do `on_change=State.user.set_email`, instead of needing an interim `State.set_user_email`, seems even better.

<sub>[REF-1356](https://linear.app/reflex-dev/issue/REF-1356/updates-custom-var-with-collection-does-not-trigger-re-render)</sub>

masenf:
When a helper method manipulates fields on a Base instance that is associated with a state, recursively wrap the helper method with the MutableProxy so it also tracks changes.

If the helper method returns a mutable object, this object will also be tied back to the same field on the state to allow for "get_*" type functions that still track changes.

Fix #2239


reflex/state.py
```diff
@@ -2836,6 +2836,10 @@ class MutableProxy(wrapt.ObjectProxy):
         ]
     )
 
+    __never_wrap_base_attrs__ = set(Base.__dict__) - {"set"} | set(
```
picklelo, line 2839: Can we add a comment on what this means

reflex/state.py
```diff
@@ -2885,7 +2889,9 @@ def _wrap_recursive(self, value: Any) -> Any:
         Returns:
             The wrapped value.
         """
-        if isinstance(value, self.__mutable_types__):
+        if isinstance(value, self.__mutable_types__) and not isinstance(
```
picklelo, line 2892: Also can we add comments here

masenf: Fixed both

picklelo (approved):

PR #2246, opened 2023-12-01
[REF-1365] Radix + Tailwind Compatibility (and allow tailwind to be completely disabled)

masenf:
A recent refactoring in App.compile started
unconditionally compiling tailwind, even when it is explicitly disabled by the user via tailwind=None
in rxconfig.

This patch will now fix REF-1365 or allows apps to correctly disable tailwind


picklelo (approved):

masenf:
I moved this back to draft because not compiling tailwind seems to cause problems in other parts of the app. Need a unit test for disabling tailwind so we don't regress this functionality. Will take out of draft when a test can be added.

picklelo (approved):

PR #2255, opened 2023-12-04
[REF-1404] router_data not updated when processing events

masenf:
always clean dirty vars _after_ getting delta, never before

(previously this didn't matter, because hydrate would always send the diff before any events were processed).


picklelo (approved):
Works for me now, nice fix

PR #2267, opened 2023-12-06
use redis-py url syntax for redis_url

### LINKED ISSUE #2253: Feature request: Redis Auth
I would like to configure Redis Authentication with reflex. Currently reflex only allows to set redis host and port: https://github.com/reflex-dev/reflex/blob/eb79da853834eae0e2c027c90c8b70688369d6e6/reflex/utils/prerequisites.py#L146

Maybe reflex can use https://redis-py.readthedocs.io/en/stable/connections.html#redis.Redis.from_url which allows urls with passwords and db's other than the hardcoded db 0.

benedikt-bartscher:
Closes #2253

### DIFF
--- a/.github/workflows/integration_app_harness.yml
+++ b/.github/workflows/integration_app_harness.yml
@@ -45,7 +45,7 @@ jobs:
     - name: Run app harness tests
       env:
         SCREENSHOT_DIR: /tmp/screenshots
-        REDIS_URL: ${{ matrix.state_manager == 'redis' && 'localhost:6379' || '' }}
+        REDIS_URL: ${{ matrix.state_manager == 'redis' && 'redis://localhost:6379' || '' }}
       run: |
         poetry run pytest integration
     - uses: actions/upload-artifact@v3
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -70,6 +70,6 @@ jobs:
       if: ${{ matrix.os == 'ubuntu-latest' }}
       run: |
         export PYTHONUNBUFFERED=1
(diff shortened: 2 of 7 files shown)


masenf (commented):
Not sure why the redis integration tests are partly failing now, will need to investigate... but the code looks good. Thanks for your contribution!

masenf:
I wonder if there's something to do with that db=0 parameter in the old code, not sure what that does.

benedikt-bartscher:
Where do you see the `db=0` parameter? It should only select the redis database with index 0 which is the default value anyway iirc.

masenf:
i did some more testing on this PR and the difference appears to be the `from_url` method passes in a `ConnectionPool` object to the `Redis` constructor. This changes the behavior of the `auto_close_connection_pool` parameter:

https://github.com/redis/redis-py/blob/714adf928381cae4eb8ab2d9aa7ced93f3d14a4a/redis/asyncio/client.py#L192-L198

I haven't actually traced out a possible fix yet, but i believe this is the source of the discrepancy.

masenf:
`Redis.close` has a keyword argument to force close the connection pool, which makes the integration tests pass. So I think we're good to bring this in if the tests are passing on here (except benchmark, that one just wont work)

benedikt-bartscher:
Thanks @masenf i did not find time to dig further into this.

I was not expecting that using `from_url` changes anything regarding connection pools. But it seems like redis-py internally passes the connection url to `ConnectionPool.from_url`.
However i do not understand why the pool isn't closed automatically if the client is closed, cause `from_url` sets this option on the client: `client.auto_close_connection_pool = True`

masenf (approved):

PR #2271, opened 2023-12-07
Warning if newer reflex/reflex-hosting-cli available 

Alek99:
Add check_latest_package_version when given a package this checks if there is a newer python package and throws a warning to upgrades. Used in run and init to check for reflex and reflex-hosting-cli in deploy

Screenshot 2023-12-06 at 5 53 13 PM

### DIFF
--- a/reflex/reflex.py
+++ b/reflex/reflex.py
@@ -15,7 +15,7 @@ from reflex_cli.utils import dependency
 
 from reflex import constants
 from reflex.config import get_config
-from reflex.utils import console, telemetry, prerequisites
+from reflex.utils import console, telemetry
 
 # Disable typer+rich integration for help panels
 typer.core.rich = False  # type: ignore
@@ -78,7 +78,7 @@ def _init(
     app_name = prerequisites.get_default_app_name() if name is None else name
     console.rule(f"[bold]Initializing {app_name}")
 
-    prerequisites.check_latest_package_version("reflex")    
+    prerequisites.check_latest_package_version("reflex")
 
     # Set up the web project.
     prerequisites.initialize_frontend_dependencies()
@@ -173,7 +173,7 @@ def _run(
 
     console.rule("[bold]Starting Reflex App")
 
-    prerequisites.check_latest_package_version("reflex")  
+    prerequisites.check_latest_package_version("reflex")
 
     if frontend:
         prerequisites.update_next_config()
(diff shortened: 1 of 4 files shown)


reflex/reflex.py
```diff
@@ -78,6 +78,8 @@ def _init(
     app_name = prerequisites.get_default_app_name() if name is None else name
     console.rule(f"[bold]Initializing {app_name}")
 
+    prerequisites.check_latest_package_version("reflex")
```
picklelo, line 81: Instead of "reflex" can use `constants.Reflex.MODULE_NAME` (below also). Maybe we can make a new constant for the "reflex-hosting-cli" within there also

reflex/utils/prerequisites.py
```diff
@@ -29,6 +29,26 @@
...
 
+def check_latest_package_version(package_name: str):
+    """Check if the latest version of the package is installed.
+
+    Args:
+        package_name: The name of the package.
+    """
+    url = f"https://pypi.org/pypi/{package_name}/json"
+
+    response = httpx.get(url)
```
picklelo, line 40: We may want to put this in a try/except so that they can still run their app if not connected to the internet

picklelo (approved):
nice

reflex/utils/prerequisites.py
```diff
@@ -29,6 +29,29 @@
...
+        package_name: The name of the package.
+    """
+    try:
+        url = f"https://pypi.org/pypi/{package_name}/json"
+
+        response = httpx.get(url)
+        data = response.json()
+
+        latest_version = data["info"]["version"]
+        current_version = version.parse(constants.Reflex.VERSION)
```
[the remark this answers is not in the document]
Alek99: Good catch will fix

picklelo (approved):
works well for me

reflex/utils/prerequisites.py
```diff
@@ -29,6 +30,26 @@
...
+        current_version = pkg_resources.get_distribution(package_name).version
+        url = f"https://pypi.org/pypi/{package_name}/json"
+        response = httpx.get(url)
+        latest_version = response.json()["info"]["version"]
+        if version.parse(current_version) < version.parse(latest_version):
+            console.warn(
+                f"Your version ({current_version}) of {package_name} is out of date. Upgrade to {latest_version} with 'pip install {package_name} --upgrade'"
+            )
+    except Exception:
+        pass
```
martinxu9, line 50: nit: there's a `contextlib.suppress` if the intent here is we do not care about any exceptions.

martinxu9 (approved):

PR #2279, opened 2023-12-11
color_scheme props added and updated pyi files

Lendemor:
Use `color_scheme` as an alias for the radix `color` props.

Allow mapping `color` to the default CSS property instead.


scripts/pyi_generator.py
```diff
@@ -95,6 +95,10 @@ def _get_type_hint(value, type_hint_globals, is_optional=True) -> str:
                 res = ", ".join(types)
                 res = f"Union[{res}]"
     elif isinstance(value, str):
+        if value == "Optional[str]":
+            from typing import Optional
+
+            type_hint_globals["Optional"] = Optional
```
masenf, line 101: i think we should leverage the `DEFAULT_TYPING_IMPORTS` set for this, so that way it's extensible if we need to add more and not such a special case.

technically all `pyi` files will already get `Optional` as an import, so really those `default_typing_imports` that get passed in should be part of the type_hint_globals already.

masenf (changes requested):

scripts/pyi_generator.py
Lendemor: I tried to add it there, but it didn't work.
Let me know if you manage to make it work with `DEFAULT_TYPING_IMPORTS`

masenf (approved):

PR #2284, opened 2023-12-13
[REF-1464] Handle requirements.txt encoding

martinxu9:
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

1. Previously the handling of `requirements.txt` in user local environment assumes `utf-8` the default encoding used in builtin open method. If user has a different encoding, we open and append will cause the file to have inconsistent encoding on different lines and make part of the content unreadable using any given encoding
2. Add UT to cover the non utf-8 files.
3. For hosting: now that we re-check/re-generate requirements.txt when user deploys, it is safe to simply pin the reflex version instead of >=.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,6 +58,7 @@ wrapt = [
 packaging = "^23.1"
 pipdeptree = "^2.13.0"
 reflex-hosting-cli = ">=0.1.2"
+charset-normalizer = "^3.3.2"
 
 [tool.poetry.group.dev.dependencies]
 pytest = "^7.1.2"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,7 +83,6 @@ pre-commit = {version = "^3.2.1", python = ">=3.8,<4.0"}
 selenium = "^4.11.0"
 types-tabulate = "^0.9.0.3"
 pytest-benchmark = "^4.0.0"
-sqlalchemy = "^2.0.23"
 
 [tool.poetry.scripts]
 reflex = "reflex.reflex:cli"
--- a/poetry.lock
+++ b/poetry.lock
@@ -220,6 +220,105 @@ files = [
 ]
 
 [[package]]
+name = "charset-normalizer"
+version = "3.3.2"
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -58,6 +58,7 @@ wrapt = [
 packaging = "^23.1"
 pipdeptree = "^2.13.0"
 reflex-hosting-cli = ">=0.1.2"
+charset-normalizer = "^3.3.2"
```
masenf, line 61: just out of curiosity, was `chardet` considered for this? it's what i've mostly seen used in the past, but i see it hasn't had a release for quite some time.

masenf (dismissed):

pyproject.toml
martinxu9: Not really, I didn't look into that. I saw some online answers using this, and this seems to be a dependency of some of current packages already.
But I will double check what people say about those two before merging.

masenf: oh nice, charset-normalizer _already_ being in our installed payload makes this an easy decision

martinxu9: Oh ok. Let's use this package for now. It's in a rather tiny area of our code

pyproject.toml
```diff
@@ -82,6 +83,7 @@ pre-commit = {version = "^3.2.1", python = ">=3.8,<4.0"}
 selenium = "^4.11.0"
 types-tabulate = "^0.9.0.3"
 pytest-benchmark = "^4.0.0"
+sqlalchemy = "^2.0.23"
```
martinxu9, line 86: got a pyright error about connection not having a commit() function, but this makes it go away

martinxu9: removing it seems to not cause a problem. I guess local system pre-commit probably doesn't sync up with poetry?

martinxu9:
@masenf thumbs up one more time please

masenf (approved):
sorry, thought this already merged

martinxu9:
> sorry, thought this already merged

No worries. I think our branch is set up to dismiss approvals if there's any new changes. I needed to merge upstream

PR #2285, opened 2023-12-13
Add high-level API for accordion

picklelo:
* Add high level Accordion API
* Handle `apply_theme` within `rx.foreach`

Test code:

```python
from rxconfig import config

import reflex as rx
import reflex.components.radix.primitives as rdxp



class State(rx.State):
    """The app state."""

    items: list[tuple[str, str]] = [
        ("Is it accessible?", "Yes. It's built with accessibility in mind."),
        ("Is it unstyled?", "Yes. It's unstyled by default, giving you freedom over the look and feel."),
        ("Is it finished?", "It's still in beta, but it's ready to use in production."),
    ]

def index() -> rx.Component:
    return rdxt.theme(
        rdxt.container(
            rdxt.flex(
                rdxp.accordion(
                    rdxp.accordion_item("Is it accessible?", rdxt.button("Test button"), disabled=True, font_size="3em"),
                    rdxp.accordion_item("Is it unstyled?", "Yes. It's unstyled by default, giving you freedom over the look and feel."),
                    rdxp.accordion_item("Is it finished?", "It's still in beta, but it's ready to use in production."),
                    type="single",
                    collapsible=True,
                    width="300px",
                ),
                rdxp.accordion(
                    rx.foreach(
                        State.items,
                        lambda item, i: rdxp.accordion_item(item[0], item[1], value=f"item-{i}"),
                    ),
                    type="single",
                    collapsible=True,
                    width="300px",
                ),
            ),
            padding_y="5em",
            font_size="2em",
            text_align="center",
        )
    )


app = rx.App(
    theme=rdxt.theme(
        appearance="dark", has_background=True, radius="none", accent_color="crimson"
    ),
)

app.add_page(index)
app.compile()
```

### DIFF
--- a/reflex/components/layout/foreach.py
+++ b/reflex/components/layout/foreach.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 
 import inspect
 from hashlib import md5
-from typing import Any, Callable, Iterable
+from typing import Any, Callable, Iterable, Optional
 
 from reflex.components.component import Component
 from reflex.components.layout.fragment import Fragment
@@ -22,6 +22,17 @@ class Foreach(Component):
 
     # A function from the render args to the component.
     render_fn: Callable = Fragment.create
+
+    # The theme if set.
+    theme: Optional[Component] = None
+
+    def _apply_theme(self, theme: Component):
+        """Apply the theme to this component.
+
+        Args:
+            theme: The theme to apply.
+        """
+        self.theme = theme
 
     @classmethod
     def create(cls, iterable: Var[Iterable], render_fn: Callable, **props) -> Foreach:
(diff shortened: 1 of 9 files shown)


Alek99:
Api looks nice but how would one use the low level accordion items when we override it here with the high level accordion_item

Lendemor (changes requested):
Need an example where the primitives render properly even if we are not using any theme.

picklelo:
> Api looks nice! How would one use the low-level accordion items when we override them here with the high-level accordion_item
> 
> I guess the broader question is will the low and high level share the same namespace and how to accommodate this if so

Discussed offline - but we're thinking that we shouldn't let the user combine high/low level APIs. Not sure if we should detect this (maybe we can through the `valid_children`) but we can at least document that they shouldn't be mixed.

The high level will be in the `rx.` namespace and the low-level will be in the `rx.radix` namespace so we can have a general rule of thumb to not mix components from different namespaces.

Alek99 (approved):

PR #2289, opened 2023-12-13
add radix primitive for slider

Lendemor:
implementation of `@radix-ui/react-slider` 

also add default base class for all radix primitives.

```python
class State(rx.State):
    vals: list[int] = [0,0]

rdxp.slider(
    default_value=[0, 0]
    value=State.vals,
    on_value_change=State.handler_change,
),        
```


reflex/app.py
```diff
@@ -667,8 +667,8 @@ def mark_complete(_=None):
                 # Merge the component style with the app style.
                 component.add_style(self.style)
 
-                if self.theme is not None:
-                    component.apply_theme(self.theme)
+                # if self.theme is not None:
```
masenf, line 670:
```suggestion
```

reflex/components/radix/primitives/base.py
```diff
@@ -0,0 +1,23 @@
...
+    # Change the default rendered element for the one passed as a child.
+    as_child: Var[bool]
+
+    def _render(self) -> Tag:
+        return (
+            super()
+            ._render()
+            .add_props(
+                **{
+                    "class_name": format.to_title_case(self.tag or ""),
```
masenf, line 20: i think the way this is written does not let the user pass their own `class_name` to the component.

probably we should have some code above like

```python
component_class = format.to_title_case(self.alias or self.tag or "")
class_name = (self.class_name or "") + component_class
```

and then call `add_props` with _that_ `class_name`

reflex/components/radix/primitives/slider.py
```diff
@@ -0,0 +1,170 @@
...
+
+    disabled: Var[bool]
+
+    orientation: Var[LiteralSliderOrientation]
+
+    dir: Var[LiteralSliderDir]
+
+    inverted: Var[bool]
+
+    min: Var[int]
```
masenf, line 40: in the chakra components we have added underscore suffix to names that shadow python builtins, will we do the same here?

personally i think _NOT_ changing the names is better, as long as they will never be used as unqualified names, then it's better to have consistency with the upstream component, and the shadowing won't actually matter (it also looks better than having a bunch of underscore suffixes)

reflex/components/radix/primitives/slider.py
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+    def _apply_theme(self, theme: Component):
+        self.style = Style(
+            {
+                "position": "relative",
+                "display": "flex",
+                "align-items": "center",
+                "user-select": "none",
+                "touch-action": "none",
```
masenf, line 67: for consistency with other parts of reflex, should we use snake case here?

```suggestion
                "align_items": "center",
                "user_select": "none",
                "touch_action": "none",
```

masenf (changes requested):
need to import these names in reflex.components.radix.primitives to make them visible

do you have some sample code? i can't seem to get it working. it says "TypeError: values.map is not a function This error happened while generating the page. Any console logs will be displayed in the terminal window."


maybe `value` should take a `list` of `int` like `default_value` does?

reflex/components/radix/primitives/base.py
Lendemor: I got this code from nihkil PR on the accordion, but I'm not even sure we need the component_class in the end.
Need to verify this.

reflex/components/radix/primitives/slider.py
Lendemor: We need to discuss with the team to decide on this yeah.

reflex/components/component.py
```diff
@@ -446,25 +446,24 @@ def __str__(self) -> str:
 
         return _compile_component(self)
 
-    def _apply_theme(self, theme: Component):
+    def _apply_theme(self, theme: Optional[Component]):
```
picklelo, line 449: I think this should not be Optional, or else we will have to do `None` checks within every component's `_apply_theme` method.

reflex/components/component.py
```diff
@@ -446,25 +446,24 @@ def __str__(self) -> str:
...
         """Apply the theme to this component.
 
         Args:
             theme: The theme to apply.
         """
         pass
 
-    def apply_theme(self, theme: Component):
+    def apply_theme(self, theme: Optional[Component]):
```
picklelo, line 457: Maybe just do the `None` check here and don't recurse if it's none. Alternatively, can we keep it where it was above?

picklelo (changes requested):

reflex/components/component.py
Lendemor: This PR need it because apply_theme was never called at all when theme=None otherwise.

masenf:
did this already merge? looks like the code is already in main

PR #2291, opened 2023-12-14
apps should no longer call "app.compile()"

### LINKED ISSUE #2293: Remove `app.compile()` boilerplate line in apps.
**Describe the bug**
Reflex apps require `app.compile()` line.  It would be better to not require that.

**To Reproduce**
Any existing reflex app in the wild.

**Expected behavior**
All existing reflex functionality works as usual, without `app.compile()` in apps.

jackie-pc:
Closes #2293.

Two main tricks:
* To handle backend running, `gunicorn` or `uvicorn` to load a special shim `app_module_for_backend.py`, instead of the real app (e.g. `counter.py`).
* A new `get_compiled_app` function, which essentially does `get_app() + compile` - to account for the fact that `get_app()` will now return an app module with an **uncompiled** app.  We update existing call sites of `get_app()` with `get_compiled_app` accordingly, with some exceptions (when we really don't want to compile).

If we merge, release this, we need to clean up reflex-web and reflex-examples.


reflex/app.py
```diff
@@ -619,6 +619,24 @@ def _should_compile(self) -> bool:
...
+        Existing call sites will get a deprecation message.
+
+        We will eventually remove app.compile() method entirely.
+        """
+        console.deprecate(
+            feature_name="app.compile()",
+            reason="Explicit calls to app.compile() are not needed."
+            " Method will be removed in 0.3.9",
+            deprecation_version="0.3.7",
+            removal_version="0.3.9",
```
Lendemor, line 635: Mark the removal to 0.4.0 to sync with other planned removals.

reflex/app.py
```diff
@@ -619,6 +619,24 @@ def _should_compile(self) -> bool:
...
+        Reflex framework will perform compile_() as needed.
+
+        Existing call sites will get a deprecation message.
+
+        We will eventually remove app.compile() method entirely.
+        """
+        console.deprecate(
+            feature_name="app.compile()",
+            reason="Explicit calls to app.compile() are not needed."
+            " Method will be removed in 0.4.0",
```
Lendemor, line 633: This line is not needed, `removal_version` already add this part of the message.

picklelo (approved):
Works well for me, will work on updating the main website / examples before we cut this

Issue #2282, opened 2023-12-12
Running Reflex via systemd with --env prod param only runs the frontend and not the backend.


rushout09:
**Describe the bug**
in systemd: 
env DEV work
env PROD fails

running manually : 
both works

**To Reproduce**

Create a systemd file
```
[Unit]
Description=Databees Website App
After=network.target

[Service]
ExecStart=/home/ubuntu/databees_website/venv/bin/reflex run --env prod
WorkingDirectory=/home/ubuntu/databees_website
User=ubuntu
Group=ubuntu
Restart=always

[Install]
WantedBy=multi-user.target
```

sudo systemctl daemon-reload
sudo systemctl start test.service

**Expected behavior**
ubuntu@ip-172-31-41-251:~$ sudo lsof -i -P -n | grep LISTE
systemd-r  325 systemd-resolve   14u  IPv4  15785      0t0  TCP 127.0.0.53:53 (LISTEN)
sshd       565            root    3u  IPv4  17438      0t0  TCP *:22 (LISTEN)
sshd       565            root    4u  IPv6  17449      0t0  TCP *:22 (LISTEN)
reflex    6894          ubuntu   13u  IPv4  68388      0t0  TCP *:8000 (LISTEN)
python3   6927          ubuntu   13u  IPv4  68388      0t0  TCP *:8000 (LISTEN)
next-serv 6957          ubuntu   20u  IPv6  69656      0t0  TCP *:3000 (LISTEN)

**Screenshots**

**Specifics (please complete the following information):**
 - Python Version: 3.10.8
 - Reflex Version:
reflex==0.3.6
reflex-hosting-cli==0.1.3
 - OS: ubuntu 22
 - Browser (Optional):
 
**Additional context**
https://discord.com/channels/1029853095527727165/1184109557946003576

jackie-pc:
Thanks for the report.

When backend prod runs, it calls out to the program `gunicorn`.  However gunicorn program is only available once the virtual env is activated by `source venv/bin/activate`.  Simply calling `venv/bin/reflex` is insufficient as that takes care only of libraries within the Python process.

This exposes a bug in the way to propagate errors though - it should have been straightforward from logs ("gunicorn not found").   #2296  above will address this.

rushout09:
@jackie-pc is there a way to properly initialize venv using systemd?

jackie-pc:
You would initialize like how you normally would.  `source ...../activate`.  E.g. 

[...]

PR #2296, opened 2023-12-14
capture exceptions thrown by any process in run_concurrently_context

jackie-pc:
Related to https://github.com/reflex-dev/reflex/issues/2282

In that issue, the prod backend failed to come up, very likely due to `gunicorn` program not being available (because venv not initialized properly).

However logs are silent.  This is because when `task.result()` raises an exception, we try to exit the `with ... as executor:` context, which does a `shutdown(wait=True)` under the covers.  Because the `run_frontend_prod` task is still running fine, we never manage to exit that context, and therefore the exception never makes it further up the stack.

Here we do a manual `try ... finally` block instead of context manager.  We do a `shutdown(wait=False)` instead.  Note this is a logic change... if any concurrent task raises, `run_concurrently_context` will raise.  I think this is more expected, based on current usage (e.g. to install node and bun concurrently, and to run backend/frontend concurrently.).  The improvement here is only that the logs will show the raised exception now.

As before if any task within the thread pool does not terminate, then that will hold up the Python interpreter from going down.  One more Ctrl+C will do it.  This is existing behavior that is unchanged (we just see any exceptions now).

### DIFF
--- a/reflex/utils/processes.py
+++ b/reflex/utils/processes.py
@@ -171,7 +171,9 @@ def run_concurrently_context(
     fns = [fn if isinstance(fn, tuple) else (fn,) for fn in fns]  # type: ignore
 
     # Run the functions concurrently.
-    with futures.ThreadPoolExecutor(max_workers=len(fns)) as executor:
+    executor = None
+    try:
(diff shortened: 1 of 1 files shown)


masenf (approved):
looks like a good fix to me

PR #2297, opened 2023-12-14
High level api graphing

tgberkeley:
Example of API:

```
from reflex.components.graphing.recharts import bar_chart_high_api

def index():
    return rx.vstack(
        bar_chart_high_api(data=chart_data, x="col1", y=["col2", "col3", "col4"], color=["#FF0000", "#CC0000", "AA0000"], legend=True, tooltip=True, cartesian_axis=True, x_axis_brush=True),
        height="30em",
    ) 
```

Have used a lot of functionality to allow the function to take in a `pd.DataFrame | np.ndarray | Dict[str, List[str | int]] | List[Dict[str, Any]]`

### DIFF
--- a/reflex/components/graphing/recharts/utils.py
+++ b/reflex/components/graphing/recharts/utils.py
@@ -354,6 +354,11 @@ def bar_chart_high_api(
     # must write a test here to make sure that list_of_dicts is of type List[Dict[str, Any]]
 
     return rx.recharts.bar_chart(
+        rx.cond(
+            tooltip,
+            rx.recharts.graphing_tooltip(),
+        ),
+        
         *[rx.recharts.bar(data_key=key[0], stroke=key[1], fill=key[1])
         for key in y_and_color
         ],
@@ -365,10 +370,7 @@ def bar_chart_high_api(
             rx.recharts.legend(),
         ),
 
-        rx.cond(
-            tooltip,
-            rx.recharts.graphing_tooltip(),
-        ),
+        
 
         rx.cond(
             cartesian_axis,
--- a/reflex/components/graphing/recharts/utils.py
+++ b/reflex/components/graphing/recharts/utils.py
@@ -1,0 +1,387 @@
+import reflex as rx
+from pandas import DataFrame
+import pandas as pd
+import numpy as np
+import re
+import contextlib
+from typing import Any, Dict, List, Sequence, Tuple, cast, Union
+
+from reflex.components.component import Component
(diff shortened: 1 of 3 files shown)


reflex/components/graphing/recharts/utils.py
```diff
@@ -0,0 +1,389 @@
...
+    return y_column_list
+
+
+def _maybe_reset_index_in_place(
+    df: pd.DataFrame, x_column: str | None, y_column_list: List[str]
+) -> str | None:
+    if x_column is None and len(y_column_list) > 0:
+        if df.index.name is None:
+            # Pick column name that is unlikely to collide with user-given names.
+            x_column = "index--p5bJXXpQgv"
```
masenf, line 159: seems like this should be dynamic, perhaps using `rx.vars.get_unique_variable_name`, or at least pulled out to a top level variable for reuse without repetition... however, if this will end up in the compiled code, it's probably better to have a stable value that doesn't change on every compile. In other parts of the code, i hash the output object/code/whatever and use the hash as the identifier.

reflex/components/graphing/recharts/utils.py
```diff
@@ -0,0 +1,389 @@
...
+    melted_data, y_column = _maybe_melt(selected_data, x_column, y_column_list)
+
+    # Return the data, but also the new names to use for x, y, and color.
+    return melted_data, x_column, y_column
+
+
+def prepare_y_and_color(y, color):
+    
+    # Set default color if color is None
+    default_color = "#8884d8"
```
masenf, line 295: this should be at the top level, so it can be more easily reused where needed.

reflex/components/graphing/recharts/utils.py
```diff
@@ -0,0 +1,389 @@
...
+        return zip(y, color)
+
+    # If lists are of different lengths, raise an error
+    else:
+        raise ValueError("y and color must be of the same length")
+    
+
+
+def bar_chart_high_api(
+        data: pd.DataFrame | np.ndarray | Dict[str, List[str | int]] | List[Dict[str, Any]],
```
masenf, line 316: the fact that this doesn't work with state seems like a big incongruity in the API.

it seems like it mostly could work with state as long as the data is in the right format, except for `convert_anything_to_dataframe` really wanting a python value, not a Var

masenf (changes requested):
docstrings should use the google style like we have elsewhere throughout the library

PR #2306, opened 2023-12-18
Lendemor/radix form primitive

Lendemor:
```python
rdxp.form_root(
    rdxp.form_field(
        rdxp.form_label("Email: ", style={"margin-right": "10px"}),
        rdxp.form_control(
            rx.el.input(type="email", required=True, name="email" ),
            as_child=True,
        ),
        rdxp.form_message("Value is missing", match="valueMissing"),
        rdxp.form_message("Value is wrong type", match="typeMismatch"),
    ),
    rdxp.form_field(
        rdxp.form_label("Age :", style={"margin-right": "10px"}),
        rdxp.form_control(
            rx.el.input(type="number", min=13, name="age"),
            as_child=True,
        ),
        rdxp.form_message("Too young", match="rangeUnderflow"),
    ),
    rdxp.form_submit(
        rx.el.button("Submit", type="submit"),
        as_child=True,
    ),
),
```


picklelo (approved):
Tested all the components and they work well for me. I think we don't need the slider for this initial version since it is in the Radix themes already: https://www.radix-ui.com/themes/docs/components/slider which we somehow missed. But now we should have all the components we need wrapped

PR #2314, opened 2023-12-19
Add support for custom components starter

martinxu9:
### Summary
1. Add support for user contributed custom component starter template. Design doc here: https://www.notion.so/reflex-dev/3rd-Party-Component-Ecosystem-live-9d35d4a9d36245db969dd714522f4776?pvs=4
2. Add a new sub-command series named `component`
3. `reflex component init`: use the enclosing folder name to figure out "library name". We use variations of this for different purposes, e.g.: class name would be the Title case of those parts/words. See design doc for more details. As a result, we create package config files (including pyproject.toml), a folder `custom_components` with template py files prefilled with wrapping react code/tutorials, a folder `<new_componet>_demo` a reflex app with imports of the newly defined component. Also install the package locally in editable mode for convenience of dev.
4. `reflex component build`: this builds the distribution files `dist/*`.
5. `reflex component publish --token <your python package repo API token> -r [pypi|testpypi]`
### Tests
- [x] create works
- [x] build/publish works, published `reflex-google-auth` on test pypi
- [x] verify the package published is functional


reflex/.templates/jinja/custom_components/pyproject.toml.jinja2
```diff
@@ -0,0 +1,35 @@
+[build-system]
```
martinxu9, line 1: The general folder has limitation in certain ways using poetry. E.g., if we create new_component folder in reflex/examples and just the reflex in poetry environment, when we are in the new_component folder, it detects the pyproject.toml but no poetry section, it'll complain. Could just use `poetry shell` instead. Our user likely wont use it like this, so hopefully not interference with their choice of package manager.

reflex/.templates/jinja/custom_components/src.py.jinja2
```diff
@@ -0,0 +1,59 @@
...
+class {{ component_class_name }}(rx.Component):
+    """{{ component_class_name }} component."""
+
+    # The React library to wrap.
+    library = "Fill-Me"
+
+    # The React component tag.
+    tag = "Fill-Me"
+
+    # If the tag is the default export from the module, you can set is_default = True.
```
tgberkeley, line 26: "you can set" should change to "you must set" as otherwise it throws a server side error

tgberkeley:
add a small comment that we must run the `reflex component build` command from the base folder, where the pyproject.toml file is

reflex/.templates/jinja/custom_components/README.md.jinja2
```diff
@@ -0,0 +1,3 @@
+# {{ module_name }}
```
picklelo, line 1: Let's include the example app in the README as well like gradio: https://huggingface.co/spaces/freddyaboulton/gradio_calendar

We can include the installation code as well as the demo app code. Perhaps if we want to be smart, we can auto-generate it on `reflex component build`, or we can just let them do it. But should have an installation / usage examples.

We should also list the props in the readme, but they may have to do that manually. There's way to auto-generate it similar to how we do on our main site, so we could do that also..

tgberkeley:
maybe have a few quick lines on how to set up pypi, as not super obvious

picklelo (commented):
After publishing, we should be able to import the component as `from reflex_component import component` (rather than having to do `from reflex_component.component import component` - pretty much move it up to the __init__.py also

martinxu9:
> add a small comment that we must run the `reflex component build` command from the base folder, where the pyproject.toml file is

Sounds good. I'm adding that into the command help menu. This applies to the publish command as well.

martinxu9:
> maybe have a few quick lines on how to set up pypi, as not super obvious

I think we'll add this to the docs on reflex-web together with the tutorial how to run these new commands.

pyproject.toml
```diff
@@ -59,6 +59,9 @@ packaging = "^23.1"
 pipdeptree = "^2.13.0"
 reflex-hosting-cli = ">=0.1.2"
 charset-normalizer = "^3.3.2"
+wheel = "^0.42.0"
```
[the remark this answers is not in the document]
martinxu9: This is required for the build backend. As discussed, this is also already in the install payload. I would prefer to explicitly include this in case other packages depending on wheel get removed and we'll get unexpected failure.

reflex/.templates/jinja/custom_components/README.md.jinja2
picklelo: Maybe the only thing we need to add is the `pip install` command, agreed everything else would be a pain to auto-generate

reflex/constants/custom_components.py
```diff
@@ -0,0 +1,30 @@
...
+from __future__ import annotations
+
+from types import SimpleNamespace
+
+
+class CustomComponents(SimpleNamespace):
+    """Constants for the custom components."""
+
+    # The name of the custom components source directory.
+    SRC_DIR = "custom_components"
```
picklelo, line 12: Maybe lets make the src_dir match the module name

picklelo: never mind, this is the top level

picklelo (approved):

PR #2315, opened 2023-12-20
Refactor chakra components into separate folder

picklelo:
This change moves most of the core components into the `reflex/components/chakra` namespace, in order to prepare for the switch to the Radix components.

All imports for end-users should work exactly as is, this is just an internal restructuring.


reflex/components/chakra/forms/debounce.py
Lendemor, line 1: Debounce is not chakra specific, it can probably go to `core` or it's own folder.

reflex/components/chakra/forms/password.py
Lendemor, line 1: since `Password` doesn't actually exist in Chakra, this should probably become a high level component only, that specialize an input with `type="password"`

reflex/components/chakra/layout/fragment.pyi
Lendemor, line 1: Fragment moved to core, this pyi should be removed.

reflex/components/chakra/navigation/nextlink.pyi
Lendemor, line 1: should be removed

reflex/components/chakra/overlay/banner.pyi
Lendemor, line 1: should be removed

reflex/components/radix/themes/components/textarea.pyi
```diff
@@ -9,8 +9,8 @@ from reflex.event import EventChain, EventHandler, EventSpec
 from reflex.style import Style
 from typing import Any, Dict, Literal
 from reflex import el
+from reflex.components.chakra.forms.debounce import DebounceInput
```
Lendemor, line 12: fix those imports when debounce is moved to `core`

Lendemor (changes requested):
A few changes needed but I'm glad this PR is finally here 😄

reflex/components/chakra/forms/debounce.py
picklelo: Moving this

reflex/components/chakra/forms/password.py
picklelo: Agreed, I think these ones are tied to the Chakra Input component, maybe we can leave it for now and abstract it later.

requirements.txt
masenf, line 1: why is this here? and why is it considered a binary file?

picklelo: Removed and added go gitignore (I think it comes from running the unit tests for me, we should probably see why)

reflex/components/chakra/forms/multiselect.py
masenf, line 1: why no `pyi` file for this component?

masenf (dismissed):

reflex/components/chakra/forms/multiselect.py
picklelo: I'm not sure, I ran the generator

Lendemor (approved):

reflex/components/chakra/forms/multiselect.py
Lendemor: multiselect.py is black-listed in the generator, probably a remnant of when it was broken ?

picklelo: We can bring it back now I suppose

PR #2318, opened 2023-12-20
rx.match component

ElijahAhianyo:
nested `rx.cond` can get very intricate and messy. This PR introduces `rx.match` to fix this problem. 

`rx.match` takes in a condition which should be a var, match case tuples and a non tuple argument that will be treated as the default or catchall case.
 
```python
import reflex as rx

class State(rx.State):
    value: int = 0

def index():
    return rx.vstack(
        rx.match(
            State.value,
            (1, rx.vstack(rx.text('first value'))),
            (2, rx.text("second value")),
            rx.text("default value")
        )
    )
```
This renders as(formatted manually) :

```javascript

/** @jsxImportSource @emotion/react */import { Fragment, useContext } from "react"
import { Fragment_fd0e7cb8f9fb4669a6805377d925fba0 } from "/utils/stateful_components"
import { Text, VStack } from "@chakra-ui/react"
import { StateContexts } from "/utils/context"
import "focus-visible/dist/focus-visible"
import NextHead from "next/head"



  const state__state = useContext(StateContexts.state__state)

  return (
    <Fragment>
      <Fragment_fd0e7cb8f9fb4669a6805377d925fba0 />
      <VStack>
        <Fragment>
          {
            (() => {
              switch (state__state.value) {
                case 1:
                  return <VStack>
                    <Text>
                      {`first value`}
                    </Text>
                  </VStack>;
                  break;
                case 2:
                  return <Text>
                    {`second value`}
                  </Text>;
                  break;
                default:
                  return <Text>
                    {`default value`}
                  </Text>;
                  break;
              }
            })()
          }
        </Fragment>
      </VStack>
      <NextHead>
        <title>
          {`Reflex App`}
        </title>
        <meta content={`A Reflex app.`} name={`description`} />
        <meta content={`favicon.ico`} property={`og:image`} />
      </NextHead>
    </Fragment>
  )
}


```

## Match cases
These are tuples containing cases to be matched as well as their return values. The first `n-1` values are treated as the cases to be matched and the last value in the tuple is treated as the return value.
 Note that, a match case tuple MUST have a minimum of 2 elements (the condition and the return value).

## Default/catchall case
A default case is a non-tuple argument that is matched when none of the match cases are matched. Below are the caveats to know about the default case:
- The default case MUST be the last argument(the app will throw a `ValueError` if not)
(not accepted) ❌ 
```python
 rx.match(
            State.value,
            (1, rx.vstack(rx.text('first value'))),
            rx.text("default value"),
            (2, rx.text("second value")),
        )
```
(not accepted) ❌ 
```python
 rx.match(
            State.value,
            (1, 'first value'),
            "default value",
            (2, "second value"),
        )
```

- There can only be one default case.
(not accepted) ❌ 
```python
 rx.match(
            State.value,
            (1, rx.vstack(rx.text('first value'))),
            (2, rx.text("second value")),
            rx.text("default value"),
            rx.text("Another default value")
        )
```

(not accepted) ❌ 
```python
 rx.match(
            State.value,
            (1, 'first value'),
            (2, "second value"),
            "default value",
            "another default value",
        )
```
- The default case is optional when the type of the return value of the cases is a component (a default case `Fragment.create()` is implicitly created), however, when the return type is a Var (or when `rx.match` is used as a prop) a default case MUST be provided.

(accepted)✅ 
```python
rx.match(
            State.value,
            (1, rx.vstack(rx.text('first value'))),
            (2, rx.text("second value")),
        )
```

(not accepted) ❌ 
```python
 rx.match(
            State.value,
            (1, 'first value'),
            (2, "second value"),
        )
```

## `rx.match` as prop
```python
...
class State(rx.State):
    value: int = 0
    num: int = 1
...
rx.text(
            "Random text",
            color=rx.match(
                State.value,
                (1, "red"),
                (2, "blue"),
                (State.num + 1, "orange"),
                "yellow"
            ),
        ),

```

## support for multiple cases
```python
...
rx.match(
            State.value,
            (1, rx.text("first case")),
            (2, rx.text("second case")),
            (3, 4, 5, rx.text("third case")),
	    ((1,2,3), 6, 7, rx.text("fourth case")),
            ({"foo": "bar"}, [10, 11, 12], "random_case" ,rx.text("fifth case")),
             rx.text("default case"),
        ),
```

As mentioned above, the first `n-1` elements of the tuple are treated as cases to be matched and the last as the reurn value.

**Note** : Match cases must not have different return types:

(not accepted) ❌ 
```python
...
rx.match(
            State.value,
            (1, rx.text("first case")),
            (2, rx.text("second case")),
            (1, "red"),
            (2, "blue"),
             rx.text("default case"),
        ),

```

### DIFF
--- a/reflex/.templates/jinja/web/pages/utils.js.jinja2
+++ b/reflex/.templates/jinja/web/pages/utils.js.jinja2
@@ -8,6 +8,8 @@
     {{- component }}
   {%- elif "iterable" in component %}
     {{- render_iterable_tag(component) }}
+  {%- elif component.name == "match"%}
+    {{- render_match_tag(component) }}
   {%- elif "cond" in component %}
     {{- render_condition_tag(component) }}
   {%- elif component.children|length %}
@@ -76,6 +78,30 @@
 {% macro render_props(props) %}
 {% if props|length %} {{ props|join(" ") }}{% endif %}
 {% endmacro %}
+
+{# Rendering Match component. #}
+{# Args: #}
+{#     component: component dictionary #}
+{% macro render_match_tag(component) %}
+{
+    (() => {
+        let codeToRender;
+        switch ({{ component.cond._var_full_name }}) {
+        {% for case in component.match_cases %}
+            {% for condition in case[:-1] %}
+                case {{ condition._var_full_name }}:
+            {% endfor %}
+                codeToRender = {{ case[-1] }};
+                break;
+        {% endfor %}
+            default:
+                codeToRender = {{ component.default }};
+                break;
+        }
+        return codeToRender;
+    })()
+  }
+{%- endmacro %}
 
 
 {# Rendering content with args. #}
--- a/reflex/.templates/jinja/web/pages/utils.js.jinja2
+++ b/reflex/.templates/jinja/web/pages/utils.js.jinja2
@@ -88,7 +88,7 @@
         switch (JSON.stringify({{ component.cond._var_full_name }})) {
         {% for case in component.match_cases %}
             {% for condition in case[:-1] %}
-                case JSON.stringify({{ condition._var_string_without_curly_braces }}):
+                case JSON.stringify({{ condition._var_name_unwrapped }}):
             {% endfor %}
                 return {{ case[-1] }};
                 break;
--- a/reflex/.templates/jinja/web/pages/utils.js.jinja2
+++ b/reflex/.templates/jinja/web/pages/utils.js.jinja2
@@ -88,7 +88,7 @@
         switch ({{ component.cond._var_full_name }}) {
         {% for case in component.match_cases %}
             {% for condition in case[:-1] %}
-                case {{ condition._var_full_name }}:
+                case {{ condition._var_string_without_curly_braces }}:
             {% endfor %}
                 return {{ case[-1] }};
                 break;
--- a/reflex/.templates/jinja/web/pages/utils.js.jinja2
+++ b/reflex/.templates/jinja/web/pages/utils.js.jinja2
@@ -85,20 +85,18 @@
 {% macro render_match_tag(component) %}
 {
     (() => {
-        let codeToRender;
         switch ({{ component.cond._var_full_name }}) {
         {% for case in component.match_cases %}
             {% for condition in case[:-1] %}
                 case {{ condition._var_full_name }}:
             {% endfor %}
-                codeToRender = {{ case[-1] }};
+                return {{ case[-1] }};
                 break;
         {% endfor %}
             default:
-                codeToRender = {{ component.default }};
+                return {{ component.default }};
                 break;
         }
-        return codeToRender;
     })()
   }
 {%- endmacro %}
--- a/reflex/.templates/jinja/web/pages/utils.js.jinja2
+++ b/reflex/.templates/jinja/web/pages/utils.js.jinja2
@@ -85,10 +85,10 @@
 {% macro render_match_tag(component) %}
 {
     (() => {
-        switch ({{ component.cond._var_full_name }}) {
+        switch (JSON.stringify({{ component.cond._var_full_name }})) {
         {% for case in component.match_cases %}
             {% for condition in case[:-1] %}
-                case {{ condition._var_string_without_curly_braces }}:
+                case JSON.stringify({{ condition._var_string_without_curly_braces }}):
             {% endfor %}
                 return {{ case[-1] }};
                 break;
--- a/reflex/components/core/match.py
+++ b/reflex/components/core/match.py
@@ -5,7 +5,7 @@ from reflex.components.base import Fragment
 from reflex.components.component import BaseComponent, Component, MemoizationLeaf
 from reflex.components.tags import MatchTag, Tag
 from reflex.utils import format, imports, types
-from reflex.vars import BaseVar, Var
+from reflex.vars import BaseVar, Var, VarData
 
 
 class Match(MemoizationLeaf):
@@ -201,6 +201,13 @@ class Match(MemoizationLeaf):
         ) or not types._isinstance(default, BaseVar):
             raise ValueError("Return types of match cases should be Vars.")
 
+        # match cases and default should all be Vars at this point.
+        # Retrieve var data of every var in the match cases and default.
+        var_data = [
+            *[el._var_data for case in match_cases for el in case],
+            default._var_data,  # type: ignore
+        ]
+
         return match_cond_var._replace(
             _var_name=format.format_match(
                 cond=match_cond_var._var_full_name,
@@ -210,6 +217,7 @@ class Match(MemoizationLeaf):
             _var_type=default._var_type,  # type: ignore
             _var_is_local=False,
             _var_full_name_needs_state_prefix=False,
+            merge_var_data=VarData.merge(*var_data),
         )
 
     def _render(self) -> Tag:
--- a/reflex/components/core/match.py
+++ b/reflex/components/core/match.py
@@ -91,7 +91,7 @@ class Match(MemoizationLeaf):
         if not isinstance(cases[-1], tuple):
             default = cases.pop()
             default = (
-                Var.create(default)
+                Var.create(default, _var_is_string=type(default) is str)
                 if not isinstance(default, BaseComponent)
                 else default
             )
@@ -127,7 +127,7 @@ class Match(MemoizationLeaf):
             for element in case:
                 # convert all non component element to vars.
                 el = (
-                    Var.create(element)
+                    Var.create(element, _var_is_string=type(element) is str)
                     if not isinstance(element, BaseComponent)
                     else element
                 )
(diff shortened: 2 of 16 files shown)


reflex/components/layout/match.py
```diff
@@ -0,0 +1,103 @@
...
+
+            case_list = []
+            for element in case:
+                el = Var.create(element) if not isinstance(element, BaseComponent) else element
+                assert isinstance(el, (BaseVar, BaseComponent)), "case element must be a var or component"
+                case_list.append(el)
+
+            match_cases.append(case_list)
+
+        return_type = BaseComponent if types._isinstance(match_cases[0][-1], BaseComponent) else BaseVar if types._isinstance(match_cases[0][-1], BaseVar) else type(match_cases[0][-1])
```
masenf, line 50: this double ternary is kind of a readability problem IMO

how about

```python
first_case_return = match_cases[0][-1]
return_type = type(first_case_return)

# Generalize expected return type for known base classes.
if types._isinstance(first_case_return, BaseComponent):
    return_type = BaseComponent
elif types._isinstance(first_case_return, BaseVar):
    return_type = BaseVar
```

jackie-pc:
Looks really cool.  What types of values are supported for case values?  (examples above show "int").

Alek99:
> Looks really cool. What types of values are supported for case values? (examples above show "int").



> Looks really cool. What types of values are supported for case values? (examples above show "int").

Should work with any regular python type 

https://github.com/reflex-dev/reflex/blob/1adfeafb0e1910ec6329c5e3b97f608e8017b04f/tests/components/layout/test_match.py#L216-L237

Alek99:
```python

from rxconfig import config

import reflex as rx

docs_url = "https://reflex.dev/docs/getting-started/introduction"
filename = f"{config.app_name}/{config.app_name}.py"

class State(rx.State):
    value: str = 'a'

    def increment(self):
        self.value += 'a'
    

def index():
    return rx.vstack(
        rx.text("Hello, world!"),
        rx.text(State.value),
        rx.match(
            State.value,
            ('a', rx.vstack(rx.text('first value'))),
            ('aa', rx.text("fourth value")),
            ('aaa', rx.text("fifth value")),
            rx.text("default value")
        ),
    )

app = rx.App()
app.add_page(index)
```

reflex/components/core/match.py
```diff
@@ -0,0 +1,243 @@
...
+        return match_cond_var._replace(
+            _var_name=format.format_match(
+                cond=match_cond_var._var_full_name,
+                match_cases=match_cases,  # type: ignore
+                default=default,  # type: ignore
+            ),
+            _var_type=default._var_type,  # type: ignore
+            _var_is_local=False,
+            _var_full_name_needs_state_prefix=False,
+        )
```
masenf, line 213: we also need to `merge_var_data=VarData.merge(v._var_data for v in match_cases)` to sweep up any hooks/imports from stateful vars that are part of the match set.

tests/components/layout/test_match.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def test_match_vars():
+    """Test matching cases with return values as Vars."""
+    match_case_tuples = (
+        (1, "first"),
+        (2, 3, "second value"),
+        ([1, 2], "third-value"),
+        ("random", "fourth_value"),
+        ({"foo": "bar"}, "fifth value"),
```
masenf, line 90: i think at least some of the return values should be stateful Var and we should make sure that the resulting match code includes the necessary `useContext` to pull in access to the given substate.

reflex/components/core/match.py
```diff
@@ -0,0 +1,243 @@
...
+            ValueError: If the default case is not the last case or the tuple elements are less than 2.
+        """
+        match_cases = []
+        for case in cases:
+            if not isinstance(case, tuple):
+                raise ValueError(
+                    "rx.match should have tuples of cases and a default case as the last argument."
+                )
+            # There should be at least two elements in a case tuple(a condition and return value)
+            if len(case) < 2:
```
masenf, line 121: shouldn't this `if len(case) != 2`... i'd consider it a programming error if the case tuple has multiple return values, unless i'm missing something in the implementation.

reflex/components/core/match.py
```diff
@@ -0,0 +1,243 @@
...
+        return_type = type(first_case_return)
+
+        if types._isinstance(first_case_return, BaseComponent):
+            return_type = BaseComponent
+        elif types._isinstance(first_case_return, BaseVar):
+            return_type = BaseVar
+
+        for case in match_cases:
+            if not types._issubclass(type(case[-1]), return_type):
+                raise TypeError("match cases should have the same return types")
```
masenf, line 162: Include the checked types here, otherwise it might be hard to track down the exact issue.

```python
raise TypeError(f"Match cases should have the same return types. Case {index} of type {type(case[-1])!r} is not {return_type}")
```

masenf (changes requested):
haven't had a chance to run it in detail and play with it yet, but a couple of things stand out.

ElijahAhianyo:
> ```python
> """Welcome to Reflex! This file outlines the steps to create a basic app."""
> from rxconfig import config
> 
> import reflex as rx
> 
> docs_url = "https://reflex.dev/docs/getting-started/introduction"
> filename = f"{config.app_name}/{config.app_name}.py"
> 
> class State(rx.State):
>     value: str = 'a'
> 
>     def increment(self):
>         self.value += 'a'
>     
> 
> def index():
>     return rx.vstack(
>         rx.text("Hello, world!"),
>         rx.text(State.value),
>         rx.match(
>             State.value,
>             ('a', rx.vstack(rx.text('first value'))),
>             ('aa', rx.text("fourth value")),
>             ('aaa', rx.text("fifth value")),
>             rx.text("default value")
>         ),
>         rx.button("Increment", on_click=State.increment),
>     )
> 
> # Add state and page to the app.
> app = rx.App()
> app.add_page(index)
> ```
> 
> Screenshot 2023-12-21 at 4 12 10 PM

This is fixed now

reflex/components/core/match.py
ElijahAhianyo: The first `n-1` elements in the tuple are treated as the cases to be matched and the last value as the return value, which means there can only be one return value in a tuple. A minimum of two elements means, the first value is treated as the match case the second as the return value

reflex/components/core/match.py
ElijahAhianyo: Got it. Whats the best way to test this out?

Alek99:
> ```python
> ```python
> """Welcome to Reflex! This file outlines the steps to create a basic app."""
> from rxconfig import config
> 
> import reflex as rx
> 
> docs_url = "https://reflex.dev/docs/getting-started/introduction"
> filename = f"{config.app_name}/{config.app_name}.py"
> 
> class State(rx.State):
>     value: str = 'a'
> 
>     def increment(self):
>         self.value += 'a'
>     
> 
> def index():
>     return rx.vstack(
>         rx.text("Hello, world!"),
>         rx.text(State.value),
>         rx.match(
>             State.value,
>             ('a', rx.vstack(rx.text('first value'))),
>             ('aa', rx.text("fourth value")),
>             ('aaa', rx.text("fifth value")),
>             rx.text("default value")
>         ),
>         rx.button("Increment", on_click=State.increment),
>     )
> 
> # Add state and page to the app.
> app = rx.App()
> app.add_page(index)
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> 
> 
> This is fixed now
> ```

Ok nice testing

reflex/components/core/match.py
masenf: ohhh yeah that's right; we talked about that.... we should make sure that's clearly documented though so people know what to expect

reflex/vars.py
```diff
@@ -1530,6 +1530,26 @@ def _var_state(self) -> str:
         """
         return self._var_data.state if self._var_data else ""
 
+    @property
+    def _var_string_without_curly_braces(self) -> str:
```
masenf, line 1534: i think this is an important function/property, but i don't really like the name.

some suggestions:

```suggestion
    def _var_name_unwrapped(self) -> str:
```

```suggestion
    def _var_str_bare(self) -> str:
```

I think this allows the function to better express the intent of the returned value; it's not that we want the var stringified without curly braces (because indeed it does include curly's when the value is a dict literal), what we actually want is the value of the Var without it being wrapped for formatting as a prop or value (i.e. unwrapped / bare value can be directly formatted into code and used as an identifier)

Alek99 (approved):

reflex/.templates/jinja/web/pages/utils.js.jinja2
```diff
@@ -77,6 +79,28 @@
 {% if props|length %} {{ props|join(" ") }}{% endif %}
 {% endmacro %}
 
+{# Rendering Match component. #}
+{# Args: #}
+{#     component: component dictionary #}
+{% macro render_match_tag(component) %}
+{
+    (() => {
+        switch (JSON.stringify({{ component.cond._var_full_name }})) {
```
masenf, line 88: should this also be `_var_name_unwrapped`? not sure what the difference is between this and line 91 below?

masenf (approved):
maybe there's a bug? but lets get this in to unblock the theming work and if a bug arises we can fix it later

Alek99:
Seems to be working alright for me when I test it

PR #2320, opened 2023-12-20
Apply default theme to apps

picklelo:
Use a radix theme by default so the user doesn't have to specify. Chose "blue" for now.


Lendemor (approved):

masenf (approved):

masenf:
should address https://linear.app/reflex-dev/issue/REF-1360/if-the-app-doesnt-define-a-theme-then-radix-themes-components-do-not

PR #2322, opened 2023-12-21
CLI script to maintain Chakra backed components in rx namespace in older apps

jackie-pc:
On `reflex init` path, instructions to maintain Chakra component usage in an app will be provided in these scenarios:
* It's the very first `reflex init`.
* Previous `reflex init` was done with a pre-0.4 reflex, and the current reflex version is 0.4.

The instructions basically talks about `reflex script keep-chakra`. It will look for references to `rx.<component>` in the current app and change the reference to `rx.chakra.<component>`.  Whilst the actual transform is a simple string replace, there are safeguards in place, such as never transforming `rx.S` => `rx.chakra.T`, if S was not a `ChakraComponent` to begin with, or T is not a valid `ChakraComponent` under `rx.chakra`.

### DIFF
--- a/reflex/reflex.py
+++ b/reflex/reflex.py
@@ -100,9 +100,8 @@ def _init(
     # Migrate Pynecone projects to Reflex.
     prerequisites.migrate_to_reflex()
 
-    # IF WE WANT TO PROMPT USER TO DO RUN AUTO-MIGRATION SCRIPT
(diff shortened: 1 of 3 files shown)


masenf (approved):
@picklelo i'm good bringing this in as-is so we can more easily convert existing reflex code and test it without checking out an alternate branch.

i created REF-1902 to update the TODO link below

PR #2331, opened 2023-12-22
add center / spacer / hstack / vstack in radix/themes/layout

Lendemor:
```python

rdxt.center(
    rdxt.hstack(
        rx.text("A"),
        rx.text("B"),
        rdxt.spacer(),
        rx.text("C"),
    )
)

```


reflex/components/core/layout/center.py
```diff
@@ -0,0 +1,24 @@
...
+
+from reflex.components.component import Component
+from reflex.components.el.elements.typography import Div
+from reflex.style import Style
+
+
+class Center(Div):
+    """A center component."""
+
+    tag = "div"
```
masenf, line 11: does this even need to be set if you inherit from `Div`?

reflex/components/core/layout/stack.py
```diff
@@ -0,0 +1,108 @@
...
+        Args:
+            *children: The children of the stack.
+            **props: The properties of the stack.
+
+        Returns:
+            The stack component.
+        """
+        style = props.setdefault("style", {})
+        align = props.pop("align", "center")
+        spacing = props.pop("spacing", "0.5rem")
```
masenf, line 36: where does the 0.5rem come from? just what "feels nice"?

more broadly, extending these from `Div` and not radix `Flex` means we dont get the common layout and common margin props. I can see the benefits of both ways, but as it is now, if you want consistency with radix themes components, then you probably end up reimplementing these yourself anyway, but based on radix `Flex` or `Box`

reflex/components/core/layout/stack.py
```diff
@@ -0,0 +1,108 @@
...
+    """A stack component."""
+
+    # The direction of the stack.
+    justify: Var[Literal["start", "center", "end"]]
+
+    # The alignment of the stack.
+    align: Var[Literal["start", "center", "end", "stretch"]]
+
+    # The spacing between each stack item.
+    spacing: Var[str]
```
masenf, line 21: a bit unexpected that these are given as props, considering they wouldn't render correctly if they were set after the fact (outside of create).

i don't think anything should be a prop unless it is expected to be rendered into the tag for the frontend code... there are some exceptions throughout the codebase, but in those cases, we explicitly strip out the "fake props" inside `_render`

reflex/components/core/layout/stack.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class HStack(Stack):
+    """A horizontal stack component."""
+
+    def _apply_theme(self, theme: Component | None):
+        self.style.update({"flex_direction": "row"})
+
+
+# class Stack(rdxt.Flex):
```
masenf, line 63: 🤔 maybe _this_ stack should go under the radix themes hierarchy?

reflex/components/core/layout/stack.py
Lendemor: same default as chakra for the "0.5rem"

reflex/components/core/layout/center.py
Lendemor: no i'll remove it.

reflex/components/core/layout/stack.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+        style = props.setdefault("style", {})
+        style.update(
+            {
+                "display": "flex",
+                "alignItems": align,
+                "justifyContent": justify,
+                "gap": spacing,
+            }
+        )
```
masenf, line 45:
```suggestion
        props["style"] = {
                "display": "flex",
                "alignItems": align,
                "justifyContent": justify,
                "gap": spacing,
                **style,
        }
```

we should allow the user to override the styles right?

masenf (changes requested):

picklelo (approved):

PR #2355, opened 2024-01-04
Bare sqlalchemy metadata

### LINKED ISSUE #2340: [REF-1560] Bare SQLAlchemy support
Hey reflex team, I hope you all enjoyed your free time during Christmas :)

I stumbled about some issues with sqlmodel and decided to go with bare sqlalchemy without sqlmodel for now. A simple serializer function with reflex's `@serializer` decorator seems to work fine, and my sqlalchemy models even work when exposed directly as a reflex state variable.
However, since my sqlalchemy models do not inherit from `rx.Model` they are not discovered by `reflex db` commands and hence not included in my alembic migrations etc. I cannot inherit from `rx.Model` since sqlalchemy has many compatibility problems with pydantic (see f.e. https://github.com/sqlalchemy/sqlalchemy/discussions/10243 and https://github.com/pydantic/pydantic/issues/7129).

Would you accept a PR which tries to make bare sqlalchemy models work better with reflex? Maybe 2 different `rx.Model` subclasses, one with pydantics BaseModel for sqlmodel and one without for sqlalchemy? Or a simple bare minimal Mixin class which just "marks" all reflex models to be discovered (registered) for migrations etc.?

<sub>[REF-1560](https://linear.app/reflex-dev/issue/REF-1560/bare-sqlalchemy-support)</sub>

benedikt-bartscher:
#2340

### DIFF
--- a/reflex/model.py
+++ b/reflex/model.py
@@ -121,7 +121,7 @@ class ModelRegistry:
             metadata = next(iter(models)).metadata
         else:
             # Merge the metadata from all the models.
-            # This allows mixin bare sqlalchemy models with sqlmodel models in one database.
+            # This allows mixing bare sqlalchemy models with sqlmodel models in one database.
             metadata = sqlalchemy.MetaData()
             for model in cls.get_models():
                 for table in model.metadata.tables.values():
--- a/reflex/model.py
+++ b/reflex/model.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import os
 from collections import defaultdict
 from pathlib import Path
-from typing import Any, Optional
+from typing import Any, ClassVar, Optional, Type, Union
 
 import alembic.autogenerate
 import alembic.command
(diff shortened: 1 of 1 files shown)


reflex/model.py
```diff
@@ -51,7 +51,99 @@ def get_engine(url: str | None = None):
...
+            model: The model to check.
+
+        Returns:
+            True if the model metadata is empty, False otherwise.
+        """
+        return len(model.metadata.tables) == 0
+
+    @classmethod
+    @property
+    def metadata(cls) -> sqlalchemy.MetaData:
```
picklelo, line 107: Looks like joining these isn't supported in 3.8 (and won't be supported in 3.13 again later) which is why the unit tests are failing: https://stackoverflow.com/questions/128573/using-property-on-classmethods/64738850#64738850

benedikt-bartscher: Thanks for the hint, I refactored it to not use chained classmethods and propertys anymore. I also refactored the `Model.select` method, which chained those decorators as well. However, that was not the reason for the tests to fail.  

I cleaned everything up now, fixed the tests and rebased on main.

picklelo (approved):
tested that existing functionality works. this will be a great extension to our model thanks for adding!

reflex/model.py
```diff
@@ -300,7 +382,6 @@ def migrate(cls, autogenerate: bool = False) -> bool | None:
         return True
 
     @classmethod
-    @property
```
masenf, line 303: i think this might be a breaking change... my example app [rx_shout](https://github.com/masenf/rx_shout/tree/main) was using `select` as a class property on a model.

And now it throws stack at runtime

```python
  File "/Users/masenf/code/reflex-dev/reflex/reflex/state.py", line 1454, in _process_event
    events = fn(**payload)
             ^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/repro-datetime-model/rx_shout/state.py", line 48, in on_load
    self.load_entries()
  File "/Users/masenf/code/reflex-dev/repro-datetime-model/rx_shout/state.py", line 45, in load_entries
    self.entries = session.exec(Entry.select.order_by(Entry.ts.desc())).all()
                                ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'function' object has no attribute 'order_by'
```

Workaround is to call the function instead of treating it like a property: `Entry.select().order_by(...)`

masenf (commented):
Discovered this potential issue to document while testing `main`

reflex/model.py
benedikt-bartscher: Yes, that's a breaking change. However, I still think it was the right decision to drop those deprecated decorator chains.

PR #2358, opened 2024-01-04
Enable real app AppHarness tests to not specify `state=`

masenf:
Move the `self.state = State` assignment outside of the `if not is_testing_env` block to allow default state to be assigned during testing.

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -169,7 +169,8 @@ class App(Base):
                     deprecation_version="0.3.5",
                     removal_version="0.4.0",
                 )
-            self.state = State
+            if len(State.class_subclasses) > 0:
+                self.state = State
         # Get the config
         config = get_config()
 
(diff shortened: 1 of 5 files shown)


reflex/app.py
```diff
@@ -181,6 +180,9 @@ def __init__(self, *args, **kwargs):
         self.add_cors()
         self.add_default_endpoints()
 
+        if self.state is None and State.class_subclasses:
```
picklelo, line 183: can we do `len(State.class_subclasses) > 0` for clarity

reflex/app.py
```diff
@@ -701,6 +706,16 @@ def mark_complete(_=None):
                 stateful_components_code,
                 page_components,
             ) = compiler.compile_stateful_components(self.pages.values())
+
+            # Catch "static" apps (that do not define a rx.State subclass) which are trying to access rx.State.
+            if (
```
picklelo, line 711: Can we do a follow-up to abstract this out into a function `if uses_state(stateful_component_code)`

picklelo (dismissed):

picklelo (approved):

PR #2359, opened 2024-01-05
[REF-1586] Use bun as a package manager on windows

masenf:
## Bun is finally on windows! 🎉 
This PR aims to use bun on windows. Bun is primarily used to install packages. Note that we still run Reflex using npm(just like we do for mac and linux). 

## Caveats
Although `bun` is now supported on windows, it does not work on all architectures. We primarily support `AMD64` (and `x86_64`) which is what bun [supports](https://github.com/oven-sh/bun/blob/main/src/cli/install.ps1)
For cases where `bun` is not supported(ARM, ARM64, etc),` npm` will be used by default. 

## NPM Fallback
Bun on windows is not extremely stable, so to avoid the weird stuff on windows, we fallback to `npm`  when `bun install ` fails


masenf:
Currently, still broken

```console
(VENV-dev) PS C:\Users\masenf\code\repro-counter> reflex run --loglevel debug
───────────────────────────────────────────────────── System Info ──────────────────────────────────────────────────────
Debug: Config file: 'C:\\Users\\masenf\\code\\VENV-dev\\Lib\\site-packages\\reflex\\config.py'
Debug: Config: app_name='repro_counter' loglevel=<LogLevel.INFO: 'info'> frontend_port=3000 frontend_path=''
backend_port=8000 api_url='http://localhost:8000' deploy_url='http://localhost:3000' backend_host='0.0.0.0'
db_url='sqlite:///reflex.db' redis_url=None telemetry_enabled=True
bun_path='C:\\Users\\masenf\\AppData\\Local\\reflex\\bun\\bin\\bun' cors_allowed_origins=['*'] tailwind={} timeout=120
next_compression=True event_namespace=None frontend_packages=[] cp_backend_url='https://rxcp-prod-control-plane.fly.dev'
cp_web_url='https://control-plane.reflex.run' gunicorn_worker_class='uvicorn.workers.UvicornH11Worker'
Debug: Running command:
['C:\\Users\\masenf\\AppData\\Local\\reflex\\fnm\\node-versions\\v18.17.0\\installation\\node.exe', '-v']
Debug: Running command: ['C:\\Users\\masenf\\AppData\\Local\\reflex\\fnm\\fnm.exe', '--version']
Debug: [Reflex 0.4.3 with Python 3.11.4 (PATH: C:\Users\masenf\code\VENV-dev\Scripts\python.exe)]
Debug: [Node 18.17.0 (Expected: 18.17.0)
(PATH:C:\Users\masenf\AppData\Local\reflex\fnm\node-versions\v18.17.0\installation\node.exe)]
Debug: [FNM 1.35.1 (Expected: 1.35.1) (PATH: C:\Users\masenf\AppData\Local\reflex\fnm\fnm.exe)]
Debug: [OS Windows 10.0.22621]
Debug: Using package installer at: C:\Users\masenf\AppData\Local\reflex\bun\bin\bun.exe
Debug: Using package executer at: C:\Users\masenf\AppData\Local\reflex\fnm\node-versions\v18.17.0\installation\npm
Warning: Windows Subsystem for Linux (WSL) is recommended for improving initial install times.
───────────────────────────────────────────────── Starting Reflex App ──────────────────────────────────────────────────
DeprecationWarning: app.compile() has been deprecated in version 0.3.8 Explicit calls to app.compile() are not needed.
Method will be removed in 0.4.0. It will be completely removed in 0.5.0
Compiling: ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 9/9 0:00:00
Debug: Running command: ['C:\\Users\\masenf\\AppData\\Local\\reflex\\bun\\bin\\bun.exe', 'install', '--loglevel',
'silly']
Debug: Installing base frontend packages
Debug: bun add v1.0.30-canary.17 (f6d5325d)
Debug:  Resolving dependencies
Debug: Illegal instruction at address 0x7ff76fa191d1
Debug: ???:?:?: 0x7ff76fa1946f in ??? (bun.exe)
Debug: ???:?:?: 0x7ff76ef8c55e in ??? (bun.exe)
Debug: ???:?:?: 0x7ff76f1205e2 in ??? (bun.exe)
Debug: ???:?:?: 0x7ff76ef26d94 in ??? (bun.exe)
Debug: ???:?:?: 0x7ff76f0d0d61 in ??? (bun.exe)
Debug: ???:?:?: 0x7ffa72400a9b in ??? (KERNEL32.DLL)
Debug: ???:?:?: 0x7ffa72397c0f in ??? (KERNEL32.DLL)
Debug: ???:?:?: 0x7ffa7313b8d7 in ??? (ntdll.dll)
```

reflex/utils/processes.py
```diff
@@ -282,3 +283,40 @@ def show_progress(message: str, process: subprocess.Popen, checkpoints: List[str
...
+        show_status_message: The status message to be displayed in the console.
+        fallback: The fallback command to run.
+        kwargs: Kwargs to pass to new_process function.
+    """
+
+    def execute_process(process):
+        if not constants.IS_WINDOWS:
+            show_status(show_status_message, process)
+        else:
+            process.wait()
```
ElijahAhianyo, line 302: Not a fan of this but this was a way to get it to work. Here's the use case for this:
`subprocess.Popen` is non-blocking which is what we use conventionally, `show_status` prints the output of the spawned process. 
Using `show_status` in this case means errors will still be displayed when the process throws one in INFO mode (I find it intuitive to have this error in debug mode instead).
In the case of running with a fallback, we want to know whether the process failed or not, so we wait(which is blocking) for completion of the process. I tested this against reflex-web and the perf difference (between the case of waiting and not waiting) was not noticeable. Open to suggestions, though.

reflex/utils/prerequisites.py
```diff
@@ -766,40 +772,44 @@ def install_frontend_packages(packages: set[str], config: Config):
...
+            show_status_message="Installing tailwind",
             cwd=constants.Dirs.WEB,
             shell=constants.IS_WINDOWS,
         )
-        processes.show_status("Installing tailwind", process)
 
     # Install custom packages defined in frontend_packages
     if len(packages) > 0:
-        process = processes.new_process(
+        processes.run_process_with_fallback(
```
picklelo, line 806: We may need to add a print statement here. npm somehow gave out a log of everything but with bun it looks like it's hanging because it says "App Running" and then the port doesn't show up for around 10s.

picklelo (approved):
This is great, tested it on Windows and it's much faster than npm

PR #2400, opened 2024-01-16
Skip frontend packages install if previously done

jackie-pc:
We introduce a decorator `@cached_procedure` for this functionality.  "Procedure" because we are not caching any results, only the inputs to a previous call.  It is up to the decoratee to ensure its args capture the unique call under cache sufficiently.

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -595,7 +595,7 @@ class App(Base):
                 continue
             _frontend_packages.append(package)
         page_imports.update(_frontend_packages)
-        prerequisites.install_frontend_packages(page_imports)
+        prerequisites.install_frontend_packages(page_imports, get_config())
 
     def _app_root(self, app_wrappers: dict[tuple[int, str], Component]) -> Component:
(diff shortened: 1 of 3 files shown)


masenf (approved):
as i'm reading it, this will also result in frontend packages not being installed for `export` and `run --env prod` if the packages have already been installed.

seems the only way to get package installation to occur now is to change something in the app the requires a different set of packages, or re-init the project.

jackie-pc:
Is it necessary or desirable to perform real frontend install steps under those conditions?  It's really a case of capturing all such requirements in the cache payload (e.g. by including them as args to the function).

masenf:
> Is it necessary or desirable to perform real frontend install steps under those conditions?

i dont think it's necessary, just a potentially notable change in how those commands work.

i mark approve, i think we should take it; but also note in the docs that if you want to actually reinstall packages to use `reflex init` first.

picklelo (approved):

PR #2401, opened 2024-01-16
radiogroup high level api

tgberkeley:
Must import as shown" `from reflex.components.radix.themes.components import * `

Most basic usage, pass in a list of strings :

`radio_group(["1", "2", "3", "4", "5"])`

Can set several props such as `direction`, `gap`:

`radio_group(["1", "2", "3", "4", "5"], direction="row", gap="6"))`

and any other prop that the `RadioGroupRoot` component takes:

`radio_group(["1", "2", "3", "4", "5"], direction="row", gap="6", variant="soft", size="3", color_scheme="red", default_value="3")`

The list of strings passed can also be a state var and this will still work

```
from reflex.components.radix.themes.components import *


class State(rx.State):
    """The app state."""

    rad: list[str] = ["Name", "Age", "Location"]


def index():
    """The main view."""
    return rx.center(
        rx.vstack(
            radio_group(["1", "2", "3", "4", "5"]),
            radio_group(State.rad),
            radio_group(["1", "2", "3", "4", "5"], direction="row", gap="6")),
            radio_group(["1", "2", "3", "4", "5"], direction="row", gap="6", variant="soft", size="3", color_scheme="red", default_value="3"),
        ),
    )

app = rx.App()
app.add_page(index)
```

### DIFF
--- a/reflex/components/radix/themes/components/radiogroup.py
+++ b/reflex/components/radix/themes/components/radiogroup.py
@@ -89,13 +89,13 @@ class HighLevelRadioGroup(RadioGroupRoot):
     items: Var[List[str]]
 
     # The direction of the radio group.
-    direction: Var[LiteralFlexDirection] = "column"
+    direction: Var[LiteralFlexDirection] = Var.create_safe("column")
 
     # The gap between the items of the radio group.
-    gap: Var[LiteralSize] = "2"
+    gap: Var[LiteralSize] = Var.create_safe("2")
 
     # The size of the radio group.
-    size: Var[Literal["1", "2", "3"]] = "2"
+    size: Var[Literal["1", "2", "3"]] = Var.create_safe("2")
 
     @classmethod
     def create(cls, items: Var[List[str]], **props) -> Component:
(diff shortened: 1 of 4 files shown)


reflex/components/radix/themes/components/radiogroup.py
```diff
@@ -73,3 +80,47 @@ class RadioGroupItem(CommonMarginProps, RadixThemesComponent):
...
+    Returns:
+        The created radio group component.
+    """
+
+    def radio_group_item(value: str) -> Component:
+        return Text.create(
+            Flex.create(
+                RadioGroupItem.create(value=value),
+                value,
+                gap="2",
```
masenf, line 108: should this also use the passed `gap` for consistency?

tgberkeley: This is the `gap=2` is the distance between the radio group and the text it is associated with. The gap that we allow the user to pass is the distance between the different radio group items. I therefore do not think they should be passed the same value, and think we should keep the value between the radio group and the text fixed.
Screenshot 2024-01-17 at 9 58 14 AM

picklelo (approved):

PR #2403, opened 2024-01-17
Remove default radix theme

picklelo:
Fixes the issue of global styles being overridden.


masenf:
Do we know the root cause here? Or have a regression test we can add?

Because we're going to need a default theme at some point, probably, and need to avoid breaking this again.

picklelo:
No don't know the root cause yet. Agreed we need a fix and to reset this in the future, we only need to merge this if we want an interim fix and cut a release while we figure that out.

masenf (approved):

PR #2405, opened 2024-01-17
Individually Import Chakra Packages

Alek99:
Continue using chakra layout components as base layout, radix layout will override for duplicates.

For all chakra component import the package directly to reduce .web folder size when components are not used.

https://www.npmjs.com/package/@chakra-ui/layout?activeTab=readme


masenf (approved):
Initial check looks good

Alek99:
> Initial check looks good

Do you see a problem with how my change will affect the failed unit tests switching the level of the chakra provider.

masenf (changes requested):
the app wrap ordering become order dependent since the test uses priority `50` for the `Text` App Wrap and `ChakraColorModeProvider` is also at priority `50`.

I would probably change the test to use `return {(45, "Text"): Text.create()}` on line 1254, that way the sorting is deterministic instead of depending on the order the components were added to the tree.

masenf (dismissed):

picklelo (approved):

PR #2408, opened 2024-01-18
Allow app harness tests to import State subclasses

masenf:
Otherwise importing with `reload=True` would remove all known state subclasses, even those defined in modules that would NOT be reloaded, which resulted in test failures for AppHarness tests that relied on state defined in another module.

https://discord.com/channels/1029853095527727165/1061874061250150441/1197312746224615514


benedikt-bartscher:
Works great, thank you very much!

picklelo (approved):

PR #2427, opened 2024-01-21
Allow app harness tests States to inherit from imported States

benedikt-bartscher:
Followup of #2408 

Inheriting from an imported State (f.e. from a AuthState provided by an auth library) is now possible in AppHarness tests:

```python
from examplelib.state import AuthState


class Test(AuthState):
    pass
```


benedikt-bartscher:
While this PR does solve the problem mentioned above, it still causes problems if 2 different tests use the same imported parent state in a single pytest run. Here is a testrepo: https://github.com/benedikt-bartscher/reflex-integration-import-state

```python
# this works fine
pytest tests/test_input_module.py
pytest tests/test_input.py

# this does fail with the usual "shadowing subclasses is not allowed" exception
pytest tests
```

benedikt-bartscher:
superseded by #2441

PR #2429, opened 2024-01-21
Track state variable and attribute usage

benedikt-bartscher:
Track if `rx.Var`s and their attributes are used (directly or via var operations) and only include used variables in the state dict. This should result in reduced initial state payloads and network traffic for var changes, which are not really needed in the frontend.  

My plan is to write a second PR based on this one, which includes "used" properties (normal python properties with `@property` decorator) in the serialization. This allows users to use those properties in the frontend without always computing all properties for a given class. (related to my comment [here](https://github.com/reflex-dev/reflex/pull/2422#issuecomment-1901195257))

Co-Author: @macmoritz


tests/states/upload.py
```diff
@@ -80,6 +80,9 @@ async def bg_upload(self, files: List[rx.UploadFile]):
         pass
 
 
+rx.fragment(FileUploadState.img_list)
```
masenf, line 83: this would be more readable with a helper function, maybe defined in reflex.vars:

```
rx.vars.mark_used(FileUploadState.img_list)
```

This helper function should take arbitrary `*args` and just set `_var_is_used = True` on each of them.

benedikt-bartscher: I agree, i had a bad feeling writing those lines

masenf (commented):
I think i fixed the integration test failures, pretty sure we can merge if CI is passing.

reflex/vars.py
```diff
@@ -400,6 +400,9 @@ def _replace(self, merge_var_data=None, **kwargs: Any) -> Var:
             _var_is_local=kwargs.pop("_var_is_local", self._var_is_local),
             _var_is_string=kwargs.pop("_var_is_string", self._var_is_string),
             _var_is_used=kwargs.pop("_var_is_used", self._var_is_used),
+            _var_used_attributes=kwargs.pop(
```
benedikt-bartscher, line 410: This may be wrong, i will check later

masenf:
Will have to continue holding off on this one for the time being, because it breaks with the reflex.dev website.

Flexdown currently redefines State classes multiple times per doc file, so they lose their _is_used markers.

benedikt-bartscher:
Needs #3214 for flexdown

benedikt-bartscher:
> Will have to continue holding off on this one for the time being, because it breaks with the reflex.dev website.
> 
> Flexdown currently redefines State classes multiple times per doc file, so they lose their _is_used markers.

@masenf I just rebased onto main and tested this locally with flexdown branch benedikt-bartscher/per-block-module, and it seems to work fine. However, reflex-web ci is still failing

benedikt-bartscher:
@masenf did you release a new flexdown version yet?
Shall we stall this one until the new var system is merged? Did you have usage tracking in mind while designing it?

picklelo:
@benedikt-bartscher I just cut a new version of flexdown 0.1.5a10 off the latest main

benedikt-bartscher:
Thank you, @picklelo. I just raised https://github.com/reflex-dev/reflex-web/pull/855

benedikt-bartscher:
Finally, reflex-web ci is passing, thanks again @picklelo and @masenf.  
Now we just need to figure out how we handle this with the new var system in #3743 cc @adhami3310

benedikt-bartscher:
@masenf @adhami3310 do you have any hints/ideas on how to implement this for ImmutableVar's?

adhami3310:
> ideas on how to implement this for ImmutableVar

Since immutable vars can't have such fields, it could be possible to have a state-level mapping between vars and if they are used or not, although it will be fairly tricky to implement.

I'm not sure however how this will be that strong of an optimization considering that in an ideal application (almost) all vars will be used somewhere in the code, rendering all of the operations we do to figure out if a variable is being used unnecessary. Another issue is that it would slow down compile times (even further). Is there unseen advantages I'm missing?

benedikt-bartscher:
Thanks for your feedback. You are right, it might introduce some additional compile time. But I think it should be possible without too much overhead, and (at least us) runtime performance is more important.

> an ideal application (almost) all vars will be used somewhere in the code
While this should always be the goal, there are many cases where it is not true.
Some examples:
- Reflex internal vars like router or dynamic router vars can not be disabled and may not be needed at all.
- Libraries may share reusable State classes or Mixins, and the user might not want to use all vars.

Another advantage of usage tracking is automatic serialization of python/sqla/... properties if they are used in the frontend.

benedikt-bartscher:
@adhami3310 I completely forgot about one of the biggest benefits. Sorry, it has been a long time since I started this PR. Reflex currently always serialize complete root-level vars. Meaning, if you have f.e. a `rx.Base` class with 10 fields, reflex always serializes all of them if you don't provide a custom serializer.

benedikt-bartscher:
Closing this, continue here: https://github.com/reflex-dev/reflex/pull/3845

PR #2430, opened 2024-01-21
auto enable /_upload endpoint only if Upload component is used

benedikt-bartscher:
https://github.com/orgs/reflex-dev/discussions/2415  

This also changes the pyi generator to ignore `ClassVar`s in the `Component.create` method and do not delete `ClassVar`s in the class attributes


reflex/components/core/upload.py
```diff
@@ -92,12 +92,33 @@ def cancel_upload(upload_id: str) -> EventSpec:
     return call_script(f"upload_controllers[{upload_id!r}]?.abort()")
 
 
+class UploadComponentUsed:
+    """A class to indicate whether an Upload component is used."""
+
+    is_used: ClassVar[bool] = False
```
masenf, line 98: can this just be a ClassVar on the `UploadFilesProvider`?

reflex/components/core/upload.py
```diff
@@ -148,6 +169,8 @@ def create(cls, *children, **props) -> Component:
         Returns:
             The upload component.
         """
+        UploadComponentUsed.is_used = True
```
masenf, line 172: instead of also setting this here, lets call `UploadFilesProvider.create()` down in `_get_app_wrap_components` and let the `UploadFilesProvider.create` set the flag.

reflex/components/core/upload.py
benedikt-bartscher: My first implementation was using a `ClassVar` but pyright did not like the generated pyi: https://github.com/reflex-dev/reflex/actions/runs/7603957212/job/20706298762

masenf: oh gotcha; the pyi_generator script is broken 😢 

can you add `_used` or `_is_used` to the `EXCLUDED_PROPS` list around line 39 of the `pyi_generator.py` script. that will avoid putting the `ClassVar` into the signature of `.create` method. 

> `error: "ClassVar" is not allowed in this context`

This is basically saying you cannot annotate a function argument as a `ClassVar`, which makes perfect sense to me (but definitely was confusing when i first saw it).

reflex/components/core/upload.py
benedikt-bartscher: Like this?
```patch
diff --git a/reflex/components/core/upload.py b/reflex/components/core/upload.py
index 389e25a8..30856140 100644
--- a/reflex/components/core/upload.py
+++ b/reflex/components/core/upload.py
@@ -169,8 +169,6 @@ class Upload(Component):
         Returns:
             The upload component.
         """
-        UploadComponentUsed.is_used = True
-
         # get only upload component props
         supported_props = cls.get_props()
         upload_props = {
@@ -215,5 +213,5 @@ class Upload(Component):
     @staticmethod
     def _get_app_wrap_components() -> dict[tuple[int, str], Component]:
         return {
-            (5, "UploadFilesProvider"): UploadFilesProvider(),
+            (5, "UploadFilesProvider"): UploadFilesProvider.create(),
         }
```
With this patch the upload endpoint is always disabled for me

reflex/components/core/upload.py
benedikt-bartscher: Yes i know, i even tried `EXCLUDED_PROPS` but then the ClassVar is excluded everywhere and hence pyright complains about the missing `_is_used` class attribute

benedikt-bartscher: `  /home/bb/dev/reflex/github-bb-reflex/reflex/app.py:248:32 - error: Cannot access member "is_used" for type "type[UploadFilesProvider]"`

benedikt-bartscher: @masenf i adjusted the pyi generator to exclude `ClassVars` and migrated `is_used` to `UploadFilesProvider`

reflex/components/core/upload.py
benedikt-bartscher: fixed by moving the endpoint setup at the end of `compile_`

masenf (approved):
Looks good to me! Thanks for wrangling the `pyi_generator`, that script can be a pain sometimes.

PR #2436, opened 2024-01-22
[REF-1692] re-enable partial reflex web windows

masenf:
Re-enable reflex-web on windows integration test, with a limited number of routes.


picklelo (approved):
nice!

PR #2441, opened 2024-01-23
Track state usage

benedikt-bartscher:
Based on #2427  

## Stateless detection

Currently, reflex detects if an app is stateless by only checking if `State.class_subclasses` is empty. If you just import a python module which has a `rx.State` defined, your app wouldn't be stateless anymore. This PR changes the behavior and only activates stateful mode if the app really uses any State variables (like #2429). A further (much smaller) PR could extend this detection to exclude all unused States completely at runtime. This is especially useful when importing other packages which provide States.

## App Harness State reloading

Allow app harness tests States to inherit from imported States (see #2427 on which this PR is based) by keeping all imported States in the `class_subclasses` and recursively cleaning up all States from the test module from `class_subclasses` before each test.

Tested with this repo: https://github.com/benedikt-bartscher/reflex-integration-import-state


reflex/app.py
```diff
@@ -654,6 +703,9 @@ def compile_(self):
         if not self._should_compile():
             return
 
+        self._setup_overlay_component()
+        self._setup_dynamic_routes()
```
masenf, line 707: I think this needs to happen _above_ the `if not self._should_compile()` line, because the backend needs to setup the dynamic route vars, even if we're not compiling the frontend.

masenf (changes requested):
Seems like a good approach.

reflex/app.py
benedikt-bartscher: But we only need to update them if something changed or am I missing something? I tested it with the integration tests and one of my apps which uses dynamic routes.

masenf: did you test it with an export and `reflex run --backend-only`?

When running in backend only mode, the compile step is skipped.

benedikt-bartscher: You got me, I did not test a prod deployment. We should think about adding a test for this.

masenf (changes requested):
Moving the `_setup_dynamic_routes` to compile time has introduced a regression

```python
import reflex as rx


class State(rx.State):
    @rx.var
    def post_id(self):
        return self.router.page.params.get("pid", "no pid")


def index():
    return rx.link("Post 123", href="/post/123")


def post():
    """A page that updates based on the route."""
    return rx.vstack(
        rx.heading(State.post_id),
        rx.heading(State.pid),
    )


app = rx.App()
app.add_page(index)
app.add_page(post, route="/post/[pid]")
```

This app works fine in 0.3.9, but does not work with this patch

Traceback

```console
  File "/Users/masenf/code/reflex-dev/VENV-dev/bin/reflex", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/Users/masenf/code/reflex-dev/VENV-dev/lib/python3.11/site-packages/typer/main.py", line 214, in __call__
    return get_command(self)(*args, **kwargs)
  File "/Users/masenf/code/reflex-dev/VENV-dev/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/VENV-dev/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/VENV-dev/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/masenf/code/reflex-dev/VENV-dev/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/VENV-dev/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/VENV-dev/lib/python3.11/site-packages/typer/main.py", line 532, in wrapper
    return callback(**use_params)  # type: ignore
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/reflex/reflex/reflex.py", line 250, in run
    _run(env, frontend, backend, frontend_port, backend_port, backend_host, loglevel)
  File "/Users/masenf/code/reflex-dev/reflex/reflex/reflex.py", line 181, in _run
    prerequisites.get_compiled_app()
  File "/Users/masenf/code/reflex-dev/reflex/reflex/utils/prerequisites.py", line 184, in get_compiled_app
    app_module = get_app(reload=reload)
                 ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/reflex/reflex/utils/prerequisites.py", line 161, in get_app
    app = __import__(module, fromlist=(constants.CompileVars.APP,))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/repro-dynamic-route/repro_dynamic_route/repro_dynamic_route.py", line 24, in <module>
    app.add_page(post, route="/post/[pid]")
  File "/Users/masenf/code/reflex-dev/reflex/reflex/app.py", line 427, in add_page
    component = self._generate_component(component)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/reflex/reflex/app.py", line 371, in _generate_component
    return component if isinstance(component, Component) else component()
                                                              ^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/repro-dynamic-route/repro_dynamic_route/repro_dynamic_route.py", line 18, in post
    rx.heading(State.pid),
               ^^^^^^^^^
AttributeError: type object 'State' has no attribute 'pid'
```

benedikt-bartscher:
@masenf Good catch, thanks for testing! I move the dynamic route initialization out of the compile process again. This needed a small adjustment for tests using the deprecated `state` kwarg of `App`. We can drop that line after it gets removed in 0.4.0.

benedikt-bartscher:
@masenf Friendly remainder, is there anything I can do to get this one in?

reflex/components/radix/primitives/form.py
```diff
@@ -160,7 +160,15 @@ def _apply_theme(self, theme: Component):
...
+    def _get_vars(self, include_children: bool = True) -> Iterator[Var]:
+        """Walk all Vars used in this form.
+
+        Args:
+            include_children: Whether to include Vars from children.
+
+        Yields:
+            Each var referenced by the form (props, styles, event handlers).
+        """
         yield from super()._get_vars()
```
masenf, line 172: need to pass through the new `include_children` kwarg here

masenf (commented):
i generally like this change. will try to test it out locally and get it in soon.

reflex/app.py
```diff
@@ -187,8 +193,6 @@ def __init__(self, *args, **kwargs):
             # 2 substates are built-in and not considered when determining if app is stateless.
             if len(State.class_subclasses) > 2:
                 self.state = State
```
masenf, line 195: I think this line probably needs to go, otherwise it's still enabling the state based on number of subclasses still.

benedikt-bartscher: Indeed, I removed those lines and added an unused State to the tailwind integration test

masenf (dismissed):

masenf:
Had to fix conflicts from my recent form consolidation that merged it. Will merge when CI looks good.

masenf (approved):

PR #2443, opened 2024-01-24
[WIP] rx.color

Alek99:
Initial thoughts on what it would look like 

```python
from rxconfig import config

import reflex as rx
import reflex.components.radix.primitives as rdxp
from reflex.utils.serializers import get_serializer
import random

class State(rx.State):
    pass

def index() -> rx.Component:
    return rdxt.theme(
        rx.text(
            "test",
            bg=rx.color("red"),
        ),
    )     
    
app = rx.App(
    theme=rdxt.theme(
        appearance="dark", has_background=True, radius="none", accent_color="gold" 
    ), 
) 
 
app.add_page(index) 
  ```

### DIFF
--- a/reflex/constants/colors.py
+++ b/reflex/constants/colors.py
@@ -1,0 +1,59 @@
+"""The colors used in Reflex are a wrapper around https://www.radix-ui.com/colors."""
+from typing import Literal
+
+ColorType= Literal[
+        "gray",
+        "mauve",
+        "slate",
+        "sage",
+        "olive",
+        "sand",
+        "tomato",
+        "red",
+        "ruby",
+        "crimson",
+        "pink",
+        "plum",
+        "purple",
+        "violet",
+        "iris",
+        "indigo",
+        "blue",
+        "cyan",
+        "teal",
+        "jade",
+        "green",
+        "grass",
+        "brown",
+        "orange",
+        "sky",
+        "mint",
+        "lime",
+        "yellow",
+        "amber",
+        "gold",
+        "bronze",
+        "gray",
+    ]
+
+ShadeType= Literal[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
+
+
+class Colors:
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: reflex/constants/colors.py [hunk 1]
```diff
@@ -0,0 +1,77 @@
...
+
+    Args:
+        color: The color to serialize.
+
+    Returns:
+        The serialized color.
+    """
+    print(f"--var({color.color}-{color.shade})")
+    if color.alpha:
+        return f"var(--{color.color}-{color.shade}-alpha)"
```
masenf, line 70:
```suggestion
        return f"var(--{color.color}-a{color.shade})"
```

i thought the alphas used an `a` prefix on the shade

masenf (commented):
dude, i _dig_ this approach! awesome! 🤩

reflex/constants/colors.py
Alek99: ah your correct

picklelo (commented):
Looks good!

* I think this should be accessible as `rx.colors` rather than as a separate import
* Can we add some tests in test_serializers.py

reflex/constants/colors.py
```diff
@@ -0,0 +1,77 @@
...
+        "amber",
+        "gold",
+        "bronze",
+        "gray",
+    ]
+
+ShadeType= Literal[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
+
+@dataclass
+class Colors:
```
picklelo, line 46: Let's name this class `Color`, then call it as `rx.color(...)`

Alek99:
If the app isn't in a radix theme the colors won't work as they ref radix color palette. I don't see an easy way to check this though to throw an error if this happens.

masenf:
> If the app isn't in a radix theme

i think the plan is still to always apply a default theme, so this might be a moot point. either way you should be able to get `get_app().theme` and see if it's set to an instance of `rdxt.Theme`

picklelo (approved):

PR #2444, opened 2024-01-24
add lucide icons as a candidate for core icon library

Lendemor:
`size`, `strokeWidth` and `color` can be customized like in the following example:

```python
import reflex as rx

rx.lucide.icon(tag="moon", size=24, strokeWidth="1"),
rx.lucide.icon(tag="moon", size=48, strokeWidth="2"),
rx.lucide.icon(tag="moon", size=24, strokeWidth="2", color="red"),
rx.lucide.icon(tag="moon", size=48, strokeWidth="2", color="blue"),
```

image


reflex/components/lucide/icon.py
Alek99, line 1: Hard  pin version here

Alek99 (changes requested):
### Comments from discord 
he only comment I have with this is that if they are using the theme provider I would like this to default to using it so it can look good in a button like so https://www.radix-ui.com/themes/docs/components/button
We may need to make a method that can check if a theme is being used
We can also get rid of radix icons as we only need one icon lib

reflex/components/lucide/icon.py
Alek99: This is the only one needing to be addressed now

picklelo (approved):
awesome

PR #2447, opened 2024-01-24
Add git precommit hook that generates pyi files

martinxu9:
### All Submissions:

- [ ] Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/reflex-dev/reflex/pulls ) for the desired changed?

### Type of change

### New Feature Submission:

### Changes To Core Features:

- Add git precommit hook that generates pyi files


benedikt-bartscher:
@martinxu9 https://github.com/reflex-dev/reflex/actions/runs/7646912485/job/20836748665?pr=2447  
Precommit CI now also runs pyi generator and fails

martinxu9:
> @martinxu9 https://github.com/reflex-dev/reflex/actions/runs/7646912485/job/20836748665?pr=2447 Precommit CI now also runs pyi generator and fails

oh yes, that's a very good point, the pre-commit github action runs all pre-commit checks, so the separate check-generated-pyi action can be removed

benedikt-bartscher:
You would have to use the correct virtual environment then, that's why the pre-commit job is currently failing. But I would rather disable pyi generator in pre-commit for CI. 2 separate Jobs are faster, and you can directly see if a failing job is related to pyi generation or not.

picklelo (commented):
For some reason the pre-commit is failing - change looks good though

benedikt-bartscher:
That's because pre-commit runs pyi_generator against every file with its full path. Which makes no sense for the following reasons:
- Does not work for deleted .py component files
- Does not leverage multiprocessing of pyi_generator
- pyi_generator has its own logic on which files to handle/ignore

Try with `always_run: true` and without `files: '.*\.py$'`

martinxu9:
> That's because pre-commit runs pyi_generator against every file with its full path. Which makes no sense for the following reasons:
> 
> * Does not work for deleted .py component files
> * Does not leverage multiprocessing of pyi_generator
> * pyi_generator has its own logic on which files to handle/ignore
> 
> Try with `always_run: true` and without `files: '.*\.py$'`
> 
> Btw, could you merge main to test against [08524e2](https://github.com/reflex-dev/reflex/commit/08524e22aaefbfb3e2dc95a998f7cbb3e157c591) ?

That didn't work either. I guess like you explained, pyi generator is no ordinary script. Per pre-commit hook docs, the system type hook does not have access to the python virutual environment, so probably takes more changes to make it work. I like the previous suggestion to keep it separate in the CI. Now it (pyi gen) is not run on the pre-commit CI, only user environment locally.

picklelo (approved):

PR #2452, opened 2024-01-25
[REF-1742] Radio group prop types fix

ElijahAhianyo:
Radio group component items and default value to support types other than `str`

```python
from typing import List, Tuple, Any, Union
import reflex as rx

class State(rx.State):
    selected: str | int | list | dict = ""
    default: list = [1,2,3]
    items: list = [ 1, 2, 2.578,"hey there", [1,2,3], {1:2, 3: 4}]

def index():
    return rdxt.flex(
            rdxt.flex(rdxt.text(f"selected is {State.selected}"),),
            rdxt.flex(
                rdxt.radio_group(
                items= [ 1, 2, 2.578,"hey there", [1,2,3], {1:2, 3: 4}], #  State.items,
                on_value_change=State.set_selected,
                direction="row",
                default_value=State.default,
                gap="3"
            ),

            ),
        gap="3",
        direction="column",
        align_items="center"
    )

app = rx.App(
theme=rdxt.theme(
        has_background=True, radius="none", accent_color="grass", appearance="light",
    ),
)
app.add_page(index)
```

### DIFF
--- a/reflex/vars.py
+++ b/reflex/vars.py
@@ -887,7 +887,7 @@ class Var:
         """
         for python_types, js_type in PYTHON_JS_TYPE_MAP.items():
             if not isinstance(other, Var) and other in python_types:
-                return self.compare("===", Var.create(js_type, _var_is_string=True))  # type: ignore
+                return self.compare("!==", Var.create(js_type, _var_is_string=True))  # type: ignore
         return self.compare("!==", other)
 
     def __gt__(self, other: Var) -> Var:
--- a/reflex/vars.py
+++ b/reflex/vars.py
@@ -87,6 +87,15 @@ REPLACED_NAMES = {
     "deps": "_deps",
 }
 
+PYTHON_JS_TYPE_MAP = {
+    (int, float): "number",
+    (str,): "string",
+    (bool,): "boolean",
+    (list, tuple): "Array",
+    (dict,): "Object",
+    (None,): "undefined",
+}
+
 
 def get_unique_variable_name() -> str:
     """Get a unique variable name.
@@ -843,7 +852,17 @@ class Var:
             _var_is_string=False,
         )
 
-    def __eq__(self, other: Var) -> Var:
+    def type(self) -> Var:
+        """Get the type of the Var in Javascript.
+
+        Returns:
+            A var representing the type check.
+        """
+        return self._replace(
+            _var_name=f"typeof {self._var_name}", _var_type=str, _var_is_string=False
+        )
+
+    def __eq__(self, other: Union[Var, Type]) -> Var:
         """Perform an equality comparison.
 
         Args:
@@ -852,9 +871,12 @@ class Var:
         Returns:
             A var representing the equality comparison.
         """
+        for python_types, js_type in PYTHON_JS_TYPE_MAP.items():
+            if not isinstance(other, Var) and other in python_types:
+                return self.compare("===", Var.create(js_type))  # type: ignore
         return self.compare("===", other)
 
-    def __ne__(self, other: Var) -> Var:
+    def __ne__(self, other: Union[Var, Type]) -> Var:
         """Perform an inequality comparison.
 
         Args:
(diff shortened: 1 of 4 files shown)


masenf:
What other types could I put in here and not break? I think we should be more restrictive.

reflex/vars.py
```diff
@@ -843,7 +852,17 @@ def length(self) -> Var:
             _var_is_string=False,
         )
 
-    def __eq__(self, other: Var) -> Var:
+    def type(self) -> Var:
```
masenf, line 855:
```suggestion
    def _type(self) -> Var:
```

Recommend using an under prefix name for any non-operator function defined in the `Var` class to avoid shadowing potential field names in objects represented by the Var that are accessed via `__getattr__`. This is why `name` and friends were renamed to `_var_name`

```python
class Foo(rx.Base):
    name: str = ""
    type: str = ""


class State(rx.State):
    f: Foo = Foo()

...

rx.cond(State.f.type == "pasta", ...)
```

☝️ this (potentially existing) code will break with this change

---------------

yes, shadowing underscore names is also not ideal, but these are less likely to be used in user-facing code.

masenf (changes requested):

reflex/vars.py
ElijahAhianyo: > Recommend using an under prefix name for any non-operator function defined in the `Var` class to avoid shadowing potential field names in objects represented by the Var that are accessed via `__getattr__`. This is why `name` and friends were renamed to `_var_name`
> 
> ```python
> class Foo(rx.Base):
>     name: str = ""
>     type: str = ""
> 
> 
> class State(rx.State):
>     f: Foo = Foo()
> 
> ...
> 
> rx.cond(State.f.type == "pasta", ...)
> ```
> 
> ☝️ this (potentially existing) code will break with this change
> 
> yes, shadowing underscore names is also not ideal, but these are less likely to be used in user-facing code.

got it, What does that mean for existing methods like `reverse`, `length`, etc though? Do those need to converted to prefixed-underscores as well?

masenf: > What does that mean for existing methods like `reverse`, `length`, etc though?

They probably should be renamed, but we haven't really gotten any complaints about shadowing those yet and they're already in the public API, so a frivolous rename would just break code.

I _do_ think before 1.0, we need a general solution to this problem, because the underscore names don't look great in IDE (like pycharm underlines them super ugly), and people might not think they can use the methods safely due to the underscore prefix.

I don't know what that solution looks like yet.

For now, lets just try to avoid shadowing additional names and plan to sort it out after the radix dust settles.

reflex/vars.py
```diff
@@ -87,6 +87,15 @@
     "deps": "_deps",
 }
 
+PYTHON_JS_TYPE_MAP = {
+    (int, float): "number",
+    (str,): "string",
+    (bool,): "boolean",
+    (list, tuple): "Array",
+    (dict,): "Object",
+    (None,): "undefined",
```
masenf, line 96: suggest adding a test for the `None` case. i was under the impression that in most circumstances a `None` value mapped to JS `null` rather than `undefined`, but I haven't tested it personally.

reflex/vars.py
```diff
@@ -843,7 +852,17 @@ def length(self) -> Var:
...
 
-    def __eq__(self, other: Var) -> Var:
+    def _type(self) -> Var:
+        """Get the type of the Var in Javascript.
+
+        Returns:
+            A var representing the type check.
+        """
+        return self._replace(
+            _var_name=f"typeof {self._var_name}", _var_type=str, _var_is_string=False
```
masenf, line 862:
```suggestion
            _var_name=f"typeof {self._var_full_name}", _var_type=str, _var_is_string=False, _var_full_name_needs_state_prefix=False
```

I think we should use `_var_full_name` when formatting the `typeof` expression and then explicitly indicate that the resulting var should no longer use the state prefix (since now it starts with `typeof` instead of the `_var_name`)

masenf (commented):
couple more suggests

masenf (approved):

PR #2453, opened 2024-01-25
use multiprocessing to speed up pyi generation

benedikt-bartscher:
```bash
# main
python scripts/pyi_generator.py  23,71s user 2,06s system 107% cpu 23,960 total

# this PR
python scripts/pyi_generator.py  79,24s user 94,34s system 2111% cpu 8,220 total
```

PR #2456, opened 2024-01-25
Pyi generator checksums

benedikt-bartscher:
Based on #2453

@masenf fast & efficient is even better

```bash
# first run
python scripts/pyi_generator.py  80,61s user 92,46s system 2084% cpu 8,301 total

# second run
python scripts/pyi_generator.py  1,06s user 1,99s system 290% cpu 1,051 total
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ examples/
 .venv
 venv
 requirements.txt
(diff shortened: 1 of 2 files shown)


masenf (commented):
definitely like the `git` approach more 🙏

PR #2458, opened 2024-01-26
wip event handlers for state mixins

benedikt-bartscher:
```python
import reflex as rx


class Mixin(rx.Base):
    counter: int = 0

    def increment(self):
        self.counter += 1

    def decrement(self):
        self.counter -= 1


class State(Mixin, rx.State):
    #  counter: int = 0
    counter2: int = 0

    def increment2(self):
        self.counter2 += 1

    def decrement2(self):
        self.counter2 -= 1


def index() -> rx.Component:
    return rx.fragment(
        rx.vstack(
            rx.heading(State.counter),
            rx.button("Decrement", on_click=State.decrement),
        ),
        rx.vstack(
            rx.heading(State.counter2),
            rx.button("Increment2", on_click=State.increment2),
            rx.button("Decrement2", on_click=State.decrement2),
        ),
    )


app = rx.App()
app.add_page(index)
```


benedikt-bartscher:
@masenf the example above does work now.
However, I think there is still one bug with more complicated inheritance structures. I added a simple failing test in `integration/test_state_inheritance.py`. Any ideas?

PR #2460, opened 2024-01-26
pyi generator git diff

benedikt-bartscher:
Alternative implementation of #2456  
You can choose which one you like more


scripts/pyi_generator.py
```diff
@@ -658,6 +660,12 @@ def scan_all(self, targets):
                     for file in files:
                         if file in EXCLUDED_FILES or not file.endswith(".py"):
                             continue
+                        if (
+                            changed_files is not None
+                            and os.path.relpath(os.path.join(root, file), ".")
```
masenf, line 665: overall i'd prefer to see _less_ usage of the older `os` APIs which aren't as nice to work with as `pathlib.Path`.

This can be written as `(Path(root) / file).relative_to(Path(".").resolve())`.

Which similar join already appears on line 669 so that could become a loop-local var and `Path(".").resolve()` isn't going to change at runtime, so that could be done once outside the loop.

scripts/pyi_generator.py
```diff
@@ -681,8 +689,32 @@ def generate_init():
     logging.basicConfig(level=logging.DEBUG)
     logging.getLogger("blib2to3.pgen2.driver").setLevel(logging.INFO)
 
+    try:
+        with open(LAST_RUN_COMMIT_SHA_FILE, "r") as f:
+            last_run_commit_sha = f.read().strip()
```
masenf, line 694:
```python
Path(LAST_RUN_COMMIT_SHA_FILE).read_text().strip()
```

scripts/pyi_generator.py
```diff
@@ -681,8 +689,32 @@ def generate_init():
...
     targets = sys.argv[1:] if len(sys.argv) > 1 else ["reflex/components"]
     logger.info(f"Running .pyi generator for {targets}")
     gen = PyiGenerator()
-    gen.scan_all(targets)
+    gen.scan_all(targets, changed_files)
     generate_init()
+
+    current_commit_sha = os.popen("git rev-parse HEAD").read().strip()
+    with open(LAST_RUN_COMMIT_SHA_FILE, "w") as f:
+        f.write(current_commit_sha)
```
masenf, line 720:
```python
Path(LAST_RUN_COMMIT_SHA_FILE).write_text(current_commit_sha)
```

masenf (changes requested):
i think the approach here is sound, just some comments on api choice

benedikt-bartscher:
@masenf thanks for pointing me to those new apis! I changed those `os` calls and cleaned up a bit.

benedikt-bartscher:
@masenf i found one edge case when a `.py` file is first edited and then restored. This is now handled aswell.

scripts/pyi_generator.py
```diff
@@ -638,27 +687,44 @@ def _scan_file(self, module_path: Path):
...
-            path = Path(target)
-            if target.endswith(".py") and path.is_file():
-                file_targets.append(path)
-            elif path.is_dir():
-                for root, _, files in os.walk(path):
+            target_path = Path(target)
+            if target_path.is_file() and target_path.suffix == ".py":
+                file_targets.append(target_path)
+            elif target_path.is_dir():
+                for root, _, files in os.walk(target_path):
```
benedikt-bartscher, line 707: https://docs.python.org/3/library/pathlib.html#pathlib.Path.walk

masenf: yeah, i'm slowly getting the whole team on board with the new APIs, we currently have a mixture in use, depending on which file you're in.

masenf (commented):
i thought i left a comment on this yesterday, but i'm not seeing it now...

consider `subprocess.run` over `os.popen`; yes you'll have to specify `capture_output=True` and `encoding="utf-8"`, but you can specify the command as an arg list instead of a string, which alleviates some potential shell quoting concerns.

The other thing about `os.popen` is that it returns files handles which must be closed, but they are not being closed here. Additionally, if the output is large, using this method can deadlock the process if the pipe fills up (i think).

scripts/pyi_generator.py
benedikt-bartscher: sadly, `Path.walk` is only available since python 3.12. However, I still migrated this line to pathlib using a different approach

PR #2474, opened 2024-01-29
improve sqlalchemy type parsing

benedikt-bartscher:
Part of #2340  

This PR addresses 3 issues:
- If using sqlalchemy>2.0 one would probably want to benefit from its excellent typing features. However, for many-to-many relations across modules, one would need to use PEP 563 (`from __future__ import annotations`). This breaks reflex's `get_attribute_access_type` because it cannot evaluate those postponed annotations with `get_type_hints`. This PR makes use of sqlalchemys `inspect` which allows us to access the correct python types with ease.  
- support for nested types with sqlalchemy relations
- general support to parse type hints of properties

### DIFF
--- a/reflex/utils/types.py
+++ b/reflex/utils/types.py
@@ -173,10 +173,6 @@ def get_attribute_access_type(cls: GenericType, name: str) -> GenericType | None
             if isinstance(type_, ModelField):
                 return type_.type_  # SQLAlchemy v1.4
             return type_
-        if name in cls.__dict__:
-            value = cls.__dict__[name]
-            if hint := get_property_hint(value):
-                return hint
     elif is_union(cls):
         # Check in each arg of the annotation.
         for arg in get_args(cls):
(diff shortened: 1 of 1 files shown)


reflex/utils/types.py
```diff
@@ -129,7 +148,21 @@ def get_attribute_access_type(cls: GenericType, name: str) -> GenericType | None
             # Ensure frontend uses null coalescing when accessing.
             type_ = Optional[type_]
         return type_
-    elif isinstance(cls, type) and issubclass(cls, (Model, DeclarativeBase)):
+    elif isinstance(cls, type) and issubclass(cls, DeclarativeBase):
+        insp = inspect(cls)
```
masenf, line 152: can we actually `import sqlalchemy` and then access this via `sqlalchemy.inspect` for readability

masenf (changes requested):
just a quick suggestion, then we can take this. everything else looks good. thanks for the contribution 😀

benedikt-bartscher:
Yes, especially in that context, someone might think it is another inspect. Thanks for your quick review!

masenf (approved):

PR #2480, opened 2024-01-29
Rework telemetry to support installation_id

jackie-pc:


reflex/reflex.py
```diff
@@ -80,6 +81,10 @@ def _init(
 
     prerequisites.check_latest_package_version(constants.Reflex.MODULE_NAME)
 
+    prerequisites.initialize_reflex_user_directory()
+
+    ensure_reflex_distinct_id()
```
picklelo, line 86: Can this just go inside of `initialize_reflex_user_directory` to keep everything in the prerequisites file?

reflex/utils/prerequisites.py
```diff
@@ -824,10 +824,44 @@ def validate_frontend_dependencies(init=True):
         validate_bun()
 
 
-def initialize_frontend_dependencies():
-    """Initialize all the frontend dependencies."""
+def ensure_reflex_distinct_id() -> int:
+    """Ensures that a reflex distinct id has been generated and stored in the reflex directory.
+
+    Returns:
+        Distinct id (int).
```
picklelo, line 831: nit: don't need to annotate return types since we use annotations

reflex/utils/prerequisites.py
```diff
@@ -824,10 +824,44 @@ def validate_frontend_dependencies(init=True):
...
+    """
+    initialize_reflex_user_directory()
+    distinct_id_file = os.path.join(constants.Reflex.DIR, "distinct_id")
```
picklelo, line 834: Maybe call the file `reflex.json` to match how we do it in the `.web` folder, we can reuse the file parsing/writing functions for it too then

reflex/reflex.py
jackie-pc: It is in the prerequisite file.  Removed the separate import.  It will say `prerequisites.ensure_reflex_distinct_id()` now.

reflex/utils/prerequisites.py
jackie-pc: There isn't really any existing logic, it's all raw json.loads into dictionary, etc.

The reason I was leaning to a flat, single purpose file is for simplicity.  Without any further things to be stashed, it would be better to keep it minimal.

We can discuss the naming further - we can finalize a name on Discord and just update it to that.

reflex/utils/prerequisites.py
```diff
@@ -824,10 +824,44 @@ def validate_frontend_dependencies(init=True):
...
+        except Exception:
+            # If anything goes wrong at all... just regenerate.
+            # Like what? Examples:
+            #     - file not exists
+            #     - file not readable
+            #     - content not parseable as an int
+            pass
+
+    if installation_id is None:
+        installation_id = random.getrandbits(128)
```
picklelo, line 850: Should we use uuid4 instead of random bits in both places?

picklelo (dismissed):

reflex/reflex.py
```diff
@@ -80,6 +80,10 @@ def _init(
 
     prerequisites.check_latest_package_version(constants.Reflex.MODULE_NAME)
 
+    prerequisites.initialize_reflex_user_directory()
+
+    prerequisites.ensure_reflex_installation_id()
```
masenf, line 85: should we do this, even if the user has disabled telemetry? my thought is `TELEMETRY_DISABLED=1` and we don't generate a marker file at all

reflex/utils/prerequisites.py
```diff
@@ -824,10 +824,44 @@ def validate_frontend_dependencies(init=True):
...
+        with open(installation_id_file, "w") as f:
+            f.write(str(installation_id))
```
masenf, line 852: If we cannot write out the file for some reason, i don't think reflex should just raise an exception and exit, that seems like a potentially poor UX

masenf (changes requested):

reflex/utils/prerequisites.py
jackie-pc: The thinking is that this is not supposed to be a UX at all.  What is the benefit in interrupting the user's workflow for telemetry?

From this viewpoint, the most we should do is perhaps add a `console.log.debug`.

reflex/reflex.py
jackie-pc: Currently, `project_hash` has exactly the same nature - it is only used for telemetry, and yet we always generate it even if `TELEMETRY_DISABLED=1`.  I.e. we persist something, but we just never send out.

Checking for the flag costs some incremental complexity.  Note even if we did do it, it would not be clean.  Suppose telemetry was only disabled after initial use.  The file would have already been created.

Don't really feel strongly either way about it - if others do, of course we can just add the check!

reflex/utils/prerequisites.py
jackie-pc: I went with what was there already.  It seems acceptable for our purpose.  What would drive us to change it to uuid?

reflex/reflex.py
picklelo: I think we can keep the file, as long as we never send the data when the flag is enabled

picklelo: Since `ensure_reflex_installation_id` already calls `initialize_reflex_user_directory()` do we need both calls? that's what I was thinking originally - just calling `ensure_reflex_installation_id` at the end of `initialize_` every time

picklelo (approved):

masenf (approved):

PR #2481, opened 2024-01-29
[REF-1632] Apply `rx.App` styles to Radix Themes root `div`

masenf:
Restore default radix themes theme specification without regressing global style rule functionality.

There were two problems before:
  * global styles were only applied by the `ChakraProvider`, so if chakra components were not used, then no global style rules were applied
  * radix applies default rules to the `.radix-themes` class, which has a higher specificity than "body type" rules originating from the global styles; so even if they are set (like `body { font-family: monospace }`), the `.radix-themes` container will override the `font-family` with the `--default-font-family` defined by radix anyway.

This fix works around the issue by using emotion to apply the global css rules for the `:root` and `body` selectors directly to the radix `Theme` component via the `css` prop.

This has one unfortunate side effect w.r.t tailwind compatibility: global style rules, like `p { color: blue }` can no longer be overridden by tailwind utility classes, because the single utility class has a lower specificity than the "theme class + type" selector that is generated to override radix themes defaults.

The workaround for that is to avoid mixing global styles and tailwind or to explicitly set `rx.App(theme=None)` to disable radix themes.

### DIFF
--- a/integration/test_tailwind.py
+++ b/integration/test_tailwind.py
@@ -10,14 +10,13 @@ from reflex.testing import AppHarness
 
 PARAGRAPH_TEXT = "Tailwind Is Cool"
 PARAGRAPH_CLASS_NAME = "text-red-500"
-GLOBAL_PARAGRAPH_COLOR = "rgba(0, 0, 242, 1)"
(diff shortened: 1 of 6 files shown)


picklelo (approved):
nice!

PR #2484, opened 2024-01-30
Radix props literals as ints and strs

ElijahAhianyo:
With this PR, props of radix components with string ints should accept ints in the component API. 

```python
rdxt.heading("This ia a heading", size="8")
rdxt.heading("This ia another heading", size=8)

```

This PR doesnt handle the case of Vars as values of literal prop types yet, typical eg is the case of `rx.cond` and `rx.match`
```python
rdxt.heading("This ia a heading", size=rx.cond(True, "8", 9))
```
There should be a follow up PR to look into this


reflex/components/radix/themes/base.py
```diff
@@ -210,3 +212,61 @@ class RadixThemesColorModeProvider(Component):
     library = "/components/reflex/radix_themes_color_mode_provider.js"
     tag = "RadixThemesColorModeProvider"
     is_default = True
+
+
+class RadixThemesComponentPropsOverride(Component):
```
masenf, line 217: maybe we can make this just a plain mixin class (i.e. not a subclass of Component)

and then override `__init__` to do the conversions before passing off to `super().__init__`.

That way we wouldn't need to redefine `create`.

The RadixThemesComponentPropsOverride would have to be named first in the list of parent classes though, to ensure an actual `Component` `__init__` didn't get called first.

reflex/components/radix/themes/base.py
```diff
@@ -11,7 +11,9 @@
 
 LiteralAlign = Literal["start", "center", "end", "baseline", "stretch"]
 LiteralJustify = Literal["start", "center", "end", "between"]
-LiteralSize = Literal["1", "2", "3", "4", "5", "6", "7", "8", "9"]
+LiteralSize = Literal[
+    "1", "2", "3", "4", "5", "6", "7", "8", "9", 1, 2, 3, 4, 5, 6, 7, 8, 9
```
masenf, line 15: i don't really like this, because now the autocomplete list seems to have duplicates in it; the generated docs will show both the str and the int values, and it's kind of clunky.

i would suggest just keeping the str values, but doing the conversion to `str` under the hood if the user provides an `int`

masenf (changes requested):

reflex/components/radix/themes/base.py
ElijahAhianyo: I had it this way because I thought the idea was to support strs and ints in the API?

masenf:
i'm thinking we should close this with radix 3.0 on the horizon that allows regular css values for these props as well. maybe we can rethink it then?

PR #2489, opened 2024-01-30
simple pytest benchmark for measuring event <=> state update round trip time

jackie-pc:
This is an integration test that sets up a minimal one-button app, with varying number of vars on State.

Each button click increments one of the vars.  The var update round trips back to the button as the button's content.

Spoiler alert - there is no material impact on round trip times from having large states (e.g. benchmark tries up to 10000 vars). This is with Redis state manager backed by local redis server.

image

### DIFF
--- a/integration/test_large_state.py
+++ b/integration/test_large_state.py
@@ -1,0 +1,90 @@
+"""Test large state."""
+import time
+
+import jinja2
+import pytest
+from selenium.webdriver.common.by import By
+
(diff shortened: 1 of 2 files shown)


picklelo (approved):
Awesome, good to hear for our use cases

masenf (approved):
appharness instance and selenium itself have some polling helpers that might be cleaner than the polling loops, but i'm fine with this as it is.

PR #2490, opened 2024-01-31
Support reflex app creation from templates from github

martinxu9:
## Summary
- Support reflex app creation from templates from github
  - When reflex init: fetch from reflex backend server: the list of Reflex apps that can be used as a starting template  (name and their URL). When no internet or backend is unreachable, only `blank` is shown to user.
  - Prompt user for template selection: default is `blank`.
  - Upon valid selection, download the source files as a zip. The Zip structure must contain a top level directory, which contains the rxconfig. This is the standard structure downloaded from github repos.
  - Extract all files from zip, copy files into current directory
  - Prompt for name change, by default the name is the valid reflex app name constructed by current directory name (e.g. no dashes allowed).
  - Rename folders/files, fix imports
- Not supported:
  - Download a sub-directory in a github repo as standalone zip.

Usage:
`reflex init --template https://github.com/reflex-dev/reflex-chat`

## Tests
- [x] Tested with CP running https://github.com/reflex-dev/reflex-hosting/pull/232 locally with following (**below are not the final/real templates**):
   - `blank_template`:  https://github.com/reflex-dev/blank-template/archive/main.zip
   - `chat_app`: https://api.github.com/repos/reflex-dev/reflex-chat/zipball/ee5b472


reflex/utils/prerequisites.py
```diff
@@ -942,3 +968,151 @@ def migrate_to_reflex():
...
+        template_name_to_url: The dictionary of template names to download URLs.
+
+    Raises:
+        Exit: If template is directly provided in the command flag and is invalid.
+
+    Returns:
+        A valid template name.
+    """
+    if template is not None:
+        # If user selects a template, it needs to exist
```
picklelo, line 1015: nit: end comments with periods for consistency

reflex/utils/prerequisites.py
```diff
@@ -942,3 +968,151 @@ def migrate_to_reflex():
...
+        raise typer.Exit(1) from uze
+
+    # Move the rxconfig file here first
+    path_ops.mv(str(template_dir / constants.Config.FILE), constants.Config.FILE)
+    new_config = get_config(reload=True)
+    # Get the template app's name from rxconfig in case it is different than
+    # the source code repo name on github
+    template_name = new_config.app_name
+
+    # Prompt for Reflex App name change
```
picklelo, line 1092: Should we move this section to its own function? I think this may in the future be useful as its own subcommand `reflex init --rename <new_name>`

picklelo (commented):
Works well for me (though I wasn't able to connect to get the list of templates). Maybe for now we can leave out the interactive init and support initialization with the `--template` flag only. I think I they do `--template` without any args, that's where we could show all the templates we offer through the CLI.

reflex/utils/prerequisites.py
martinxu9: remove this section for now. Let me do that when we add --rename. The rename is slightly differently I suppose. Here, we basically generate a new rxconfig file instead regex replace any strings. I suspect we might have to do string substitution when it a generic app. There might be other attributes added in the template app.

martinxu9:
> Works well for me (though I wasn't able to connect to get the list of templates). Maybe for now we can leave out the interactive init and support initialization with the `--template` flag only. I think I they do `--template` without any args, that's where we could show all the templates we offer through the CLI.

do you know how to achieve that in typer option? From what I experimented, if it's not a boolean flag, it'll expect something set for that flag or no flag at all for None.

reflex/utils/prerequisites.py
```diff
@@ -1370,12 +1350,17 @@ def initialize_app(app_name: str, template: str | None = None):
         return
 
     # Get the available templates
-    templates = fetch_app_templates() or {}
-    if template is None:
-        template = prompt_for_template(templates)
+    templates: dict[str, Template] = fetch_app_templates() or {}
```
martinxu9, line 1353: nit: `fetch_app_templates` always return a dict, no need for `or {}`

reflex/utils/prerequisites.py
```diff
@@ -1178,3 +1216,166 @@ def migrate_to_reflex():
...
+    else:
+        # Fetch App templates from the backend server.
+        console.debug(f"Available templates: {templates}")
+
+        # If user selects a template, it needs to exist.
+        if template in templates:
+            template_url = templates[template].code_url
+        else:
+            # Check if the template is a github repo.
+            if "github.com" not in template:
```
martinxu9, line 1371: should it be a more strict check something like it must start with `https://github.com/` or something?

reflex/utils/prerequisites.py
```diff
@@ -1178,3 +1216,166 @@ def migrate_to_reflex():
...
+                console.error(f"Template `{template}` not found.")
+                raise typer.Exit(1)
+            repo = template.split("github.com/")[1].strip("/")
+            template_url = f"https://github.com/{repo}/archive/main.zip"
```
martinxu9, line 1375: looks like we expect the template_url to be of `https://github.com/owner/repo/` format, do we want to just right strip `/` then append `main.zip` which also means we can't specify a branch for now. Likely ok

reflex/utils/prerequisites.py
picklelo: good catch, fixed

picklelo (approved):

PR #2492, opened 2024-01-31
[0.4.0] Namespace the Drawer primitive subcomponents

masenf:
Ensure `Drawer` component uses a `SimpleNamespace` like the other primitives.

### DIFF
--- a/reflex/components/radix/primitives/drawer.py
+++ b/reflex/components/radix/primitives/drawer.py
@@ -3,6 +3,7 @@
 # Style based on https://ui.shadcn.com/docs/components/drawer
 from __future__ import annotations
 
+from types import SimpleNamespace
 from typing import Any, Dict, List, Literal, Optional, Union
 
 from reflex.components.radix.primitives.base import RadixPrimitiveComponentWithClassName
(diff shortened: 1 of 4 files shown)


reflex/components/radix/primitives/drawer.py
```diff
@@ -230,11 +231,17 @@ def _get_style(self) -> dict:
...
 
-drawer_root = DrawerRoot.create
-drawer_trigger = DrawerTrigger.create
-drawer_portal = DrawerPortal.create
-drawer_content = DrawerContent.create
-drawer_overlay = DrawerOverlay.create
-drawer_close = DrawerClose.create
-drawer_title = DrawerTitle.create
-drawer_description = DrawerDescription.create
+class Drawer(SimpleNamespace):
```
picklelo, line 234: I think we can make a subclass of SimpleNamespace such that we can define these like:

```python
class Drawer(NamespaceSubclass):
    root = DrawerRoot
    trigger = DrawerTrigger
    ...
```

picklelo (approved):
LGTM - an idea on how we can make defining the namespaces cleaner

reflex/components/radix/primitives/drawer.py
masenf: that would break static analysis and autocomplete i think. i tried making a fancy class first

masenf: to be clear, i'm not opposed to the idea per-se, but it might be trickier to get the typing right and would definitely need some help from `pyi_generator`.

i'd prefer to go with the simpler, expliciter route for now, but after we get this all out the door, i'm not opposed to a refactor or cleanup of the namespace classes

PR #2493, opened 2024-01-31
rx.download accepts `data` arg as either str or bytes

tgberkeley:
```
class State(rx.State):
    """The app state."""


    #Define the data as a dictionary
    data = {
        'Column1': [1, 2, 3, 4],
        'Column2': ['A', 'B', 'C', 'D'],
        'Column3': [True, False, True, False]
    }

    # Create a DataFrame
    df = pd.DataFrame(data)

    my_csv = df.to_csv()

def index() -> rx.Component:
      return (
rdxt.button("Download!", on_click=rx.download(filename="datafile.csv", data=State.my_csv,)),
)
```

# Update

now supports

```python
from pathlib import Path

from PIL import Image, ImageDraw
import pandas as pd
import reflex as rx

width, height = 200, 200


class State(rx.State):
    #Define the data as a dictionary
    data = {
        'Column1': [1, 2, 3, 4],
        'Column2': ['A', 'B', 'C', 'D'],
        'Column3': [True, False, True, False]
    }

    df = pd.DataFrame(data)

    my_csv = df.to_csv()

    img: Image.Image = Image.new("RGB", (width, height), "blue")

    def download_text(self):
        return rx.download(data="Hello, world!", filename="hello.txt")

    def download_image(self):
        return rx.download(
            data=Path("/Users/masenf/Downloads/reflex_banner.png").read_bytes(),
            filename="reflex_banner.png",
        )

    def download_url(self):
        return rx.download(url="/cool.webp")



def index() -> rx.Component:
    return rx.fragment(
        rx.vstack(
            rx.link("Download text file", on_click=State.download_text),
            rx.link("Download image", on_click=State.download_image),
            rx.link("Download image from URL", on_click=State.download_url),
            rx.link("Download data as CSV", on_click=rx.download(filename="datafile.csv", data=State.my_csv)),
            rx.link("Download data as JSON", on_click=rx.download(filename="data.json", data=State.data)),
            rx.link("Download generated image", on_click=rx.download(filename="generated_image.png", data=State.img)),
            spacing="1.5em",
            font_size="2em",
            padding_top="10%",
        ),
    )


app = rx.App()
app.add_page(index)
```

### DIFF
--- a/reflex/event.py
+++ b/reflex/event.py
@@ -552,7 +552,11 @@ def set_clipboard(content: str) -> EventSpec:
     )
 
 
-def download(url: str | Var | None = None, filename: Optional[str | Var] = None, data: str | None = None) -> EventSpec:
+def download(
+    url: str | Var | None = None,
+    filename: Optional[str | Var] = None,
+    data: str | None = None,
+) -> EventSpec:
     """Download the file at a given path.
 
     Args:
@@ -566,9 +570,6 @@ def download(url: str | Var | None = None, filename: Optional[str | Var] = None,
     Returns:
         EventSpec: An event to download the associated file.
     """
-    if isinstance(url, Var) and filename is None:
-        filename = ""
-
     if isinstance(url, str):
         if not url.startswith("/"):
             raise ValueError("The URL argument should start with a /")
@@ -576,32 +577,17 @@ def download(url: str | Var | None = None, filename: Optional[str | Var] = None,
         # if filename is not provided, infer it from url
         if filename is None:
             filename = url.rpartition("/")[-1]
-    
-    if url is None:
-        if filename is None:
-            filename = "download.txt"
 
-        return server_side(
-            "_download",
-            get_fn_signature(download),
-            filename=filename,
-            data=data,
-        )
+    if filename is None:
+        filename = ""
 
-    if data is None:
-        return server_side(
-            "_download",
-            get_fn_signature(download),
-            url=url,
-            filename=filename,
-        )
+    args = {"url": url, "filename": filename, "data": data}
+    filtered_args = {k: v for k, v in args.items() if v is not None}
 
     return server_side(
         "_download",
         get_fn_signature(download),
-        url=url,
-        filename=filename,
-        data=data,
+        **filtered_args,
     )
 
 
--- a/reflex/event.py
+++ b/reflex/event.py
@@ -552,12 +552,13 @@ def set_clipboard(content: str) -> EventSpec:
     )
 
 
-def download(url: str | Var, filename: Optional[str | Var] = None) -> EventSpec:
+def download(url: str | Var | None = None, filename: Optional[str | Var] = None, data: str | None = None) -> EventSpec:
     """Download the file at a given path.
 
     Args:
         url : The URL to the file to download.
         filename : The name that the file should be saved as after download.
+        data : The data to download.
 
     Raises:
         ValueError: If the URL provided is invalid.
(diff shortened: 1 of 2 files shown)


reflex/event.py
```diff
@@ -575,12 +576,32 @@ def download(url: str | Var, filename: Optional[str | Var] = None) -> EventSpec:
...
+            data=data,
+        )
+
+    if data is None:
+        return server_side(
+            "_download",
+            get_fn_signature(download),
+            url=url,
+            filename=filename,
+        )
```
masenf, line 597: why are these multiple calls needed? if you pass `None` through the event how does it come out on the js side? if it comes as a `null`, then maybe you can change the `if` in `state.js` to check for that.

I think this function would be cleaner if there was only a single call to `server_side`.

reflex/event.py
```diff
@@ -575,12 +576,32 @@ def download(url: str | Var, filename: Optional[str | Var] = None) -> EventSpec:
         # if filename is not provided, infer it from url
         if filename is None:
             filename = url.rpartition("/")[-1]
+    
+    if url is None:
+        if filename is None:
+            filename = "download.txt"
```
masenf, line 582: is this necessary? up above on line 570, if the filename isn't specified, it's just passed as an empty string.

maybe this whole check could be moved up (outside the `if url is None`) and replace the `if isinstance(url, Var) and filename is None:` bit?

what i'm thinking is remove that check on line 569/570, and move these two lines just above the `if url is None`. That way we'll set `filename` if we have a URL, otherwise, if it remains unspecified, then it can be set to `""` (or `"download.txt"`, i guess, but i'm still not a big fan of making up a name like that)

masenf (changes requested):

reflex/event.py
tgberkeley: Nice spot, changed this

tgberkeley: and removed the download.txt to just have "" now

reflex/event.py
tgberkeley: have changed this to use a dict that only passes the arguments through if they are not none

reflex/event.py
tgberkeley: turns out having an empty string leads to the download failing on the javascript side as an if statement is not run through to do the download so will stick with download.txt

tgberkeley: changed so now can be "" again by removing some of the JS code

reflex/event.py
```diff
@@ -552,21 +553,25 @@ def set_clipboard(content: str) -> EventSpec:
...
+    url: str | Var | None = None,
+    filename: Optional[str | Var] = None,
+    data: str | bytes | Var | None = None,
+) -> EventSpec:
+    """Download the file at a given path or with the specified data.
 
     Args:
         url : The URL to the file to download.
         filename : The name that the file should be saved as after download.
+        data : The data to download.
```
picklelo, line 566: nit: we shouldn't have spaces before the colons

reflex/event.py
```diff
@@ -576,6 +581,27 @@ def download(url: str | Var, filename: Optional[str | Var] = None) -> EventSpec:
         if filename is None:
             filename = url.rpartition("/")[-1]
 
+    if filename is None:
```
picklelo, line 584: Can we add a couple comments in this function saying what the different url/encodings mean

picklelo (approved):

picklelo (approved):
awesome!

masenf (approved):

PR #2495, opened 2024-01-31
Radix Components Valid children/parents

ElijahAhianyo:
This PR updates the necessary radix components with their valid children or parents.


picklelo (approved):
Looks good to me

PR #2498, opened 2024-01-31
Support component create methods as keys in global styles

picklelo:
Before you had to use the class name, such as `rx.Text` to specify a global style. With this change, you can use the create method, i.e. `rx.text`

Example code:

```python
import reflex as rx


def content():
    return rx.box(
        rx.heading("Welcome to My App"),
        rx.text("This is the main content of the page."),
    )


def sidebar():
    return rx.vstack(
        rx.image(src="/favicon.ico", width="3em"),
        rx.heading("Sidebar", margin_bottom="1em"),
        position="fixed",
        height="100%",
        left="0px",
        top="0px",
        z_index="5",
        padding_x="2em",
        padding_y="1em",
        background_color="lightgray",
        align_items="left",
        width="250px",
    )

def index():
    return rx.fragment(
        sidebar(),
        rx.container(content(), max_width="60em", margin_left="250px"),
    )

app = rx.App(
    style={
        rx.text: {
            "color": "blue",
        },
        rx.heading: {
            "font_size": "2em",
            "_hover": {
                "color": "red",
            }
        },
        rx.container: {
            "padding": "2em",
        }
    }
)
app.add_page(index)
```

### DIFF
--- a/reflex/compiler/utils.py
+++ b/reflex/compiler/utils.py
@@ -2,7 +2,7 @@
 from __future__ import annotations
 
 import os
-from typing import Any, Type
+from typing import Any, Callable, Type
 from urllib.parse import urlparse
 
 from pydantic.fields import ModelField
@@ -290,7 +290,7 @@ def create_theme(style: ComponentStyle) -> dict:
         The base style for the app.
     """
     # Get the global style from the style dict.
-    style_rules = Style({k: v for k, v in style.items() if not isinstance(k, type)})
+    style_rules = Style({k: v for k, v in style.items() if not isinstance(k, Callable)})
 
     root_style = {
         # Root styles.
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -115,7 +115,7 @@ class BaseComponent(Base, ABC):
 
 
 # Map from component to styling.
-ComponentStyle = Dict[Union[str, Type[BaseComponent]], Any]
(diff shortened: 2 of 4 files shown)


reflex/compiler/utils.py
```diff
@@ -290,7 +290,7 @@ def create_theme(style: ComponentStyle) -> dict:
         The base style for the app.
     """
     # Get the global style from the style dict.
-    style_rules = Style({k: v for k, v in style.items() if not isinstance(k, type)})
+    style_rules = Style({k: v for k, v in style.items() if not isinstance(k, Callable)})
```
masenf, line 293:
```suggestion
    style_rules = Style({k: v for k, v in style.items() if not isinstance(k, (type, Callable))})
```

lets support both, so we don't break existing apps

we can even have a thing like

```python
for style_key in style_rules:
    if isinstance(style_key, type):
        console.deprecate(
            feature_name="Component classes in global styles",
            reason="Use the lowercase component function, like rx.text",
            deprecation_version="0.3.10",
            removal_version="1.0",
        )
```

masenf (changes requested):

reflex/compiler/utils.py
picklelo: Agreed that we don't want to break any existing code

masenf: ah okay, sounds good 👍

masenf (approved):

PR #2504, opened 2024-01-31
radix Theme panel_background prop: transparent -> translucent

martinxu9:
## Summary
- radix Theme panel_background prop: transparent -> translucent


masenf (approved):
ohhhh, is that why this prop never seemed to work? good catch

PR #2511, opened 2024-02-01
Map Accordion color schemes to radix colors

ElijahAhianyo:
Currently, there are only two color schemes for the accordion component: `primary` and `accent`. The goal of this PR 
 is to map the color schemes for the accordion components to radix color_schemes.

### DIFF
--- a/reflex/components/radix/primitives/accordion.py
+++ b/reflex/components/radix/primitives/accordion.py
@@ -336,7 +336,7 @@ class AccordionRoot(AccordionComponent):
     variant: Var[LiteralAccordionRootVariant] = "classic"  # type: ignore
 
     # The color scheme of the accordion.
-    color_scheme: Var[LiteralAccentColor] = "green"  # type: ignore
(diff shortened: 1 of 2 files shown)


reflex/components/radix/primitives/accordion.py
```diff
@@ -409,7 +336,7 @@ class AccordionRoot(AccordionComponent):
     variant: Var[LiteralAccordionRootVariant] = "classic"  # type: ignore
 
     # The color scheme of the accordion.
-    color_scheme: Var[LiteralAccordionRootColorScheme] = "primary"  # type: ignore
+    color_scheme: Var[LiteralAccentColor] = "green"  # type: ignore
```
masenf, line 339:
```suggestion
    color_scheme: Var[LiteralAccentColor]  # type: ignore
```

i dont think there should be a default here. if `color_scheme` is unspecified, then it should use the theme color right?

masenf (changes requested):

masenf (approved):

PR #2517, opened 2024-02-02
[REF-1827]Meta data for conditional Vars

ElijahAhianyo:
## PROBLEM STATEMENT

Currently for Vars, once an `rx.cond` is computed, there's no clean way to extract the constituent parts of the `cond` should you need it later as they will be lost. 
Eg.
`cond_var = rx.cond(True, "true-condition", "false-condition")`
becomes
```
cond_var= BaseVar(_var_name='isTrue(true) ? `true-condition` : `false-condition`', _var_type=<class 'str'>, _var_is_local=False, _var_is_string=False, _var_full_name_needs_state_prefix=False, _var_data=VarData(state='', imports={'/utils/state': [ImportVar(tag='isTrue', is_default=False, alias=None, install=True, render=True), ImportVar(tag='isTrue', is_default=False, alias=None, install=True, render=True)]}, hooks=set()))
```

A typical use case on why we need the constituent parts of the cond var after its been converted to a BaseVar is checking the validity of literal prop types when a Var is passed.

Take in the example:

```python
class Heading(RadixThemesComponent):
      size: Literal["1", "2", "3", "4", "5", "6", "7", "8", "9"]
```
```python
...
heading("Header value", size=rx.cond(True, "5", "9"))
```
The true and false values are lost after conversion to a `BaseVar` which makes it harder to validate values against the accepted literal values of the `size` prop.

## SOLUTION

This PR adds an extra `_var_cond_data` to the `Var` data type. Which stores meta data for conditional vars(`rx.match` and `rx.cond`)

### DIFF
--- a/reflex/components/core/cond.py
+++ b/reflex/components/core/cond.py
@@ -8,7 +8,7 @@ from reflex.components.component import BaseComponent, Component, MemoizationLea
 from reflex.components.tags import CondTag, Tag
 from reflex.constants import Dirs
 from reflex.utils import format, imports
-from reflex.vars import BaseVar, Var, VarData
+from reflex.vars import BaseVar, Var, VarData, ConditionalVar
 
 _IS_TRUE_IMPORT = {
     f"/{Dirs.STATE_PATH}": {imports.ImportVar(tag="isTrue")},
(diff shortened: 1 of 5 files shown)


masenf:
I think we can close this one for now and bring it back later if we end up needing it.

PR #2523, opened 2024-02-03
Top-level namespace tweaks

masenf:
Add missing names to top-level `rx.` namespace.

Move code_block out of chakra subdir.

pyi_generator ignores files starting with `tests`


reflex/__init__.py
```diff
@@ -16,6 +16,7 @@
 
 _ALL_COMPONENTS = [
     # Core
+    "Cond",
```
picklelo, line 19: Why do we need these? I thought we decided to keep the component classes out of the `rx.` namespace

reflex/__init__.py
```diff
@@ -26,6 +27,7 @@
     "debounce_input",
     # Base
     "fragment",
+    "Fragment",
```
picklelo, line 30: Same here

reflex/__init__.py
```diff
@@ -82,26 +85,30 @@
...
     "spacer",
     # "strong" (bold?)
+    "stack",
     "switch",
     "table",
     "tabs",
     "text",
     "text_area",
-    # "text_field" (input)
+    "text_field",
```
picklelo, line 94: This should be `input`, we shouldn't have `rx.text_field`

reflex/__init__.py
masenf: fragment is very useful as a base class for custom functionality and we use it in the wrapping react docs, i think it needs to stay in the top level

reflex/__init__.py
masenf: i had this for the doc page, because `rx.cond` isn't a `create` function, so i couldn't tie it back to the class `rx.Cond`

```console
masen@asmbp21 reflex-web % head docs/library/layout/cond.md
---
components:
    - rx.Cond
---
```

i can try removing it though and getting that reference via `rx.components.core.Cond` instead

reflex/__init__.py
masenf: it is done

picklelo (approved):

PR #2524, opened 2024-02-04
CodeAnt AI: Making Antipattern fixes

Amartyajha:
[CodeAnt AI](https://codeant.ai/): Making Antipattern fixes


Alek99 (changes requested):
It seems to have converted some lambdas to list which is not right there.

PR #2526, opened 2024-02-05
[REF-1743] Implement radix-native color mode switch and button

masenf:
Move `color_mode_cond` to core.

Expose `color_mode_*` names at top-level.


reflex/__init__.py
```diff
@@ -20,7 +20,10 @@
     "foreach",
     "html",
     "match",
-    # "color_mode_cond",
+    "color_mode_cond",
```
picklelo, line 23: Should we namespace these as `color_mode.cond`, `color_mode.switch`, etc. ? Also, maybe we need to standardize on our terminology, since we call it "color_mode" here, but in the radix theme it's called "appearance". I'm in favor of aliasing the radix theme one to match this one.

reflex/components/chakra/forms/__init__.py
```diff
@@ -47,8 +46,6 @@
 from .textarea import TextArea
 from .time_picker import TimePicker
 
-helpers = [
-    "color_mode_cond",
-]
+helpers = []
```
picklelo, line 49: Can remove this now

reflex/components/core/cond.py
```diff
@@ -180,3 +181,20 @@ def cond(condition: Any, c1: Any, c2: Any = None):
         _var_full_name_needs_state_prefix=False,
         merge_var_data=VarData.merge(*var_datas),
     )
+
+
+def color_mode_cond(light: Any, dark: Any = None) -> Var | Component:
```
picklelo, line 186: This could go into `colormode.py` also

reflex/components/radix/themes/colormode.py
```diff
@@ -0,0 +1,100 @@
+"""A switch component for toggling color_mode.
```
picklelo, line 1: I think we should call this `color_mode.py` for consistency with the name

reflex/style.py
```diff
@@ -12,6 +12,9 @@
 
 VarData.update_forward_refs()  # Ensure all type definitions are resolved
 
+LIGHT_COLOR_MODE: str = "light"
+DARK_COLOR_MODE: str = "light"
```
picklelo, line 16: should be "dark"

masenf: lol yup

reflex/components/core/cond.py
masenf: i specifically moved it out here so it could be shared by chakra and radix or used standalone with any component library. `colormode.py` is radix specific

masenf:
@picklelo ready to go

picklelo (approved):

PR #2534, opened 2024-02-06
fix EventHandlers and ComputedVars inherited by mixin classes

benedikt-bartscher:


reflex/state.py
```diff
@@ -352,17 +350,88 @@ def __init_subclass__(cls, **kwargs):
...
+                    cls.computed_vars[newcv._var_name] = newcv
+                    cls.vars[newcv._var_name] = newcv
+                    continue
+                if events.get(name) is not None:
+                    continue
+                if not cls._item_is_event_handler(name, value):
+                    continue
+                if parent_state is not None and parent_state.event_handlers.get(name):
+                    continue
+                value = cls._copy_fn(value)
```
masenf, line 372: what is the value is not a function?

benedikt-bartscher: Should be handled by the `_item_is_event_handler` condition in line 368

PR #2538, opened 2024-02-06
remove radix icons

Lendemor:
remove all remaining use of `radix.icon`.
not only `lucide.icon` are used, accessible with `rx.icon`


masenf (approved):
Looks good, will take into the branch, even with breakage. reflex-web reflex-0.4.0 branch is not passing CI right now anyway.

PR #2542, opened 2024-02-07
update slider so width automatically set and default value automatica…

tgberkeley:
Example:
The `slider_value` state var can equal [40] or 40 or [40, 50] or we can pass these values through directly to the `default_value` prop of slider as non state vars.
```
class State(rx.State):
    """The app state."""

    slider_value = [40]


def index() -> rx.Component:
    return rdxt.theme(
        rdxt.slider(default_value=State.slider_value),
    )

app = rx.App(
    theme=rdxt.theme(
        appearance="light", has_background=True, radius="large", accent_color="teal"
    ),
)
app.add_page(index)
app.compile()

### DIFF
--- a/reflex/components/radix/themes/components/slider.py
+++ b/reflex/components/radix/themes/components/slider.py
@@ -92,8 +92,14 @@ class Slider(RadixThemesComponent):
             The component.
         """
         default_value = props.pop("default_value", [50])
-        if isinstance(default_value, (int, float)):
+
+        if isinstance(default_value, Var):
+            if issubclass(default_value._var_type, (int, float)):
+                default_value = [default_value]
+
+        elif isinstance(default_value, (int, float)):
             default_value = [default_value]
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by masenf — DISMISSED** [review]

File: reflex/components/radix/themes/components/slider.py [hunk 1]
```diff
@@ -35,7 +36,7 @@ class Slider(RadixThemesComponent):
     radius: Var[LiteralRadius]
 
     # The value of the slider when initially rendered. Use when you do not need to control the state of the slider.
-    default_value: Var[List[Union[float, int]]]
+    default_value: Var[Union[List[Union[float, int]], Union[float, int]]]
```
picklelo, line 39: nit: don't need the last `Union`, this can be `Var[Union[List[Union[float, int]], float, int]]`

picklelo (approved):

PR #2546, opened 2024-02-07
rx.color to work with conditional vars

ElijahAhianyo:
The code snippets below should work with this PR:

```python
import reflex as rx


class State(rx.State):
    """Test color state."""

    color: str = "mint"
    shade: int = 4
```

```python
rx.cond(True, rx.color(State.color), rx.color(State.color, 5))
   
```

```python
rx.match(
   "condition",
    ("first", rx.color("mint")),
     ("second", rx.color("tomato", 5)),
     rx.color(State.color, 2), 
 ),
 

 rx.match(
     "condition",
      ("first", rx.color(State.color)), 
      ("second", rx.color(State.color, 5)), 
      rx.color(State.color, 2),  
 ),
```

### DIFF
--- a/reflex/utils/format.py
+++ b/reflex/utils/format.py
@@ -256,9 +256,9 @@ def format_cond(
 
     def create_cond_part_var(cond_part):
         return Var.create_safe(
-                cond_part,
-                _var_is_string=type(cond_part) is str or isinstance(cond_part, Color),
-            )
+            cond_part,
+            _var_is_string=type(cond_part) is str or isinstance(cond_part, Color),
+        )
 
     # Format prop conds.
     if is_prop:
--- a/reflex/utils/format.py
+++ b/reflex/utils/format.py
@@ -10,6 +10,7 @@ import sys
 from typing import TYPE_CHECKING, Any, List, Union
 
 from reflex import constants
+from reflex.constants.colors import Color
 from reflex.utils import exceptions, serializers, types
 from reflex.utils.serializers import serialize
 from reflex.vars import BaseVar, Var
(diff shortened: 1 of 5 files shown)


reflex/utils/format.py
```diff
@@ -253,24 +254,24 @@ def format_cond(
     # Use Python truthiness.
     cond = f"isTrue({cond})"
 
+    def create_cond_part_var(cond_part):
+        return Var.create_safe(
+            cond_part,
+            _var_is_string=type(cond_part) is str or isinstance(cond_part, Color),
```
masenf, line 260: is this to ensure that if someone passes an instance of `Color` directly (instead of using the `color` function) that it still works?

masenf (approved):
on board with this as a specific solution to the new `Color` class, although a more general solution might come later, lets get this in to ensure we have good functionality for the 0.4.0 release

reflex/utils/format.py
ElijahAhianyo: This is actually not needed since rx.color would be a Var by the time it gets here. Will push an update to take this out

reflex/vars.py
```diff
@@ -1612,8 +1612,10 @@ def _var_name_unwrapped(self) -> str:
             if types.is_generic_alias(self._var_type)
             else self._var_type
         )
-
         wrapped_var = str(self)
+
+        if type_ == Any:
```
picklelo, line 1617: Can we add a comment on why this logic is needed

reflex/utils/format.py
```diff
@@ -253,24 +253,21 @@ def format_cond(
     # Use Python truthiness.
     cond = f"isTrue({cond})"
 
+    def create_cond_part_var(cond_part):
```
picklelo, line 256: To simplify we can just call this function `def create_var(value)`

reflex/utils/format.py
```diff
@@ -253,24 +253,21 @@ def format_cond(
...
+        return Var.create_safe(cond_part, _var_is_string=type(cond_part) is str)
+
     # Format prop conds.
     if is_prop:
         if not isinstance(true_value, Var):
-            true_value = Var.create_safe(
-                true_value,
-                _var_is_string=type(true_value) is str,
-            )
+            true_value = create_cond_part_var(true_value)
```
picklelo, line 262: We can do this unconditionally (don't have to do is instance checks) since create_safe will just return the Var itself it its not already a var

reflex/utils/format.py
```diff
@@ -253,24 +253,21 @@ def format_cond(
...
         prop1 = true_value._replace(
             _var_is_local=True,
         )
         if not isinstance(false_value, Var):
-            false_value = Var.create_safe(
-                false_value,
-                _var_is_string=type(false_value) is str,
-            )
+            false_value = create_cond_part_var(false_value)
```
picklelo, line 267: same here don't need the if statement anymore

Alek99 (dismissed):
nice working for me

picklelo (approved):

PR #2548, opened 2024-02-07
height width required for next_image

Lendemor:
`width` and `height` are now required
```python
rx.image(src="/favicon.ico", width=150, height=100) 
```


masenf (approved):
we can either take this as an interim, or leave it open to add the fallback logic discussed in the sync up today.

reflex/components/next/image.py
```diff
@@ -80,6 +83,27 @@ def create(
         Returns:
             _type_: _description_
         """
+        if "width" not in props or "height" not in props:
+            # props.pop("width", None)
+            # props.pop("height", None)
+            if "width" in props:
+                props["width"] = f"{props['width']}px"
```
picklelo, line 90: We should only do this if they're not already setting a unit

reflex/components/next/image.py
```diff
@@ -80,6 +83,27 @@ def create(
...
+            if "height" in props:
+                props["height"] = f"{props['height']}px"
+                # return Box.create(
```
picklelo, line 93: What's the issue in using a Box?

reflex/components/next/image.py
Lendemor: next_image only accept int anyway so they can't set unit

reflex/components/next/image.py
Lendemor: We've have to know both width and height if we want the box to constrain an image set to 100%, it won't work.

masenf (commented):
Do we still want/need this? I know next/image isn't the default `rx.image` anymore, but is it usable enough through `rx.next.image` to include? Is next/image even desirable for us, given that it does not work with static exports?

PR #2549, opened 2024-02-07
(dev mode) start backend without waiting for page compilation

jackie-pc:
Previous logic when `reflex run` (dev mode) performs hot-reload:
* uvicorn detects file change (.py)
* uvicorn begins reloads the main asgi app (by re-importing the app module, and retrieving the `app` object within).
* as a blocking step of the import process, Reflex pages (written in Python) are compiled into NextJs pages to be written out to `.web/pages/`.
* After the end of import process, uvicorn brings up new backend server on port `:8000`.

New logic:
* uvicorn detects file change (.py)
* uvicorn begins reloads the main asgi app (by re-importing the app module, and retrieving the `app` object within).
* During import process, we kick off the page compilation process in a thread, without waiting for its completion.
* Import completes while compile is still going.
* Uvicorn brings up new backend server on port `:8000` much faster (by compilation duration).

Note during dev/test frequent hot reloads that are closer together than compile duration means uvicorn tries to shutdown the previous backend process before it is done with page compilation.  Hanging seemed possible in some cases.

Upgrading uvicorn to 0.27 seemed to make the problem go away (anecdotal!).

If this hanging occurs in future, another thing to try is to set `--timeout-graceful-shutdown` https://www.uvicorn.org/settings/#timeouts to ensure that the previous backend pid goes down in a timely manner (even if compile is ongoing).


reflex/app_module_for_backend.py
```diff
@@ -1,13 +1,21 @@
...
-from reflex.utils.prerequisites import get_compiled_app
+from reflex.utils.prerequisites import get_app, get_compiled_app
 
 if "app" != constants.CompileVars.APP:
     raise AssertionError("unexpected variable name for 'app'")
-app = getattr(get_compiled_app(), constants.CompileVars.APP)
+
+app_module = get_app(reload=False)
+app = getattr(app_module, constants.CompileVars.APP)
+ThreadPoolExecutor(max_workers=1).submit(app.compile_)
```
masenf, line 14: i kind of assumed this would swallow exceptions, but it seems to have this behavior:

1. on first launch, if compile raises an exception then the backend exits
2. on subsequent hot reload, if compile raises an exception, then it prints it and backend stays up

is this the expected behavior?

masenf (changes requested):

reflex/app_module_for_backend.py
jackie-pc: Good point!  I understand why it behaves that way.

On first launch, compilation happens [here](https://github.com/reflex-dev/reflex/blob/eabbbea171415e0d9743c565fdf3a5a7d47018e5/reflex/reflex.py#L185) (blocking, still), and everything is aborted - nothing will come up (front or back).  Which is good, I think.

We can decide what we want to expect for hot reload case, and make it so.

Option A (current PR state):
* Compile fails in a thread. Because we never check `result()` from the task future, there are no other effects on the rest of the program.  Frontend will continue to serve last compiled pages.  Backend stays up (assuming only errors were from the threaded page compiles).

Option B - Shut it all down
* Compile fails in a thread.  We keep a reference to the task future somewhere, so we handle it.  Handling really could only sensibly be to shutdown everything.  For example through the [add_done_callback](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.add_done_callback) mechanic.  Note this callback will run in a child pid (started by uvicorn, depends on specific uvicorn config). If we just kill the child pid - not sure what happens (crash loop vs just give up - latter is better).

Option B is good in that it feels consistent with first launch behavior.  But more complex implementation.

Option A is less consistent with first launch.  But much simpler.  UX would be: they notice that UI didn't change, and they see blurb in terminal (even if it didn't shutdown).

I say we attempt option B, see if it is hairy at all.  If not, then great!

Alek99: I agree, option b feel more consistent with current behavior.

jackie-pc: Thinking a bit more, consider the code edit workflow.  It is quite common for py files to be in an inconsistent state (depending on the IDE - does it autosave).  For example if each character typed is autosaved immediately by the IDE, it is guaranteed that compile errors will happen sometimes.  Do we want to break developer flow by forcing them to restart `reflex run` in that case?

Alek99: I thought most text editors, that do auto saving, have some sort of debouncing to ensure it only saves after the user is done typing. In my opinion this seems like more of an end user problem to configure (or disable) themselves in their own editor. 

Breaking the flow and having the user restart reflex run seems like a hassle for an end user.

picklelo: Yeah if a compile error happens it shouldn't end the program, it should just print the compile error and then resolve it on the next hot reload. I believe that is the current behavior so we should try to keep it.

jackie-pc: How about this as the desired workflow.

When background compile fails during hot reload:
* backend will be down
* browser will show "Connection Error".
* user checks terminal, sees the compile error as printed
* user fixes code
* hot reload / hot re-compile kicks in
* app recovers fully

Current PR does the above.

We could add an extra blurb when compile error happens (e.g. "Backend is down (page compilation failure). It will be brought back up on next code edit").

picklelo: That sounds good to me, that's similar to what we have now. If they see the connection banner they will check the terminal and see the error, which seems like a good workflow.

masenf (approved):
lets do it!

Alek99:
Working well for me too

Alek99 (approved):

PR #2553, opened 2024-02-08
Decorator to validate `rx.color` prop fields

ElijahAhianyo:
This PR validates literal fields on `rx.color`. The code below should throw an error:

```python
rx,color("mint", 3000)
```

It also refactors the literal validation logic in its own function and provides a decorator to be applied to functions whose arguments need validation for literal fields


picklelo (approved):
awesome!

PR #2558, opened 2024-02-08
Use rx.el.img as rx.image

masenf:
Image optimization from next/image is breaking `reflex export`. Here we alias rx.image to rx.el.img (and remove the `height`/`width` props, so those can handle responsive lists of css values instead of being HTML attributes)


Alek99 (approved):
working for me on reflex web

masenf:
Fixed the test and cleaned up the top level import

PR #2566, opened 2024-02-09
[REF-1921] Remove HTML attributes that shadow CSS props

masenf:
Most of these attributes are older and less flexible than the equivalent named CSS prop, so removing the props from the Element allows them to be automatically passed in the style dict.

Probably want to filter out pyi changes for the review.


picklelo (approved):
🚢

masenf:
Oh i totally spaced that the radix components based on these HTML components also needed to have their `pyi` files rewritten.

That's good though, otherwise radix components wouldn't work with stuff like css `translate` 😮

picklelo (approved):

PR #2574, opened 2024-02-10
[REF-1885] Shard Substates when serializing to Redis

### LINKED ISSUE #2554: [REF-1885] Explore: substate sharding
reflex-web has too large of state to serialize and deserialize for every request. This ticket is to explore handling substates individually in redis.

[https://github.com/reflex-dev/reflex/pull/2574](https://github.com/reflex-dev/reflex/pull/2574)

<sub>From [SyncLinear.com](https://synclinear.com) | [REF-1885](https://linear.app/reflex-dev/issue/REF-1885/explore-substate-sharding)</sub>

masenf:
This is a performance optimization for apps that have many smaller states that end up taking excessive time to serialize/deserialize for every event, like reflex-web.

Instead of only serializing the top-level `rx.State` and implicitly all substates by virtue of their presence in the `self.substates` dictionary, the `StateManagerRedis` applies special logic to persist substates into their own keys. When retrieving a particular state, it's parent and all of it's substates (recursively) are fetched from redis and assembled into a structure resembling the previous implementation. The main difference is that when using redis, sibling states are not directly accessible. This will be addressed in a followup patch, that allows arbitrary access to any state defined in the app.

Fixes #2554


reflex/state.py
```diff
@@ -1405,6 +1419,24 @@ def on_load_internal(self) -> list[Event | EventSpec] | None:
             type(self).set_is_hydrated(True),  # type: ignore
         ]
 
+    def __getstate__(self):
```
abulvenz, line 1422: This leaves `router_data` on each substate. Is that the intended behavior?

masenf: For now, yes. But I might be able to get away with stripping it out.

abulvenz: Maybe `router_data` turns out to be a substate on its own in the end :smile:

picklelo: @abulvenz not a bad idea actually.. it may involve having to save two states each time as a downside, but it might be a cleaner separation in the future.

reflex/app.py
```diff
@@ -1002,7 +1002,8 @@ async def upload_file(request: Request, files: List[UploadFile]):
             )
 
         # Get the state for the session.
-        state = await app.state_manager.get_state(token)
+        substate_token = token + "_" + handler.rpartition(".")[0]
```
picklelo, line 1005: nit: maybe we can pull this out into a function `get_substater_token(token, handler)`

reflex/event.py
```diff
@@ -41,6 +41,16 @@ class Event(Base):
...
     payload: Dict[str, Any] = {}
 
+    @property
+    def substate_token(self) -> str:
+        """Get the substate token for the event.
+
+        Returns:
+            The substate token.
+        """
+        substate = self.name.rpartition(".")[0]
```
picklelo, line 51: if we pull it out, then we can reuse the function here

reflex/state.py
```diff
@@ -1479,6 +1511,8 @@ async def __aenter__(self) -> StateProxy:
         """
         self._self_actx = self._self_app.modify_state(
             self.__wrapped__.router.session.client_token
+            + "_"
```
picklelo, line 1514: same comment about pulling this out

reflex/state.py
```diff
@@ -1737,35 +1775,118 @@ class StateManagerRedis(StateManager):
...
+            if parent_state is None:
+                # Retrieve the parent state from redis.
+                parent_state_name = state_path.rpartition(".")[0]
+                if parent_state_name:
+                    parent_state_key = token.rpartition(".")[0]
+                    parent_state = await self.get_state(
+                        parent_state_key, top_level=False, get_substates=False
+                    )
+            # Set up Bidirectional linkage between this state and its parent.
+            if parent_state is not None:
```
picklelo, line 1826: this can just be an `else:`

picklelo (approved):
minor comments

reflex/state.py
masenf: i don't think so, because we may or may not set `parent_state` in the previous condition (for example, when retrieving the base state of the app, `parent_state` will still be None as this point, even though we entered the previous condition)

reflex/state.py
masenf: adding a utility for this in the followup PR

PR #2585, opened 2024-02-12
keep-chakra: whitelist is always whitelist

masenf:
Regardless of whether the same name exists in the top level rx.<x> namespace, always convert whitelisted names to rx.chakra.<x>

Prior to this change, `rx.color_mode_icon()` was not getting converted to `rx.chakra.color_mode_icon()` because `rx.color_mode_icon` no longer exists (in radix, it has become `rx.color_mode.icon()`).

Remove the `code_block` and `color_mode_cond` from the whitelist, since those were moved out of the chakra namespace.

Remove `base` from the whitelist, because `rx.base` still exists and never pointed into the chakra namespace.


jackie-pc (approved):
nice!

PR #2587, opened 2024-02-12
Expose `get_upload_url` and `get_upload_dir` at top level

masenf:
Make these names easier to access from the top level rx. namespace.


reflex/__init__.py
```diff
@@ -41,6 +41,8 @@
     "clear_selected_files",
     "selected_files",
     "upload",
+    "get_uploaded_files_dir",
```
picklelo, line 44: nit: should we name this `get_upload_dir` to make it a bit shorter. And `get_upload_url` below?

masenf: do you want me to alias it? or rename the actual functions?

masenf: @picklelo i just changed the names all the way down for consistency since there's no compat to maintain at this point.

picklelo (approved):

PR #2591, opened 2024-02-13
[REF-1919]Valid Children/parents to allow Foreach,Cond,Match and Fragment

ElijahAhianyo:
This PR marks `Foreach`, `Cond` ,`Match` and `Fragment` as Valid children or Parents of any component. It further extracts the immediate children (in the case of Fragment) and all return components and validates them as well.

### DIFF
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -673,15 +673,46 @@ class Component(BaseComponent, ABC):
             return
 
         comp_name = type(self).__name__
+        allowed_components = ["Fragment", "Foreach", "Cond", "Match"]
 
-        def validate_invalid_child(child_name):
+        def validate_invalid_child(child):
+            child_name = type(child).__name__
+
+            if child_name == "Fragment":
+                for c in child.children:
+                    validate_invalid_child(c)
+
+            if child_name == "Cond":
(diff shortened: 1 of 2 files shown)


masenf (approved):

reflex/components/component.py
```diff
@@ -668,47 +668,63 @@ def _validate_component_children(self, children: List[Component]):
...
+        no_valid_parents_defined = all(child._valid_parents == [] for child in children)
+        if (
+            not self._invalid_children
+            and not self._valid_children
+            and no_valid_parents_defined
+        ):
             return
 
         comp_name = type(self).__name__
+        allowed_components = ["Fragment", "Foreach", "Cond", "Match"]
```
picklelo, line 680: I think it would be better to actually import these components, and then do issubclass checks on those, rather than checking the name. Also, perhaps we could move the validate_child method to the sub components, so we just do `child.validate_children()` and we don't have to have all the if cases here (we can do this in the future)

masenf: Good point

picklelo: Let's do this in a follow-up - we can set the default values to `None` within cond.

PR #2597, opened 2024-02-13
fix text_field other event when debounce

Lendemor:
using other event like on_key_down when rx.input() is fully controlled should now work.


reflex/components/radix/themes/components/text_field.py
```diff
@@ -55,7 +55,15 @@ def create(cls, *children, **props) -> Component:
         """
         if props.get("value") is not None and props.get("on_change"):
             # create a debounced input if the user requests full control to avoid typing jank
-            return DebounceInput.create(super().create(*children, **props))
+            debounce_props = {
+                "on_change": props.pop("on_change"),
+                "value": props.pop("value"),
```
masenf, line 60: we also need `debounce_timeout` and the other config props... basically, 

```python
debounce_props = {
    prop: props.pop(prop)
    for prop in DebounceInput.get_props()
    if prop in props
}
debounce_props["on_change"] = props.pop("on_change")
```

masenf (commented):
i'm still overall skeptical about this one... 🤔 

I guess i need to just check it out and try it.

PR #2598, opened 2024-02-13
[REF-1925]Accordion foreach fix

ElijahAhianyo:
When a foreach is used to create accordion items( like in the example code below) and a value not supplied in the resulting render function, the str of the header value supplied is inferred to be the accordion value (this is what the accordion trigger uses in determining which item to trigger). However, for foreach, the header is passed as a Var. The `str` of the var renders improperly. This PR fixes that.

```python
import reflex as rx


class State(rx.State):
    """The app state."""
    items: list[tuple[str, str]] = [
        ("Item 1", "Content 1"),
        ("Item 2", "Content 2"),
        ("Item 3", "Content 3"),
    ]

def index():
    return rx.accordion.root(
        rx.foreach(
            State.items, lambda item: rx.accordion.item(item[0], item[1])
        ),
        collapsible=True,
    )

app = rx.App(
    theme=rx.radix.themes.theme(
        has_background=True, radius="none", accent_color="orange", appearance="light",
    ),
)
app.add_page(index)

```

### DIFF
--- a/reflex/components/radix/primitives/accordion.py
+++ b/reflex/components/radix/primitives/accordion.py
@@ -546,10 +546,8 @@ def accordion_item(
     """Create an accordion item.
 
     Args:
-        header: The header of the accordion item. This could be a regular component
-                or a Var when you use a foreach
-        content: The content of the accordion item. This could be a regular component
-                or a Var when you use a foreach.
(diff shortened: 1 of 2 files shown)


picklelo (approved):
nice catch

PR #2600, opened 2024-02-13
Fix AccordionItem interactive docs not showing up

invrainbow:
### Type of change

### New Feature Submission:

### Description

Move `accordion_item` into `AccordionItem.create` so it gets picked up as a component by reflex-web (which looks for a `__self__` attribute).

### DIFF
--- a/reflex/components/radix/primitives/accordion.py
+++ b/reflex/components/radix/primitives/accordion.py
@@ -469,7 +469,7 @@ class AccordionItem(AccordionComponent):
         )
 
     @classmethod
-    def create(
+    def create_high_level(
         cls, header: Component | Var, content: Component | Var, **props
     ) -> Component:
         """Create an accordion item.
@@ -485,7 +485,7 @@ class AccordionItem(AccordionComponent):
         # The item requires a value to toggle (use the header as the default value).
         value = props.pop("value", header if isinstance(header, Var) else str(header))
 
-        return super().create(
+        return cls.create(
             AccordionHeader.create(
                 AccordionTrigger.create(
                     header,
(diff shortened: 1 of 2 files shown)


picklelo (approved):
nice catch

masenf (commented):
Does this block usage of the low level API now?

I mean, i guess it was blocked before, by virtue of the `accordion.item` pointing to the high level API, but if we want to retain access to the low level API, then `header` and `content` should be allowed to be None, in which case, the `*children` and `**props` should be passed directly to the contructor.

invrainbow:
What if I call it `create_high_level` or something and leave `create` available for use as a low-level API?

masenf:
> What if I call it create_high_level or something and leave create available for use as a low-level API?

I still would argue that the low level isn't available in this case, because the name we expose, `accordion.item` doesn't point to the low level API.

If the `AccordionItem.create` function accepted `*children` and `**props` and only did the high level stuff when `header` and `content` are provided, then I think we're okay.

@picklelo @tgberkeley any comment here?

masenf:
I think that's what we need to figure out before we ship 0.4.0; if we leave `accordion.item` as it is, then we will need another name to expose the low level `AccordionItem` down the road.

In my opinion, forcing `header` and `content` to be passed by keyword is aligned with other parts of our API and gives us the best flexibility.

masenf (approved):
cool. thanks for making the changes. some of the logic in there looks a bit questionable, like "what if i provide `header` and not `content`?"

But that's not an API breaking change, and we can fix that / throw an exception and improve the validation down the road.

PR #2605, opened 2024-02-14
Add replace var operation 

picklelo:
Operation for Python `str.replace`


masenf (commented):
we should add a test in `integration/test_var_operations.py`

PR #2615, opened 2024-02-14
Update docstrings for Radix components based on upstream descriptions

masenf:
These are better than copying the generic description of Button in many many many components.


picklelo (approved):

Alek99 (approved):
lgtm

PR #2616, opened 2024-02-14
fix value/default_value in accordion

Lendemor:
fix type annotation for value/default_value in `AccordionRoot`

fix default for value in `AccordionItem`


reflex/components/radix/primitives/accordion.py
```diff
@@ -488,14 +488,18 @@ def create(
             The accordion item.
         """
         # The item requires a value to toggle (use the header as the default value).
```
masenf, line 490: update comment while you're here

masenf (approved):

PR #2620, opened 2024-02-15
Text `as_` prop literals update

ElijahAhianyo:
Support more literal values

Screenshot 2024-02-15 at 4 40 13 PM (2)

### DIFF
--- a/reflex/components/radix/themes/typography/text.py
+++ b/reflex/components/radix/themes/typography/text.py
@@ -22,7 +22,26 @@ from .base import (
     LiteralTextWeight,
 )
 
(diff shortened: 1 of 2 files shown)


masenf (commented):
I thought this was something we wanted too, but radix says this field only supports     "p",
    "label",
    "div",
    "span",

https://www.radix-ui.com/themes/docs/components/text

image

picklelo:
@masenf What does supported mean in this case? Chatted with Eli about this, and it seems the other tags still work, though they're not listed in that Literal

Lendemor (approved):

PR #2628, opened 2024-02-15
accordion default classname

Lendemor:
```
rx.accordion.root(
    rx.accordion.item(
        rx.accordion.header(
            rx.accordion.trigger(
                "Item 1",
                rx.accordion.icon(),
            ),
        ),
        rx.accordion.content(
            rx.text("Item 1"),
        ),
    ),
    rx.accordion.item(
        header="Item 2",
        content=rx.text("Item 2"),
    ),
)
```

### DIFF
--- a/reflex/components/radix/primitives/accordion.py
+++ b/reflex/components/radix/primitives/accordion.py
@@ -525,15 +525,10 @@ class AccordionItem(AccordionComponent):
                 AccordionHeader.create(
                     AccordionTrigger.create(
                         header,
-                        Icon.create(
-                            tag="chevron_down",
-                            class_name="AccordionChevron",
-                            display="inline-block",
-                        ),
-                        class_name="AccordionTrigger",
+                        AccordionIcon.create(),
                     ),
                 ),
-                AccordionContent.create(content, class_name="AccordionContent"),
+                AccordionContent.create(content),
             ]
 
         return super().create(*children, value=value, **props, class_name=cls_name)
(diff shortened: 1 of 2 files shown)


picklelo (approved):

masenf (commented):
Can we get the same treatment for AccordionHeader and AccordionItem

Also, can you move the definition of `slideUp` and `slideDown` to AccordionRoot

Lendemor:
> Can we get the same treatment for AccordionHeader and AccordionItem
> 
> Also, can you move the definition of `slideUp` and `slideDown` to AccordionRoot

can do

reflex/components/radix/primitives/accordion.py
```diff
@@ -493,6 +515,11 @@ def create(
...
         value = props.pop("value", get_unique_variable_name())
 
+        if "AccordionItem" not in (
+            cls_name := props.pop("class_name", "AccordionItem")
+        ):
+            cls_name = f"{cls_name} AccordionItem"
+
         if (header is not None) and (content is not None):
             children = [
                 AccordionHeader.create(
```
masenf, line 525: we can simplify the following code now: removing the `class_name` specification and replacing `Icon` with the new `AccordionIcon`.

people looking to use the low level API will probably copy this as an example

masenf (approved):

reflex/components/radix/primitives/accordion.py
Lendemor: sure let's do that

PR #2630, opened 2024-02-15
Handle component namespaces in global styles

picklelo:
Before this, having something like `style = {rx.text: ...}` didn't work, as `rx.text` is now a namespace.

This PR creates a `ComponentNamespace` class that is hashable, and converts all of these namespaces to create methods when applicable.

Test app:

```python
import reflex as rx

def index():
    """The main view."""
    return rx.text("hi")


style = {rx.text: {"color": "green"}}
app = rx.App(style=style)
app.add_page(index)
```


reflex/components/component.pyi
```diff
@@ -0,0 +1,393 @@
+"""Stub file for reflex/components/component.py"""
```
masenf, line 1: remove this file; not needed

masenf (approved):

PR #2631, opened 2024-02-15
Update new readme to match radix api

Alek99:
The api for input is not the best right now but this is how it works with our current functionality

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,20 +101,17 @@ def index():
         rx.vstack(
             rx.heading("DALL-E", font_size="1.5em"),
             rx.input(
-                    placeholder="Enter a prompt..",
-                    on_blur=State.set_prompt,
-                    width="25em",
+                placeholder="Enter a prompt..",
+                on_blur=State.set_prompt,
+                width="25em",
             ),
-            rx.button("Generate Image", on_click=State.get_image,width="25em"),
+            rx.button("Generate Image", on_click=State.get_image, width="25em"),
             rx.cond(
                 State.processing,
                 rx.chakra.circular_progress(is_indeterminate=True),
                 rx.cond(
                     State.complete,
-                    rx.image(
-                        src=State.image_url,
-                        width="20em"
-                    ),
+                    rx.image(src=State.image_url, width="20em"),
                 ),
             ),
             align="center",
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -88,41 +90,51 @@ class State(rx.State):
...
+                    ),
+                ),
             ),
-            padding="2em",
-            shadow="lg",
-            border_radius="lg",
+            width="25em",
+            padding="1em",
+            spacing="2",
+            align_items="center",
```
picklelo, line 126: This should be `align` now

README.md
```diff
@@ -88,41 +90,51 @@ class State(rx.State):
...
-                "Generate Image",
-                on_click=State.get_image,
-                is_loading=State.processing,
+            rx.heading("DALL-E", font_size="1.5em"),
+            rx.radix.text_field.root(
+                rx.input(
+                    placeholder="Enter a prompt..",
+                    on_blur=State.set_prompt,
+                    width="100%",
+                ),
```
masenf, line 109: why is this wrapped in a root? maybe we can just set the width to some relative unit, like `width="25em"`?

Alek99: It has to because there is no way to affect the percentage width otherwise

Alek99: Ok I can do that for now

README.md
```diff
@@ -72,10 +72,12 @@ Here is the complete code to create this. This is all done in one Python file!
 import reflex as rx
 import openai
 
-openai.api_key = "YOUR_API_KEY"
+openai_client =  openai.OpenAI(api_key="YOUR_API_KEY")
```
picklelo, line 75: I think in the new api you don't need to pass in the key, it just reads it from the env var. (Maybe we can just put a comment telling them to make sure its set)

README.md
```diff
@@ -88,41 +90,50 @@ class State(rx.State):
...
+                ),
             ),
-            padding="2em",
-            shadow="lg",
-            border_radius="lg",
+            width="25em",
+            padding="1em",
+            spacing="2",
+            align_items="center",
+            border_radius="10px",
```
picklelo, line 127: Potentially confusing to newcomers that we're mixing px, em, and % all together for. maybe we can unify for the intro app

picklelo (commented):
looks good, can we run `black` on the code example, then we can merge this

picklelo (approved):

PR #2636, opened 2024-02-16
Component: translate underscore suffix for props supported by chakra

masenf:
* type_ becomes type
* min_ becomes min
* max_ becomes max

The deprecation warning is only displayed when the underscore suffix prop is passed and the non-underscore suffix prop is defined on the given component.

Existing chakra-based code is unaffected.

Rename type_ to type in accordion and scroll_area: All of the new radix components avoid the underscore suffix names where possible.

### DIFF
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -599,10 +599,14 @@ class Component(BaseComponent, ABC):
         from reflex.components.base.bare import Bare
 
         # Translate deprecated props to new names.
-        known_props = cls.get_props()
-        for prop in ["type", "min", "max", "id"]:
+        new_prop_names = [
+            prop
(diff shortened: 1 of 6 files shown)


picklelo (approved):

reflex/components/component.py
```diff
@@ -598,6 +598,19 @@ def create(cls, *children, **props) -> Component:
         # Import here to avoid circular imports.
         from reflex.components.base.bare import Bare
 
+        # Translate deprecated props to new names.
+        known_props = cls.get_props()
```
picklelo, line 602: nit: this can be `known_props = [prop for prop in cls.get_props() if prop in ["type", ...]]`

picklelo (approved):

PR #2651, opened 2024-02-17
[REF-1993] link: respect `is_external` prop and other attributes on A tag

### LINKED ISSUE #2648: [REF-1993] Can't set is_external on rx.link in 0.4.0
**Describe the bug**
The new radix based link component does not have is_external prop available to open in a new tab.

This functionality should be restored.

<sub>[REF-1993](https://linear.app/reflex-dev/issue/REF-1993/cant-set-is-external-on-rxlink-in-040)</sub>

masenf:
Instead of passing all props to NextLink by default, only pass props that NextLink understands, placing the remaining props on the Radix link

Add a test case to avoid regression of `is_external` behavior.

Fix #2648

### DIFF
--- a/integration/test_navigation.py
+++ b/integration/test_navigation.py
@@ -1,0 +1,89 @@
+"""Integration tests for links and related components."""
+from typing import Generator
+from urllib.parse import urlsplit
+
+import pytest
+from selenium.webdriver.common.by import By
+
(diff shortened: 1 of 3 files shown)


Alek99 (approved):
working well for me

PR #2663, opened 2024-02-19
Accordion Items unique Value

### LINKED ISSUE #2652: [REF-1995] Radix Accordion generated with foreach and external function does not have unique value
**Describe the bug**
It shows up correctly, but after clicking on any item, all items open at the same time and the whole accordion freezes. No error message on console or backend.
This has worked with Chakra accordion.

**To Reproduce**

```
    @rx.cached_var
    def accordion(self) -> list[str]: 
        return ['item1','item2','item3']

def accordion_level1(name) -> rx.Component:
    return rx.accordion.item(
        header=name,
        content=name,
    )

rx.accordion.root(
    rx.foreach(State.accordion,accordion_level1)),
),
```

**Expected behavior**
Working like chakra accordion.

**Screenshots**

* Python Version: 3.9.2
* Reflex Version: 0.4.0
* OS: debian11
* Browser (Optional):

**Additional context**

[https://github.com/reflex-dev/reflex/pull/2663](https://github.com/reflex-dev/reflex/pull/2663)

<sub>From [SyncLinear.com](https://synclinear.com) | [REF-1995](https://linear.app/reflex-dev/issue/REF-1995/radix-accordion-generated-with-foreach-and-external-function-does-not)</sub>

ElijahAhianyo:
use the  `generateUUID` function in `.web/utils/state.js` to generate accordion item value at runtime instead of compile time

fixes #2652


masenf (dismissed):
interesting solution. i like that the code output is stable too

reflex/vars.py
```diff
@@ -2003,3 +2003,9 @@ def __call__(self, *args, **kwargs) -> BaseVar:
             The Var returned from calling the function.
         """
         return self.fn(*args, **kwargs)
+
+
+unique_string_var = BaseVar(
+    _var_name="crypto.randomUUID()",
```
masenf, line 2009: I was doing some more research into this and apparently it's only available "secure contexts" in some browsers. I think more testing is needed around this, or we might get browser errors in some cases.

ElijahAhianyo: Right, an alternative may be to use a third-party library like https://www.npmjs.com/package/uuid , and even they advise using `crypto.randomUUID()` if you only need to generate uuids and I'm also very reluctant on the idea of installing a new package just to generate random UUIDs. Another option would be having to do this ourselves, i.e writing a helper function to generate random strings. What do you think?

masenf: we already have our own function in `state.js`, maybe we can just export and use that?

https://github.com/reflex-dev/reflex/blob/6384c62e51cc354b0c1071c8a7ffa66cabd51a17/reflex/.templates/web/utils/state.js#L43-L57

ElijahAhianyo: yes, that really should work. Will update the the PR to use this instead

ElijahAhianyo: fixed

masenf (approved):

PR #2664, opened 2024-02-19
Accordion Root Exclude `color_scheme` and `variant` props in tag

### LINKED ISSUE #2657: [REF-1996] Setting radix accordion.root.color_scheme gives javascript warning
**Describe the bug**
client.js:2 Warning: React does not recognize the `colorScheme` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `colorscheme` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

**To Reproduce**

```
rx.accordion.root(
        rx.accordion.item(
            header="First Item",
            content=The First accordion item's content",
            font_size="3em",
            value="item_1",
        ),
        rx.accordion.item(
            header="Second Item",
            content="The second accordion item's content",
            font_size="3em",
            value="item_2",
        ),
        rx.accordion.item(
            header="Third item",
            content="The third accordion item's content",
            font_size="3em",
            value="item_3",
        ),
        collapsible=True,
        color_scheme='indigo',
    )
```

**Expected behavior**

**Screenshots**

* Python Version: 0.9.2
* Reflex Version: 0.4.0
* OS: debian 11
* Browser (Optional):

**Additional context**

[https://github.com/reflex-dev/reflex/pull/2664](https://github.com/reflex-dev/reflex/pull/2664)

<sub>From [SyncLinear.com](https://synclinear.com) | [REF-1996](https://linear.app/reflex-dev/issue/REF-1996/setting-radix-accordionrootcolor-scheme-gives-javascript-warning)</sub>

ElijahAhianyo:
This PR removes the `color_scheme` and `variant` props before adding props to the accordion root tag. This is because the aforementioned props are only used to dynamically curate a theme style and do not have any effect being added to the rendered Tag. It also fixes the warning in the console log reported in the linked issue to this pr.

Fixes #2657


reflex/components/radix/primitives/base.py
```diff
@@ -13,6 +13,9 @@ class RadixPrimitiveComponent(Component):
     # Change the default rendered element for the one passed as a child.
     as_child: Var[bool]
 
+    # "Fake" prop color_scheme is used to avoid shadowing CSS prop "color".
+    _rename_props: Dict[str, str] = {"colorScheme": "color"}
```
masenf, line 17: i dont think we want this here, because no radix primitive will understand either `color_scheme` or `color` without our help.

i think having the `_excluded_props` is all we really need.

masenf (changes requested):

reflex/components/radix/primitives/base.py
ElijahAhianyo: Yeah that makes sense

ElijahAhianyo: fixed

reflex/components/component.py
```diff
@@ -544,6 +552,10 @@ def _render(self, props: dict[str, Any] | None = None) -> Tag:
         props.update(self._get_style())
         props.update(self.custom_attrs)
 
+        # remove excluded props from prop dict before adding to tag.
+        for prop_to_exclude in self._exclude_props():
+            del props[prop_to_exclude]
```
masenf, line 557: This should be a .pop(prop_to_exclude, None)

In case the named prop didn't get added for whatever reason.

masenf (changes requested):
Didn't notice this earlier

masenf (approved):

PR #2665, opened 2024-02-19
fix upload clear bug

abulvenz:
- feat: Synchronizing localStorage between tabs using browser events
- test_client_storage: Test sync'd local storage vars
- update_vars_internal: generic handler to apply var changes to state tree
- fix: File upload error after using rx.clear_selected_files


abulvenz:
Sorry, wrong branch.

PR #2666, opened 2024-02-19
fix: File upload error after using rx.clear_selected_files

abulvenz:
After rx.clear_selected_files events, uploading the empty list fails with an error. I have tested it manually in reflex.examples, where the clear_selected_files has been removed when moving to 0.4.0.


masenf (approved):
looks reasonable to me

PR #2668, opened 2024-02-19
Fix SQLAlchemy list types

benedikt-bartscher:


reflex/utils/types.py
```diff
@@ -164,7 +164,11 @@ def get_attribute_access_type(cls: GenericType, name: str) -> GenericType | None
             prop = descriptor.property
             if not isinstance(prop, Relationship):
                 return None
-            return prop.mapper.class_
+            class_ = prop.mapper.class_
+            if prop.uselist:
+                return list[class_]
```
masenf, line 169:
```suggestion
                return List[class_]
```

I think we should use cap `List` to keep py3.8 and py3.9 compatibility, if that works.

masenf (changes requested):
a test case would be cool too, but not strictly necessary since most of this code is already untested.

benedikt-bartscher:
@masenf I wanted to add tests later, if I have a fully working bare sqlalchemy with reflex. There are still some rough edges, f.e. with serialization, recursive/self-referencing relations and AssociationProxys. But I am getting close, most Issues are fixed in my reflex builds.  
There's also #2355 which needs some love before reflex can really support bare sqlalchemy

masenf (approved):

PR #2670, opened 2024-02-19
Initial values for computed vars

benedikt-bartscher:
```python
import reflex as rx


class State(rx.State):
    """The app state."""

    @rx.var(initial_value="initial")
    def test(self) -> str:
        return "Hello, world!"


cv = State.test
cv_initial = cv._initial_value
print(f"{cv=}, {cv_initial=}")

state = State()
state_d = state.dict(initial=True)
initial_d_state = state_d["state.state"]
initial_test = initial_d_state["test"]
print(f"Initial state: {initial_test=}")


def index() -> rx.Component:
    return rx.center(
        rx.vstack(
            rx.heading(State.test),
            rx.button(
                "Check out our docs!",
                size="4",
            ),
            align="center",
            spacing="7",
            font_size="2em",
        ),
        height="100vh",
    )


app = rx.App()
app.add_page(index)
```

### DIFF
--- a/reflex/vars.py
+++ b/reflex/vars.py
@@ -1800,7 +1800,7 @@ class ComputedVar(Var, property):
     # Whether to track dependencies and cache computed values
     _cache: bool = dataclasses.field(default=False)
 
-    _initial_value: Any | None = dataclasses.field(default=None)
+    _initial_value: Any | types.Unset = dataclasses.field(default_factory=types.Unset)
 
     def __init__(
         self,
@@ -1808,7 +1808,7 @@ class ComputedVar(Var, property):
         fset: Callable[[BaseState, Any], None] | None = None,
         fdel: Callable[[BaseState], Any] | None = None,
         doc: str | None = None,
-        initial_value: Any | None = None,
+        initial_value: Any | types.Unset = types.Unset(),
         **kwargs,
     ):
         """Initialize a ComputedVar.
--- a/reflex/vars.py
+++ b/reflex/vars.py
@@ -1800,12 +1800,15 @@ class ComputedVar(Var, property):
     # Whether to track dependencies and cache computed values
     _cache: bool = dataclasses.field(default=False)
 
+    _initial_value: Any | None = dataclasses.field(default=None)
+
     def __init__(
         self,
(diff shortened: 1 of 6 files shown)


tests/test_var.py
```diff
@@ -712,6 +752,65 @@ def test_computed_var_with_annotation_error(request, fixture, full_name):
...
+):
+    """Test that the initial and runtime values of a computed var are correct.
+
+    Args:
+        request: Fixture Request.
+        fixture: The state fixture.
+        var_name: The name of the computed var.
+        expected_initial: The expected initial value of the computed var.
+        expected_runtime: The expected runtime value of the computed var.
+        runtime_only: Whether the computed var is runtime only.
```
masenf, line 799: Seems like `runtime_only` might be more accurately named `raises_at_runtime`

reflex/vars.py
```diff
@@ -1958,6 +1963,43 @@ def _determine_var_type(self) -> Type:
         return Any
 
 
+def computed_var(
+    fget: Callable[[BaseState], Any] | None = None,
+    fset: Callable[[BaseState, Any], None] | None = None,
+    fdel: Callable[[BaseState], Any] | None = None,
+    doc: str | None = None,
```
masenf, line 1970: can we remove these from the signature, computed vars never have setters or deleters, they're just in the property signature... actually probably should remove them from `__init__` as well if we're cleaning this up

reflex/vars.py
```diff
@@ -1958,6 +1963,43 @@ def _determine_var_type(self) -> Type:
...
+    initial_value: Any | None = None,
+    **kwargs,
```
masenf, line 1972: lets explicitly define `cache: bool = False` so users can create a cached var like `@rx.var(cached=True)`

then update the `cached_var` down below to essentially be a `functools.partial` of this with `cached=True` already set?

that would allow users to also write `@rx.cached_var(initial_value="foo")` and keep both decorators consistent

masenf (changes requested):
looks good, a few suggestions

benedikt-bartscher:
Thanks for the quick review, your suggestions are great. Done!

masenf (approved):

PR #2671, opened 2024-02-20
Connection state

benedikt-bartscher:
```python
import asyncio
import datetime

import reflex as rx
from reflex.state import SessionStatusEnum


class State(rx.State):
    count: int = 0
    first_event: datetime.datetime = datetime.datetime.now()
    # events: list[tuple[SessionStatusEnum, datetime.datetime]] = []

    # def on_load(self) -> None:
    #     last_status = self.events[-1][0] if self.events else None
    #
    #     if self.status == last_status:
    #         return
    #
    #     new_event = (self.status, datetime.datetime.now())
    #     self.events.append(new_event)

    def increment(self) -> None:
        self.count += 1

    @rx.var
    def last_event(self) -> datetime.datetime:
        return self._session_status.last_event

    @rx.var
    def status(self) -> SessionStatusEnum:
        return self._session_status.status

    @rx.background
    async def background(self) -> None:
        while True:
            await asyncio.sleep(1)
            async with self:
                if self.status == SessionStatusEnum.DISCONNECTED:
                    print("Session disconnected")
                    break
                self.count += 1
            print(f"background loop: {self.count}")


def index() -> rx.Component:
    return rx.fragment(
        rx.button("Increment", on_click=State.increment),  # pyright: ignore[reportUnknownMemberType]
        rx.button("Start Background", on_click=State.background),  # pyright: ignore[reportUnknownMemberType]
        rx.text(State.count),
        rx.text(f"session_id: {State.router.session.session_id}"),
        rx.text(f"client_token: {State.router.session.client_token}"),
        rx.text(f"last_event: {State.last_event}"),
        rx.text(f"first_event: {State.first_event}"),
        rx.text(f"status: {State.status}"),
        # rx.text("events:"),
        # rx.foreach(State.events, lambda event: rx.text(f"{event[0]}: {event[1]}")),
    )


app = rx.App()
app.add_page(
    index,
    # on_load=State.on_load,
)
```


Lendemor:
I missed that PR before.

I kinda like the idea of having a SessionStatus field.

Do you think you could later update this PR to use the change from #3388 once they are merged ?

benedikt-bartscher:
@Lendemor no worries, it wasn't really finished yet. I will rebase once your PR is merged.

benedikt-bartscher:
@Lendemor I rebased it and added a small use example in the issue description.
The functionality is ready for review, but I'll need to adjust some tests which do not expect the new session state to be there (with #2429 fewer tests would need to be adjusted).

Lendemor:
@benedikt-bartscher Got it, I'll review whenever the PR is marked as ready 👍

reflex/app.py
```diff
@@ -1138,6 +1139,9 @@ async def process(
                 # assignment will recurse into substates and force recalculation of
                 # dependent ComputedVar (dynamic route variables)
                 state.router_data = router_data
+            if state.router:
+                state.router.update(router_data)
+            else:
```
masenf, line 1144: if we update the `router` outside of the `if state.router_data != router_data:` block, then that means _every_ event will trigger a change and delta to the `router` Var (and any dependent computed vars); this could increase network traffic pretty significantly since the serialized `router` object is relatively large.

I wonder if it makes sense to keep the connection `status` and `last_event` fields on a separate object that is not sent to the frontend. I don't really see either of these being super useful for frontend rendering, they're mainly for checking from event handlers from my perspective.

masenf (changes requested):
i started looking at the test changes, but i think if we just avoid sending these new fields in the delta then the tests can stay the way they are.

reflex/app.py
benedikt-bartscher: Thanks for your review and suggestions. Indeed, we do not really need `status` and `last_event` in the frontend. I adjusted the session state to be a separate backend var.
If someone really needs `status` or `last_event` in the frontend, it's still possible to wrap them in a `@rx.var` as seen in the updated PR description.

benedikt-bartscher:
Anything else missing here?

masenf (approved):

masenf (commented):
hitting a couple of issues with this

1. all states seem to share the same _session_status; i think this needs to be declared as a `pydantic.PrivateAttr` so each instance gets its own copy
2. the `RECONNECTED` state does not seem to "stick". it shows up once when the initial reconnection happens, but then immediately changes to `CONNECTED` again, so detecting the reconnected state is not really possible if there is an on_load or some other event processed before the handler that is supposed to check the status.

We might need some integration tests for this feature.

masenf (changes requested):
Needs a test case and the above issues addressed (accidentally clicked approve earlier)

benedikt-bartscher:
Thanks for looking into this!

1. is fixed in https://github.com/reflex-dev/reflex/pull/2671/commits/39b5658bcc061874be5f7193691685c577407d4a
2. yes, I noticed this as well. I will drop reconnected for now

Lendemor:
Marking this as draft until ready for review.

adhami3310:

PR #2675, opened 2024-02-20
ThemePanel is toggleable with bool in Theme

Lendemor:
Can now use `rx.theme_panel` like this :

```python
app = rx.App(
    theme=rx.theme(use_panel=True)
)
```


masenf (dismissed):
good ux

any thoughts on `use_theme_panel` vs `use_panel`? i know it's more to type...

down the road when we make our own panel, we can just swap out the radix one with our own

Alek99:
I prefer use_theme_panel its a little more clear

Alek99:
or just theme_panel

masenf (approved):

PR #2676, opened 2024-02-20
add pulser for connection + adjust condition for connnection_modal

Lendemor:
Add a small `wifi_off` icon that pulse on bottom left when connection is down.

connection_pulser is included in the `default_overlay_component`

apparition of connection_banner is delayed a little, appear after a few failed attemps.


reflex/.templates/web/utils/state.js
```diff
@@ -312,23 +325,22 @@ export const connect = async (
...
   // Once the socket is open, hydrate the page.
   socket.current.on("connect", () => {
-    setConnectError(null)
+    setConnectErrors([]);
   });
 
-  socket.current.on('connect_error', (error) => {
-    setConnectError(error)
+  socket.current.on("connect_error", (error) => {
+    setConnectErrors((connectErrors) => [...connectErrors, error]);
```
masenf, line 332: i think we should slice out the existing connectErrors to < 10 so we don't balloon the memory of the browser if the backend doesn't come back up for a long period.

reflex retries pretty aggressively and if there's a tab in the background with a disconnected backend (for days) i could see this causing problems.

masenf (changes requested):

masenf (approved):

masenf (approved):

PR #2678, opened 2024-02-21
[REF-1988] API to Get instance of Arbitrary State class

masenf:
Implements a new api on `rx.State`:

## `other = await self.get_state(MyOtherState)`

This is called from an event handler to get access to the latest instance of `MyOtherState` associated with the same token as the currently running event handler.

## How it Works

With the recent Redis state sharding PR #2574, instead of fetching the entire state tree to process an event, only the "branch" of the state tree pertaining to the event handler is fetched from redis and deserialized.

For the given state containing the event handler the following are considered a "branch":

* all parent states (up to rx.State)
  * any substates of any parent with computed vars (so that the values can be recomputed after the event handler runs)
* all substates (recursively)

To access other "sibling" / "cousin" states, a new async API `State.get_state` is provided which can dynamically fetch the requested state from redis and patch it into the state tree as another "branch".

Practically, this means that state can scale horizontally without significant performance overhead (aside from hydration which occurs when the websocket comes up and on_load which occurs on page navigation). It also minimizes the changes needed to the existing State API, retaining compatibility with existing apps (that don't use `parent_state` chasing).

Delta calculations are performed in exactly the same way as before: by finding the top-level rx.State and recursively calling `get_delta` on all its substates... the difference is that not _all_ substates are included anymore, so the calculation can be performed more quickly.

Similarly during serialization, the top-level `rx.State` and its substates are persisted to redis, but only the substates that were previously fetched are serialized. If an event only targets a single substate in a mostly flat state hierachy, this has significant performance benefits for rapidly occuring events, like controlled sliders.

## Limitations and Caveats

Computed vars and cached computed vars cannot use this API, because properties do not work with `async` / `await`. This is probably fine: just define the computed var within the state containing the vars it will use.

Avoid computed vars in a deeply nested substate, because all of the parents will have to be fetched and instantiated for every event 😱. Any substate containing a computed var will be fetched for every event, to ensure it always gets recalculated.

Using an `rx.cached_var` is nice here, because those only get fetched when the "branch" containing it is fetched, since we already know what vars and substates the cached var depends on, and can be assured that it can only depend on vars in itself or one of its parent states. **Breaking Change**: When calculating the deps of a computed var, we no longer allow access to `parent_state`, `substates`, `get_substate`, or `get_state` (and it wouldn't work anyway). This ensures that all vars can be properly accounted for, and avoids the user shooting themselves in the foot with a cached_var that won't update properly.

Background events must first call `async with self` before accessing this API. However _after_ the block exits, there is no protection against subsequent modifications to the instance of the state, although these changes cannot be persisted and will not be considered for delta calculation (i.e. it just wont work). Down the road, we might implement some guardrails that will raise an exception when attempting to modify a stale state instance.

All of this only applies to **redis** use. The in-memory state manager used in dev mode works the same way it always has: keeping everything in memory all the time.

# Bonus

* [REF-2009] Instantiating a State class directly now raises an exception pointing to the state docs
* [CR Feedback] New helper methods for constructing and deconstructing "substate_key" (the combination of client_token and state full name)
* State sharding is now using `create_task` to achieve better concurrency. In reflex-web, this results in a 2x speedup on initial state creation (for a new user) and a 10x speedup deserializing existing app state.
* All the goodness from #2725 
  * [REF-2038] Do not send update_vars_internal as on_load when no cookies are set
  * More efficient `on_load_internal` handling.
  * More efficient state class initialization when the state does not exist in redis for the given token.

#### Some example code i was playing with

Don't ask me what it does 😹 

```python
import asyncio
import datetime

import reflex as rx


class State(rx.State):
    foo: str = ""


class Sub(rx.State):
    bar: str = ""

    def do_reset(self):
        self.reset()

    async def do_mod_other(self):
        other = await self.get_state(Other)
        other.quuc = str(len(self.bar))


class Child(Sub):
    baz: str = ""

    async def get_in_other_tree(self):
        sub = await self.get_state(Sub)
        print(sub.bar)
        gc = await self.get_state(GChild)
        print(gc.bar_cap)
        state = await self.get_state(rx.State)
        print(state.is_hydrated)


class GChild(Child):
    @rx.cached_var
    def bar_cap(self) -> str:
        return self.bar.upper()

    @rx.cached_var
    def baz_cap(self) -> str:
        return self.baz.upper()


class Other(rx.State):
    quuc: str = ""

    async def set_quuc(self, value: str):
        self.quuc = value
        child = await self.get_state(Child)
        child.baz = value


class DTState(rx.State):
    @rx.var
    def last_ts(self) -> str:
        return datetime.datetime.now().isoformat()


class DTChild(Other):
    @rx.cached_var
    def last_quuc(self) -> str:
        return self.quuc + " at " + datetime.datetime.now().isoformat()


class BackgroundState(rx.State):
    @rx.background
    async def background(self):
        # The following line raises an exception
        # print(self.substates)
        await asyncio.sleep(1)
        async with self:
            other = await self.get_state(Other)
            other.quuc = "From background"
            self.get_substate(tuple(BGSub.get_full_name().split(".")[1:])).bg_foo = "From background2"
        await asyncio.sleep(5)
        async with self:
            bg_sub = await self.get_state(BGSub)
            bg_sub.bg_foo = "Task Complete"


class BGSub(BackgroundState):
    bg_foo: str


class RouterDepState(BGSub):
    @rx.cached_var
    def router_dep(self) -> str:
        return self.router.page.full_raw_path


def index() -> rx.Component:
    return rx.fragment(
        rx.color_mode.button(rx.color_mode.icon(), float="right"),
        rx.vstack(
            rx.input(value=State.foo, on_change=State.set_foo),
            rx.input(value=Sub.bar, on_change=Sub.set_bar),
            rx.input(value=Child.baz, on_change=Child.set_baz),
            rx.input(value=Other.quuc, on_change=Other.set_quuc),
            rx.button("Reset", on_click=Sub.do_reset),
            rx.button("Get in other tree", on_click=Child.get_in_other_tree),
            rx.button("Measure bar", on_click=Sub.do_mod_other),
            rx.button(f"Background {BGSub.bg_foo}", on_click=BackgroundState.background),
            rx.heading("Quuc: ", Other.quuc),
            rx.heading("Last TS: ", DTState.last_ts),
            rx.heading("Last Quuc: ", DTChild.last_quuc),
            rx.text("GChild.bar_cap: ", GChild.bar_cap),
            rx.text("GChild.baz_cap: ", GChild.baz_cap),
            rx.text("Full path raw: ", RouterDepState.router_dep),
            rx.link("Go to another page", href="/foo"),
            gap="1.5em",
            padding_top="10%",
            width="50%",
        ),
    )


app = rx.App()
app.add_page(index)
app.add_page(index, route="/foo")
```


reflex/state.py
```diff
@@ -153,6 +153,50 @@ def __init__(self, router_data: Optional[dict] = None):
...
+        substate_key: The substate key.
+
+    Returns:
+        Tuple of token and state name.
+    """
+    token, _, state_name = substate_key.partition("_")
+    return token, state_name
+
+
+def _is_testing_env() -> bool:
```
picklelo, line 191: this one may be better in the `utils` directory

reflex/state.py
```diff
@@ -153,6 +153,50 @@ def __init__(self, router_data: Optional[dict] = None):
...
+    """Check if the app is running in a testing environment.
+
+    Returns:
+        True if the app is running in a testing environment, False otherwise.
```
picklelo, line 195: nit: can simplify these comments to "Whether the app is running in a testing environment."

reflex/state.py
```diff
@@ -218,24 +262,39 @@ def __init__(
...
             *args: The args to pass to the Pydantic init method.
             parent_state: The parent state.
             init_substates: Whether to initialize the substates in this instance.
+            _reflex_internal_init: A flag to indicate that the state is being initialized by the framework.
             **kwargs: The kwargs to pass to the Pydantic init method.
 
+        Raises:
+            RuntimeError: If the state is instantiated directly by end user.
         """
+        if not _reflex_internal_init and not _is_testing_env():
```
picklelo, line 282: I wonder if instead of checking `_is_testing_env` we should check the inverse, `is_reflex_run`. So that when external users work with it, they don't have to set this flag (similar to the hack I had to do in flexdown)

reflex/state.py
```diff
@@ -218,24 +262,39 @@ def __init__(
         *args,
         parent_state: BaseState | None = None,
         init_substates: bool = True,
+        _reflex_internal_init: bool = False,
```
picklelo, line 265: Can we handle this with only the flag to avoid adding another argument here.

reflex/state.py
```diff
@@ -1032,6 +1093,79 @@ def get_substate(self, path: Sequence[str]) -> BaseState:
             raise ValueError(f"Invalid path: {path}")
         return self.substates[path[0]].get_substate(path[1:])
 
+    def _get_state_manager(self) -> StateManager:
```
picklelo, line 1096: this can be a static method. Also, should this go inside `app.py` maybe?

reflex/state.py
```diff
@@ -1032,6 +1093,79 @@ def get_substate(self, path: Sequence[str]) -> BaseState:
...
+
+        Returns:
+            The state.
+
+        Raises:
+            RuntimeError: If the state is not cached and redis is not used.
+        """
+        target_state_parts = tuple(state_cls.get_full_name().split("."))
+
+        # fast case - maybe we already have this state instance cached
```
picklelo, line 1116: We should break these separate cases up into functions

reflex/state.py
```diff
@@ -1823,14 +2018,34 @@ async def get_state(
...
+            else:
+                # Only _always_dirty_substates need to be fetched to recalc computed vars.
+                fetch_substates = [
+                    state.get_class_substate(tuple(substate_name.split(".")))
+                    for substate_name in state_cls._always_dirty_substates
+                ]
+
+            tasks = {}
+            # Retrieve necessary substates from redis.
+            for substate_cls in fetch_substates:
```
picklelo, line 2035: Can this function be broken up?

picklelo (commented):
The syntax for accessing the state looks good.

On a broader note, I'm scared the code is getting too complicated, particularly now the state class has so many methods and that file is huge. It 

The methods are also a bit intimidating - for a newcomer I'm not sure they'd be able to reason around our codebase very easily now. We should have a design meeting to look over how we're doing anything to see if there's some things we can simplify.

reflex/state.py
masenf: The sucky thing about putting anything inside app.py is that `app.py` essentially imports the world and causes circular imports.

I'll move it to the top level of `state.py` for now.

reflex/state.py
masenf: lets talk about this on monday, i'm not really grokking the suggestion. how does `is_reflex_run` get determined?

masenf (commented):
> On a broader note, I'm scared the code is getting too complicated, particularly now the state class has so many methods and that file is huge.

I agree, there's too much going on inside the `BaseState`. I have some ideas for simplifying it, particularly via the use of descriptors for Var and EventHandler access, could clean up a lot of the `__init_subclass__` logic where we treat the same attribute differently on the class vs the instance.

> The methods are also a bit intimidating

I split up `get_state` into 6 smaller methods with hopefully more descriptive readable names to help paint a picture of what's going on. I don't see an easy way to get away from the intimidation of the `BaseState` generally though... I was thinking we could split up some of the distinct functionality into a few mixin classes (i.e. one for var/dirty/delta management, one for substate/parent state management, and one for event processing), but i'm not sure that would be more enlightening or worth the time at this point.

I think a bigger state refactoring is on the horizon, but I don't think we can justify further investment on the `BaseState` after getting this API in.

masenf (commented):
There's a problem here if a substate has a cached_var that depends on something in a parent state, but that substate isn't loaded because the event is being processed on a sibling state, then the cached var doesn't update (and it throws an exception).

reflex/.templates/jinja/web/utils/context.js.jinja2
```diff
@@ -25,10 +25,20 @@ export const clientStorage = {}
 
 {% if state_name %}
 export const state_name = "{{state_name}}"
-export const onLoadInternalEvent = () => [
-    Event('{{state_name}}.{{const.update_vars_internal}}', {vars: hydrateClientStorage(clientStorage)}),
-    Event('{{state_name}}.{{const.on_load_internal}}')
-]
+export const onLoadInternalEvent = () => {
```
picklelo, line 28: Nice, so we avoid sending the event if no cookies are set, is that right? Can we add some comments to this explaining high level

reflex/state.py
```diff
@@ -1292,20 +1531,40 @@ def _mark_dirty(self):
...
                 substate = substates[substate_name]
                 substate.dirty_vars.add(var)
                 substate._mark_dirty()
 
+    def _get_was_touched(self) -> bool:
+        """Check current dirty_vars and flag to determine if state instance was modified.
+
+        If any dirty vars belong to this state, mark _was_touched.
+
+        This flag determine whether this state instance should be persisted to redis.
```
picklelo, line 1551: nit: determines

reflex/state.py
```diff
@@ -1292,20 +1531,40 @@ def _mark_dirty(self):
...
         for var in self.dirty_vars:
             for substate_name in self._substate_var_dependencies[var]:
                 self.dirty_substates.add(substate_name)
-                if substate_name not in substates:
-                    continue
                 substate = substates[substate_name]
                 substate.dirty_vars.add(var)
                 substate._mark_dirty()
 
+    def _get_was_touched(self) -> bool:
```
picklelo, line 1546: Should this be called `set_was_touched`? I got confused below on line 1566 when you called it, as I didn't realize this also sets the value. The `get` is just the attribute itself right?

reflex/state.py
```diff
@@ -214,29 +250,47 @@ class BaseState(Base, ABC, extra=pydantic.Extra.allow):
...
             init_substates: Whether to initialize the substates in this instance.
+            _reflex_internal_init: A flag to indicate that the state is being initialized by the framework.
             **kwargs: The kwargs to pass to the Pydantic init method.
 
+        Raises:
+            RuntimeError: If the state is instantiated directly by end user.
         """
+        if not _reflex_internal_init and not is_testing_env():
+            raise RuntimeError(
+                "State classes should not be instantiated directly. The rx.App and its StateManager "
```
picklelo, line 280: Let's not mention `StateManager` it may just add confusion for a user who hits this error. Just "State classes should not be instantiated directly within a Reflex app. See .... "

reflex/state.py
```diff
@@ -923,8 +979,12 @@ def __getattribute__(self, name: str) -> Any:
             **super().__getattribute__("inherited_vars"),
             **super().__getattribute__("inherited_backend_vars"),
         }
-        if name in inherited_vars:
-            return getattr(super().__getattribute__("parent_state"), name)
+
+        # For now, handle router_data updates as a special case.
```
picklelo, line 983: Is this because router_data is only attached to the top level state?

reflex/state.py
```diff
@@ -1036,6 +1093,168 @@ def get_substate(self, path: Sequence[str]) -> BaseState:
...
+            target_state_cls: The class of the state to populate parent states for.
+
+        Returns:
+            The parent state instance of target_state_cls.
+
+        Raises:
+            RuntimeError: If the state is not cached and redis is not used.
+        """
+        state_manager = get_state_manager()
+        if not isinstance(state_manager, StateManagerRedis):
```
picklelo, line 1168: Shouldn't this work for MemoryStateManager also (though no caching etc would have to be done, could this just be a no-op?)

reflex/state.py
```diff
@@ -1036,6 +1093,168 @@ def get_substate(self, path: Sequence[str]) -> BaseState:
...
+        Returns:
+            The instance of state_cls associated with this state's client_token.
+        """
+        if self.parent_state is None:
+            root_state = self
+        else:
+            root_state = self._get_parent_states()[-1][1]
+        return root_state.get_substate(state_cls.get_full_name().split("."))
+
+    async def _get_state_from_redis(self, state_cls: Type[BaseState]) -> BaseState:
```
picklelo, line 1210: In the future should we move these to another class to keep things modular, so there's no explicit Redis dependencies within this file?

reflex/state.py
```diff
@@ -1238,6 +1457,28 @@ def _dirty_computed_vars(self, from_vars: set[str] | None = None) -> set[str]:
             for cvar in self._computed_var_dependencies[dirty_var]
         )
 
+    @classmethod
+    def _potentially_dirty_substates(cls) -> set[Type[BaseState]]:
```
picklelo, line 1461: So does this mean any substate with a computed var will always be refetched on each event?

reflex/state.py
masenf: Yeah i went back and forth on what to call this one... The flag is called `_was_touched`, and this function `_get_was_touched` checks both the flag and other conditions (and then updates the flag when those conditions indicate touching has occurred). The reason you would call this function is to decide whether the state has been modified since it was instantiated. I'll split the getting and updating logic out to clarify.

reflex/state.py
masenf: Before we used to push `router_data` into all substates, but that was a kludge that i introduced a long time ago while working on cached_var. Ultimately we only want to keep one copy of `router_data`, and we keep that on the base `rx.State` now.

The reason why `router_data` needs this special case is because it's present in `get_skip_vars`, so it doesn't get the normal "inherited var" tracking because it's treated as a non-var. Contrast this with `router`, which is always considered an inherited var and handled in the regular way.

The long term fix to remove this special case is to separate "internal" vars from "non-serialized" vars. I think that might actually happen sooner than later when the "Remove Cloudpickle" PR comes in and refactors how serialization is handled.

reflex/state.py
masenf: If we hit this code path in the non-redis case, it's certainly a bug (or mismatching frontend and backend code). We could try to proceed, but `get_state` first checks for the requested class instance from the top-level state. If that fails, then hitting the state manager again isn't going to help.

I updated the exception message indicating that this is likely a bug to hit this point, and hopefully they will report it on github/discord.

Otherwise this check is here primarily for `pyright`, which won't let me call `state_manager.get_state` with the extra args unless i confirm that it's indeed an instance of `StateManagerRedis`

reflex/state.py
masenf: not sure how it would be structured... we only want downstream users to inherit from a single `rx.State` type, and have it work regardless of the state manager that's backing it.

reflex/state.py
masenf: yes that's a correct read.

currently we promise that computed vars will always update and cached_vars will update when their deps change. to keep that promise, we have to fetch the states containing those vars.

it's not really "nice", but it's better than getting all the states all the time, and users can optimize their own apps by limiting the use of computed vars or replacing them with cached vars in a shallow hierarchy.

this keeps the mental model simple for downstream users.

reflex/state.py
picklelo: not sure either, we can think later if it's possible. Because it would be nice if people could use their own StateManagers later e.g. memcachd or whatever else they want

reflex/state.py
picklelo: sounds good - I suppose in reflex-web we don't have many of these which is good. We should document all this thoroughly later in the substate section

picklelo (approved):

PR #2690, opened 2024-02-22
docker-example overhaul

masenf:
Update docker-example with a more realistic multi-compose deployment, and also a more simplistic single-image deploy.

* Use `uv` for faster bootstrapping
* Separate simple and production-ready docker files
* Split compose.yaml into 3 parts
  * Persist sqlite db and tls keys
  * Include postgres and redis
  * Include Adminer and redis-commander for adminstration
  * Suppose upload persistence
* Update documentation
* Update Caddyfile for compatibility with new Upload API

Have a go with my test app: https://github.com/masenf/repro-datetime-model/tree/main

### DIFF
--- a/docker-example/Dockerfile
+++ b/docker-example/Dockerfile
@@ -1,39 +1,25 @@
-# Stage 1: init
-FROM python:3.11 as init
+# This Dockerfile is used to deploy a simple single-container Reflex app instance.
+FROM python:3.11
 
-# Pass `--build-arg API_URL=http://app.example.com:8000` during build
-ARG API_URL
+ARG uv=/root/.cargo/bin/uv
+
+# Install `uv` for faster package boostrapping
+ENV VIRTUAL_ENV=/usr/local
+ADD --chmod=755 https://astral.sh/uv/install.sh /install.sh
+RUN /install.sh && rm /install.sh
 
 # Copy local context to `/app` inside container (see .dockerignore)
 WORKDIR /app
 COPY . .
 
-# Create virtualenv which will be copied into final container
-ENV VIRTUAL_ENV=/app/.venv
-ENV PATH="$VIRTUAL_ENV/bin:$PATH"
-RUN python3.11 -m venv $VIRTUAL_ENV
-
-# Install app requirements and reflex inside virtualenv
-RUN pip install -r requirements.txt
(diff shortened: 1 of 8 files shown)


docker-example/Dockerfile
```diff
@@ -1,39 +1,28 @@
-# Stage 1: init
-FROM python:3.11 as init
+# This Dockerfile is used to deploy a simple single-container Reflex app instance.
+FROM python:3.11
 
-# Pass `--build-arg API_URL=http://app.example.com:8000` during build
-ARG API_URL
+ARG uv=/root/.cargo/bin/uv
```
picklelo, line 4: Should we focus on simplicity for this Dockerfile and use just the basic tools? I think it would be nice if the base one we had just minimized lines of code to get it working, and we can have another more optimized one.

masenf: changed to just using `pip`

docker-example/Dockerfile
```diff
@@ -1,39 +1,21 @@
...
-RUN adduser --disabled-password --home /app reflex
-COPY --chown=reflex --from=init /app /app
-USER reflex
-ENV PATH="/app/.venv/bin:$PATH" API_URL=$API_URL
+# Needed until Reflex properly passes SIGTERM on backend.
+STOPSIGNAL SIGKILL
 
-CMD reflex db migrate && reflex run --env prod --backend-only
+# Always apply migrations before starting the backend.
+CMD [ -d alembic ] && reflex db migrate; reflex run --env prod
```
picklelo, line 21: This ends up exporting the app again. Maybe we need a way to run an exported app? Otherwise we should remove the export step above, as it does it twice.

picklelo (changes requested):

docker-example/Dockerfile
masenf: i agree it's not ideal... but if i don't do the export during build time, then it has to fetch all of the node modules at runtime, which can be very very slow when running the container.

i'm thinking of making the "simple" docker file also use a reverse proxy and serve the frontend statically; then we could `reflex run --env prod --backend-only` and it would be really fast to start up, but there's some extra complexity since there would be another service and config file to make that happen.

picklelo (approved):

PR #2696, opened 2024-02-22
use direction props from radix.Flex

Lendemor:
new usage for `rx.stack`

```python
rx.stack(..., direction="row")
rx.hstack()

rx.stack(..., direction="column")
rx.vstack()

rx.stack(..., direction="row-reverse")
rx.stack(..., direction="column-reverse")
```

### DIFF
--- a/reflex/components/radix/themes/layout/stack.py
+++ b/reflex/components/radix/themes/layout/stack.py
@@ -42,7 +42,7 @@ class Stack(Flex):
 class VStack(Stack):
     """A vertical stack component."""
 
-    direction: Var[LiteralFlexDirection] = Var.create_safe("column")
+    direction: Var[LiteralFlexDirection] = "column"  # type: ignore
 
 
 class HStack(Stack):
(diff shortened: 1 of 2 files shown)


reflex/components/radix/themes/layout/stack.py
```diff
@@ -41,12 +41,32 @@ def create(
 class VStack(Stack):
     """A vertical stack component."""
 
-    def _apply_theme(self, theme: Component):
-        self.style.update({"flex_direction": "column"})
+    @classmethod
```
picklelo, line 44: Couldn't we just event set it as a prop:

`direction: Var[Literal[...]] = "column"`

reflex/components/radix/themes/layout/stack.py
```diff
@@ -41,12 +41,32 @@ def create(
...
+        """
+        return super().create(*children, direction="column", **props)
 
 
 class HStack(Stack):
     """A horizontal stack component."""
 
-    def _apply_theme(self, theme: Component):
-        self.style.update({"flex_direction": "row"})
+    @classmethod
```
picklelo, line 61: and same here

picklelo (dismissed):

picklelo (approved):

PR #2697, opened 2024-02-22
pass lang and custom_attrs from app to html root

macmoritz:
The HTML root element should have a `lang=en` attribute, mostly for SEO reasons. Also passing custom attributes, can be used for example to add version or xmlns tags.
  
See HTML docs:
- https://www.w3.org/TR/2010/WD-html-markup-20100624/html.html#html.attrs.manifest
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/html?retiredLocale=de 
  
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -73,7 +73,6 @@ from reflex.state import (
 )
 from reflex.utils import console, exceptions, format, prerequisites, types
 from reflex.utils.imports import ImportVar
-from reflex.vars import Var
 
 # Define custom types.
 ComponentCallable = Callable[[], Component]
@@ -135,7 +134,7 @@ class App(Base):
     html_lang: Optional[str] = None
 
     # Attributes to add to the html root tag of every page.
-    html_custom_attrs: Optional[Dict[str, Union[Var, str]]] = None
+    html_custom_attrs: Optional[Dict[str, str]] = None
 
     # A component that is present on every page.
     overlay_component: Optional[
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -132,7 +132,7 @@ class App(Base):
     head_components: List[Component] = []
 
     # The language to add to the html root tag of every page.
-    html_lang: Optional[Union[Var[str], str]] = None
+    html_lang: Optional[str] = None
 
     # Attributes to add to the html root tag of every page.
     html_custom_attrs: Optional[Dict[str, Union[Var, str]]] = None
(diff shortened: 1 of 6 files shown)


macmoritz:
Thanks for the quick review @masenf!
Currently the tests are working but it does not get rendered. Thats the reason this PR is in draft status. 
I will change the status as soon as the feature is completed and come back to you!

macmoritz:
Hey @masenf!   
This is now finally ready, sorry for the delay. Feel free to review and thanks in advance.
Hope the commits get squashed :)

reflex/app.py
```diff
@@ -130,6 +131,12 @@ class App(Base):
     # Components to add to the head of every page.
     head_components: List[Component] = []
 
+    # The language to add to the html root tag of every page.
+    html_lang: Optional[Union[Var[str], str]] = None
+
+    # Attributes to add to the html root tag of every page.
+    html_custom_attrs: Optional[Dict[str, Union[Var, str]]] = None
```
masenf, line 138: For now, accessing state Var won't work in the _document.js file; so i think we should just accept `str` type instead of `Union[Var[str], str]`

masenf (changes requested):

reflex/app.py
macmoritz: Fixed it for `html_lang`. How should we handle the `custom_attrs`? This attribute is inherited from the `Component` class.

masenf: i think just accept it as a `str` in the App, and then you can convert it to a Var in the compile function, like you're already doing.

the main problem is that at the time `_document.js` is rendered by Next, the state hasn't been set up yet. So Vars in general are fine, but State Vars are not.

macmoritz: Sounds logical.
But due to possible nested dicts this wouldn't be so easy to implement.

Shouldn't it be possible to dirty fix `custom_attrs` with a `# type: ignore`?
In the outer layer (app) we accept `Dict[str, str]` which can be "casted" to Html's `Dict[str, Union[Var, str]]` with the type ignoration. I want to achieve, that we can fully stick to the inherited `custom_attrs`.

masenf: i dont see the need for the `# type: ignore`. When i change `html_lang` to `Optional[Dict[str, str]]` i don't get any type checking errors from pyright.

i would assume that `Dict[str, str]` is "compatible" with `Dict[str, Union[Var, str]]`

macmoritz: Thanks for the detailed review. Is fixed.

masenf (approved):

PR #2700, opened 2024-02-22
add event triggers for rx.video

Lendemor:


reflex/components/react_player/react_player.py
```diff
@@ -43,3 +45,28 @@ class ReactPlayer(NoSSRComponent):
...
     height: Var[str]
+
+    def get_event_triggers(self) -> Dict[str, Any]:
+        """Get the event triggers.
+
+        Returns:
+            The event triggers.
+        """
+        return {
+            "onReady": lambda: [],
```
picklelo, line 56: we should use the snake_case names

picklelo (dismissed):

masenf:
reflex-web really needs a better method for getting docs of event triggers; right now this is causing chicken-egg failures because reflex-web doesn't know about "on_start", "on_ended", etc

reflex/components/react_player/audio.pyi
```diff
@@ -32,49 +32,52 @@ class Audio(ReactPlayer):
         class_name: Optional[Any] = None,
         autofocus: Optional[bool] = None,
         custom_attrs: Optional[Dict[str, Union[Var, str]]] = None,
-        on_blur: Optional[
+        onBuffer: Optional[
```
masenf, line 35: are these supposed to be camel case? or did they just not get regenerated?

picklelo (approved):

PR #2715, opened 2024-02-24
fix inherited backend var deps

### LINKED ISSUE #2714: Bug: cached ComputedVars depending on parent state backend vars
**Describe the bug**
I think I discovered a bug while trying to migrate reflex away from cloudpickle.
Cached ComputedVars depending on parent state backend vars do not get recomputed.

**To Reproduce**
```python
import reflex as rx

class Parent(rx.State):
    _backend: int = 0

class State(Parent):
    _backend_child: int = 0

    @rx.cached_var
    def backend_var(self) -> int:
        return self._backend

    @rx.cached_var
    def backend_var_child(self) -> int:
        return self._backend_child

    def increase(self) -> None:
        self._backend += 1
        self._backend_child += 1
        print(f"{self._backend=}, {self._backend_child=}")


def index() -> rx.Component:
    return rx.vstack(
        rx.text(f"backend_var: {State.backend_var}"),
        rx.text(f"backend_var_child: {State.backend_var_child}"),
        rx.button(
            "Increase",
            on_click=State.increase,  # type: ignore
            size="4",
        ),
    )

app = rx.App()
app.add_page(index)
```

**Expected behavior**
cached ComputedVars get recomputed if a dependent backend var from a parent state changes.

 - Python Version: 3.12.2
 - Reflex Version: 0.4.1

benedikt-bartscher:
Closes #2714

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -950,9 +950,14 @@ class BaseState(Base, ABC, extra=pydantic.Extra.allow):
             # unwrap proxy objects when assigning back to the state
             value = value.__wrapped__
 
-        # Set the var on the parent state.
-        inherited_vars = {**self.inherited_vars, **self.inherited_backend_vars}
-        if name in inherited_vars:
+        # Set backend var on the parent state.
+        if name in self.inherited_backend_vars:
+            setattr(self.parent_state, name, value)
+            self.dirty_vars.add(name)
+            return
+
+        # Set var on the parent state.
+        if name in self.inherited_vars:
             setattr(self.parent_state, name, value)
             return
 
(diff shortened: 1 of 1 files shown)


masenf (commented):
Interesting, i wonder why it wasn't getting marked dirty in the parent state and being pushed down in the substates via this block

```python
        if (
            types.is_backend_variable(name, self.__class__)
            and name not in RESERVED_BACKEND_VAR_NAMES
        ):
            self._backend_vars.__setitem__(name, value)
            self.dirty_vars.add(name)
            self._mark_dirty()
            return
```

masenf (approved):

Issue #2399, opened 2024-01-16
Serialization issue when using REDIS


PasqualePuzio:
When we run the application storing the state in memory, everything works fine. However, when we use Redis, we get the following exception.

We're not storing any custom or complex object in our states, we only use primitive types such dict, list, int and str.
I'm not sure if it's related to this issue, but the some of these variables can be set to None sometimes.

Do you have any clue on what might be the root cause here?

Task exception was never retrieved
future: <Task finished name='Task-27' coro=<AsyncServer._handle_event_internal() done, defined at /root/fantacalcio-frontend-web/.venv/lib/python3.10/site-packages/socketio/async_server.py:605> exception=NotImplementedError('object proxy must define __reduce_ex__()')>
Traceback (most recent call last):
  File "/root/fantacalcio-frontend-web/.venv/lib/python3.10/site-packages/socketio/async_server.py", line 607, in _handle_event_internal
    r = await server._trigger_event(data[0], namespace, sid, *data[1:])
  File "/root/fantacalcio-frontend-web/.venv/lib/python3.10/site-packages/socketio/async_server.py", line 643, in _trigger_event
    return await handler.trigger_event(event, *args)
  File "/root/fantacalcio-frontend-web/.venv/lib/python3.10/site-packages/socketio/async_namespace.py", line 37, in trigger_event
    ret = await handler(*args)
  File "/root/fantacalcio-frontend-web/.venv/lib/python3.10/site-packages/reflex/app.py", line 1079, in on_event
    async for update in process(self.app, event, sid, headers, client_ip):
  File "/root/fantacalcio-frontend-web/.venv/lib/python3.10/site-packages/reflex/app.py", line 871, in process
    async with app.state_manager.modify_state(event.token) as state:
  File "/usr/lib/python3.10/contextlib.py", line 206, in __aexit__
    await anext(self.gen)
  File "/root/fantacalcio-frontend-web/.venv/lib/python3.10/site-packages/reflex/state.py", line 1697, in modify_state
    await self.set_state(token, state, lock_id)
  File "/root/fantacalcio-frontend-web/.venv/lib/python3.10/site-packages/reflex/state.py", line 1682, in set_state
    await self.redis.set(token, cloudpickle.dumps(state), ex=self.token_expiration)
  File "/root/fantacalcio-frontend-web/.venv/lib/python3.10/site-packages/cloudpickle/cloudpickle_fast.py", line 73, in dumps
    cp.dump(obj)
[...]

Issue #2189, opened 2023-11-17
[REF-1192] Cannot use openai>1 API with State when using redis (hosting service)


masenf:
**Describe the bug**
The openai>1 client has internal non-picklable objects.

When running in prod mode with redis available, Reflex attempts to cloudpickle the State.

If any event handlers in the State depend on the OpenAI client, then cloudpickle will attempt to pickle the OpenAI client since it is used in the closure of a function defined on an object being pickled.

This fails due to containing a _thread.RLock and SSLContext (among other things)

**To Reproduce**
Minimal app

```python
from openai import OpenAI

import reflex as rx


client = OpenAI()


class State(rx.State):
    def bar(self):
        print(client)  # cloudpickle attempts to serialize values used in the closure


def index() -> rx.Component:
    return rx.text("Nothing too exciting")


app = rx.App()
app.add_page(index)
app.compile()
```

MUST RUN WITH REDIS and `--env prod`

```console
docker run -d -p 6379:6379 redis
REDIS_URL=localhost:6379 reflex run --env prod
```

```console
  File "/opt/homebrew/Cellar/python@3.11/3.11.5/Frameworks/Python.framework/Versions/3.11/lib/python3.11/contextlib.py", line 204, in __aenter__
    return await anext(self.gen)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/VENV-release-0.3.4a1/lib/python3.11/site-packages/reflex/state.py", line 1605, in modify_state
    state = await self.get_state(token)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/VENV-release-0.3.4a1/lib/python3.11/site-packages/reflex/state.py", line 1567, in get_state
    await self.set_state(token, self.state())
  File "/Users/masenf/code/reflex-dev/VENV-release-0.3.4a1/lib/python3.11/site-packages/reflex/state.py", line 1592, in set_state
    await self.redis.set(token, cloudpickle.dumps(state), ex=self.token_expiration)
                                ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/VENV-release-0.3.4a1/lib/python3.11/site-packages/cloudpickle/cloudpickle_fast.py", line 73, in dumps
    cp.dump(obj)
[...]

PR #2717, opened 2024-02-24
[WIP] Drop cloudpickle v2

benedikt-bartscher:
In prod mode (`StateManagerRedis`) reflex cloudpickles the whole State which causes problems f.e. with SQLAlchemy, but also with many other libraries. In many cases this also results in unnecessary big redis payloads (all functions, classes, attributes, etc. get pickled for every session).

This PR is still a draft because there are some design decisions to be made.

Essentially, I see 3 options for cloudpickle:
- completely drop cloudpickle
- auto-fallback to cloudpickle (if no serializer could be found)
- explicit cloudpickle only (f.e. with `@rx.var(cloudpickle=True)`)

And the question if we should only rely on pydantics "deserialization" (which the code in this PR currently does) or implement custom "deserialization". Reflex only has an api for custom serializers, but we need both ways to store/restore states from redis.  

We should also think about either disabling pydantic validation for `rx.BaseState` by somehow patching it (pydantic v1 does not allow to disable validation) or using pydantics `construct` instead of `parse_obj` [here](https://github.com/reflex-dev/reflex/pull/2717/files#diff-704efe440192781e072231490ac1d12defd1aba5f6debafbc7907eebcc2abcc7R1948). Using `construct` does require manual construction of sub-models.  

You have to set a `REDIS_URL` env variable to test this PR. Do not forget to clear redis or create a new reflex session, else you will experience problems.  

#2399  
#2189  
#1824

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -932,7 +932,7 @@ async def process(
             # assignment will recurse into substates and force recalculation of
             # dependent ComputedVar (dynamic route variables)
             state.router_data = router_data
-            state.router = RouterData(router_data)
+            state.router = RouterData.from_router(router_data)
 
         # Preprocess the event.
         update = await app.preprocess(state, event)
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -28,7 +28,6 @@ from typing import (
     Type,
 )
 
-import cloudpickle
 import pydantic
 import wrapt
 from redis.asyncio import Redis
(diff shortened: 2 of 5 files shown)


Lendemor:
I think this PR is outdated with recent changes to var system / serialization, so closing for now.
Let us know if it wasn't the case.

PR #2736, opened 2024-02-27
fix telemetry for init event

Lendemor:
since `telemetry.send` rely on `.web/reflex.json` folder existing, send the event after the creation of the folder `.web`.


picklelo (approved):
nice catch

picklelo:
Can we add some unit tests, either here or in a follow up, that actually tries to `init` the app and verifies the correct event is trying to send

PR #2746, opened 2024-02-28
add initial_value for base vars

benedikt-bartscher:
Followup of #2670 which implemented `initial_value` for computed vars.

```python
import reflex as rx

from pydantic import Field


class State(rx.State):
    basevar: str = Field(default="basevar", initial_value="initial")

    @rx.var(initial_value="initial")
    def test(self) -> str:
        return "Hello, world!"


#  cv = State.test
#  cv_initial = cv._initial_value
#  print(f"{cv=}, {cv_initial=}")

#  state = State()
#  state_d = state.dict(initial=True)
#  initial_d_state = state_d["state.state"]
#  initial_test = initial_d_state["test"]
#  print(f"Initial state: {initial_test=}")


def index() -> rx.Component:
    return rx.center(
        rx.vstack(
            rx.heading(State.test),
            rx.heading(State.basevar),
            rx.button(
                "Check out our docs!",
                size="4",
            ),
            align="center",
            spacing="7",
            font_size="2em",
        ),
        height="100vh",
    )


app = rx.App()
app.add_page(index)
```

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -1616,7 +1616,15 @@ class BaseState(Base, ABC, extra=pydantic.Extra.allow):
             self._mark_dirty()
 
         base_vars = {
-            prop_name: self.get_value(getattr(self, prop_name))
+            prop_name: self.__fields__[prop_name].field_info.extra.get("initial_value")
+            if initial
+            and not isinstance(
+                self.__fields__[prop_name].field_info.extra.get(
+                    "initial_value", types.Unset()
+                ),
+                types.Unset,
+            )
+            else self.get_value(getattr(self, prop_name))
             for prop_name in self.base_vars
         }
         if initial:
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -1596,6 +1596,25 @@ class BaseState(Base, ABC, extra=pydantic.Extra.allow):
             return super().get_value(key.__wrapped__)
         return super().get_value(key)
 
+    def _get_base_var_value(self, prop_name: str, initial: bool) -> Any:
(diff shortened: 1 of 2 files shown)


reflex/state.py
```diff
@@ -1616,7 +1616,15 @@ def dict(
             self._mark_dirty()
 
         base_vars = {
-            prop_name: self.get_value(getattr(self, prop_name))
+            prop_name: self.__fields__[prop_name].field_info.extra.get("initial_value")
+            if initial
+            and not isinstance(
```
picklelo, line 1621: Can we add a comment on what this is checking, and potentially move it to a helper function?

masenf:
@benedikt-bartscher can you elaborate on the use case for this feature? I'm trying to understand where the "initial_value" would show up? Seems like a pre-hydration placeholder maybe?

benedikt-bartscher:
@masenf  you are right, it allows setting a different value in the initial state dict (pre-hydration). This can f.e. be used to prevent `default_factorys` to be executed at compile time.

reflex/state.py
benedikt-bartscher: Yes, I will clean it up later, thanks for the quick review!

masenf:
> This can f.e. be used to prevent `default_factorys` to be executed at compile time.

At compile time, the State class is still instantiated, so i believe that would call all of the default_factory functions regardless. There would need to be some mechanism to delay the execution of these functions in the Base itself i think.

Lendemor:
Marking this as draft until re-synced with main (can be closed if it's no longer needed)

PR #2745, opened 2024-02-28
Add config to rx.plotly component

amirmolavi:
Address issue #2744

PR #2750, opened 2024-02-28
fix black+ruff

benedikt-bartscher:
#2745 broke pre-commit (black+ruff) on main


picklelo (approved):
thanks for fixing, I missed that

PR #2760, opened 2024-03-01
app_module_for_backend: wait for _compile in prod mode

masenf:
Prod mode uses separate worker processes that fork from the main process.

If the app is not fully compiled when the fork occurs, any further changes to the app (like mounting the `_upload` endpoint) will not be reflected in the workers. This is not a performance hit because compile is skipped anyway for backend processes and hot reload is not in the picture for prod mode.

### DIFF
--- a/reflex/compiler/compiler.py
+++ b/reflex/compiler/compiler.py
@@ -18,6 +18,7 @@ from reflex.components.component import (
 from reflex.config import get_config
 from reflex.state import BaseState
 from reflex.style import LIGHT_COLOR_MODE
+from reflex.utils.exec import is_prod_mode
 from reflex.utils.imports import ImportVar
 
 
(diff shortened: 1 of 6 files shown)


reflex/compiler/compiler.py
```diff
@@ -66,7 +67,7 @@ def _compile_theme(theme: dict) -> str:
 
 
 def _is_dev_mode() -> bool:
-    return os.environ.get("REFLEX_ENV_MODE", "dev") == "dev"
+    return not is_prod_mode()
```
picklelo, line 70: nit: should we put `_is_dev_mode` in the same file as `is_prod_mode`, and also make their naming consistent with the initial underscore

picklelo (approved):

picklelo (approved):

PR #2762, opened 2024-03-01
[REF-2087] Better rx.progress styling integration with radix themes

masenf:
Support the `radius` prop on `ProgressRoot`, via data-radius and CSS tokens

Support the `color_scheme` on `ProgressIndicator`, via data-accent-color and CSS token

Move high-level `Progress` to a real `Component` subclass to get better pyi hinting

Allow overriding the background color of the `ProgressIndicator` via low-level api

Remove `value` and `max` props from `ProgressRoot` (these only apply to `ProgressIndicator`)

# Sample code

```python
import reflex as rx


def index() -> rx.Component:
    return rx.center(
        rx.theme_panel(),
        rx.vstack(
            # custom height
            rx.progress(value=25, height="5px"),
            # custom radius
            rx.progress(value=50, height="15px", radius="full"),
            # override color scheme
            rx.progress(value=75, color_scheme="tomato"),
            # override radius (low-level)
            rx.progress.root(
                rx.progress.indicator(
                    value=88,
                    max=100,
                ),
                radius="none",
            ),
            # override color scheme (low-level)
            rx.progress.root(
                rx.progress.indicator(
                    value=88,
                    max=100,
                    color_scheme="amber",
                ),
            ),
            # override progress background (low-level)
            rx.progress.root(
                rx.progress.indicator(
                    value=88,
                    max=100,
                    background_color="papayawhip",
                ),
                height="50px",
            ),
            rx.button("foo"),
            width="650px",
        ),
        height="100vh",
    )


app = rx.App(theme=rx.theme(accent_color="plum"))
app.add_page(index)
```

### DIFF
--- a/reflex/components/radix/primitives/progress.py
+++ b/reflex/components/radix/primitives/progress.py
@@ -71,7 +71,7 @@ class ProgressIndicator(ProgressComponent):
 
         self.style = Style(
             {
-                "background-color": color("accent", 9),
+                "background_color": color("accent", 9),
                 "width": "100%",
                 "height": "100%",
                 f"transition": f"transform {DEFAULT_ANIMATION_DURATION}ms linear",
--- a/reflex/components/radix/primitives/progress.py
+++ b/reflex/components/radix/primitives/progress.py
@@ -116,8 +116,8 @@ class Progress(ProgressRoot):
 
         return ProgressRoot.create(
             ProgressIndicator.create(
-                value=props.get("value"),
-                max=props.get("max", 100),
+                value=props.pop("value", 0),
+                max=props.pop("max", 100),
                 **progress_indicator_props,
             ),
             **props,
--- a/reflex/components/radix/primitives/progress.py
+++ b/reflex/components/radix/primitives/progress.py
@@ -2,13 +2,13 @@
 
 from __future__ import annotations
 
-from typing import Optional, Union
+from typing import Optional
 
 from reflex.components.component import Component, ComponentNamespace
 from reflex.components.core.colors import color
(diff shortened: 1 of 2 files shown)


reflex/components/radix/primitives/progress.py
```diff
@@ -63,22 +66,12 @@ class ProgressIndicator(ProgressComponent):
...
-            self.color_scheme if self.color_scheme is not None else global_color_scheme,  # type: ignore
-            9,
-        )
+        if self.color_scheme is not None:
+            self.custom_attrs["data-accent-color"] = self.color_scheme
 
         self.style = Style(
             {
-                "background-color": color_scheme,
+                "background-color": color("accent", 9),
```
picklelo, line 74: We should make the key `background_color`. Also, I'm a bit unclear on how `color_scheme` plays together with the theme. Is it right that we should always use the theme color, but if color_scheme is set it overrides it? In that case, why do we make it `accent` here?

picklelo (commented):
Big improvement it works with theme now - just one comment, maybe we need to document better the interaction between `theme` and `color_scheme`

picklelo (approved):

reflex/components/radix/primitives/progress.py
masenf: we set the `data-accent-color` attribute on the element to the `color_scheme` provided by the user. and `var(--accent-9)` picks up the "nearest" `data-accent-color`.

this is essentially how all of the existing radix themes components handle overriding the `color`, `radius`, etc props: they don't change the css, instead they set the `data-*` attribute and the CSS tokens respond to that.

picklelo (approved):

Issue #2737, opened 2024-02-27
Feature Request: A new component-specific event trigger "on_file_select"


Anga205:
**Describe the feature**
this event would trigger as soon as a new file(s) as been selected in an rx.upload component, this would make things possible like, instantly uploading files that have been dragged and dropped into the website

masenf:
I think this would be a good one for an outside contributor to pick up; essentially we just need to plumb through the `onDropAccepted` callback of ReactDropzone as an event trigger and allow the user to specify their upload handler.

In the meantime, here is a workaround using `rx.moment`: https://github.com/masenf/rx_shout/blob/94a9fb9c8e91ebb21e52e43614c40462e55beb33/rx_shout/components/image_upload.py#L42-L51

```python
        # Auto-upload after half second.
        rx.moment(
            interval=rx.cond(
                rx.selected_files(UPLOAD_ID) & ~UploadState.is_uploading,
                auto_upload_delay_ms,
                0,
            ),
            on_change=lambda _: UploadState.handle_upload(
                rx.upload_files(
                    upload_id=UPLOAD_ID,
                    on_upload_progress=UploadState.on_upload_progress,
                ),
            ),
            display="none",
        )
```

emmakodes:
I am picking this up

Anga205:
well on_drop is pretty much exactly what this thread asked for, and now that its been added, ill close this issue

PR #2766, opened 2024-03-03
Expose `on_drop` event trigger for rx.upload component.

masenf:
If `on_drop` is provided, it should be an EventSpec accepting the special rx.upload_files() arg.

When `on_drop` is provided, it will be called immediately when files are selected. The default functionality of saving a file list to be uploaded later will not be available.

#2737

### DIFF
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -374,21 +374,12 @@ class Component(BaseComponent, ABC):
 
         arg_spec = triggers.get(event_trigger, lambda: [])
 
-        wrapped = False
         # If the input is a single event handler, wrap it in a list.
         if isinstance(value, (EventHandler, EventSpec)):
-            wrapped = True
             value = [value]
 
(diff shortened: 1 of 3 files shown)


picklelo (commented):
Works well, but I get the Warning

```
DeprecationWarning: EventChain has been deprecated in version 0.2.8 to avoid confusion, only use yield API. It will be completely removed in 0.5.0
```

Should we remove the event chain warning?

masenf:
> Should we remove the event chain warning?

I'm thinking probably yes, undep this feature and remove the warning. We've gotten at least 2 users on discord that have expressed concern about losing the ability to pass multiple event handlers for a single trigger, and for whatever reason don't want to explicitly tie them together via the `yield` syntax. For example, if I have two handlers that otherwise aren't related, I don't necessarily want to have one always call the other, in case they are used in other situations as well.

picklelo (approved):

PR #2775, opened 2024-03-04
[REF-2117]:`rx.color_mode_cond` to work in f-strings

ElijahAhianyo:
The code below should now work with this PR change:

```python
def index():
    return rx.box(
        rx.text("random"),
        bg="red",
        border_left=f'1px {rx.color_mode_cond("var(--mauve-11)", "var(--mauve-10)")} solid',
    )
```


picklelo (approved):
sorry for the delay - works well for me

PR #2779, opened 2024-03-04
Perf improvements (part 1)

masenf:
Low hanging fruit in reflex itself:

* avoid imports and exception handlers in hot paths inside reflex.utils.types
* avoid recreating regex pattern in _decode_var
* avoid complex logic in _decode_var if the var does not have the special tag


picklelo (approved):
nice

Issue #2788, opened 2024-03-05
Problem when running ```reflex init``` in Windows 10, .gitignore overwrite.


cllatser:
**Describe the bug**

Hello, I use Windows for programming, both natively and through WSL2. After going crazy because git was ignoring the .gitignore file, I realized that when performing the `reflex init` command in Windows 10 native, it overwrites the file and changes its encoding type, so Windows git doesn't recognize it and ignores it. Is there any option to prevent it from overwriting the .gitignore file if it already exists?

**Expected behavior**

If the .gitignore file already exists, it should not be overwritten when executing `reflex init`.

**Screenshots**

![git001](https://github.com/reflex-dev/reflex/assets/65849393/89c8d210-2bcb-42f6-88a1-56aa0626808f)
![git002](https://github.com/reflex-dev/reflex/assets/65849393/ad933fbc-ecfc-41b3-9415-4a77fe1f5d76)

**Specifics (please complete the following information):**
 - Python Version: 3.11.8
 - Reflex Version: 0.4.3
 - OS: Windows 10
 - Browser (Optional):

picklelo:
We can definitely support that. Only thing is then users may forget to manually add `.web` and other folders in. Maybe there's a solution where we can keep it overriding, but fix the final encoding type?

Would welcome any contributions to this issue.

cllatser:
I have put in the `initialize_gitignore()` function, when it opens the file to write the parameter` newline='\n'` and it works fine, as expected. I can do a PR but as we know that on Mac there will be no problems.

![newline](https://github.com/reflex-dev/reflex/assets/65849393/cc0e760d-fbe6-4823-b4dc-4e6133ba34bd)

picklelo:
Yes please put a PR - we have integration tests that will catch if it breaks anything

picklelo:
Merged in #2797

PR #2797, opened 2024-03-06
Updating initialize_gitignore() to avoid modifying the encoding of the .gitignore file

cllatser:
With the parameter newline=' \n' we avoid that the encoding of the .gitignore file is modified and works correctly in Windows.

Adding to initialize_gitignore() the parameter newline='\n' to avoid modifying the encoding of the .gitignore file and make it work correctly in Windows.

closes[#2788]


picklelo (approved):
thanks for fixing!

PR #2801, opened 2024-03-07
[REF-1368] Move common form functionality to rx.el.forms

masenf:
Allow plain HTML Form element to have magic on_submit event handler.

### Chakra and Radix forms inherit on_submit functionality from rx.el.form
Consolidate logic in the basic HTML form and use it in both Radix and Chakra form wrappers.

### Update `test_form_submit.py`

Run the integration test with all three types of form roots to ensure consistent behavior.

### DIFF
--- a/reflex/components/chakra/forms/form.pyi
+++ b/reflex/components/chakra/forms/form.pyi
@@ -7,32 +7,85 @@ from typing import Any, Dict, Literal, Optional, Union, overload
 from reflex.vars import Var, BaseVar, ComputedVar
 from reflex.event import EventChain, EventHandler, EventSpec
 from reflex.style import Style
-from hashlib import md5
-from typing import Any, Dict, Iterator
-from jinja2 import Environment
 from reflex.components.chakra import ChakraComponent
 from reflex.components.component import Component
-from reflex.components.tags import Tag
-from reflex.constants import Dirs, EventTriggers
-from reflex.event import EventChain
-from reflex.utils import imports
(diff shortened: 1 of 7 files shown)


reflex/components/chakra/forms/form.pyi
```diff
@@ -7,32 +7,85 @@ from typing import Any, Dict, Literal, Optional, Union, overload
...
-
-class Form(ChakraComponent):
+class Form(ChakraComponent, HTMLForm):
     @overload
     @classmethod
     def create(  # type: ignore
         cls,
         *children,
         as_: Optional[Union[Var[str], str]] = None,
+        accept: Optional[
```
picklelo, line 22: Do we want to keep all these other props also, do they work? We could delete them in the base el.Form, I think it may add more noise/clutter than benefit

picklelo (approved):
I think the auto-docs don't show the extra props, but I'm leaning in favor of removing some of the base html props to keep our api simple (can do later on)

reflex/components/chakra/forms/form.pyi
masenf: The radix form for sure supports all of the native HTML Form props... chakra, i'm less sure about. At any rate, i don't think these will show up in the API docs, but I'll check before merging

masenf: The base form props are not rendering in the API docs. going ahead with the merge.

cllatser:
Hi, the `on_submit` works perfectly in the `rx.el.form`.
Thank you so much. 
💯

PR #2808, opened 2024-03-07
Get `client_ip` from `asgi.scope`

masenf:
It seems like REMOTE_ADDR is always 127.0.0.1, which is not super useful when trying to figure out where the websocket connection is originating from.

Of course this isn't a silver bullet because most-likely the WS will be passed through a reverse proxy anyway... in that case, the client IP is likely in the headers under `x_forwarded_for`

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -1146,7 +1146,7 @@ class EventNamespace(AsyncNamespace):
         }
 
         # Get the client IP
-        client_ip = environ["REMOTE_ADDR"]
+        client_ip = environ["asgi.scope"]["client"][0]
 
         # Process the events.
         async for update in process(self.app, event, sid, headers, client_ip):


reflex/app.py
```diff
@@ -1146,7 +1146,7 @@ async def on_event(self, sid, data):
         }
 
         # Get the client IP
-        client_ip = environ["REMOTE_ADDR"]
+        client_ip = environ["asgi.scope"]["client"][0]
```
picklelo, line 1149: Do we want to add any `get` or `try/catch` here, or will these key lookups always work? We should avoid crashing outside the event loop is my concern

masenf: yeah i can try/catch it. i didn't because we didn't before, but this is multiple levels of derefs, so more could go wrong

picklelo (approved):

picklelo (approved):

PR #2809, opened 2024-03-07
add support for sqlalchemy AssociationProxys

benedikt-bartscher:
#2340

### DIFF
--- a/reflex/utils/types.py
+++ b/reflex/utils/types.py
@@ -196,7 +196,7 @@ def get_attribute_access_type(cls: GenericType, name: str) -> GenericType | None
             else:
                 return class_
         if isinstance(attr, AssociationProxyInstance):
-            return list[
+            return List[
                 get_attribute_access_type(
                     attr.target_class,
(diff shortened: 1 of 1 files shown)


reflex/utils/types.py
```diff
@@ -194,6 +195,13 @@ def get_attribute_access_type(cls: GenericType, name: str) -> GenericType | None
                 return List[class_]
             else:
                 return class_
+        if isinstance(attr, AssociationProxyInstance):
+            return list[
```
masenf, line 199: do we need cap `List` here or will this work fine on 3.8?

Not sure how to test it, but I assume you have the environment to confirm.

benedikt-bartscher: I can not test with 3.8 easily. But I will test if it works with `List` as well. Sorry, I always forget about 3.8

benedikt-bartscher: I changed it to cap `List` and it still works fine with recent python versions

masenf (approved):

PR #2810, opened 2024-03-07
Make @rx.memo work with state vars passed as props

masenf:
Seems like this was a regression from the StatefulComponent refactor, because trying to pass a state Var to a CustomComponent gave undefined, likely due to `_get_vars` not accounting for `self.props` in CustomComponents.

With this change, it works.

Integration test added to `test_var_operations.py`


picklelo (approved):
LGTM

PR #2815, opened 2024-03-08
[REF-2158] Enable state when `on_load` or event_triggers are set.

masenf:
Basically all events in a reflex app require the backend to be up, even the built-in server side events round trip to the backend and require "state" even if no user-defined state classes are declared.

test_app_state_determination: checking that state is enabled at the right time

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -453,14 +453,18 @@ class App(Base):
         # Generate the component if it is a callable.
         component = self._generate_component(component)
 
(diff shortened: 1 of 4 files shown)


picklelo (approved):
LGTM - on a more broad not, I feel we have a LOT of places where we descend through the component tree to determine / set things. In the future maybe we can have some kind of hook system where the traversal is only done once and all these operations can run together

PR #2818, opened 2024-03-09
improve event handler state references

benedikt-bartscher:
Fixes some issues with event handlers inherited from mixins across modules.  
Should also fix potential issues with event handlers and States with the same name in the same module.  
In general, this should be a much safer way to format event handlers.

### DIFF
--- a/reflex/event.py
+++ b/reflex/event.py
@@ -147,6 +147,10 @@ class EventHandler(EventActionsMixin):
     # The function to call in response to the event.
     fn: Any
 
+    # The full name of the state class this event handler is attached to.
+    # Emtpy string means this event handler is a server side event.
+    state_full_name: str = ""
+
     class Config:
         """The Pydantic config."""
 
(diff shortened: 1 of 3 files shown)


reflex/event.py
```diff
@@ -147,6 +147,10 @@ class EventHandler(EventActionsMixin):
     # The function to call in response to the event.
     fn: Any
 
+    # The full name of the state class this event handler is attached to.
+    # Emtpy string means this event handler is a server side event.
```
picklelo, line 151: nit: Empty

masenf (approved):

reflex/utils/format.py
```diff
@@ -470,18 +469,18 @@ def get_event_handler_parts(handler: EventHandler) -> tuple[str, str]:
     if len(parts) == 1:
         return ("", parts[-1])
 
-    # Get the state and the function name.
-    state_name, name = parts[-2:]
+    # Get the state full name
+    state_full_name = handler.state_full_name
```
picklelo, line 473: This is nicer, good call

picklelo (approved):

PR #2827, opened 2024-03-10
added name of Var to error message

### LINKED ISSUE #2826: include the name of the Var so it's easier to track down `No JSON serializer` error
It will be nice to also include the name of the Var causing the ```TypeError(f"No JSON serializer found for var {value} of type {type(value)}.")```  so it's easier for a user to track down and maybe convert the Var to a string

emmakodes:
-  Added the name of the Var to the error message for ```TypeError(f"No JSON serializer found for var {value} of type {type(value)}.")``` closes #2826 

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

    b. Describe your changes.


emmakodes:
@masenf

reflex/utils/format.py
```diff
@@ -598,7 +600,14 @@ def format_state(value: Any) -> Any:
     """
     # Handle dicts.
     if isinstance(value, dict):
-        return {k: format_state(v) for k, v in value.items()}
+        result = {}
+        for k, v in value.items():
+            if isinstance(v, (pathlib.PosixPath, functools.partial)):
```
picklelo, line 605: I think we can be more general here by either:
1. Using `reflex.utils.serailizers.has_serializer` to check
2. Just putting a try/except around the setting.

Actually I think the easiest fix would be to pass down the key to `format_state` and just update the error message thrown there.

picklelo (changes requested):

reflex/utils/format.py
emmakodes: Thanks, @picklelo, your suggestion makes sense.

emmakodes: I have now updated the code by passing the key to `format_state` and updating the error message

picklelo (approved):
awesome, much better messages now. thanks for adding!

PR #2829, opened 2024-03-10
Allow access to position from HighLevelSelect

hjpr:
Access to position prop is buried in Select.content. Should leave finer control such as spacing and offset in content, but have ability to specify behavior of content window at the top-level.


Lendemor (changes requested):
`black` complain on the formatting, once it's solved we can take this PR.
Running `pre-commit` or making sure that the version of `black` match the expected version should be enough.

Lendemor (approved):

PR #2837, opened 2024-03-12
ruff-format: unify Black with Ruff

Borda:
### All Submissions:

### Type of change

As Ruff lint is already used I would suggest using also `ruff-format` for better consistency between lint and formatted


Borda:
well while the Black was back from 2022, it seems the diff is quite large...

picklelo:
@Borda seems most of the large diff is adding a blank line after the final `Returns:` - is there a way to disable that? Then it may be much smaller

Borda:
> most of the large diff is adding a blank line after the final `Returns:` - is there a way to disable that? Then it may be much smaller

I found that these changes would be skipped case-by-case, but that would help with diff size at all :(
ref: https://docs.astral.sh/ruff/formatter/#format-suppression

Borda:
@picklelo and the formating evolves over time, and the actual version of Ruff is before 0.1, so a fine middle ground would be just update to Ruff 0.1 with enabled formatting, and if you agree do smaller steps by updating to 0.2 and 0.3 in separate/consecutive/incremental PRs

reflex/components/chakra/media/icon.py
```diff
@@ -37,9 +37,9 @@ def create(cls, *children, **props):
             raise AttributeError(
                 f"Passing children to Icon component is not allowed: remove positional arguments {children} to fix"
             )
-        if "tag" not in props.keys():
+        if "tag" not in props:
             raise AttributeError("Missing 'tag' keyword-argument for Icon")
-        if type(props["tag"]) != str or props["tag"].lower() not in ICON_LIST:
+        if isinstance(props["tag"], str) or props["tag"].lower() not in ICON_LIST:
```
Lendemor, line 42: This change fails on almost all CI.
I don't know if this is ruff doing or a manual change, but the equivalent of `type(props["tag"] != str)` should be `not isinstance(props["tag"], str)`

```suggestion
        if not isinstance(props["tag"], str) or props["tag"].lower() not in ICON_LIST:
```

Borda: You are right, my bad...

Borda:
> Is there a huge change if we go to 0.3 of ruff now?

I do not think in ruff lint but as you saw some patterns changed in Black/Ruff formating
if pre-commit as bot is enabled it has also a feature to update its version monthly via PR

Borda:
looking to the `pyright` and not sure how to read the issues, what to fix...

masenf:
> looking to the pyright and not sure how to read the issues, what to fix...

i would assume it's because the formatter moved important `# type: ignore` comments to the wrong place and thus they stopped applying to the line they were meant to apply to

Borda:
seems this alos would need to be updated:
https://github.com/reflex-dev/reflex/blob/036afa951a5700b04722a0bcd7d3143dde07a12a/scripts/pyi_generator.py#L19-L20
or is it fine to drop formatting for the pyi as they seem to be generated anyway?

picklelo:
Yeah less concerned about formatting the `pyi` files we can add that in later if needed.

Lendemor:
We could drop the formatting of `.pyi` files but we have to make sure that the output of the script is deterministic, to avoid false positive when detecting is pyi were updated.

pyproject.toml
```diff
@@ -96,16 +96,11 @@ build-backend = "poetry.core.masonry.api"
 [tool.pyright]
 
 [tool.ruff]
```
Lendemor, line 98: Actually, as an intermediary step, I'd suggest to use
 ```suggestion
[tool.ruff]
exclude = ["*.pyi"]
```
This way the pyi files can still be generated as it is now, (using black in script mode)
but we can use ruff for everything else, reducing the amount of changes we have to go through.

Once we are up to date with latest ruff we can consider dropping the use of black in `.pyi` files

Comment (Borda) [reply pyproject.toml:100]: updated and resolved conflicts :)

Comment (Lendemor) [reply pyproject.toml:100]: Thanks, seems like the change to the pyi (from earlier commits?) are still present, if we remove them I think we'll be good.

**Comment by Lendemor** [comment]
Can we sync this PR on main again since there is quite a few conflicts?

**Comment by Lendemor** [comment]
> > Can we sync this PR on main again since there is quite a few conflicts?
> 
> @Lendemor, merged with accepting all from `main` and re-run precommit... 🦩

I think you still need to remove some changes to pyi files, that's why the pre-commit check is failing.

File: reflex/components/radix/themes/components/icon_button.py [hunk 3]
```diff
@@ -68,7 +68,7 @@ def create(cls, *children, **props) -> Component:
                 "IconButton requires a child icon. Pass a string as the first child or a rx.icon."
             )
         if "size" in props:
-            if type(props["size"]) == str:
+            if isinstance(props["size"]):
```
Lendemor, line 71:
```suggestion
            if isinstance(props["size"], str):
```

reflex/components/radix/themes/components/radio_group.py
```diff
@@ -146,7 +146,7 @@ def create(
         # convert only non-strings to json(JSON.stringify) so quotes are not rendered
         # for string literal types.
         if (
-            type(default_value) is str
+            type(default_value) is str  # noqa: E721
```
Lendemor, line 149: why not use `isinstance` here like in other files?
if there is a specific reason we should add a comment why.

Lendemor (changes requested):

reflex/components/radix/themes/components/radio_group.py
Borda: i had some problem recently with bool/int and was not sure if sudden failing test was related...

pyproject.toml
Borda: revered and fixing some mnors

Lendemor:
> @picklelo @Lendemor mind check it now, pls 🦩

I've reran the CI, it seems like some of the `# type: ignore` pragma have been moved by the formatting, so they are not on the line they should be.

reflex/components/radix/themes/components/radio_group.py
```diff
@@ -145,11 +145,7 @@ def create(
 
         # convert only non-strings to json(JSON.stringify) so quotes are not rendered
         # for string literal types.
-        if (
-            type(default_value) is str
-            or isinstance(default_value, Var)
-            and default_value._var_type is str
-        ):
+        if isinstance(default_value, (str, Var)) and default_value._var_type is str:
```
Lendemor, line 148: I think this line change is not equivalent to the previous line

This seems more correct :
```suggestion
        if isinstance(default_value, str) or (isinstance(default_value, Var) and default_value._var_type is str):
```

Lendemor (changes requested):

reflex/components/datadisplay/code.pyi
Lendemor, line 1: Change for this file need to be removed too.

reflex/components/datadisplay/code.pyi
```diff
@@ -1082,7 +1082,7 @@ class CodeBlock(Component):
         on_unmount: Optional[
             Union[EventHandler, EventSpec, list, function, BaseVar]
         ] = None,
-        **props
+        **props,
```
Lendemor, line 1085:
```suggestion
        **props
```

tests/compiler/test_compiler.py
```diff
@@ -208,7 +208,9 @@ def test_create_document_root():
         utils.NextScript.create(src="foo.js"),
         utils.NextScript.create(src="bar.js"),
     ]
-    root = utils.create_document_root(head_components=comps, html_lang="rx", html_custom_attrs={"project": "reflex"})  # type: ignore
+    root = utils.create_document_root(
+        head_components=comps, html_lang="rx", html_custom_attrs={"project": "reflex"}
+    )  # type: ignore
```
Lendemor, line 213:
```suggestion
        head_components=comps,  # type: ignore
        html_lang="rx",
        html_custom_attrs={"project": "reflex"},
    )
```

tests/components/core/test_colors.py
```diff
@@ -31,15 +31,19 @@ def create_color_var(color):
             "var(--${state__color_state.color}-${state__color_state.shade})",
         ),
         (
-            create_color_var(rx.color(f"{ColorState.color_part}ato", f"{ColorState.shade}")),  # type: ignore
+            create_color_var(
+                rx.color(f"{ColorState.color_part}ato", f"{ColorState.shade}")
+            ),  # type: ignore
```
Lendemor, line 36:
```suggestion
                rx.color(f"{ColorState.color_part}ato", f"{ColorState.shade}") # type: ignore
            ),
```

tests/components/core/test_colors.py
```diff
@@ -31,15 +31,19 @@ def create_color_var(color):
...
         ),
         (
             create_color_var(f'{rx.color(ColorState.color, f"{ColorState.shade}")}'),  # type: ignore
             "var(--${state__color_state.color}-${state__color_state.shade})",
         ),
         (
-            create_color_var(f'{rx.color(f"{ColorState.color}", f"{ColorState.shade}")}'),  # type: ignore
+            create_color_var(
+                f'{rx.color(f"{ColorState.color}", f"{ColorState.shade}")}'
+            ),  # type: ignore
```
Lendemor, line 46:
```suggestion
                f'{rx.color(f"{ColorState.color}", f"{ColorState.shade}")}'  # type: ignore
            ),
```

integration/test_dynamic_routes.py
```diff
@@ -55,11 +55,15 @@ def index():
...
+                "next",
+                href="/page/" + DynamicState.next_page,
+                id="link_page_next",  # type: ignore
             ),
             rx.link("missing", href="/missing", id="link_missing"),
             rx.chakra.list(
-                rx.foreach(DynamicState.order, lambda i: rx.chakra.list_item(rx.text(i))),  # type: ignore
+                rx.foreach(
+                    DynamicState.order, lambda i: rx.chakra.list_item(rx.text(i))
+                ),  # type: ignore
```
Lendemor, line 66:
```suggestion
                    DynamicState.order,  # type: ignore
                    lambda i: rx.chakra.list_item(rx.text(i)),
                ),
```

reflex/components/radix/themes/components/radio_group.py
```diff
@@ -145,11 +145,7 @@ def create(
 
         # convert only non-strings to json(JSON.stringify) so quotes are not rendered
         # for string literal types.
-        if (
-            type(default_value) is str
-            or isinstance(default_value, Var)
-            and default_value._var_type is str
-        ):
+        if isinstance(default_value, str) or (isinstance(default_value, Var) and default_value._var_type is str):
```
Lendemor, line 148:
```suggestion
        if isinstance(default_value, str) or (
            isinstance(default_value, Var) and default_value._var_type is str
        ):
```

reflex/utils/imports.py
```diff
@@ -66,7 +66,9 @@ def name(self) -> str:
             The name(tag name with alias) of tag.
         """
         if self.alias:
-            return self.alias if self.is_default else " as ".join([self.tag, self.alias])  # type: ignore
+            return (
+                self.alias if self.is_default else " as ".join([self.tag, self.alias])
+            )  # type: ignore
```
Lendemor, line 71:
```suggestion
                self.alias if self.is_default else " as ".join([self.tag, self.alias])  # type: ignore
            )
```

Lendemor (changes requested):

tests/components/core/test_colors.py
```diff
@@ -31,15 +31,19 @@ def create_color_var(color):
             "var(--${state__color_state.color}-${state__color_state.shade})",
         ),
         (
-            create_color_var(rx.color(f"{ColorState.color_part}ato", f"{ColorState.shade}")),  # type: ignore
+            create_color_var(
+                rx.color(f"{ColorState.color_part}ato", f"{ColorState.shade}") # type: ignore
```
Lendemor, line 35: missed a space in previous suggestion, should be good with this
```suggestion
                rx.color(f"{ColorState.color_part}ato", f"{ColorState.shade}")  # type: ignore
```

reflex/app.py
```diff
@@ -350,11 +350,17 @@ async def postprocess(
         for middleware in self.middleware:
             if asyncio.iscoroutinefunction(middleware.postprocess):
                 out = await middleware.postprocess(
-                    app=self, state=state, event=event, update=update  # type: ignore
+                    app=self,
+                    state=state,
+                    event=event,
+                    update=update,  # type: ignore
```
Lendemor, line 356:
```suggestion
                    app=self,  # type: ignore
                    state=state,
                    event=event,
                    update=update,
```

reflex/app.py
```diff
@@ -350,11 +350,17 @@ async def postprocess(
...
                 )
             else:
                 out = middleware.postprocess(
-                    app=self, state=state, event=event, update=update  # type: ignore
+                    app=self,
+                    state=state,
+                    event=event,
+                    update=update,  # type: ignore
```
Lendemor, line 363:
```suggestion
                    app=self,  # type: ignore
                    state=state,
                    event=event,
                    update=update,
```

benchmarks/test_benchmark_compile_pages.py
```diff
@@ -225,7 +225,12 @@ def app_with_ten_pages(
         an AppHarness instance
     """
     root = tmp_path_factory.mktemp(f"app10")
-    yield AppHarness.create(root=root, app_source=functools.partial(AppWithTenPages, render_comp=render_multiple_pages))  # type: ignore
+    yield AppHarness.create(
+        root=root,
+        app_source=functools.partial(
+            AppWithTenPages, render_comp=render_multiple_pages
+        ),
+    )  # type: ignore
```
Lendemor, line 233:
```suggestion
            AppWithTenPages, render_comp=render_multiple_pages  # type: ignore
        ),
    )
```

benchmarks/test_benchmark_compile_pages.py
```diff
@@ -225,7 +225,12 @@ def app_with_ten_pages(
         an AppHarness instance
     """
     root = tmp_path_factory.mktemp(f"app10")
-    yield AppHarness.create(root=root, app_source=functools.partial(AppWithTenPages, render_comp=render_multiple_pages))  # type: ignore
+    yield AppHarness.create(
+        root=root,
+        app_source=functools.partial(
+            AppWithTenPages, render_comp=render_multiple_pages  # type: ignore
```
Lendemor, line 231:
```suggestion
            AppWithTenPages, 
            render_comp=render_multiple_pages,  # type: ignore
```

benchmarks/test_benchmark_compile_pages.py
```diff
@@ -225,7 +225,13 @@ def app_with_ten_pages(
         an AppHarness instance
     """
     root = tmp_path_factory.mktemp(f"app10")
-    yield AppHarness.create(root=root, app_source=functools.partial(AppWithTenPages, render_comp=render_multiple_pages))  # type: ignore
+    yield AppHarness.create(
+        root=root,
+        app_source=functools.partial(
+            AppWithTenPages,
```
Lendemor, line 231:
```suggestion
            AppWithTenPages,
```

Borda:
> PR should be ready for merge, but it seems like github changed stuff about the available macos worker, we'll need to fix it first 👍

Thank you for your help and patience! 🙏

Lendemor:
We've merged a fix for macos CI, one last sync with main and we should be good to go. 🎉

Borda:
@Lendemor merged with `main` :flamingo:

Lendemor (approved):

PR #2838, opened 2024-03-12
[REF-2122] Opt-in multiprocess compile

masenf:
Based on #2810

Pass the environment variable `REFLEX_COMPILE_PROCESSES=0` to enable multiprocessing compilation engine. Pass a number other than zero to control the number of processes to spawn.

It's also possible to set `REFLEX_COMPILE_THREADS=X` where X is the number of threads to use, and zero (the default) means automatic.

The default behavior here is to continue using the ThreadPoolExecutor as the default engine. It's not as fast as multiprocessing, but since 4 previous revisions of this code have been reverted, I'm sticking with a safe default until we can get more real world testing... but no one is going to test it in a side branch, so we need to get it into main and enable it during out development processes to try to tease out any issues.

Exposing the controls this way also allows us to more easily tune and tweak the implementation without having to edit the code directly.


picklelo (dismissed):
good idea keeping behind a flag - super snappy for me with it enabled

picklelo (approved):
Works well for me - in the future we may want to break up the `compile` method as it's getting quite huge

masenf:
Personally I'd like to see if more modularized and pluggable, at least the parts we parallelize, because obviously the order does not matter for those.

PR #2839, opened 2024-03-12
Make AppHarness integration tests work with py38

masenf:
Python 3.8 is a known gap in our end-to-end test, as evidenced by the 3 bugs in the framework found by making the integration tests run on py38.

Now that things are working on py38, run the tests in CI to prevent regressions.


masenf:
included in #2841

PR #2844, opened 2024-03-12
better default title for seo (also remove default description)

Lendemor:
Adding a new page without passing a title will now use the method `make_default_page_title` by passing `app_name` and `route` .

Example: 
```python
$ format.make_default_page_title("my_beautiful_app", "third_page")
> 'My Beautiful App : Third Page'
```

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -408,7 +408,7 @@ class App(Base):
         self,
         component: Component | ComponentCallable,
         route: str | None = None,
-        title: str = constants.DefaultPage.TITLE,
+        title: str | None = None,
         description: str = constants.DefaultPage.DESCRIPTION,
         image: str = constants.DefaultPage.IMAGE,
         on_load: (
@@ -467,6 +467,10 @@ class App(Base):
                     break
 
         component = OverlayFragment.create(component)
+
+        if title is None:
+            # title = get_config().app_name + " : " + route.title()
+            title = format.make_default_page_title(get_config().app_name, route)
 
         # Add meta information to the component.
         compiler_utils.add_meta(
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -409,7 +409,7 @@ class App(Base):
         component: Component | ComponentCallable,
         route: str | None = None,
         title: str | None = None,
-        description: str = constants.DefaultPage.DESCRIPTION,
+        description: str | None = None,
         image: str = constants.DefaultPage.IMAGE,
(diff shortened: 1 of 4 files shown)


reflex/constants/route.py
```diff
@@ -50,9 +50,9 @@ class DefaultPage(SimpleNamespace):
     """Default page constants."""
 
     # The default title to show for Reflex apps.
-    TITLE = "Reflex App"
+    TITLE = "{} : {}"
```
picklelo, line 53: I think using `|` or `-` instead of `:` is more standard - but I like this approach!

reflex/utils/format.py
```diff
@@ -187,6 +187,21 @@ def to_kebab_case(text: str) -> str:
...
+
+    Args:
+        app_name: The name of the app owning the page.
+        route: The route to make the title from.
+
+    Returns:
+        The default page title.
+    """
+    title = constants.DefaultPage.TITLE.format(app_name, route)
+    title = title.replace("_", " ").replace("-", " ")
```
picklelo, line 201: Can you use `format.to_title_case` here?

reflex/constants/route.py
Alek99: I'm not sure this is the case I was looking at this SEO article and most use":"https://blog.hubspot.com/blog/tabid/6307/bid/5984/spilling-seo-juice-3-dos-and-don-ts-for-writing-great-page-titles.aspx

picklelo: If we use a colon let's remove the space before, but I personally prefer the `|`

Alek99: Ok just do the |

reflex/utils/format.py
Lendemor: `format.to_title_case` remove the `_` and `-`, but here I want to replace them with spaces.

Alek99:
Testing on the dev reflex web see if this fixes the description problem

reflex/app.py
```diff
@@ -408,7 +408,7 @@ def add_page(
         self,
         component: Component | ComponentCallable,
         route: str | None = None,
-        title: str = constants.DefaultPage.TITLE,
+        title: str | None = None,
```
picklelo, line 411: We should set `description` to None also (instead of empty string)

Alek99 (approved):

Alek99 (approved):

PR #2860, opened 2024-03-14
convert text inside list_item to span and set icon display to inline

Lendemor:
```python

rx.list_item(
    rx.text("Web app with Reflex: "),
    rx.icon("key"),
    rx.link("🔗 Reflex Site", href="https://reflex.dev/docs/"),
)
```
The above  code will now behave the same as 
```python
rx.list_item(
    rx.text.span("Web app with Reflex:  "),
    rx.icon("key", display="inline),
    rx.link("🔗 Reflex Site", href="https://reflex.dev/docs/")
)
```

### DIFF
--- a/reflex/components/radix/themes/layout/list.py
+++ b/reflex/components/radix/themes/layout/list.py
@@ -41,12 +41,15 @@ LiteralListStyleTypeOrdered = Literal[
 class BaseList(Flex, LayoutComponent):
     """Base class for ordered and unordered lists."""
 
+    list_style_type: Var[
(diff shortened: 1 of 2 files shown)


masenf (approved):
I like that this only applies to direct children... so if somebody _did_ want the broken behavior, for whatever reason, they could wrap the contents in an `rx.fragment` to bypass this prop forcing

masenf (approved):
CI issues are known problems that are fixed in `main`; will merge

PR #2862, opened 2024-03-14
Read and Write Cookie with httpOnly=true.

Fabricio-T:
### All Submissions:

- [ ] Have you followed the guidelines stated in [CONTRIBUTING.md](https://github.com/reflex-dev/reflex/blob/main/CONTRIBUTING.md) file? **A little bit, but not all.**
- [X] Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/reflex-dev/reflex/pulls ) for the desired changed?

### Type of change

### New Feature Submission:

### Changes To Core Features:

### Description
Now its possible to set and read [httpOnly Cookies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#httponly).

# What I did?
- Added 'cookie' to HeaderData(state.py).
- Added new default_endpoint(app.py) as _/_set_cookie/_
- Added new _set_http_only_cookie_ event(event.py) to easy create a httpOnly cookie.

# How to use it?
1. Create a State
2. Use 'self.router.headers.cookie'
3. Create a function and 'return rx.set_http_only_cookie(...)'

Like that:
```
class MyCustomState(rx.State):
    def get_my_cookies(self):
        myCookies = self.router.headers.cookie
        logger.info(myCookies)

    def set_cookie(self):
        return rx.set_http_only_cookie(
            key="Test",
            value="5Test5",
            max_age=1500,
        )

...

rx.button(
            "Get All Cookie",
            width="100%",
            size="3",
            radius="large",
            variant="solid",
            on_click=MyCustomState.get_my_cookies,
        ),

rx.button(
            "Set A Cookie",
            width="100%",
            size="3",
            radius="large",
            variant="solid",
            on_click=MyCustomState.set_cookie,
        ),
```

# What can be improved?
1. self.router.headers.cookie should be a dictionary... Possible but need more work...
2. self.router.headers.cookie doesnt returns any cookie that has a 'path' different to '/'. It only works with a path='/'. I dont know why...
3. Maybe can it be built-in with rx.Cookie in state.py? I couldnt find a good way to integrate with it...


reflex/event.py
```diff
@@ -652,6 +655,58 @@ def _callback_arg_spec(eval_result):
     return [eval_result]
 
 
+def set_http_only_cookie(
```
masenf, line 658: i think it would be ideal to handle these kinds of cookies in the same way we handle `rx.Cookie` currently, instead of having a kind of separate parallel API.

In `state.py`, we can add the `http_only` field to the existing `Cookie` class.

Then in `state.js` if the `http_only` flag is set, instead of setting the cookie via JS DOM, we set it using the `fetch` mechanism. Right about here: https://github.com/reflex-dev/reflex/blob/fb2c3606d951aab77e4f8b61a9d97ddc43a6a87d/reflex/.templates/web/utils/state.js#L528

The last piece of the puzzle is applying any cookies that get set when the websocket connects to the state so they can be easily accessed... i'll have to think about that one a little more to give a proper recommendation. so accessing them through the new `cookie` field in `HeaderData` is probably fine for now.

reflex/app.py
```diff
@@ -273,6 +274,7 @@ def add_default_endpoints(self):
         """Add default api endpoints (ping)."""
         # To test the server.
         self.api.get(str(constants.Endpoint.PING))(ping)
+        self.api.post("/_set_cookie/")(set_http_only_cookie)
```
watfordsuzy, line 277: I don't believe it is safe/secure to have an API which allows a frontend caller unrestricted access to setup HTTP Only cookies, or at least goes against the security rationale which envisioned HTTP Only cookies in the first place. (I could be wrong)

Lendemor:
Marking this as draft until resynced with main and conflict are addressed.

adhami3310:

PR #2863, opened 2024-03-14
Account for imports of @rx.memo components for frontend package installation 

masenf:
More long standing bugs in the CustomComponent (@rx.memo) that were not exposed until other change was made to allow components as props to CustomComponent and we started using that functionality combined with windows only compiling a subset of reflex-web.

See the newly added test case for how simple this is to repro.

This has been an issue for a long time, but I suppose due to various factors around how and where `rx.memo` is used, we just never really noticed that it wasn't installing package deps for components wrapped in `rx.memo` because most likely _somewhere else_ in reflex-web (or large project using `rx.memo`), the package was referenced and installed via another route.

The following app exposes the issue with 0.4.4

```python
import reflex as rx


@rx.memo
def wrapper(child: rx.Component) -> rx.Component:
    return rx.el.div(child, style={"border": "2px solid red"})

@rx.memo
def inner(child: str) -> rx.Component:
    return rx.drawer.root(child)

def index() -> rx.Component:
    return rx.vstack(
        inner(child="foo"),
        wrapper(
            child=inner(child="foo"),
        ),
    )


app = rx.App()
app.add_page(index)
app.add_page(index, route="/index2")
```

This has problems in either dev or prod mode.

```console
./utils/components.js:6:0
Module not found: Can't resolve 'vaul'
  4 | import { memo } from "react"
  5 | import { E, isTrue } from "/utils/state"
> 6 | import { Drawer as VaulDrawer } from "vaul"
  7 |
  8 |
  9 |

https://nextjs.org/docs/messages/module-not-found

Import trace for requested module:
./pages/index.js
```

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -818,6 +818,7 @@ class App(Base):
             compile_results.append((stateful_components_path, stateful_components_code))
 
             result_futures = []
+            custom_components_future = None
 
             def submit_work(fn, *args, **kwargs):
                 """Submit work to the thread pool and add a callback to mark the task as complete.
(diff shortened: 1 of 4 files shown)


masenf:
Yikes this is a major compile time perf regression. Going from 12s on my mac to over 2 minutes 😱 

More tweaks needed 😔

masenf:
this will conflict with #2838,  but i have already resolved the conflicts in `masenf/multiprocess-compile-fx-conflict`

picklelo (approved):

PR #2867, opened 2024-03-16
added feature to automatically change port

### LINKED ISSUE #2857: [REF-2220] reflex run should automatically select new port if default is taken
Currently if you have one app running, and try to `reflex run` on another app, it asks you to change the port. Instead, we should automatically find the next free frontend and backend ports (3001, 8001) and run it on that.

<sub>[REF-2220](https://linear.app/reflex-dev/issue/REF-2220/reflex-run-should-automatically-select-new-port-if-default-is-taken)</sub>

wassafshahzad:
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

### Description
Added feature to automatically change port if currently one is used.

### Linked Issue
Closes #2857

### DIFF
--- a/reflex/utils/processes.py
+++ b/reflex/utils/processes.py
@@ -100,7 +100,7 @@ def change_port(port: str, _type: str) -> str:
     """
     new_port = str(int(port) + 1)
     if is_process_on_port(new_port):
-        return change_port(new_port)
+        return change_port(new_port, _type)
     console.info(
         f"The {_type} will run on port [bold underline]{new_port}[/bold underline]."
     )
--- a/reflex/utils/processes.py
+++ b/reflex/utils/processes.py
@@ -92,6 +92,7 @@ def change_port(port: str, _type: str) -> str:
 
     Args:
         port: The port.
+        _type: The type of the port.
 
     Returns:
         The new port.
--- a/reflex/utils/processes.py
+++ b/reflex/utils/processes.py
@@ -87,7 +87,7 @@ def kill_process_on_port(port):
             get_process_on_port(port).kill()  # type: ignore
 
 
-def change_port(port: str) -> str:
+def change_port(port: str, _type: str) -> str:
     """Change the port.
 
     Args:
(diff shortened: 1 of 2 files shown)


srikary12:
I've implemented something similar, but added just auto select functionality after 5 sec instead of by default. You can find the changes here https://github.com/srikary12/reflex-dev

wassafshahzad:
> I've implemented something similar, but added just auto select functionality after 5 sec instead of by default. You can find the changes here https://github.com/srikary12/reflex-dev

Are you shadowing the repo ? if not you should first have the issue  assigned to your self before starting and then create a pull request. Other then that I have taken a look at your changes. You are using sockets to check if the port is connected or not, which is not needed since we already have a function called `is_process_on_port` which informs us if the port is already occupied.

I do like you idea of informing the user before auto selecting the ports but I believe it would be better to have a property in the config something like "auto_select_port: Bool". Since asking the user `change_or_terminate_port` is a bit redundant in most case as the user can just manually enter a port.
Still good effort.

reflex/utils/processes.py
```diff
@@ -87,41 +87,20 @@ def kill_process_on_port(port):
...
-                f"The {_type} will run on port [bold underline]{new_port}[/bold underline]."
-            )
-            return new_port
-    else:
-        console.log("Exiting...")
-        raise typer.Exit()
+    new_port = str(int(port) + 1)
+    if is_process_on_port(new_port):
+        return change_port(new_port)
+    return new_port
```
masenf, line 103: i think we should print an info message here, similar to the one that was removed

```python
            console.info(
                f"The {_type} will run on port [bold underline]{new_port}[/bold underline]."
            )
```

masenf (changes requested):
lgtm

reflex/utils/processes.py
wassafshahzad: @masenf Done

picklelo (approved):
Works great for me, this is a sweet improvement!

PR #2871, opened 2024-03-17
[REF-2229]Dedupe deprecation warnings

ElijahAhianyo:
When you use a deprecated warning n times, we print the warning message `n` times. Ideally we would only want to print the message once. This PR does that.

### DIFF
--- a/reflex/utils/__init__.py
+++ b/reflex/utils/__init__.py
@@ -1,1 +1,2 @@
 """Reflex utilities."""
+DEPRECATED_FEATURES = set()
--- a/reflex/utils/__init__.py
+++ b/reflex/utils/__init__.py
@@ -1,2 +1,1 @@
 """Reflex utilities."""
-DEPRECATED_FEATURES = set()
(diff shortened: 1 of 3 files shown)


reflex/utils/console.py
```diff
@@ -119,15 +120,21 @@ def deprecate(
         feature_name: The feature to deprecate.
         reason: The reason for deprecation.
         deprecation_version: The version the feature was deprecated
-        removal_version: The version the deprecated feature will be removed.
+        removal_version: The version the deprecated feature will be removed
+        dedupe: Whether to allow multiple console logs of deprecation message.
```
masenf, line 124:
```suggestion
        dedupe: If True, suppress multiple console logs of deprecation message.
```

reflex/utils/__init__.py
```diff
@@ -1 +1,2 @@
 """Reflex utilities."""
+DEPRECATED_FEATURES = set()
```
masenf, line 2: Is there a reason to not just define this inside `console.py` itself? I dont think it will be used elsewhere.

I also suggest a different name, something like `_EMITTED_DEPRECATION_WARNINGS`

masenf (commented):
Just a var name and docstring rewording suggestion for clarity

reflex/utils/__init__.py
ElijahAhianyo: fixed

masenf (approved):

PR #2874, opened 2024-03-18
Revert "Remove width prop from rx.select (#2835)"

masenf:
This reverts commit d14292dc9b2c4fe7dfe5e1cebce42cb5a3a317a0.

This change prevented setting the width on high level select at all, we need to find a different approach to solve the overflowing container issue.


masenf:
Brought in with #2876

PR #2877, opened 2024-03-18
Escape " in PR titles in benchmark CI

martinxu9:


.github/workflows/benchmarks.yml
```diff
@@ -119,8 +119,10 @@ jobs:
       - name: Upload benchmark results
         # Only run if the database creds are available in this context.
         if: ${{ env.DATABASE_URL }}
-        run: poetry run python scripts/simple_app_benchmark_upload.py --os "${{ matrix.os }}"
-          --python-version "${{ matrix.python-version }}" --commit-sha "${{ github.sha }}" 
-          --benchmark-json "${{ env.OUTPUT_FILE }}" --pr-title "${{ github.event.pull_request.title }}"
+        run: |
+          escaped_pr_title=$(echo ${{ github.event.pull_request.title }} | sed 's/"/\\"/g')
```
masenf, line 123: i think we should use single quotes to avoid shell expansion in PR titles... like if i called my pr `Cool new trick $(rm -rf /)`... that might not be great for the runner

masenf (changes requested):

.github/workflows/benchmarks.yml
martinxu9: Not sure what you had in mind with no variable expansion if we were to keep processing the env and passing it around, don't think I figured out a way. But of course we don't have to deal with that, I've uploaded a version that runs

masenf:
We had an issue in the `main` branch with this PR. It's causing a failure because it requires a PR title, but the push event is not a PR.

PR #2880, opened 2024-03-19
[REF-1682][REF-1683][REF-1684][REF-2283]Benchmark reflex package size and .web folder

ElijahAhianyo:
This Pr we benchmark the following:

- **Reflex build size**: We achieve this by running `poetry build` on the PR and measuring the size of the generated `dist` folder
- **Reflex install size**: This measures the total size of installing reflex(`pip install reflex`) with all of the dependencies it installs. we achieve this by installing reflex in `venv` and then calculating the size of the `site-packages` directory.
- **.web folder size of a base app**: We measure the size of the `.web` folder after running the [counter app](https://github.com/reflex-dev/reflex-examples/tree/main/counter).
- **.web folder size of reflex web**: We also measure the size of the `.web` folder after running [reflex-web](https://github.com/reflex-dev/reflex-web).

We then publish the data obtained to a postgres db. 
**Note that the sizes calculated are published/stored as megabytes.**


scripts/benchmarks/benchmark_reflex_size.py
```diff
@@ -0,0 +1,204 @@
+"""Checks the size of a specific directory and uploads result."""
+import argparse
+import os
+import subprocess
+from datetime import datetime
+
+import psycopg2
+
+
+def get_directory_size(directory):
```
ElijahAhianyo, line 10: I think this is accurate, just tested this against a few dirs locally. The discrepancies in the db come from the CI, perhaps some packages are added by github. Will need to check that out

ElijahAhianyo: fixed

martinxu9 (approved):

PR #2881, opened 2024-03-19
When __REFLEX_SKIP_COMPILE == "yes" allow telemetry to fail

masenf:
Allow running `--backend-only` without a .web directory

Creates a new `reflex.utils.exec.should_skip_compile()` check, since this is now checked in multiple places.


Lendemor (dismissed):
Looks good.

Lendemor (approved):

PR #2883, opened 2024-03-19
[REF-2168] Add share options to custom component commands

martinxu9:
## Summary
- Use tomlkit to process pyproject.toml, when publishing, ask user if want to interactively update the info (such as source URL, author email)
- Add a `share` flag for the `publish` command. By default after a successful publish, the command ask the user if they want to share information such as demo app link. This information goes to Reflex backend services.
- Add a `share` command so the same share logic can happen independent of the publish.
## Tests
- [x] Tested locally together with https://github.com/reflex-dev/reflex-hosting/pull/240 
## How to test
- Use the reflex in this PR. Use an existing custom component project or follow the docs to create a new one. The actual implementation does not matter. 
- Point to control plane dev by setting
```
export CP_WEB_URL=https://control-plane.dev.reflexcorp.run
export CP_BACKEND_URL=https://rxcp-dev-control-plane.fly.dev
```
- Run `reflex component share` (it also works when not at the top level of the project, user will be prompted to enter name)
- Answer all the questions (feel free to leave out the optional ones)
- Finally expect to see a new entry on the DB (if you have supabase dev access, look at the `customcomponent` table)
- Expect to see a new entry when querying the public endpoint. One way to easily do that is through swagger https://rxcp-dev-control-plane.fly.dev/docs#/default/get_custom_components_gallery_data_custom_components_gallery_get


reflex/custom_components/custom_components.py
```diff
@@ -657,3 +673,211 @@ def publish(
...
+    if (
+        console.ask(
+            "Would you like to include your published component on our gallery?",
+            choices=["n", "y"],
+            default="y",
+        )
+        == "n"
+    ):
+        console.print(
+            "If you decide to do this later, you can run `reflex custom-components share` command. Thank you!"
```
masenf, line 691:
```suggestion
            "If you decide to do this later, you can run `reflex component share` command. Thank you!"
```

pyproject.toml
```diff
@@ -62,6 +62,7 @@ build = "^1.0.3"
 setuptools = "^69.1.1"
 httpx = ">=0.25.1"
 twine = ">=4.0.0"
+tomlkit = "^0.12.4"
```
picklelo, line 65: Is this a new dependency or already included by one of the packages? Is there a reason for tomlkit rather than the base toml library?

picklelo (commented):
Tested the workflow with my image-zoom app and it seemed to work well. I see it entered in the dev database.

Will wait for masen for final approval

pyproject.toml
martinxu9: tomllib was introduced in 3.11 and it intentionally only includes read/parse. There's no write.

martinxu9: tomllib was introduced in 3.11 and it intentionally only includes read/parse. There's no write.

masenf (approved):
i like the toml parsing much better. nice

i wasn't able to actually test it with my new library yet, due to deploy issues. but the code looks good.

martinxu9:
> i like the toml parsing much better. nice
> 
> 
> 
> i wasn't able to actually test it with my new library yet, due to deploy issues. but the code looks good.

I'll try to invite the whole team to share and seed our db for the gallery

masenf (approved):

Lendemor (approved):

PR #2891, opened 2024-03-20
Update LUCIDE_ICON_LIST with newest Icon names

### LINKED ISSUE #2882: Fix some of the lucide icons allowed names
Some of the tags allowed for use in lucide `Icon` are missing/wrong.

We need to make sure the values in  `LUCIDE_ICON_LIST` are correct.

luccavp12:
Attempt to fix #2882 

Compared to this newest version of Lucide, there have been some icon name changes. For example, a previous version using `rx.icon("activity_square")` will no longer work and need to be changed to `rx.icon("square_activity")`.


masenf (commented):
Thanks for your contribution!

The main issue I see with this (and apparent from the CI failures) is that these name changes will break existing code that already works...

If possible I think we should maintain a mapping of the old names to the new names so that we can warn users about the removed names and hopefully automatically map them to the new names.

Maybe something like

```python
RENAMED_ICONS_05 = {
    "activity_square": "square_activity",
    "alert_circle": "circle_alert",
    "alert_octagon": "octagon_alert",
    ...
}

def map_deprecated_icon_names_05(tag: str) -> str:
    new_tag = RENAMED_ICONS_05.get(tag)
    if new_tag is not None:
        console.deprecate(
            feature_name=f"icon {tag}",
            reason=f"it was renamed upstream. Use {new_tag} instead.",
            deprecation_version="0.4.6",
            removal_version="0.5.0",
        )
        return new_tag
    return tag
```

Then in the `create` method, we can call this function to check if the name was changed.

@Lendemor curious what you think about this approach; lucide seems kind of brazen about breaking things at the moment.

luccavp12:
That makes a ton of sense, thank you @masenf! 

Could I have ran the reflex-web CI tests locally before pushing?

Also, what do you think we should do with regards to the names that have been completely dropped by Lucide? I guess we just have to find the closest looking replacement.

masenf:
You can check out reflex-dev/reflex-web and install the requirements.txt in a new venv and then install your local version of reflex as well. Then `reflex init` and `reflex run --env prod` and you would simulate the CI run.

masenf:
As for the ones that are totally removed... if there's a close enough one, that's ideal, otherwise we can just fail.

Lendemor:
I quite like with the idea of mapping names, anything that prevent existing apps from breaking is good, and failing is a "last resort".

Also another thing, but even if we fix the lucide version to keep a "stable" version, most people will go pickup the tags from lucide home page, so there will be discrepancy between what we accept and what's offered on `lucide.dev`

masenf:
> most people will go pickup the tags from lucide home page, so there will be discrepancy between what we accept and what's offered on lucide.dev

I actually hit this myself quite often. I think the lucide library is still getting stable.

luccavp12:
Just added in the suggested mapping function.

Lendemor (approved):

PR #2895, opened 2024-03-21
[REF-2306] Include twine in dependencies on pyproject.toml

martinxu9:


Lendemor (approved):

reflex/custom_components/custom_components.py
```diff
@@ -633,12 +633,6 @@ def publish(
     # Validate the distribution directory.
     _ensure_dist_dir(version_to_publish=version_to_publish, build=build)
 
-    # We install twine on the fly if required so it is not a stable dependency of reflex.
-    try:
-        import twine  # noqa: F401  # type: ignore
-    except (ImportError, ModuleNotFoundError) as ex:
-        if not _pip_install_on_demand("twine"):
```
masenf, line 640: why not `"twine>=4.0.0"` here?

Lendemor: If a twine below 4.0.0 is already installed we won't enter the `except`

martinxu9: I think we kinda touched on this in the standup. I feel the main reason we wanted to do like this is for the installed payload from reflex. Twine is not a big package (not sure how to measure it), but the files in the site-packages directory seem to only mount to around a couple 100KBs. That's not a good enough reason to dance around an already established mechanism of specifying dependencies. Please feel free to share more thoughts. I guess some of the earlier messages likely preceded our standup discussion.

masenf (approved):

PR #2906, opened 2024-03-23
fix hook order to use ref inside user hooks

Lendemor:
First attempt : hooks render seems non-deterministic, we need to ensure the ref hook is added before others hooks.

### DIFF
--- a/reflex/.templates/jinja/web/pages/index.js.jinja2
+++ b/reflex/.templates/jinja/web/pages/index.js.jinja2
@@ -8,6 +8,10 @@
 
 {% block export %}
 export default function Component() {
+  {% for ref_hook in ref_hooks %}
+  {{ ref_hook }}
+  {% endfor %}
+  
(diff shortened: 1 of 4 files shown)


Lendemor:
> i think the core problem comes from this line
> 
> ```python
> component.get_hooks_internal() | component.get_hooks(),
> ```
> 
> instead of special casing refs, can we just move the internal hooks to always precede the user hooks?

I guess we could, but we have to move away from `set()` because they are not ordered, (my first attempt was just shuffling around the ref hooks to try to make them come first, which didn't always work.

masenf:
Lets add a ticket to the Component API issue that changes to using an underlying dict to keep the hooks, that way we can get the ordering _and_ the uniqueness

masenf (commented):
remove the prints and we can take this as it is for now

masenf (approved):

PR #2916, opened 2024-03-24
fix AppHarness reloading

benedikt-bartscher:
Fixes issues with `AppHarness` running against reflex projects which import `rx.State` subclasses from modules.


benchmarks/test_benchmark_compile_pages.py
```diff
@@ -209,10 +209,11 @@ def app_with_one_page(
     """
     root = tmp_path_factory.mktemp(f"app1")
 
-    yield AppHarness.create(root=root, app_source=AppWithOnePage)  # type: ignore
+    with AppHarness.create(root=root, app_source=AppWithOnePage) as harness:  # type: ignore
```
ElijahAhianyo, line 212: I dont think we'd want to use a context manager here because this starts/runs the app. We're only interested in measuring the compile time i.e the time it takes to run `AppHarness.app_instance._compile()`

benedikt-bartscher: @ElijahAhianyo thanks for reviewing! I will revert that commit. 
Maybe we could move the new state cleanup in a separate method of `AppHarness` and call it manually in the benchmark tests?

benedikt-bartscher: I just addressed the failing benchmark jobs by manually calling the state cleanup after benchmarking.

masenf (approved):
Excellent! 😀

PR #2922, opened 2024-03-25
[REF-2089] Use dill instead of cloudpickle for serialization

masenf:
* smaller size pickles
* support dynamically defined states [REF-2265]
* avoid issues with unpickleable globals [REF-1192]

### [Dynamically convert EventHandler to functools.partial](https://github.com/reflex-dev/reflex/commit/319ff17f4c4ea42a2163d0e0f3408d7400571d61)

* smaller size pickles!!!
* serialization of the state instance no longer needs to worry about serializing partials
* initialization of state does not need to unbox EventHandler functions up front

This gets us a step closer to EventHandler as descriptor without breaking any tests or existing functionality.


picklelo (approved):

reflex/state.py
```diff
@@ -2159,6 +2187,25 @@ async def modify_state(self, token: str) -> AsyncIterator[BaseState]:
             await self.set_state(token, state)
 
 
+# Workaround https://github.com/cloudpipe/cloudpickle/issues/408 for dynamic pydantic classes
+if not isinstance(State.validate.__func__, FunctionType):
+    cython_function_or_method = type(State.validate.__func__)
```
martinxu9, line 2192: so what type is a cython function?

martinxu9 (approved):
LGTM

reflex/state.py
masenf: a cython function is compiled to C via the cython library. there's no importable name for it though, because it masquerades as a builtin, but it's not accessible as a builtin name.

masenf: similar to the builtin `function` or `method` types, but those types are exposed via `types.FunctionType` and `types.MethodType` respectively

Lendemor (approved):

PR #2937, opened 2024-03-27
[REF-2269] Add `add_imports` API for component class

martinxu9:
## Summary
- Add new API `add_imports` for `Component` class. When user defines custom components, they can override this method to return the desired imports to be added.
- Add UT for new code.
## Examples
 ```
class Test(Component):
        def add_imports(self) -> Dict[str, str | ImportVar | List[str | ImportVar]]:
            return {"react": "Component"}
```
```
class Test(Component):
        def add_imports(self) -> Dict[str, str | ImportVar | List[str | ImportVar]]:
            return {"react": ["Component", "useEffect"]}
```
```
class Test(Component):
        def add_imports(self) -> Dict[str, str | ImportVar | List[str | ImportVar]]:
            return {"react": [ImportVar(tag="Component"), ImportVar(tag="useEffect")]}
```
```
class Test(Component):
        def add_imports(self) -> Dict[str, str | ImportVar | List[str | ImportVar]]:
            return {"react": ["Component", ImportVar(tag="useEffect")]}
```

## Tests
- [x] CI green

### CODE REVIEW

File: reflex/components/component.py [hunk 1]
```diff
@@ -1007,6 +1007,14 @@ def _get_hooks_imports(self) -> imports.ImportDict:
             )
         return _imports
 
+    def add_imports(self) -> Dict[str, Union[str, List[str]]]:
```
masenf, line 1010:
```suggestion
    def add_imports(self) -> Dict[str, Union[str, List[str], List[ImportVar]]]:
```

I think we should also allow the user to return a list of ImportVar as well, in case they need the more advanced functionality (like installation, but not rendering)

masenf (changes requested):

reflex/components/component.py
Lendemor: Wouldn't it be better to have `List[Union[str, ImportVar]` then ?
If I want to use ImportVar for a specific tag I don't necessarily want to have to wrap all the tags in ImportVar.

Lendemor (approved):

masenf (approved):
This is great, thanks

PR #2945, opened 2024-03-27
generate pyi files when building/publishing 3rd party component

Lendemor:


martinxu9:
will try to look at the PR, but it works great with the newly published `reflex-calendar` with pyi files.

reflex/utils/pyi_generator.py
```diff
@@ -17,8 +17,11 @@
 from types import ModuleType, SimpleNamespace
 from typing import Any, Callable, Iterable, Type, get_args
 
-import black
-import black.mode
+try:
+    import black
+    import black.mode
+except Exception:
+    pass
```
masenf, line 24:
```suggestion
except ImportError:
    black = None
```

then down below, instead of catching blanket Exception, check for `if black is not None`

masenf (changes requested):

masenf (approved):
i appreciate your dedication to test coverage 😅

Lendemor:
> i appreciate your dedication to test coverage 😅

yeah they are in need of attention 😛

picklelo (dismissed):
works well on my custom components

masenf (approved):
Good to merge without the macos runs

PR #2949, opened 2024-03-28
Set is_hydrated=False at route onChangeStart

masenf:
When navigation event starts, set is_hydrated=False on the client side before any on_load event is dispatched. This avoids a flickering problem where the client browser navigates and briefly shows content on the page before processing on_load events associated with the page.

Fix https://github.com/orgs/reflex-dev/discussions/2885

-----------

Bonus: [Connection pulser only depends on has_connection_errors](https://github.com/reflex-dev/reflex/commit/d47f0532447001204d915811ee35e2e03d766bd7)

Avoid showing the broken WiFi icon until there is actually a connection error.

### DIFF
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -646,9 +646,12 @@ export const useEventLoop = (
 
   // Route after the initial page hydration.
   useEffect(() => {
+    const change_start = () => dispatch["state"]({is_hydrated: false})
     const change_complete = () => addEvents(onLoadInternalEvent());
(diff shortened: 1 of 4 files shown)


picklelo (approved):
Nice, works for me!

My test code FWIW:

```python
import reflex as rx

class State(rx.State):
    @rx.var
    def thinking(self):
        import time
        time.sleep(1)
        return "hi"


def index() -> rx.Component:
    return rx.heading(State.thinking)


app = rx.App()
app.add_page(index)
```

PR #2952, opened 2024-03-28
[REF-2272] Support declaring EventHandlers directly in component

martinxu9:
## Summary
- Instead of needing to implement/override `get_event_triggers`, allow declaring event handlers by type annotation.
- Add UT for new code.
## Usage
```
class MyComponent(Component):
        on_a: EventHandler[lambda e0: [e0]]
        on_b: EventHandler[lambda e0: [e0.target.value]]
        on_c: EventHandler[lambda e0: []]
        on_d: EventHandler[lambda: []]
``` 
Is equivalent to 
```
class MyComponent(Component):
        def get_event_triggers(self) -> dict[str, Any]:
            return {
                **super().get_event_triggers(),
                "on_a": lambda e: [e],
                "on_b": lambda e: [e.target.value],
                "on_c": lambda e: [],
                "on_d": lambda: [],
            }
```
## Tests
- [x] CI green.


reflex/components/component.py
```diff
@@ -471,6 +471,11 @@ def get_event_triggers(self) -> Dict[str, Any]:
             EventTriggers.ON_MOUNT: lambda: [],
             EventTriggers.ON_UNMOUNT: lambda: [],
         }
+        # Look at the fields for EventHandler types.
+        for field in self.get_fields().values():
+            if types._issubclass(field.type_, EventHandler):
+                default_triggers[field.name] = field.type_.args_spec
```
masenf, line 477:
```suggestion
                default_triggers[field.name] = getattr(field.type_, "args_spec", lambda: [])
```

In case someone just passes `on_click: rx.EventHandler`

reflex/event.py
```diff
@@ -115,6 +117,21 @@ class Config:
         # Needed to allow serialization of Callable.
         frozen = True
 
+    @classmethod
+    def __class_getitem__(cls, type_: str) -> _GenericAlias:
```
masenf, line 121:
```suggestion
    def __class_getitem__(cls, args_spec: str) -> _GenericAlias:
```

i'd probably call this `args_spec` here for consistency

reflex/event.py
```diff
@@ -115,6 +117,21 @@ class Config:
...
+        """Get a typed var.
+
+        Args:
+            type_: The type of the var.
+
+        Returns:
+            The var class item.
+        """
```
masenf, line 129: remove references to `var`

tests/components/test_component.py
```diff
@@ -1350,3 +1350,33 @@ def add_imports(
...
+                "on_b": lambda e: [e.target.value],
+                "on_c": lambda e: [],
+                "on_d": lambda: [],
+            }
+
+    class TestComponent(Component):
+        on_a: EventHandler[lambda e0: [e0]]
+        on_b: EventHandler[lambda e0: [e0.target.value]]
+        on_c: EventHandler[lambda e0: []]
+        on_d: EventHandler[lambda: []]
```
masenf, line 1375:
```suggestion
        on_d: EventHandler[lambda: []]
        on_e: EventHandler
```

Somebody is going to use a bare `EventHandler` annotation, either on purpose or on accident. We should handle it and add a test case

masenf (changes requested):

reflex/components/component.py
```diff
@@ -272,7 +274,8 @@ def __init__(self, *args, **kwargs):
 
         # Get the component fields, triggers, and props.
         fields = self.get_fields()
-        triggers = self.get_event_triggers().keys()
+        event_triggers_in_component_declaration = self.get_event_triggers()
```
picklelo, line 277: Maybe call this `component_triggers` or `component_specific_triggers`

reflex/components/component.py
```diff
@@ -471,6 +471,11 @@ def get_event_triggers(self) -> Dict[str, Any]:
             EventTriggers.ON_MOUNT: lambda: [],
             EventTriggers.ON_UNMOUNT: lambda: [],
         }
+        # Look at the fields for EventHandler types.
+        for field in self.get_fields().values():
```
picklelo, line 475: A bit confused what this portion is doing, can we add more in the comments?

tests/components/test_component.py
picklelo: Tested with a bare event handler and it works well (acts like an event handler that doesn't pass arguments)

picklelo (commented):
Syntax is working well for me

reflex/components/component.py
martinxu9: how about using the same terms "component specific triggers"

picklelo (approved):

masenf (approved):
Changes look good. We can merge when CI is clean

Issue #2851, opened 2024-03-13
[REF-2219] Computed and cached vars based on rx.Cookie and rx.LocalStorage not getting updated starting from 0.4.3


PasqualePuzio:
**Describe the bug**
In our project we have a few computed (some of which are cached) vars which are based (they perform some calculations on) on other state vars of type rx.Cookie and rx.LocalStorage.
All of them belong to the same state.
We have noticed that starting from 0.4.3 (until 0.4.2 everything is fine according to our tests) the value of these vars are not consistent with the value of the cookie or the local storage entry. It looks like they're not updated when they should.

**To Reproduce**

    class MyState(rx.State):
    
      auth_data: str = rx.Cookie('',name='auth_user',max_age=2592000)
    
      @rx.cached_var
      def auth_user(self) -> dict:
          if self.auth_data is None or self.auth_data == '':
              return None
          try:
              user = json.loads(self.decrypt(self.auth_data))
              return user
          except:
              return None
    
      @rx.cached_var
      def is_user_logged_in(self) -> bool:
          return self.auth_user is not None and len(self.auth_user) > 0

**Expected behavior**
We would expect the same behavior we had until 0.4.2, meaning that the values of these vars reflect the value of the underlying rx.Cookie or rx.LocalStorage vars and get updated upon hydration and whenever the cookie value changes.

**Specifics**
 - Python Version: 3.9.6
 - Reflex Version: 0.4.3 and 0.4.4
 - OS: Mac OS 13.3.1
 - Browser: Chrome 122.0.6261.112

<sub>[REF-2219](https://linear.app/reflex-dev/issue/REF-2219/computed-and-cached-vars-based-on-rxcookie-and-rxlocalstorage-not)</sub>

masenf:
@PasqualePuzio How is the cookie / localstorage being set in your app? I'm not able to reproduce this when setting a cookie and navigating to another page.

masenf:
Or is the problem when the cookie expires it still appears to be set in the state?

PasqualePuzio:
> @PasqualePuzio How is the cookie / localstorage being set in your app? I'm not able to reproduce this when setting a cookie and navigating to another page.

We call `self.set_auth_data(encrypted)`

masenf:
Are you using redis? Do you see this discrepancy in dev mode?

PasqualePuzio:
> Are you using redis? Do you see this discrepancy in dev mode?

[...]

PR #2953, opened 2024-03-28
[REF-2219] Avoid refetching states that are already cached

masenf:
The already cached states may have unsaved changes which can be wiped out if they are refetched from redis in the middle of handling an event.

If the root state already knows about one of the potentially missing states, then use the instance that is already cached.

Fix https://github.com/reflex-dev/reflex/issues/2851

--------

BONUS: [_determine_missing_parent_states: correctly generate state names](https://github.com/reflex-dev/reflex/commit/decdb29433b56e2dc0282d8ca80297994f0e5865) 
Prepend only the previous state name to the current relative_parent_state_name
instead of joining all of the previous state names together.

This was discovered while testing the fix and writing the regression test.

--------

[Add test_get_state_from_sibling_not_cached](https://github.com/reflex-dev/reflex/commit/e6b105cd5b25b845b8d0f578369a488f5772dade) (so this _never_ happens again 😉)

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -1262,9 +1262,18 @@ class BaseState(Base, ABC, extra=pydantic.Extra.allow):
         ) = self._determine_missing_parent_states(target_state_cls)
 
         # Fetch all missing parent states and link them up to the common ancestor.
-        parent_states_by_name = dict(self._get_parent_states())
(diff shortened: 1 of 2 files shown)


picklelo (approved):
nice optimization

PR #2965, opened 2024-03-29
adjust share command prompt

martinxu9:
## Summary
- Adjust share command prompt. We won't encourage user to upload gif types. If they deploy apps, their app URL is useful enough.
- Instead of GET, use POST to check if user has permission to update an existing package. This way, we record a package if user has difficulties with any of the remaining steps.
- Move the prompt for demo URL to last: user may not have an app ready, so this step could take long (in a new tab/terminal separately) hence keeping it last.
## Tests
- Use the command against dev CP. Check the DB columns.


martinxu9:
Just received very useful feedback. Will add a bit more changes (but very much self contained in the component command logic)

picklelo (approved):

masenf (approved):
lgtm

PR #2966, opened 2024-03-29
[REF-2425] Reduce unnecessary CI runs

martinxu9:
## Summary
- Reduce the MacOS jobs to only necessary.
- Only run benchmark jobs on merge.
- Add concurrency groups to cancel in progress.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -5,10 +5,6 @@ on:
     branches: [main]
     paths-ignore:
       - '**/*.md'
-  pull_request:
-    branches: [main]
-    paths-ignore:
-      - '**/*.md'
 
 permissions:
   contents: read
@@ -23,7 +19,6 @@ env:
   NODE_OPTIONS: '--max_old_space_size=4096'
   DATABASE_URL: ${{ secrets.DATABASE_URL }}
   PR_TITLE: ${{ github.event.pull_request.title }}
-
 
 jobs:
   reflex-web:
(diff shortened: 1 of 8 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -5,10 +5,6 @@ on:
     branches: [main]
     paths-ignore:
       - '**/*.md'
-  pull_request:
```
masenf, line 8: i thought we did want benchmarking per PR though

masenf (commented):
So the `cancel-in-progress: true` is for killing runs when there's a new push? That's pretty cool

.github/workflows/benchmarks.yml
martinxu9: how do you feel about it? I just polled inside the office. Kinda felt like `on push` is likely fine. That means if we really have done something bad with the PR in terms of benchmarks, then we catch it after the fact.

martinxu9:
> So the `cancel-in-progress: true` is for killing runs when there's a new push? That's pretty cool

Yes, I tested it with an empty commit like this. I think our tests are likely  ok to terminate middle way 
image

.github/workflows/benchmarks.yml
martinxu9: I guess more because of the mac runner situation.

masenf: i'm okay with it only doing it on push; at least we still will have a linear benchmark history on the `main` branch that we can pinpoint to a particular merge that regressed.

we probably do NOT need multiple data points within the same PR for each push.

martinxu9: > probably do NOT need multiple data points
want to clarify: are you referring to the python/os matrix or you're agreeing the reduction in this PR is fine?

masenf: i agree with the reduction here. we don't need to run for each PR

masenf (approved):

PR #2995, opened 2024-04-03
use dict instead of set to store hooks

Lendemor:
hooks are currently stored in `set()` which do not preserve their order.
This can be problematic if a given hook A rely on another hook B but A is written before B during compilation.

By using dict we can keep the uniqueness of hooks as well as keep their order.

### DIFF
--- a/reflex/.templates/jinja/web/pages/index.js.jinja2
+++ b/reflex/.templates/jinja/web/pages/index.js.jinja2
@@ -8,10 +8,6 @@
 
 {% block export %}
 export default function Component() {
-  {% for ref_hook in ref_hooks %}
-  {{ ref_hook }}
-  {% endfor %}
-  
   {% for hook in hooks %}
   {{ hook }}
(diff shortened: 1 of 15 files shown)


picklelo (commented):
Why not use a list, if we don't have any values needed?

masenf:
> Why not use a list

At the end of the day we need to dedupe the values anyway. We can't use a set for this because sets lose the order. So even if we keep it in a list, ultimately it will have to go through a value-less dict at some point anyway.

I guess the decision is just about the API, is it better to have the users deal with lists?

Lendemor:
> I guess the decision is just about the API, is it better to have the users deal with lists?

I think most users will never interact with either in the end, because they will mostly define the hooks in the _underscored method returning a single hook ?

picklelo:
Okay makes sense to me - I'm good with the dict method

picklelo (approved):

PR #2998, opened 2024-04-03
use Any for serializer fn value annotation

benedikt-bartscher:
Fixes issues with newer pyright versions


masenf:
Sorry accidentally marked this as ready.

I ran with the latest pyright and reflex has LOTS of issues `338 errors, 4 warnings, 0 informations` 😢

benedikt-bartscher:
Yes, I know :(  
But many of those issues do not arise if you use reflex in your project with a recent pyright version.  
I wanted to address some of reflex's public apis to reduce the amount of `type: ignore` annotations in projects using reflex

PR #2999, opened 2024-04-03
[REF-2296] Rename recursive functions

martinxu9:
## Summary
- Rename the current recursive functions, so it is more consistent with common naming conventions, also user unlikely to override them when implementing their components.
  - get_imports -> _get_all_imports
  - get_hooks -> _get_all_hooks
  - get_hooks_internal -> _get_all_hooks_internal
  - add_style -> add_styles_recursive
  - get_custom_code -> _get_all_custom_code
  - get_dynamic_imports -> _get_all_dynamic_imports
  - get_custom_components -> _get_all_custom_components
  - get_app_wrap_components -> _get_all_app_wrap_components
  - get_refs -> _get_all_refs


reflex/components/component.py
```diff
@@ -529,7 +529,7 @@ def _exclude_props(self) -> list[str]:
         """
         return []
 
-    def _render(self, props: dict[str, Any] | None = None) -> Tag:
+    def _render_tag(self, props: dict[str, Any] | None = None) -> Tag:
```
picklelo, line 532: I think this one we want to be a public method since users may override it `render_tag`

Should we do this one in the follow up when we change the public api?

reflex/components/component.py
```diff
@@ -1455,13 +1455,13 @@ def _get_all_custom_components(
                 )
         return custom_components
 
-    def _render(self) -> Tag:
+    def _render_tag(self) -> Tag:
```
picklelo, line 1458: Here too we can keep it for now

reflex/components/component.py
```diff
@@ -677,7 +677,7 @@ def _add_style(self, style: dict):
         """
         self.style.update(style)
 
-    def add_style(self, style: ComponentStyle) -> Component:
+    def add_style_recursive(self, style: ComponentStyle) -> Component:
```
picklelo, line 680: Let's call it `_add_all_styles` for consistency

picklelo (changes requested):
Looks good, let's hold off on the `_render` for now since that will change the public API

reflex/components/component.py
martinxu9: look at what the function does, it's not adding all styles, it's adding the style in the args to itself and children.

martinxu9: also, are we adding the prefix here? In the ticket, it was not adding that.

picklelo: Got it, add_style_recursive is good then. I think we should add the prefix because this isn't a method we expect the user to have to modify

picklelo: In a way it is adding all styles in the style dict recursively down the tree, so I think add_all_styles makes sense as well.

reflex/components/component.py
```diff
@@ -838,7 +838,7 @@ def _get_vars_from_event_triggers(
                         event_args.extend(args)
                 yield event_trigger, event_args
 
-    def _get_vars(self, include_children: bool = False) -> list[Var]:
+    def _get_all_vars(self, include_children: bool = False) -> list[Var]:
```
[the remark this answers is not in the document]
masenf: yeah this was an oversight on my part; since the recursive flag is false by default, i think we can leave this one alone.

masenf (dismissed):
This looks good to me.

My personal preference is `_add_style_recursive` but i wouldn't push too hard about it. I like `_add_style_recursive` because it describes exactly what this function is doing: adding the given style recursively down the tree.

reflex/components/chakra/forms/pininput.py
```diff
@@ -71,7 +72,7 @@ def _get_imports(self) -> ImportDict:
         range_var = Var.range(0)
         return merge_imports(
             super()._get_imports(),
-            PinInputField().get_imports(),  # type: ignore
+            PinInputField()._get_all_imports(),  # type: ignore
```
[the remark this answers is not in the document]
martinxu9: I would say not introduce the public APIs in this PR. The intention here is to pave the way for the right ones. It's easier to introduce a missing one than deprecating already public ones. So probably leave one be since we established it's a rare case anyway

reflex/components/component.py
```diff
@@ -669,7 +669,7 @@ def _add_style(self, style: dict):
         """
         self.style.update(style)
 
-    def add_style(self, style: ComponentStyle) -> Component:
+    def _add_style_recursive(self, style: ComponentStyle) -> Component:
```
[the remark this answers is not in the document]
martinxu9: I'm not hard lined on the wording here. I get the impression we like **all** and want to establish a convention here. I think this function is an exception to the other ones. The others are **getting**. One can quickly see **getting all** is to get from self **and** the children. This one, however, is an action taken on all the targets (self and children). So either **all** should describe **add** action or it describes the targets (in which case, I would say `_add_style_to_all`). The suggestion adds the `all` on `style`, which I think it's not immediately obvious. It also makes you wonder if it is a typo: shouldn't it be all **styles** instead.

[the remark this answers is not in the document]
picklelo: Sounds good let's keep this one

picklelo (approved):

PR #3007, opened 2024-04-04
[REF-2303] Update custom component template for new API

martinxu9:
## Summary
- Update custom component template for new API, mainly rx.EventHandler.

### DIFF
--- a/reflex/.templates/jinja/custom_components/src.py.jinja2
+++ b/reflex/.templates/jinja/custom_components/src.py.jinja2
@@ -43,9 +43,9 @@ class {{ component_class_name }}(rx.Component):
     # lib_dependencies: list[str] = []
 
     # Event triggers declaration if any.
-    # Below is equivalent to merging { "on_change": lambda e0: [e0] }
-    # onto the default event triggers of parent/based Component.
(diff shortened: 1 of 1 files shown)


reflex/.templates/jinja/custom_components/src.py.jinja2
```diff
@@ -42,12 +42,12 @@ class {{ component_class_name }}(rx.Component):
...
-    #     return {
-    #         **super().get_event_triggers(),
-    #         "on_change": lambda e0: [e0],
-    #     }
+    # Event triggers declaration if any.
+    # Below is equivalent to merging `{ "on_change": lambda e0: [e0] }`
+    # onto the default event triggers of parent/base Component.
+    # The function defined for the `on_change` trigger maps args for the javascript
+    # trigger to args that will be passed to the backend event handler function.
+    # on_change: rx.EventHandler[lambda e0: [e0]]
```
picklelo, line 50: small nit, but should we call this variable `e` rather than `e0`? Or maybe even something like `arg`, I'm thinking the original name may be confusing.

picklelo (approved):

PR #3011, opened 2024-04-04
Reflex run automatically inits when needed

picklelo:
Previously if you didn't have a `.web` folder or the app template had updated, `reflex run` would throw a warning saying to run `reflex init` again. This PR just automatically runs init for them, reducing the steps to get the app running.

### DIFF
--- a/reflex/reflex.py
+++ b/reflex/reflex.py
@@ -165,7 +165,8 @@ def _run(
         _skip_compile()
 
     # Check that the app is initialized.
-    prerequisites.check_initialized(frontend=frontend)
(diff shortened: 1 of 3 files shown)


martinxu9 (approved):

Sprocketer:
@picklelo Just saw this in the new release - surely if there's ever a regression this could break an app if a server running a Reflex app restarts?

picklelo:
@Sprocketer it will never update the reflex version. It will only call `reflex init` if the `.web` folder is missing or out of date.

PR #3041, opened 2024-04-08
fix download for uploaded files

Lendemor:


masenf:
I think the existing code already mostly works without adding all this extra structure. There just seems to be a problem when inlining the `rx.download` event handler:

```python
import reflex as rx

class State(rx.State):
    button_text: str = "Download"
    def do_download(self):
        filename = "foo.txt"
        (rx.get_upload_dir() / filename).write_text("test")
        return rx.download(rx.get_upload_url(filename))


@rx.page()
def index():
    return rx.vstack(
        # Works
        rx.button(State.button_text, on_click=State.do_download),
        # Works
        rx.link("Download", href=rx.get_upload_url("foo.txt")),
        # This doesn't seem to work...
        # rx.button(State.button_text, on_click=rx.download(rx.get_upload_url("foo.txt"))),
    )

app = rx.App(overlay_component=None)
```

masenf:
```diff
diff --git a/reflex/components/core/upload.py b/reflex/components/core/upload.py
index dfc62bc0..ae1e5f29 100644
--- a/reflex/components/core/upload.py
+++ b/reflex/components/core/upload.py
@@ -140,7 +140,7 @@ def get_upload_url(file_path: str) -> Var[str]:
     """
     Upload.is_used = True
 
-    return Var.create_safe(f"{uploaded_files_url_prefix}/{file_path}")
+    return Var.create_safe(f"{uploaded_files_url_prefix}/{file_path}", _var_is_string=True)
 
 
 def _on_drop_spec(files: Var):
```

This seems sufficient to get things working as expected without introducing new API

reflex/.templates/web/utils/state.js
```diff
@@ -161,8 +161,14 @@ export const applyEvent = async (event, socket) => {
...
     a.hidden = true;
-    // Special case when linking to uploaded files
-    a.href = event.payload.url.replace("${getBackendURL(env.UPLOAD)}", getBackendURL(env.UPLOAD))
+    // Special case when linking to uploaded files (when rx.download is returned from upload_files handler())
+    a.href = event.payload.url.replace(
+      "${getBackendURL(env.UPLOAD)}",
+      getBackendURL(env.UPLOAD)
+    );
+    if (event.payload.filename) {
+      a.href = a.href + "?filename=" + event.payload.filename;
```
masenf, line 170: potentially minor issue, but is there a case where the href might already have query params?

masenf (dismissed):

reflex/.templates/web/utils/state.js
Lendemor: I don't think so.

masenf (approved):

masenf (changes requested):
actually, i think this regressed the `rx.download` from backend case, that's no longer working in my sample code that i posted above

masenf:
Here is the repro

```python
import reflex as rx

class State(rx.State):
    button_text: str = "Download"
    def do_download(self):
        filename = "foo.txt"
        (rx.get_upload_dir() / filename).write_text("test")
        print("returning download event")
        #return rx.download(rx.get_upload_url(filename), filename="bar.csv")
        return rx.download(rx.get_upload_url(filename))

    def do_download_assets(self):
        return rx.download("/favicon.ico")

    def something_else(self):
        return rx.console_log("something else")


@rx.page()
def index():
    return rx.vstack(
        rx.button(State.button_text, on_click=State.do_download),
        rx.button("Download favicon.ico", on_click=State.do_download_assets),
        rx.button("Something else", on_click=State.something_else),
        rx.button("Download2", on_click=rx.download(data="test", filename="foo.csv")),
        rx.link("Download", href=rx.get_upload_url("foo.txt"), download="foo.csv"),
    )

app = rx.App(overlay_component=None)
```

If you click the top "Download" button, it works once, then the "Something Else" and "Download favicon.ico" buttons stop working until the page is refreshed. Interestingly, the "Download2" button which directly adds the `_download` event will still work after clicking the top "Download" button. It seems to only affect events that need to go to the backend first.

You can click "Download favicon.ico" or "Something Else" as many times as you want and they seem to keep working, as long as "Download" isn't clicked.

This is really weird and mysterious.

amballa:
Looks like there's still an issue with using `rx.get_upload_url()` with `rx.download()` in the latest reflex version

adhami3310:
> Looks like there's still an issue with using `rx.get_upload_url()` with `rx.download()` in the latest reflex version

Thanks for the comment! Would you please file an issue? Or is it just the code above?

amballa:
> > Looks like there's still an issue with using `rx.get_upload_url()` with `rx.download()` in the latest reflex version
> 
> Thanks for the comment! Would you please file an issue? Or is it just the code above?

Turns out I was on 0.6.7 and upgrading to 0.6.8 solved the issue !

Issue #2972, opened 2024-04-01
[REF-2444] IndexError: list index out of range when publishing custom component


masenf:
**Describe the bug**
```
Publishing custom component to pypi...
Distribution files for the version to be published already exist. Do you want to rebuild? [n/y] (n): y
Building custom component...
Traceback (most recent call last):
  File "/Users/masen/code/reflex-dev/VENV-dev311/bin/reflex", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/Users/masen/code/reflex-dev/VENV-dev311/lib/python3.11/site-packages/typer/main.py", line 328, in __call__
    raise e
  File "/Users/masen/code/reflex-dev/VENV-dev311/lib/python3.11/site-packages/typer/main.py", line 311, in __call__
    return get_command(self)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masen/code/reflex-dev/VENV-dev311/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masen/code/reflex-dev/VENV-dev311/lib/python3.11/site-packages/typer/core.py", line 778, in main
    return _main(
           ^^^^^^
  File "/Users/masen/code/reflex-dev/VENV-dev311/lib/python3.11/site-packages/typer/core.py", line 216, in _main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/masen/code/reflex-dev/VENV-dev311/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masen/code/reflex-dev/VENV-dev311/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masen/code/reflex-dev/VENV-dev311/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

PR #3048, opened 2024-04-08
Get project name from pyproject

picklelo:
`reflex component build` was throwing an error when the egg folder didn't exist. We can rely on `pyproject.toml` always existing.

### DIFF
--- a/reflex/custom_components/custom_components.py
+++ b/reflex/custom_components/custom_components.py
@@ -71,6 +71,28 @@ def _create_package_config(module_name: str, package_name: str):
         )
 
 
+def _get_package_config(exit_on_fail: bool = True) -> dict:
+    """Get the package configuration from the pyproject.toml file.
+
+    Args:
+        exit_on_fail: Whether to exit if the pyproject.toml file is not found.
+
+    Returns:
+        The package configuration.
+
+    Raises:
+        Exit: If the pyproject.toml file is not found.
+    """
+    try:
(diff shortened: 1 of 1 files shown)


Lendemor (approved):

masenf (dismissed):

Lendemor (dismissed):

reflex/custom_components/custom_components.py
```diff
@@ -416,9 +416,8 @@ def _run_commands_in_subprocess(cmds: list[str]) -> bool:
 
 def _make_pyi_files():
     """Create pyi files for the custom component."""
-    from glob import glob
-
-    package_name = glob("custom_components/*.egg-info")[0].replace(".egg-info", "")
+    with open(CustomComponents.PYPROJECT_TOML, "rb") as f:
+        package_name = tomlkit.load(f).get("project", {})["name"]
```
masenf, line 420: why the `.get` only for `"project"`... if that fails, we'll for sure get a `KeyError` looking up `"name"` in the empty dict.

masenf (changes requested):

reflex/custom_components/custom_components.py
picklelo: It was only to make the pyright happy. I'll refactor everything to a common function

masenf (approved):
nice

kshitiz305:
closes #2972

Issue #2873, opened 2024-03-18
Uploader with no files uploaded


Ernelene:
I'm currently implementing an uploader feature and need to handle the scenario where there is an error message displayed when no files are uploaded. I'm using the rx.upload_files(upload_id=id) and rx.selected_files(id) functions, but I'm unsure how to detect when no files have been uploaded. It seems that these functions don't execute when no files are uploaded. Could you please advise on how I can check for this condition and create an error message accordingly?

masenf:
Currently this is expected behavior given this code: https://github.com/reflex-dev/reflex/blob/ee1ff7f93fa5bfbeb48f296970977c9174fe7136/reflex/.templates/web/utils/state.js#L382-L385

The client side upload handler just returns if no files are selected for upload.

I suppose this code could be changed to still submit the request, but return an empty file list.

Yummy-Yums:
please assign

Yummy-Yums:
@masenf tried this and repro'd ; after clicking the upload btn , what comes through to state.js (uploadFiles ) is undefined . i logged at the call-site of* uploadFile*s and it seems 
event.payload.files
  is undefined when the user sends an "empty upload" . so it throws a parse error in the eventhandler func in uploadFiles. i think a simple log(in addition to 
return false
 to notify the user to handle this , will suffice but unless some other reason to insist on sending an empty array to the backend?

![image](https://github.com/reflex-dev/reflex/assets/77977520/21b94079-4840-4f64-9bbb-c77ac4a6ea4f)

PR #3051, opened 2024-04-09
Fix For When user selects Empty files using File Upload

Yummy-Yums:
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

    b. Describe your changes.

### DIFF
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -387,7 +387,7 @@ export const uploadFiles = async (
 ) => {
   // return if there's no file to upload
   if (files === undefined || files.length === 0) {
-    files = []
+    return false;
   }
 
   if (upload_controllers[upload_id]) {
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -244,7 +244,7 @@ export const applyRestEvent = async (event, socket) => {
 
     if (event.payload.files === undefined || event.payload.files.length === 0){
       // Submit the event over the websocket to trigger the event handler.
-      return await applyEvent(Event(event.name, event.payload.files), socket)
+      return await applyEvent(Event(event.name), socket)
     }
 
     // Start upload, but do not wait for it, which would block other events.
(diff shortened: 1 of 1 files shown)


Yummy-Yums:
#2873

reflex/.templates/web/utils/state.js
```diff
@@ -240,7 +240,13 @@ export const applyEvent = async (event, socket) => {
  */
 export const applyRestEvent = async (event, socket) => {
   let eventSent = false;
-  if (event.handler == "uploadFiles") {
+  if (event.handler === "uploadFiles") {
+
+    if (event.payload.files === undefined || event.payload.files.length === 0){
+      // Submit the event over the websocket to trigger the event handler.
+      return await applyEvent(Event(event.name, event.payload.files), socket)
```
masenf, line 247: the payload is supposed to be a object, not an array

Yummy-Yums: you mean `return await applyEvent(Event(event.name, event.payload), socket)`

masenf: the problem is that `event.payload` contains keys that do not map to an arbitrary event handler on the backend.

for example, maybe it works if the backend event handler has a `files` arg; but if that arg is named `fools`, then we get an error on the backend about unknown keyword argument.

reflex/.templates/web/utils/state.js
```diff
@@ -381,7 +387,7 @@ export const uploadFiles = async (
 ) => {
   // return if there's no file to upload
   if (files === undefined || files.length === 0) {
-    return false;
+    files = []
```
masenf, line 390: what is this change doing? if we end up hitting this code, won't it ultimately throw that "Did not find CR at end of boundary (40)" error in the upload handler?

masenf (approved):
the only thing that still makes me nervous about this is that, if the handler doesn't specify a default for the `list[rx.UploadFile]` arg, then there will be a traceback printed on the backend, whereas before, this was just a no op.

i think that's okay though. it will still be a no-op, and if they want to avoid the traceback, they can add a default value...

Yummy-Yums:
> the only thing that still makes me nervous about this is that, if the handler doesn't specify a default for the `list[rx.UploadFile]` arg, then there will be a traceback printed on the backend, whereas before, this was just a no op.
> 
> i think that's okay though. it will still be a no-op, and if they want to avoid the traceback, they can add a default value...

i think we can put it in a doc somewhere ? or leave it like that? what dyu suggest ?

PR #3062, opened 2024-04-10
[REF-2524]Tuple unpacking should apply at component level

ElijahAhianyo:
The code below should work now:

```python

def index() -> rx.Component:
    random_text = rx.text("some text here"),
    return rx.center(
        random_text
    )
    
```

Also the code below will be unwrapped in a fragment: as well
    
```python
def index() -> rx.Component:
    comp = (rx.text("some text here"), rx.text("another text"))
    return rx.center(
        comp
    )
```
Equivalent to below:
    
```python
def index() -> rx.Component:
    return rx.center(
        rx.fragment(
             rx.text("some text here"), 
            rx.text("another text")
       ),
    )
```

### DIFF
--- a/reflex/components/base/body.pyi
+++ b/reflex/components/base/body.pyi
@@ -82,8 +82,5 @@ class Body(Component):
 
         Returns:
             The component.
-
-        Raises:
-            TypeError: If an invalid child is passed.
         """
         ...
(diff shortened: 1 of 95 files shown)


Lendemor (approved):

picklelo (approved):
awesome

PR #3072, opened 2024-04-11
Support replacing route on redirect

masenf:
Support next/router `.replace` interface to change page without creating a history entry.


picklelo (approved):

Lendemor (approved):
I have no issue with the implementation.

I'm just not sure when/where this will be needed? for redirecting from login page maybe?

masenf:
> for redirecting from login page maybe?

Precisely.

There are other scenarios as well, like using URL-based state, but not wanting the back button to transition between those states.

PR #3073, opened 2024-04-11
minimal fix for upload_url in rx.download

Lendemor:
minimal fix on get_upload_url to at least avoid compilation errors.

Disclaimer: this PR ignore the fact that rx.download is not a true download for CORS situation (including the backend), this will be adressed by #3041.


masenf (approved):

masenf:
not sure how this change would result in those reflex-web errors, maybe that's something else?

picklelo (approved):
reflex-web CI is fixed now

PR #3082, opened 2024-04-12
add pathlib.Path serializer

Lendemor:
Allow to use `pathlib.Path` natively, which is very useful when handling uploaded file.

If you set a Path var in your state, you can use it without transformation for both frontend & backend:
`self.my_image = Path(upload.filename)`

## Use in backend
`image_content = (rx.get_upload_dir() / self.my_image).read_bytes()`

## Used for rendering:
`rx.image(src=rx.get_upload_url(State.my_image))`


picklelo (approved):
nice

PR #3089, opened 2024-04-15
Simplify `run_process_with_fallback` logic

masenf:
Include log streaming for initial command and fallback command. This is necessary because if the command produces significant output that is not consumed, it can fill the OS pipe and block the process from running.


picklelo (approved):
Logs are showing up for me on Windows now

Issue #1482, opened 2023-07-31
Add debouncing for slider components


picklelo:
#1383 forces inputs and textareas to be uncontrolled, and lets you use `rx.debounce_input` to make them controlled. We should support a similar method for sliders. Maybe we can rename the component to `rx.debounce` and have it generalize to all components?

masenf:
Not so sure about this one. We already expose `on_change_start` and `on_change_end` and describe those pretty well in the docs. I'm not sure what debounce would add to those existing event handlers.

If you think those are fine, then we should close this issue @picklelo.

dodeca-6-tope:
@masenf it's true that there are "discrete" handlers already but I believe generalizing this component like suggested would improve UX. I'm thinking about dropping this component in my app just because it lags on the frontend.

dodeca-6-tope:
Is there a way to continuously update slider value on the frontend and debounce on the backend? Currently there is a lag on the frontend (also on the [slider page](https://reflex.dev/docs/library/forms/slider/)). I think such enhancement can improve UX a lot.

masenf:
> Is there a way to continuously update slider value on the frontend and debounce for the backend?


There isn't a _nice_ way to do it.

We've considered an idea for first class "frontend state" where values can be stored entirely in the browser without roundtripping to the backend, but it complicates the overall design of reflex and we haven't gone forward with it intentionally.

It's hackable though, if you're okay getting your hands dirty. And thanks to recent changes in the reflex compiler, it's actually not too nasty to implement:

Essentially we have to create our own `rx.Var` which is tied to react's `setState` hook to create local state, and another `rx.Var` which can act as the event handler for `on_change` to set the local state value.

The last piece of the puzzle is this weird `IntegralFragment` which inherits from `MemoizationLeaf`, meaning that it and all it's children will NOT be separated into their own react components; this allows all of the components inside the `IntegralFragment` to access the same `local_slider_value` variable.

```python
import reflex as rx
from reflex.components.component import MemoizationLeaf
from reflex.utils import imports


class State(rx.State):
    val: int = 50
    

LOCAL_SLIDER_VALUE = "local_slider_value"
SET_LOCAL_SLIDER_VALUE = "set_local_slider_value"
LOCAL_SLIDER_VAR_DATA = rx.vars.VarData(
    imports={"react": {imports.ImportVar(tag="useState")}},
[...]

Issue #2872, opened 2024-03-18
[REF-2250] Investigate generic debounce


picklelo:
Currently the `rx.debounce` works well with inputs and textareas. Is there a way we can generalize this to work for all components such as `rx.slider` or the color picker example in the docs?

Current docs: https://reflex.dev/docs/library/forms/debounce/#debounce

<sub>[REF-2250](https://linear.app/reflex-dev/issue/REF-2250/investigate-generic-debounce)</sub>

picklelo:
Modifying `queueEvents` like this within `utils/state.js` seems to work. We should probably make this a flag though, we mayn't want to do this for all events.

```python
export const queueEvents = async (events, socket) => {
  // Filter out any events already in the queue.
  events = events.filter((event) => {
    for (const queued_event of event_queue) {
      if (queued_event.name === event.name) {
        return false;
      }
    }
    return true;
  });
  event_queue.push(...events);
  await processEvent(socket.current);
};
```

masenf:
fixed in https://github.com/reflex-dev/reflex/pull/3091

PR #3091, opened 2024-04-15
Implement `throttle` and `debounce` as event actions

masenf:
Generic client side throttling and debounce for events.

Used just like `.stop_propagation` and `.prevent_default`, so applicable to all event handler or event spec objects.

Due to how events are handled, there isn't currently a way to apply these to only a single event in the chain. If multiple event handlers are specified and any of them have `debounce` or `throttle`, then it will apply to all of the other events as well.

If the same function is throttled or debounced at the same limit/delay by multiple event triggers, these will all apply in the same domain. If different limit or delay is used, then they will be debounced or throttled separately.

## Sample Code

```python
import time


import reflex as rx


class State(rx.State):
    """The app state."""
    _t_events: int = 0
    _d_events: int = 0
    last_t_event: str
    last_d_event: str
    slider_value: list[int] = 30

    def handle_mouse_move_t(self):
        self._t_events += 1
        self.last_t_event = f"throttle moved: {self._t_events} {time.time()}"
        print(self.last_t_event)

    def handle_mouse_move_db(self):
        self._d_events += 1
        self.last_d_event = f"debounce moved: {self._d_events} {time.time()}"
        print(self.last_d_event)

    def handle_slider_change(self, value):
        self.slider_value = value[0]


def index() -> rx.Component:
    return rx.hstack(
        rx.box(
            rx.vstack(
                rx.heading("Throttle"),
                rx.box(height="50vh"),
                State.last_t_event,
                align="center",
                padding_top="10vh",
            ),
            on_mouse_move=lambda: State.handle_mouse_move_t.throttle(500),
            height="85vh",
            width="50vw",
            background="linear-gradient(#e66465, #9198e5)"
        ),
        rx.box(
            rx.vstack(
                rx.heading("Debounce"),
                rx.box(height="50vh"),
                State.last_d_event,
                align="center",
                padding_top="10vh",
            ),
            on_mouse_move=State.handle_mouse_move_db.debounce(500),
            height="85vh",
            width="50vw",
            background="linear-gradient(#9198e5, #e66465)"
        ),
    ), rx.cond(
        State.is_hydrated,
        rx.vstack(
            rx.heading(State.slider_value),
            rx.slider(
                default_value=[State.slider_value],
                min=0,
                max=100,
                on_change=State.handle_slider_change.debounce(50),
            ),
            align="center",
        )
    )


app = rx.App()
app.add_page(index)
```

https://github.com/reflex-dev/reflex/assets/1524005/5058b6bb-370d-4ef3-b752-cfea7cf22cd5

Edit 2024-04-16: added a debounced slider example. Note it's important to wrap is the `is_hydrated` cond, otherwise the default_value after refreshing will be the initial value, not the last value.

### DIFF
--- a/integration/test_event_actions.py
+++ b/integration/test_event_actions.py
@@ -134,7 +134,7 @@ def TestEventAction():
             rx.button(
                 "Throttle",
                 id="btn-throttle",
-                on_click=EventActionState.on_click_throttle.throttle(
+                on_click=lambda: EventActionState.on_click_throttle.throttle(
                     200
                 ).stop_propagation,
             ),
@@ -327,6 +327,10 @@ async def test_event_actions_throttle_debounce(
         btn_throttle.click()
         btn_debounce.click()
 
-    await poll_for_order(
-        ["on_click_throttle"] * int(exp_events) + ["on_click_debounce"]
-    )
+    try:
+        await poll_for_order(["on_click_throttle"] * exp_events + ["on_click_debounce"])
+    except AssertionError:
+        # Sometimes the last event gets throttled due to race, this is okay.
+        await poll_for_order(
+            ["on_click_throttle"] * (exp_events - 1) + ["on_click_debounce"]
+        )
--- a/integration/test_event_actions.py
+++ b/integration/test_event_actions.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import asyncio
+import time
 from typing import Callable, Coroutine, Generator
 
 import pytest
@@ -24,6 +25,12 @@ def TestEventAction():
 
         def on_click2(self):
             self.order.append("on_click2")
+
+        def on_click_throttle(self):
+            self.order.append("on_click_throttle")
+
+        def on_click_debounce(self):
+            self.order.append("on_click_debounce")
 
     class EventFiringComponent(rx.Component):
         """A component that fires onClick event without passing DOM event."""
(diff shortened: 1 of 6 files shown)


Lendemor (approved):
Nice, the UX will be much better for high fire-rate events !

Could we add a test when the event handler is a lambda? I expect the behaviour to be the same, but might as well be sure.

integration/test_event_actions.py
```diff
@@ -124,6 +131,20 @@ def index():
                     "custom-prevent-default"
                 ).prevent_default,
             ),
+            rx.button(
+                "Throttle",
+                id="btn-throttle",
+                on_click=lambda: EventActionState.on_click_throttle.throttle(
```
Lendemor, line 137: I shoud have been more specific in the test I wanted, but since lambda are used to pass pre-set parameter, when we do, do we use 
```python
lambda value: EventActionState.on_change_throttle_with_args(value, "arg1").throttle(200).stop_propagation
```
or
```python
(lambda value: EventActionState.on_change_throttle_with_args(value, "arg1")).throttle(200).stop_propagation
```
?
Though maybe it's more important to have this particular example in the docs than the tests.

Lendemor (approved):

Lendemor:
closes #1482 #2872

picklelo (approved):
works great with my color picker

PR #3093, opened 2024-04-16
do not clear all decorated pages during AppHarness tests

benedikt-bartscher:
Mitigates issues with multiple AppHarness instances running against the same app.

AppHarness currently clears the global scoped `DECORATED_PAGES` before starting the app. But reflex only reloads the "root" app module, reloading is not recursive. If you import a module with decorated pages (`@rx.page`) those pages are not applied after the first AppHarness instance is done.

This PR tries to "teardown" the changes during the test, similar to https://github.com/reflex-dev/reflex/pull/2916


masenf (approved):
I like this one. Thanks

PR #3099, opened 2024-04-17
Use Fallback command on all platforms

ElijahAhianyo:
The fallback command(`npm`) when installing packages should be used on all platforms.
Also the `--loglevel` flag is used only for npm when installing packages. Bun already runs in debug mode by default and the `--loglevel` flag has no effect

### DIFF
--- a/reflex/utils/prerequisites.py
+++ b/reflex/utils/prerequisites.py
@@ -821,18 +821,16 @@ def install_frontend_packages(packages: set[str], config: Config):
     Example:
         >>> install_frontend_packages(["react", "react-dom"], get_config())
     """
-    # unsupported archs will use npm anyway. so we dont have to run npm twice
+    # unsupported archs(arm and 32bit machines) will use npm anyway. so we dont have to run npm twice
     fallback_command = (
-        get_package_manager()
+        get_install_package_manager()
         if not constants.IS_WINDOWS
         or constants.IS_WINDOWS
         and constants.IS_WINDOWS_BUN_SUPPORTED_MACHINE
         else None
     )
     processes.run_process_with_fallback(
-        processes.get_command_with_loglevel(
-            [get_install_package_manager(), "install"], config  # type: ignore
-        ),
+        [get_install_package_manager(), "install"],  # type: ignore
         fallback=fallback_command,
         show_status_message="Installing base frontend packages",
         cwd=constants.Dirs.WEB,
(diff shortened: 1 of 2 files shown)


reflex/utils/prerequisites.py
```diff
@@ -824,11 +824,15 @@ def install_frontend_packages(packages: set[str], config: Config):
     # unsupported archs will use npm anyway. so we dont have to run npm twice
     fallback_command = (
         get_package_manager()
-        if constants.IS_WINDOWS and constants.IS_WINDOWS_BUN_SUPPORTED_MACHINE
+        if not constants.IS_WINDOWS
```
Lendemor, line 827: Shouldn't the fallback always be `npm` regardless of OS based on what we discussed? Why is there a case where it's still None ?

reflex/utils/prerequisites.py
```diff
@@ -824,11 +824,15 @@ def install_frontend_packages(packages: set[str], config: Config):
...
+        or constants.IS_WINDOWS
+        and constants.IS_WINDOWS_BUN_SUPPORTED_MACHINE
         else None
     )
     processes.run_process_with_fallback(
-        [get_install_package_manager(), "install", "--loglevel", "silly"],
+        processes.get_command_with_loglevel(
```
Lendemor, line 833: This will not work because the `run_process_with_fallback` method will pass the args of the main command to the second fallback one, but they don't expect the same log `flag`.

Lendemor (changes requested):

reflex/utils/prerequisites.py
ElijahAhianyo: Yes, I've pushed a change to resolve this as well

reflex/utils/processes.py
```diff
@@ -319,7 +339,7 @@ def run_process_with_fallback(args, *, show_status_message, fallback=None, **kwa
                 f"There was an error running command: {args}. Falling back to: {fallback_args}."
             )
             run_process_with_fallback(
-                fallback_args,
+                get_command_with_loglevel(fallback_args),
```
masenf, line 342: do we need this again here? because this is a recursive call and we handle it on line 328

masenf (changes requested):

reflex/utils/processes.py
ElijahAhianyo: Ah yes, that's absolutely right

Lendemor (approved):

PR #3113, opened 2024-04-18
allow set in var.contains

martinxu9:


masenf (dismissed):

masenf:
ah need to just remove the `set` case from `test_unsupported_types_for_contains`

martinxu9:
@masenf stamp me again please

masenf (approved):

PR #3149, opened 2024-04-23
extend rx.input allowed types

Lendemor:


reflex/components/core/debounce.py
```diff
@@ -34,7 +35,7 @@ class DebounceInput(Component):
     force_notify_on_blur: Var[bool]
 
     # If provided, create a fully-controlled input
-    value: Var[str]
+    value: Var[Union[str, int, bool, float]]
```
masenf, line 38: idk about `bool` being here. it doesn't seem to round trip properly.

If i have an input with a bool value, it renders as `true` or `false`, and if i edit the text, then my bool Var gets a string value on the backend.

masenf (changes requested):

reflex/components/core/debounce.py
Lendemor: I only put `str, int, float` initially, but `bool` was present in `el.Input` so I added it here too, but I can probably remove `bool` from `el.Input` instead ?

masenf: yeah i think all of the `rx.el` components had a prop annotation of `Var[str, int, bool]`, not sure why actually, but i think it was just hardcoded in the generator

picklelo (approved):

masenf (approved):

PR #3165, opened 2024-04-25
IconButton for color_mode with nice default and a position props to control it

Lendemor:
```python
class PositionState(rx.State):
    position: str = ""

@rx.page(route="/")
def color_icon_button():
    return rx.vstack(
        rx.select(
            ["top-left", "top-right", "bottom-left", "bottom-right"],
            value=PositionState.position,
            on_change=PositionState.set_position,
        ),
        rx.color_mode.icon_button(position=PositionState.position)
        rx.color_mode.icon_button(position="top-right"),
        rx.color_mode.icon_button(position="top-left"),
        rx.color_mode.icon_button(position="bottom-left"),
        rx.color_mode.icon_button(position="bottom-right", bottom="10rem", right="10rem")
    )

### CODE REVIEW

File: reflex/components/radix/themes/color_mode.pyi [hunk 1]
```diff
@@ -522,9 +524,269 @@ class ColorModeButton(Button):
...
+            **props: The props to pass to the component.
+
+        Returns:
+            The button component.
+        """
+        ...
+
 class ColorModeNamespace(BaseVar):
     icon = staticmethod(ColorModeIcon.create)
     switch = staticmethod(ColorModeSwitch.create)
```
masenf, line 788: i think we should deprecate the switch because it's not really doing much that you can't do with a normal switch

reflex/components/radix/themes/color_mode.pyi
```diff
@@ -522,9 +524,269 @@ class ColorModeButton(Button):
...
     button = staticmethod(ColorModeButton.create)
+    icon_button = staticmethod(ColorModeIconButton.create)
```
masenf, line 790: we should probably just replace `button` with `icon_button`.

In `ColorModeIconButton.create` we can take `*children` and if it's specified, throw a deprecation warning that passing children no longer works and to either use a regular button with `toggle_color_mode` or specify `light_component` and `dark_component` if needed

masenf (commented):
Sorry meant to make my comments on the `.py` not the `.pyi`, but oh well.

It would also be cool if the `ColorModeIcon` could take props and apply them to the icon, although i see that's not really how the code is currently written.

reflex/components/radix/themes/color_mode.py
```diff
@@ -110,69 +89,74 @@ class ColorModeIconButton(IconButton):
...
+            "position": position_values,
+            "left": ["top-left", "bottom-left"],
+            "right": ["top-right", "bottom-right"],
+            "top": ["top-left", "top-right"],
+            "bottom": ["bottom-left", "bottom-right"],
+        }
+
+        # position is used to set nice defaults for positioning the icon button
+        if isinstance(position, Var):
+            (set_var_default("position", "fixed", position),)
```
masenf, line 138: seems this is an unnecessary tuple

reflex/components/radix/themes/color_mode.py
```diff
@@ -110,69 +89,74 @@ class ColorModeIconButton(IconButton):
...
-        if isinstance(position, Var):
-            props.setdefault(
-                "position",
-                cond(find(pos_values, position), "fixed", ""),
-            )
-            props.setdefault(
-                "bottom",
-                cond(find(["bottom-left", "bottom-right"], position), "2rem", ""),
-            )
+        def set_var_default(prop, default1, default2=""):
```
masenf, line 119: These static inline methods and lookup table could be moved outside the component to avoid redefining them

reflex/components/radix/themes/color_mode.py
```diff
@@ -182,9 +166,7 @@ class ColorModeNamespace(BaseVar):
     """Namespace for color mode components."""
 
     icon = staticmethod(ColorModeIcon.create)
-    switch = staticmethod(ColorModeSwitch.create)
```
masenf, line 105: if we're removing switch from the exposed namespace, we should probably remove the `ColorModeSwitch` class as well

reflex/components/radix/themes/color_mode.py
```diff
@@ -55,44 +57,87 @@ def create(
...
+                feature_name="passing children to color_mode.button",
+                reason=", use color_mode_cond and toggle_color_mode instead to build a custom color_mode component",
+                deprecation_version="0.5.0",
+                removal_version="0.6.0",
+            )
+
+        # position is used to set nice defaults for positioning the icon button
+        if isinstance(position, Var):
+            _set_var_default(props, position, "position", "fixed", position)
+            _set_var_default(props, position, "bottom", "2rem")
```
masenf, line 120: might be nice to expose the `"2rem"` offset as a prop as well

reflex/components/radix/themes/color_mode.py
```diff
@@ -55,44 +57,87 @@ def create(
...
+
+def _set_var_default(props, position, prop, default1, default2=""):
+    props.setdefault(
+        prop, cond(_find(position_map[prop], position), default1, default2)
+    )
+
+
+def _set_static_default(props, position, prop, default):
+    if prop in position:
+        props.setdefault(prop, default)
```
masenf, line 86: maybe i should just stop nitpicking this PR, but it just looks like these two could be combined now into a single `_set_default`

```python
def _set_default(props, position, prop, default1, default2=""):
    if isinstance(prop, Var):
        props.setdefault(
            prop, cond(_find(position_map[prop], position), default1, default2)
        )
    else:
        if prop in position:
            props.setdefault(prop, default1)
```

actually w/e lets just ship it 😂

masenf (approved):

reflex/components/radix/themes/color_mode.py
Lendemor: it's a set default so they can still override it by passing `top`, `bottom`, `left`, `right`

reflex/components/radix/themes/color_mode.py
Lendemor: we can always refactor them in a follow up yeah 👍

PR #2954, opened 2024-03-28
Windows uvicorn bug fix 

sid-38:
Closes Issue #2335 

Fixes a bug that causes hot reload to not function properly in Windows with python 3.12. 

### Changes:
- In Windows, python frontend subprocess would be created with the flag CREATE_NEW_PROCESS_GROUP
- After uvicorn function returns, the frontend subprocess is killed using a SIGTERM signal. This is necessary because after CREATE_NEW_PROCESS_GROUP flag, frontend subprocess does not respond to Ctrl+C events.

For more information, check the discussion in the Issue link.

### All Submissions:

- [x] Have you followed the guidelines stated in [CONTRIBUTING.md](https://github.com/reflex-dev/reflex/blob/main/CONTRIBUTING.md) file?
- [x] Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/reflex-dev/reflex/pulls ) for the desired changed?

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

PR #3181, opened 2024-04-28
Windows `--frontend-only`  fix ctrl + c

ElijahAhianyo:
We introduced a workaround(#2954) to fix the `ctrl + c` bug  in windows affecting uvicorn which causes the frontend to exit on hot reload. This fix applied to the case of running the frontend and backend concurrently, however another issue was introduced where `ctrl + c`  does not work at all when running the frontend only. This PR makes sure the frontend is created as a subprocess group independent of its parent ONLY when the backend exists.  This should keep the frontend process in the same group as the parent process and allow SIGINT reach.

### DIFF
--- a/reflex/reflex.py
+++ b/reflex/reflex.py
@@ -212,7 +212,7 @@ def _run(
     # Run the frontend on a separate thread.
     if frontend:
         setup_frontend(Path.cwd())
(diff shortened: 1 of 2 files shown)


masenf:
It looks good, i'll test it later today

masenf (approved):

PR #3186, opened 2024-04-29
add toast component

Lendemor:
```python
class ToastState(rx.State):
    def trigger_toast(self):
        return rx.toast("Hello World 2")

@rx.page(route="/toast", title="Toast")
def index():
    return rx.center(
        rx.color_mode.button(),
        rx.toast.provider(),
        rx.vstack(
            rx.button("Toast", on_click=rx.toast("Hello World 1")),
            rx.grid(
                rx.icon_button(
                    "arrow-up-left", on_click=rx.toast("Hello World", position="top-left"),
                ),
                rx.icon_button(
                    "arrow-up-right", on_click=rx.toast("Hello World", position="top-right"),
                ),
                rx.icon_button(
                    "arrow-down-left", on_click=rx.toast("Hello World", position="bottom-left"),
                ),
                rx.icon_button(
                    "arrow-down-right", on_click=rx.toast("Hello World", position="bottom-right"),
                ),
                columns="2",
            ),
            rx.button("Toast", on_click=ToastState.trigger_toast),
            rx.button("Toast info", on_click=rx.toast.info("Hello World")),
            rx.button("Toast success", on_click=rx.toast.success("Hello World")),
            rx.button("Toast warning", on_click=rx.toast.warning("Hello World")),
            rx.button("Toast error", on_click=rx.toast.error("Hello World")),
         ),
         width="100vw",
         height="100vh",
     )
```


Alek99:
Since we plan on having an implicit toast on every page, can we get rid of needed to add `rx.toast.provider()`

Lendemor:
> Since we plan on having an implicit toast on every page via the new connection banner, can we get rid of needed to add `rx.toast.provider()`

I'm not too sure about that, maybe we can have the new connection banner include it by default, but we should still leave the option for people to use their own toast system too I think.

reflex/components/sonner/toast.py
```diff
@@ -0,0 +1,224 @@
...
+            default=serialize,
+        )
+
+
+class ToastProps(PropsBase):
+    """Props for the toast component."""
+
+    description: str = ""
+    close_button: bool = False
+    invert: bool = False
```
martinxu9, line 52: what does `invert` do?

reflex/components/sonner/toast.py
```diff
@@ -0,0 +1,224 @@
...
+    offset: Var[str]
+
+    # directionality of the toast (default: ltr)
+    dir: Var[str]
+
+    hotkey: Var[str]
+
+    invert: Var[bool]
+
+    toast_options: Var[ToastProps]
```
martinxu9, line 101: is this not used?

reflex/components/sonner/toast.py
```diff
@@ -0,0 +1,224 @@
...
+    theme: Var[str] = color_mode
+
+    # whether to show rich colors
+    rich_colors: Var[bool] = Var.create_safe(True)
+
+    # whether to expand the toast
+    expand: Var[bool] = Var.create_safe(True)
+
+    # the number of toasts that are currently visible
+    visibleToasts: Var[int]
```
martinxu9, line 83: are these supposed to be camel cases? I see another below

picklelo: +1: These should be snake_case

reflex/utils/pyi_generator.py
```diff
@@ -117,6 +117,30 @@ def _get_type_hint(value, type_hint_globals, is_optional=True) -> str:
...
+        if type(None) in value.__args__:
+            res_args = [
+                _get_type_hint(arg, type_hint_globals, rx_types.is_optional(arg))
+                for arg in value.__args__
+                if arg is not type(None)
+            ]
+            if len(res_args) == 1:
+                return f"Optional[{res_args[0]}]"
+            else:
+                print("found union with None")
```
masenf, line 134: remove print 
```suggestion
```

reflex/components/sonner/toast.py
```diff
@@ -0,0 +1,224 @@
...
+    description: str = ""
+    close_button: bool = False
+    invert: bool = False
+    important: bool = False
+    duration: int = 4000
+    position: LiteralPosition = "bottom-right"
+    dismissible: bool = True
+    icon: Optional[Icon] = None
+    action: str = ""
+    cancel: str = ""
```
masenf, line 59: it doesn't seem like the action/cancel actually works, maybe we should remove these for now until we can figure out a way to chain events from them.

reflex/components/sonner/toast.py
```diff
@@ -0,0 +1,224 @@
...
+    """Props for the toast component."""
+
+    description: str = ""
+    close_button: bool = False
+    invert: bool = False
+    important: bool = False
+    duration: int = 4000
+    position: LiteralPosition = "bottom-right"
+    dismissible: bool = True
+    icon: Optional[Icon] = None
```
masenf, line 57: if i pass a lucide icon here, it gives a frontend error and crashes

```
Unhandled Runtime Error
Error: Objects are not valid as a React child (found: object with keys {children, library, lib_dependencies, transpile_packages, tag, style, event_triggers, alias, is_default, key, id, class_name, special_props, autofocus, custom_attrs, State, size, color}). If you meant to render a collection of children, use an array instead.
```

reflex/components/sonner/toast.py
```diff
@@ -0,0 +1,224 @@
...
+        )
+
+
+class ToastProps(PropsBase):
+    """Props for the toast component."""
+
+    description: str = ""
+    close_button: bool = False
+    invert: bool = False
+    important: bool = False
```
masenf, line 53: i'm guessing `important` is supposed to show the toasts on top of any that are already open... but in my case, it pops up behind the other toasts and actually isn't visible at all

masenf (commented):
is there a good way to style the toasts?

i tried passing styles to the `toast.provider`:

```python
rx.toast.provider(
        style={
            "[data-sonner-toast][data-type=warning]": {
                "background_color": f"{rx.color('amber', 7)} !important",
                "_hover": {
                    "background_color": f"{rx.color('amber', 9)} !important",
                },
            },
        }
    )
```

But this is a bit messy and requires the use of `!important`, which is not great, and if we bundle a toast provider in the future for the connection banner, then i don't think this will be exposed outside of global styles or a custom stylesheet. Not sure if it will come up often, but worth thinking about.

reflex/components/sonner/toast.py
Lendemor: It's supposed to be for "dark toast in light mode and vice versa"
doesn't seems to work however.

Lendemor:
> is there a good way to style the toasts?
> 
> i tried passing styles to the `toast.provider`:
> 
> ```python
> rx.toast.provider(
>         style={
>             "[data-sonner-toast][data-type=warning]": {
>                 "background_color": f"{rx.color('amber', 7)} !important",
>                 "_hover": {
>                     "background_color": f"{rx.color('amber', 9)} !important",
>                 },
>             },
>         }
>     )
> ```
> 
> But this is a bit messy and requires the use of `!important`, which is not great, and if we bundle a toast provider in the future for the connection banner, then i don't think this will be exposed outside of global styles or a custom stylesheet. Not sure if it will come up often, but worth thinking about.

`toast` will support `style` props so it'll be simpler

```python
rx.button(
    "Toast",
    on_click=rx.toast(
        "Hello World 1",
        style={"background": "red", "color": "white"},
    ),
),
```

reflex/components/sonner/toast.py
Lendemor: `# Control the sensitivity of the toast for screen readers`

picklelo:
Can we allow passing in the style props as kwargs here as well:

```python
rx.toast(
    "Hello World 1",
    color="red",
)
```

reflex/components/sonner/toast.py
Lendemor: It is.

```python
rx.toast.provider(
    toast_options=rx.toast.options(
        style={"background": "red", "color": "white"}
    ),
),
```
will make all toast have this default style.

Lendemor:
> Can we allow passing in the style props as kwargs here as well:
> 
> ```python
> rx.toast(
>     "Hello World 1",
>     color="red",
> )
> ```

Can we add in a follow up? I don't think this change is trivial.

reflex/components/sonner/toast.py
```diff
@@ -0,0 +1,267 @@
...
+    dir: Var[str]
+
+    # Keyboard shortcut that will move focus to the toaster area.
+    hotkey: Var[str]
+
+    # Dark toasts in light mode and vice versa.
+    invert: Var[bool]
+
+    # These will act as default options for all toasts. See toast() for all available options.
+    toast_options: Var[ToastProps]
```
masenf, line 136: When i set defaults here, they do not seem to be respected.

I tried setting it like

```python
rx.toast.provider(toast_options=rx.toast.options(close_button=True, duration=300), close_button=True)
```

And the toasts still use the default 4000ms duration and do not have close buttons, unless i pass no options at all.

For example, i'm trying to toast like this

```python
                    yield rx.toast.warning(
                        f"Removed {item.text} from {list_name}",
                        description="Whoops, this was irreversible"
                    )
```

It seems that any options passed myself overwrite all of the toaster options with the original defaults.

masenf: so i updated ToastProps to make all of the fields optional, which seemed to help; but now i'm finding that some of the fields in `ToastProps` don't take effect when passed to the `toast_options` prop.

- description
- position
- close_button
(maybe others, didn't try them all)

masenf (commented):
Posted #3216 to get the action and cancel working, which i think are kind of essential for using toasts effectively

It also fixes some of the issues mentioned below

masenf (approved):

PR #3196, opened 2024-04-30
typed mixins and ComponentState

benedikt-bartscher:


benedikt-bartscher:
Tests are failing due to the `reflex_chat` example app which uses `ComponentState` without the new `mixin` `__init_subclass__` kwarg.

masenf:
It seems like this change requires all `ComponentState` subclasses to now pass `mixin=True` in the class definition. Is that right? What is the advantage of that?

benedikt-bartscher:
@masenf yes it currently does as mentioned in my last comment. Maybe we can find a way to auto-set this for ComponentState.

The advantage is improved typing. Currently mixins do not inherit from `rx.State` directly which leads to a bad dev experience. Consider this example:

```python
class Mixin:
  def reset_state(self):
    self.reset()

  def do_smth_with_router(self):
    print(self.router)
```

With this PR you can do this, which passes all the type checks:

```python
class Mixin(rx.State, mixin=True):
  def reset_state(self):
    self.reset()

  def do_smth_with_router(self):
    print(self.router)
```

benedikt-bartscher:
> Maybe we can find a way to auto-set this for ComponentState.

@masenf I just pushed a commit, which implements implicit `mixin=True` for `ComponentState` subclasses

masenf (dismissed):
nice

masenf (approved):

PR #3197, opened 2024-04-30
Clean up config and app API

picklelo:
Begin moving things to private methods when we don't want to expose them, and removing things that were deprecated. This is with the goal of getting our API reference to be higher quality.


reflex/app.pyi
Lendemor, line 1: Shouldn't there be more change in this file with how many functions have been renamed in `app.py` ?

masenf (commented):
This seems to revert a recent PR to remove `socket_app` and respect the user's `sio` kwarg

reflex/app.pyi
picklelo: I removed the app.pyi completely now

reflex/app.py
```diff
@@ -102,60 +102,78 @@ class OverlayFragment(Fragment):
...
-    # The Socket.IO AsyncServer.
-    sio: Optional[AsyncServer] = None
-
-    # The state class to use for the app.
-    state: Optional[Type[BaseState]] = None
+    app = rx.App(
+        # Set global level style.
+        style={...},
+        # Set the top level theme.
+        theme=rx.themes.theme(accent_color="blue"),
```
masenf, line 120:
```suggestion
        theme=rx.theme(accent_color="blue"),
```

reflex/app.py
```diff
@@ -102,60 +102,78 @@ class OverlayFragment(Fragment):
...
-    load_events: Dict[str, List[Union[EventHandler, EventSpec]]] = {}
-
-    # Admin dashboard
-    admin_dash: Optional[AdminDash] = None
+    # A list of URLs to [stylesheets](https://reflex.dev/docs/styling/custom-stylesheets) to include in the app.
+    stylesheets: List[str] = []
 
-    # The async server name space
-    event_namespace: Optional[EventNamespace] = None
+    # A component that is present on every page.
```
masenf, line 134:
```suggestion
    # A component that is present on every page (defaults to the Connection Error banner).
```

And showing the default here in the API ref doesn't seem to be that helpful because it renders as like `<function ...>`. It would be awesome if we could get it to render like "rx.fragment(rx.connection_pulser(), rx.connection_modal())"; though not sure the best way at the moment

picklelo: Yeah I have a follow up ticket to improve the actual rendering of everything on the API reference page.

masenf (approved):
sorry, thought i marked this earlier...

PR #3198, opened 2024-04-30
[REF-2370] Combine library/version into ImportVar and replace _get_imports with _get_imports_list

masenf:
Refactor the method for specifying ImportVar to a more flat format which is easier to combine and work with the intermediate format. No need to `merge_imports`, just append the new `ImportVar` to the list.

Combination occurs in the new `ImportList.collapse` function which handles version conflict detection and combination of other metadata that doesn't influence the identity of the ImportVar, like `install`, `render`, and `transpile`

--------------

2024-04-30: converted to draft to focus on component public API first.

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -627,7 +627,7 @@ class App(Base):
         Example:
             >>> get_frontend_packages({"react": "16.14.0", "react-dom": "16.14.0"})
         """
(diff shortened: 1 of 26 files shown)


masenf:
too much conflict. if we want this, we'll have to redo it

PR #3200, opened 2024-04-30
custom component install command to install current project in editable mode

martinxu9:
## Summary
- Add custom component install command to install current project in editable mode


Lendemor (approved):

Lendemor:
Nice, it will be really convenient if multiples people want to work on a same component together.

PR #3202, opened 2024-05-01
add_style api

Lendemor:
code to test accordion, who use the new API along with `data-` vars from radix-ui
```python

items = [
    rx.accordion.item(header="Item 1", content=rx.text("Item 1 content")),
    rx.accordion.item(header="Item 2", content=rx.text("Item 2 content")),
    rx.accordion.item(header="Item 3", content=rx.text("Item 3 content")),
]


@rx.page()
def index():
    return (
        rx.vstack(
            rx.hstack(
                rx.text("soft"),
                rx.text("classic"),
                rx.text("outline"),
                rx.text("ghost"),
                rx.text("surface"),
                spacing="9",
            ),
            rx.hstack(
                rx.accordion.root(*items, variant="soft"),
                rx.accordion.root(*items, variant="classic"),
                rx.accordion.root(*items, variant="outline"),
                rx.accordion.root(*items, variant="ghost"),
                rx.accordion.root(*items, variant="surface"),
            ),
            align="center",
            justify="between",
        ),
    )
```


Lendemor:
Need the pyi generator changes from #3186 before merging this

reflex/components/component.py
```diff
@@ -779,44 +781,117 @@ def validate_children(children):
...
+    def _add_style(self) -> Style:
+        """Call add_style for all bases in the MRO.
+
+        Downstream components should NOT override. Use add_style instead.
+
+        Returns:
+            The style to add.
         """
-        self.style.update(style)
+        _styles = []
```
picklelo, line 803: nit: don't need _ prefix on local vars inside a method

reflex/components/component.py
```diff
@@ -779,44 +781,117 @@ def validate_children(children):
...
+
+        # 2. User-defined style from `App.style`.
+        component_style = self._get_component_style(style)
+        if component_style:
+            _new_style.update(component_style)
+            _style_vars.append(component_style._var_data)
+
+        # 3. User-defined style from `Component.style`.
+        # Apply theme for retro-compatibility with deprecated _apply_theme API
+        if type(self)._apply_theme != Component._apply_theme:
```
picklelo, line 872: Is this mostly for our own internal code? I don't think any outside component uses this method since it didn't really do anything before.

reflex/components/component.py
```diff
@@ -779,44 +781,117 @@ def validate_children(children):
...
-            # Only add style props that are not overridden.
-            component_style = {
-                k: v for k, v in component_style.items() if k not in self.style
-            }
+        # 1. Default style from `_add_style`/`add_style`.
+        if type(self)._add_style != Component._add_style:
+            raise UserWarning(
+                "Do not override _add_style directly. Use add_style instead."
+            )
+        _new_style = self._add_style()
```
picklelo, line 861: same nit, don't need _ prefix here

reflex/components/component.py
```diff
@@ -779,44 +781,117 @@ def validate_children(children):
...
+        if type(self)._apply_theme != Component._apply_theme:
+            console.deprecate(
+                f"{self.__class__.__name__}._apply_theme",
+                reason="use add_style instead",
+                deprecation_version="0.5.0",
+                removal_version="0.6.0",
+            )
+            self._apply_theme(theme)
+
+        _new_style.update(self.style)
```
masenf, line 881:
```suggestion
        # 4. style dict and style props passed to the component instance.
        _new_style.update(self.style)
```

I think indicating that the props always override any component default styles, app styles, and component class styles is important for future readers; this is easy to gloss over, but very important line.

Lendemor: good call

reflex/components/component.py
```diff
@@ -779,44 +781,119 @@ def validate_children(children):
 
         return cls(children=children, **props)
 
-    def _add_style(self, style: dict):
-        """Add additional style to the component.
+    def add_style(self) -> Style | None:
```
picklelo, line 784: Is `None` a valid return type? I think it may mess up the `_add_style`, let's just only return `Style`

picklelo (approved):
looks good to me

reflex/components/component.py
Lendemor: returning None is already handled in `_add_style`?

picklelo: you're right my mistake

masenf (approved):

PR #3203, opened 2024-05-01
[REF-2523] Implement new public Component API

masenf:
add_imports (dict str: list of str or importvar)
add_hooks (list of str)
add_custom_code (list of str)

These functions are treated specially in that they don't have to call `super()`, ever. Internally the Component will walk up all of the base Component classes and call each `add_*` method and collect the return values.

All the downstream user has to do is implement these add_* functions and return the value they want to add.

### DIFF
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -235,7 +235,7 @@ class Component(BaseComponent, ABC):
         ```python
         return {
             "react": "useEffect",
-            "react-draggable": ["DraggableCore", ImportVar(tag="Draggable", is_default=True)],
+            "react-draggable": ["DraggableCore", rx.ImportVar(tag="Draggable", is_default=True)],
         }
         ```
         """
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -213,6 +213,63 @@ class Component(BaseComponent, ABC):
     # State class associated with this component instance
     State: Optional[Type[reflex.state.State]] = None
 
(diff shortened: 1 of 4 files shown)


reflex/components/component.py
```diff
@@ -213,6 +213,63 @@ class Component(BaseComponent, ABC):
...
+        Implementations do NOT need to call super(). The result of calling
+        add_imports in each parent class will be merged internally.
+
+        Returns:
+            The additional imports for this component subclass.
+        """
+        return {}
+
+    def add_hooks(self) -> list[str]:
+        """Add hooks inside the component function.
```
picklelo, line 230: Maybe this can include an example of how it's used in practice (along with the other ones) for the autogenerated API docs

reflex/components/component.py
```diff
@@ -932,6 +989,30 @@ def _has_event_triggers(self) -> bool:
                     return True
         return False
 
+    @classmethod
+    def _iter_parent_classes_with_method(cls, method: str) -> Iterator[Type[Component]]:
```
picklelo, line 993: Does this affect compile time noticeably at all?

picklelo (dismissed):

reflex/components/component.py
masenf: on reflex-web, there is no change in compile time 😅, i expect this iter operation to be quite fast; although it does have to look at all parent classes in the mro chain, it doesn't actually call the method unless it's been defined, and it only calls each unique method once.

picklelo (dismissed):

picklelo (approved):

PR #3206, opened 2024-05-01
Independent States

### LINKED ISSUE #3204: Independent States
**Describe the bug**
All computed vars of all states get computed for every event handler execution. Even if the state class is not used at all in your reflex app.

**To Reproduce**
```python
import reflex as rx

class StateOne(rx.State):
    @rx.var
    def computed(self) -> int:
        print(f"{type(self).__name__}.computed")
        return 1

    def do_nothing(self):
        print(f"{type(self).__name__}.do_nothing")


class StateTwo(rx.State):
    @rx.var
    def computed(self) -> int:
        print(f"{type(self).__name__}.computed")
        return 2

    def do_nothing(self):
        print(f"{type(self).__name__}.do_nothing")


def one() -> rx.Component:
    return rx.center(
        rx.button("Do nothing", on_click=StateOne.do_nothing),
        rx.text(StateOne.computed),
    )


def two() -> rx.Component:
    return rx.center(
        rx.button("Do nothing", on_click=StateTwo.do_nothing),
        rx.text(StateTwo.computed),
    )


app = rx.App()
app.add_page(one, route="/one")
# app.add_page(two, route="/two")
```

log output from pressing the "Do nothing" button:

```log
StateOne.do_nothing
StateOne.computed
StateTwo.computed
```

**Expected behavior**
`StateTwo` does not get instantiated and `StateTwo.computed` does not get called for every event handler execution in other states.

 - Python Version: 3.12.3
 - Reflex Version: main branch
 - OS: arch

benedikt-bartscher:
Closes #3204

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -1620,6 +1620,8 @@ class BaseState(Base, ABC, extra=pydantic.Extra.allow):
         # Recursively find the substate deltas.
         substates = self.substates
         for substate in self.dirty_substates.union(self._always_dirty_substates):
+            if not substates[substate]._get_was_touched():
+                continue
             delta.update(substates[substate].get_delta())
 
         # Format the delta.
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -1464,18 +1464,18 @@ class BaseState(Base, ABC, extra=pydantic.Extra.allow):
         Returns:
             The valid StateUpdate containing the events and final flag.
         """
-        # get the delta from the root of the state tree
+        # Get the delta after processing the event.
+        delta = self.get_delta(include_substates=True)
         state = self
         while state.parent_state is not None:
             state = state.parent_state
+            delta.update(state.get_delta(include_substates=False))
 
         token = self.router.session.client_token
 
         # Convert valid EventHandler and EventSpec into Event
         fixed_events = fix_events(self._check_valid(handler, events), token)
 
-        # Get the delta after processing the event.
-        delta = state.get_delta()
         state._clean()
 
         return StateUpdate(
@@ -1589,7 +1589,7 @@ class BaseState(Base, ABC, extra=pydantic.Extra.allow):
             )
         return fetch_substates
 
-    def get_delta(self) -> Delta:
+    def get_delta(self, include_substates: bool = True) -> Delta:
         """Get the delta for the state.
 
         Returns:
@@ -1618,11 +1618,12 @@ class BaseState(Base, ABC, extra=pydantic.Extra.allow):
             delta[self.get_full_name()] = subdelta
 
         # Recursively find the substate deltas.
-        substates = self.substates
-        for substate in self.dirty_substates.union(self._always_dirty_substates):
-            if not substates[substate]._get_was_touched():
-                continue
-            delta.update(substates[substate].get_delta())
+        if include_substates:
+            substates = self.substates
+            for substate_name in self.dirty_substates.union(
+                self._always_dirty_substates
+            ):
+                delta.update(substates[substate_name].get_delta())
 
         # Format the delta.
         delta = format.format_state(delta)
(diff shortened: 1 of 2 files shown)


masenf:
i think we need #2429 to solve this properly. the `_was_touched` marker is reset for every event (in redis), just because a particular substate wasn't touched, doesn't necessarily mean that it doesn't have computed vars that depend on vars in the parent states that may have changed.

benedikt-bartscher:
@masenf thanks for looking into this. While my issue description with the unused page looks like it is related to #2429 it really isn't. This is a completely different issue. Maybe my last commit makes things more clear?

benedikt-bartscher:
If my assumptions are correct, this should improve event handling performance for reflex-web quite decent. There are over 20 `ComputedVar`s in reflex-web, even one which establishes a database connection. With reflex's current behavior, this should cause every event to recompute all those `ComputedVars`.

masenf:
```python
import reflex as rx


class StateA(rx.State):
    a: str = ""

    async def handle_click(self):
        self.a = "foo"
        state_b = await self.get_state(StateB)
        state_b.b = "bar"


class StateB(rx.State):
    b: str = ""


def index() -> rx.Component:
    return rx.vstack(
        rx.hstack(
            StateA.a,
            StateB.b
        ),
        rx.button("Click me", on_click=StateA.handle_click)
    )


app = rx.App()
app.add_page(index)
```

In reflex `main` branch, the page displays "foobar" after clicking the button.

With this proposed patch, the page only displays "foo".

This is primarily why the delta is calculated from the root state, so that any states which are patched into the tree during processing of an event have their deltas also calculated.

I definitely see what you're trying to do here; `ComputedVar` is re-calculated in a somewhat overzealous manner, but the reason for that is to retain the semantics of ensuring they always have the latest value after every event. My main motivation for introducing the `@rx.cached_var` decorator last year, was to avoid the cost of recomputing these vars when the vars they depend on have not been changed.

If I just have an `@rx.var` decorated function, reflex has no way of knowing what it depends on. Since it's in a state class that is ultimately a descendent of `rx.State`, it might depend on router vars or something else that could change when an event is processed. I don't see a way around recomputing it without opting in to `@rx.cached_var` semantics... so maybe it's time to make `@rx.var` be a cached var by default?

benedikt-bartscher:
Thanks, I completely missed `get_state`. I already thought about the changing router vars...
Shouldn't it be possible to extend the computed var dependency tracking to respect `get_state`?

masenf:
> Shouldn't it be possible to extend the computed var dependency tracking to respect `get_state`?

Thankfully because computed vars use property semantics, they can only access vars in their immediate ancestors (without using hacks that we do not support). So the dependency tracking is already working i think. The problem is that for regular `@rx.var` we don't do any dependency tracking; an `@rx.var` is always considered dirty by definition.

benedikt-bartscher:
Thanks a lot, I should have looked into the dependency tracking implementation earlier...
If the tracking works fine, is there any reason not promoting it to the default?

masenf:
> If the tracking works fine, is there any reason not promoting it to the default?

There have been several bugs initially in the dependency tracking that have since been fixed. It's possible that other bugs could be lurking, but i have not uncovered them yet and use `@rx.cached_var` in all of my projects. Initially the intent was to have cached var be the default behavior, but we were hesitant to bring it in at first without some more bake time.

I think it's ready to become the default, but that means we're relying on users to report bugs that might pop up related to computed vars in the next release.

Additionally, it notably cannot track vars outside of `self`, so if someone is using a computed var to access/render data that originates from outside the state, a cached_var will essentially never update in that case.

benedikt-bartscher:
I see, makes sense. Considering the possible performance impacts of `@rx.var` I would support changing the default to `_cache = True`. I somehow was convinced that `@rx.var`s only should get updated if the associated state is dirty. But as you pointed out nicely, it's always considered dirty if it has an `@rx.var` defined.

I just migrated some codebases to cached var. If any problems arise, I will report them as usual.

PR #3211, opened 2024-05-01
Allow StatefulComponent to also be rendered via __str__

masenf:
This makes it easier to implement React "render functions" when wrapping components that require children to be a callable.


picklelo (approved):
nice

PR #3214, opened 2024-05-02
add module prefix to generated state names

benedikt-bartscher:
This PR prefixes all state names with their python module to prevent naming conflicts. Defining the same State class name multiple times in different modules is now possible. Consider this example:

myapp/pages/index.py
```python
class State(rx.State):
  pass
```

myapp/pages/about.py
```python
class State(rx.State):
  pass
```

Adjusted all tests except the actual `State.get_name` and `State.get_full_name` tests to use dynamic state names. This allows us to implement minified state names [later](https://github.com/reflex-dev/reflex/pull/3226).

Added `get_state_name` and `get_full_state_name` helpers to `AppHarness` for easy testing with parametrized app_names.

Validate state cls module names in init_subclass hook. They should not contain `___` (3x) as it's used as state name module separator.


masenf:
@benedikt-bartscher i fixed the one broken integration test that came from the `origin/main` refresh.

I also started a PR over on reflex-web to address the duplicate state names in the markdown files causing issues with this new module-prefix naming scheme. There are still a few errors that i need to work out in a couple files, but 3/5 of the failing pages are now compiling correctly.

Hopefully we can get this one in soon.

benedikt-bartscher:
@masenf I just fixed the reflex-web integration tests in https://github.com/reflex-dev/reflex/pull/3214/commits/e863da64342b7a8f3b0e83287715762560701542

masenf (approved):
looks like this is good to go.

the state names are much longer now and since they compile out to strings in the JS code that will increase the bundle size, since minifiers cannot munge strings; but i think the followup PR to minify state names within reflex will probably result in an overall decrease in the resulting code size, so i think this is the right move.

benedikt-bartscher:
Thanks for the review. I'll rebase the minified state name pr this week. Should be a relatively small one.

PR #3216, opened 2024-05-02
Get `action`, `cancel`, `on_dismiss` and `on_auto_close` working for rx.toast

masenf:
Handle callbacks into reflex event handlers.

Respect defaults set in ToastProvider toast_options when firing a toast with it's own ToastProps set.

Create new `reflex.utils.format.format_queue_events` helper method that can be used to take arbitrary eventhandler/eventspec/lambda along with an args spec, and create a callback function for use with `rx.call_script`-based APIs.


reflex/components/sonner/toast.py
```diff
@@ -27,6 +33,50 @@
 toast_ref = Var.create_safe("refs['__toast']")
 
 
+class ToastAction(Base):
+    """A toast action; rendered a button in the toast."""
```
Lendemor, line 37:
```suggestion
    """A toast action that render a button in the toast."""
```

reflex/components/sonner/toast.py
```diff
@@ -27,6 +33,50 @@
...
+        for spec in on_click_specs:
+            if isinstance(spec, EventHandler):
+                specs = [call_event_handler(spec, lambda: [])]
+            elif isinstance(spec, type(lambda: None)):
+                specs = call_event_fn(spec, lambda: [])
+            else:
+                specs = [spec]
+            events.extend(format.format_event(s) for s in specs)
+        on_click = Var.create(
+            f"() => {{queueEvents([{','.join(events)}], socket); processEvent(socket)}}",
```
Lendemor, line 67: maybe this could be exposed as a formatting method that user that wrap component can use?

reflex/components/sonner/toast.py
```diff
@@ -27,6 +33,50 @@
...
+            _var_is_string=False,
+            _var_is_local=False,
+        )
+    else:
+        on_click = Var.create("() => null", _var_is_string=False, _var_is_local=False)
```
Lendemor, line 72: If we can pass this kind of var here, it would probably be possible to allow it for built-in event triggers too right ?

Lendemor (dismissed):

reflex/components/sonner/toast.py
masenf: Little known feature; we actually _do_ allow it if you set the `_var_type` to `EventChain`, it just gets rendered as is 

https://github.com/reflex-dev/reflex/blob/db47d39979ac62ad0571dc4aca9f00e90b9ee2d1/reflex/components/component.py#L497-L499

https://github.com/reflex-dev/reflex/blob/db47d39979ac62ad0571dc4aca9f00e90b9ee2d1/reflex/utils/format.py#L572-L581

reflex/components/sonner/toast.py
masenf: moved to `rx.utils.format.format_queue_events`

i think this will be really useful for any `rx.call_script`-based API that needs to pass callback functions that trigger reflex events, so good to formalize it.

and since i formalized it, it was also easy to add in `on_dismiss` and `on_auto_close` callbacks too 😎

masenf:
@Lendemor idk if you want ot take this into your branch (it has a refresh from `origin/main`); or if you just want to take your PR into `main` and then i can rebase this one and bring it into `main` separately.

Lendemor:
> @Lendemor idk if you want to take this into your branch (it has a refresh from `origin/main`); or if you just want to take your PR into `main` and then i can rebase this one and bring it into `main` separately.

I'm fine with both, let's just go with whatever is more convenient.

reflex/components/sonner/toast.py
masenf: oh nice, i realized i was also using a variant of this code inside `call_script` itself, so i can clean that up too 😄

reflex/components/sonner/toast.py
Lendemor: I'll have to remember that one 👍

reflex/components/sonner/toast.py
```diff
@@ -27,6 +30,29 @@
...
+    Returns:
+        The serialized toast action with on_click formatted to queue the given event.
+    """
+    return {
+        "label": action.label,
+        "onClick": format.format_queue_events(action.on_click),
+    }
+
+
 class PropsBase(Base):
```
Lendemor, line 56: I think `PropsBase` should be exposed under `_x` too.
I've been playing with some libs that make extensive use of hooks, and the feature from this would be needed for them too.

Maybe in its own file since the code for it is not specific to sonner/toast

Lendemor (changes requested):

Lendemor (approved):

PR #3218, opened 2024-05-03
Fix es6 escaping

abulvenz:
I had some problems when using ES6 Code in the script tag. 
So I wrote a test that shows the problem. My futile approach to fix it duplicates some code and breaks many other tests.
Without the test passing using the script tag from the script within an example app results in the following error:
image


abulvenz:
I think this is not relevant anymore.

PR #3219, opened 2024-05-03
update blank template

Lendemor:


masenf (commented):
Is the content supposed to be left aligned now? i actually think it looks kinda cool that way.

The other issue is in chrome there is always a scroll bar now, regardless of the window height (on my screen anyway). `min_height="88vh"` or something a bit smaller might help there.

Lendemor:
> Is the content supposed to be left aligned now? i actually think it looks kinda cool that way.
> 
> The other issue is in chrome there is always a scroll bar now, regardless of the window height (on my screen anyway). `min_height="88vh"` or something a bit smaller might help there.

That's the intention yeah.

I can tweak the value for min_height yeah, didn't happens on my screen.

reflex/.templates/apps/blank/code/blank.py
```diff
@@ -11,24 +11,31 @@
...
-            rx.text("Get started by editing ", rx.code(filename)),
-            rx.button(
+            rx.text("Get started by editing ", rx.code(filename), size="5"),
+            rx.link(
                 "Check out our docs!",
-                on_click=lambda: rx.redirect(docs_url),
-                size="4",
+                href=docs_url,
+                is_external=True,
+                style={
```
picklelo, line 28: I feel we shouldn't have styles if not needed (just use the default) - or if we do, we should pass them as keyword args rather than a dict.

reflex/.templates/apps/blank/code/blank.py
```diff
@@ -11,24 +11,31 @@
...
-        rx.theme_panel(),
+    # Welcome Page (Index)
+    return rx.container(
+        rx.color_mode.button(position="top-right"),
         rx.vstack(
             rx.heading("Welcome to Reflex!", size="9"),
-            rx.text("Get started by editing ", rx.code(filename)),
-            rx.button(
+            rx.text("Get started by editing ", rx.code(filename), size="5"),
+            rx.link(
```
picklelo, line 24: Can we put the button inside the link, so we don't need the style props?

reflex/.templates/apps/blank/code/blank.py
```diff
@@ -11,24 +11,31 @@
 class State(rx.State):
     """The app state."""
 
+    ...
+
 
 def index() -> rx.Component:
-    return rx.center(
-        rx.theme_panel(),
```
picklelo, line 17: Should we include the theme_panel?

reflex/.templates/apps/blank/code/blank.py
masenf: oh yeah, that's really nice!

```python
            rx.link(
                rx.button("Check out our docs!"),
                href=docs_url,
                is_external=True,
            ),
```

masenf (changes requested):
while we're touching this, can we rearrange the imports too?

I'd rather see

```python
import reflex as rx

from rxconfig import config

filename = f"{config.app_name}/{config.app_name}.py"
```

1. this is the preferred pep8-style import order, local modules should be listed after installed modules
2. this makes it easier to delete without removing the reflex import

-------------

i also think we should just inline the `docs_url`, instead of defining it at the top of the module

Lendemor:
> while we're touching this, can we rearrange the imports too?
> 
> I'd rather see
> 
> ```python
> import reflex as rx
> 
> from rxconfig import config
> 
> filename = f"{config.app_name}/{config.app_name}.py"
> ```
> 
> 1. this is the preferred pep8-style import order, local modules should be listed after installed modules
> 2. this makes it easier to delete without removing the reflex import
> 
> i also think we should just inline the `docs_url`, instead of defining it at the top of the module

I tried, but we literally can't, because reflex is considered as a local, ruff change the import back to how they are at the moment.

benedikt-bartscher:
@Lendemor maybe you can add a ruff config overwrite for the template dir: https://docs.astral.sh/ruff/configuration/#config-file-discovery

maybe smth like this (untested):
```toml
[tool.ruff]
# Extend the `pyproject.toml` file in the parent directory...
extend = "../pyproject.toml"

[tool.ruff.isort]
known-first-party = []
```

Lendemor:
> @Lendemor maybe you can add a ruff config overwrite for the template dir: https://docs.astral.sh/ruff/configuration/#config-file-discovery
> 
> maybe smth like this (untested):
> 
> ```toml
> [tool.ruff]
> # Extend the `pyproject.toml` file in the parent directory...
> extend = "../pyproject.toml"
> 
> [tool.ruff.isort]
> known-first-party = []
> ```

Thanks for the suggestion, we went with excluding the rule `I001` for only `blank.py`, should be good enough since all the other templates live in their own repo and won't have that problem.

picklelo (approved):

masenf (approved):

PR #3220, opened 2024-05-03
External assets

abulvenz:
When referencing external modules, the files for the asset folder should be somehow shippable. This is a proposal for a simple helper function that copies the files to the assets folder of the app that is currently compiled.

```python
# library_module.index.py

# In the library module reference the 
# asset files with the relative file path to the including script.
@rx.page("/")
def index() -> rx.Component:
    return rx.center(
        rx.image(src=rx.asset("test.png",dir="subfolder"))
    )
```
During compilation those files are copied to `assets/external/library_module/index/subfolder/test.png`.

In case you want to make assets from your library includable for calling modules, you can expose them like this:
```python
class Assets:
    test_png = rx.asset("test.png", "hang_it_up")
    script_js = rx.asset("my_component_logic.js")
 ```

### CODE REVIEW

**Comment by benedikt-bartscher** [comment]
@abulvenz how about symlinking instead of copying?

File: reflex/app.py [hunk 1]
```diff
@@ -1317,3 +1320,40 @@ async def on_ping(self, sid):
...
+    cwd = Path.cwd()
+
+    module = inspect.getmodule(inspect.stack()[1][0])
+    assert module is not None
+
+    caller_module_path = module.__name__.replace(".", "/")
+    dir = caller_module_path if dir == "." else caller_module_path + "/" + dir
+    assets = constants.Dirs.APP_ASSETS
+    external = constants.Dirs.EXTERNAL_APP_ASSETS
+    Path(Path(cwd) / assets / external / dir).mkdir(parents=True, exist_ok=True)
```
benedikt-bartscher, line 1354: Create a variable for `Path(Path(cwd) / assets / external / dir)` first and use it in the next lines. Also check if the source exists with pathlib's `.resolve()` method

Lendemor:
> @abulvenz how about symlinking instead of copying? Edit: Not sure how symlinks behave on Windows

Symlink doesn't work on Windows, we had this in the beginning for syncing between `assets` and `.web/public` folder, but had to remove it for Windows.

reflex/app.py
abulvenz: @Lendemor I hope you are fine with using symlinks here, since we are not linking directories. That seems to be the only difference. https://docs.python.org/3/library/pathlib.html#pathlib.Path.symlink_to

@benedikt-bartscher I just tested what happens when `src_file` is not present. It already complains `No such file or directory`. OK, or wrap it into a an `AssetNotPresent`?

benedikt-bartscher: @abulvenz awesome, if there is already a `FileNotFound` exception, I think we are fine, no additional wrapping needed

abulvenz:
> Symlink doesn't work on Windows, we had this in the beginning for syncing between `assets` and `.web/public` folder, but had to remove it for Windows.

@Lendemor Should I change it back? :mechanic: 

Is anything else missing here?

reflex/app.py
Lendemor: @abulvenz Sorry for delay in review, was on break for a few days.

 It's passing windows CI so it should be fine I think.

I'm not sure about adding it inside `app.py` however, so could you move it to it's own file? Maybe inside `reflex/experimental` so we'll get some time to test it in real situations before making it official.

We'll also need to add unit tests for it.

abulvenz: Thank you @Lendemor, After v0.5.0 everybody earned a small break. I will move it to experimental and try to add a unit test.

reflex/experimental/assets.py
```diff
@@ -0,0 +1,57 @@
...
+
+    assets = constants.Dirs.APP_ASSETS
+    external = constants.Dirs.EXTERNAL_APP_ASSETS
+
+    if not src_file.exists():
+        raise FileNotFoundError(f"File not found: {src_file}")
+
+    # Create the asset folder in the currently compiling app.
+    cwd = Path.cwd()
+    asset_folder = Path(Path(cwd) / assets / external / subfolder)
```
benedikt-bartscher, line 48: nit pick: the outer `Path` is not needed

abulvenz: @benedikt-bartscher  Thanks for reviewing! I changed this line.

abulvenz:
@Lendemor Is it possible that I cancelled your approval by merging main? Sorry. :bowl_with_spoon:

benedikt-bartscher (approved):

PR #3221, opened 2024-05-03
prevent shadowing

Lendemor:
```python

class TestState(rx.State):
    val: bool = True

    @rx.var
    def val(self) -> bool:
        return True


@rx.page(route="/", image="settings.ico")
def index():
    return rx.container(
        rx.text(TestState.val),
    )
```

The above code will now raise an error
image

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -450,6 +450,8 @@ class BaseState(Base, ABC, extra=pydantic.Extra.allow):
         super().__init_subclass__(**kwargs)
         # Event handlers should not shadow builtin state methods.
         cls._check_overridden_methods()
+        # Computed vars should not shadow builtin state props.
+        cls._check_overriden_basevars()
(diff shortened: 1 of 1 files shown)


masenf (dismissed):

reflex/state.py
```diff
@@ -696,6 +698,19 @@ def _check_overridden_methods(cls):
...
             )
 
+    @classmethod
+    def _check_overriden_basevars(cls):
+        """Check for shadow base vars and raise error if any.
+
+        Raises:
+            NameError: When a computed var shadows a base var.
+        """
+        for name, value in cls.__dict__.items():
```
benedikt-bartscher, line 708: how about just using `_get_computed_vars` instead? It will check for conflicts in mixins as well

picklelo (approved):

PR #3227, opened 2024-05-04
[REF-3009] type transforming serializers

benedikt-bartscher:


benedikt-bartscher:
Instead of returning a tuple with the serialized type we could also add a kwarg to the `serialize` decorator.

```python
@serialize(to=str)
def serialize_datetime(dt: datetime):
    return str(dt)
```

benedikt-bartscher (commented):
@masenf your changes look good to me, thank you for jumping in!
I am thinking about dropping the `tuple` return in favor of the new `to` kwarg. If I am not missing any advantages of the `tuple` return, I think it would be better to have one way to do things.
The tuple approach would allow dynamic serialization types (type based on input value), but we can't use them in the frontend, so I see no real benefit.

benedikt-bartscher (commented):
@masenf thanks for refactoring, looks good to me!

masenf (approved):

PR #3231, opened 2024-05-04
Explicit deps and interval for computed vars

benedikt-bartscher:
```python
from datetime import datetime, timedelta

import reflex as rx


class OtherState(rx.State):
    def do_nothing(self) -> None:
        print("Doing nothing")


class State(rx.State):
    @rx.cached_var(interval=5)
    def time1(self) -> str:
        return datetime.now().strftime("%H:%M:%S")

    # same as above, different notation
    @rx.var(interval=5, cache=True)
    def time2(self) -> str:
        return datetime.now().strftime("%H:%M:%S")

    # explicit disabled auto_deps and timedelta interval
    @rx.var(interval=timedelta(minutes=5), cache=True, auto_deps=False)
    def time3(self) -> str:
        # this will not add deps, because auto_deps is False
        print(self.time1)
        print(self.time2)

        return datetime.now().strftime("%H:%M:%S")

    # explicit dependency on time var
    @rx.cached_var(deps=[time1])
    def depends_on_time1(self) -> str:
        return datetime.now().strftime("%H:%M:%S")

    # same as above
    @rx.cached_var(deps=["time1"])
    def depends_on_time1_alt(self) -> str:
        return datetime.now().strftime("%H:%M:%S")

    def do_nothing(self) -> None:
        print("Doing nothing")


def index() -> rx.Component:
    return rx.center(
        rx.vstack(
            rx.button("Do nothing", on_click=State.do_nothing),
            rx.button("Do nothing (OtherState)", on_click=OtherState.do_nothing),
            rx.text(f"Time1: {State.time1}"),
            rx.text(f"Time2: {State.time2}"),
            rx.text(f"Time3: {State.time3}"),
            rx.text(f"Depends on time1: {State.depends_on_time1}"),
        ),
    )


app = rx.App()
app.add_page(index)
```


benedikt-bartscher:
Ready for initial review. Will add more tests if you like this idea :)

masenf (commented):
This is cool. I like the new features and flexibility.

reflex/state.py
```diff
@@ -1685,6 +1697,7 @@ def get_delta(self) -> Delta:
         # and always dirty computed vars (cache=False)
         delta_vars = (
             self.dirty_vars.intersection(self.base_vars)
+            .union(self.dirty_vars.intersection(self.computed_vars))
```
benedikt-bartscher, line 1700: @masenf did I find a bug here? You mentioned that one could define a computed var without dependencies and manually mark it as dirty. If I am not mistaken, this line is needed for this to work.

benedikt-bartscher:
Needs rebase after https://github.com/reflex-dev/reflex/pull/3254 is merged

Lendemor (approved):

picklelo (approved):
Awesome - I like having `cache` as an argument rather than a separate `rx.cached_var`, perhaps we can move more to that in the future.

benedikt-bartscher:
> Awesome - I like having `cache` as an argument rather than a separate `rx.cached_var`, perhaps we can move more to that in the future.

Thanks for your feedback and the merge. Maybe we should deprecate `rx.cached_var` in favor of the `cache` argument.

Issue #3238, opened 2024-05-07
Can't create instance of `ComponentState` inside `foreach`


TimChild:
**Describe the bug**
Attempting to call the `create` method on a `ComponentState` inside of a `foreach` call results in a `ReferenceError` on the frontend (only if the ComponentState has any reference to `cls` in the returned component).

**To Reproduce**
```python
import reflex as rx

from reflex_test.templates import template


class Comp(rx.ComponentState):
    value: int = 0
    @classmethod
    def get_component(cls, v, *children, **props) -> "Component":
        return rx.card(
            rx.text(f"I'm a component with state {v}"),  # <<< This alone works
            rx.text(f"I'm a component with state {cls.value}"),  # <<< Adding this causes error
        )

@template(route='/component_state_in_foreach_issue', title='Component State in Foreach Issue')
def index() -> rx.Component:
    return rx.container(
            rx.foreach(
                [1, 2, 3],
                Comp.create,
        ),
    )
```

**Expected behavior**
Expect 3 separate components to be created without error.

**Screenshots**
![image](https://github.com/reflex-dev/reflex/assets/48571510/ff7ef26f-5c40-4717-867e-af65a5a7e120)

**Specifics (please complete the following information):**
 - Python Version: 3.12
 - Reflex Version: 4.9
 - OS: WSL2
 - Browser (Optional): Chrome
 
 
 I think I will fall back to using a state that stores a list of an `rx.Base` subclass that holds the state for each item and just include the index in any calls back to event handlers so that I can update the appropriate item.

Lendemor:
We should probably throw an error during compilation, because I don't think the current implementation is supposed to work within `rx.foreach` anyway.

TimChild:
Yea, I think that probably is a good solution, at least in the short term. I'm not sure if the `State` attribute on the `rx.Component` is exclusively for the `rx.ComponentStatate`s, but if so, then just checking for components with that set inside the foreach would be enough.

adhami3310:
closing as a duplicate of #3718 (that one is newer, but had more discussion)

PR #3243, opened 2024-05-07
throw error for componentstate in foreach

Lendemor:
Prevent using ComponentState inside rx.foreach as suggested in #3238


reflex/components/core/foreach.py
```diff
@@ -43,6 +44,15 @@ def create(cls, iterable: Var[Iterable], render_fn: Callable, **props) -> Foreac
             raise TypeError(
                 f"Could not foreach over var of type Any. (If you are trying to foreach over a state var, add a type annotation to the var.)"
             )
+
+        if (
+            hasattr(render_fn, "__qualname__")
+            and render_fn.__qualname__ == ComponentState.create.__qualname__
+        ):
+            raise TypeError(
+                "Using a ComponentState as `render_fn` inside `rx.foreach` is not supported."
```
picklelo, line 53: nit: maybe say "not supported yet" as I assume we will fix this in the future.

picklelo (approved):

picklelo (approved):

PR #3249, opened 2024-05-08
[REF-2789] Graceful deprecation of rx.input.root and rx.input.input

masenf:
Handle previously valid code where rx.input.root wrapped rx.input/rx.input.input and rx.input.slot.

Raise deprecation warnings with hints about how to refactor code.

Copy props from rx.input.root to children inputs and apply any rx.input.slot components to children inputs in an attempt to keep existing code working as best as possible.

Fix DebounceInput:
  * pass children through (for rx.input.slots)
  * pass _rename_props through (for color_scheme)

### DIFF
--- a/reflex/components/radix/themes/components/text_field.py
+++ b/reflex/components/radix/themes/components/text_field.py
@@ -114,13 +114,16 @@ class TextFieldRoot(el.Div, RadixThemesComponent):
             for child in children
             if isinstance(child, (TextFieldRoot, DebounceInput))
         ]
+        if not inputs:
+            # Old-style where no explicit child input was provided
+            return cls.create(*children, **props)
         slots = [child for child in children if isinstance(child, TextFieldSlot)]
         carry_props = {
             prop: props.pop(prop)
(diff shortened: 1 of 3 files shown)


reflex/components/radix/themes/components/text_field.py
```diff
@@ -79,10 +83,79 @@ def create(cls, *children, **props) -> Component:
...
+            deprecation_version="0.5.0",
+            removal_version="0.6.0",
+        )
+        inputs = [
+            child
+            for child in children
+            if isinstance(child, (TextFieldRoot, DebounceInput))
+        ]
+        slots = [child for child in children if isinstance(child, TextFieldSlot)]
+        carry_props = {
```
picklelo, line 118: I think the event handlers may also need to be bubbled (based on the integration tests)

masenf: hmm, i'm actually surprised that code sample in reflex-web actually worked before... i thought a root always needed an explicit input inside of it, but i guess not. will adjust

picklelo (approved):

PR #3250, opened 2024-05-08
Add skeleton to main namespace

picklelo:
Only adding the skeleton for now (the other one is data list, but I wanted to see if we want to consolidate that at all with our existing lists).


Lendemor:
I think data_list and the rx.list namespace serve different purposes, they probably shouldn't be consolidated together.

masenf:
Already merged in #3257

PR #3254, opened 2024-05-08
properly replace ComputedVars

### LINKED ISSUE #3253: `rx.cached_var` and `rx.background` don't work in mixins
**Describe the bug**

Trying to mixin an `rx.background` method causes error that "only background tasks should use async with self".

Mixing in `rx.cached_var` doesn't do any caching and instead works like a normal `rx.var`. 

**To Reproduce**
```python
import logging
import reflex as rx

from reflex_test.templates import template

logger = logging.getLogger(__name__)
logger.setLevel(logging.DEBUG)


class BasicMixin(rx.Base):
    a: int = 0
    b: int = 0
    c: int = 0

    @rx.var
    def var_a(self) -> int:
        logger.debug('var_a called')
        return self.a+10

    @rx.cached_var
    def cached_a(self) -> int:
        logger.debug('cached_a called')
        return self.a+20

    def increment_a(self):
        self.a += 1

    async def increment_b(self):
        self.b += 1

    @rx.background
    async def increment_c(self):
        async with self:
            self.c += 1


class StateWithBasicMixin(BasicMixin, rx.State):
    pass

@template(route='/mixin_cached_or_background_issue', title='Mixin Cached or Background Issue')
def index() -> rx.Component:
    return rx.container(
        rx.vstack(
            rx.heading('Mixin Cached or Background Issue', size='5'),
            rx.grid(
                rx.text('Attr a:'),
                rx.text(StateWithBasicMixin.a),
                rx.text('Var a:'),
                rx.text(StateWithBasicMixin.var_a),
                rx.text('Cached a:'),
                rx.text(StateWithBasicMixin.cached_a),
                rx.text('Attr b:'),
                rx.text(StateWithBasicMixin.b),
                rx.text('Attr c:'),
                rx.text(StateWithBasicMixin.c),
                columns='5',
                rows='2',
                flow='column'
            ),
            rx.hstack(
                rx.button('Increment a', on_click=StateWithBasicMixin.increment_a),
                rx.button('Increment b', on_click=StateWithBasicMixin.increment_b),
                rx.button('Increment c', on_click=StateWithBasicMixin.increment_c),
            ),
        ),
        padding="2em",
    )
```

**Expected behavior**
On click of `increment b` or `increment c` expect that the `rx.cached_var` is not run. This is true if the cached var is moved to the State, but not if it lives in the mixin.
On click of `increment c` expect that the value increments like a normal background task would. Instead an error is raised "TypeError: Only background task should use `async with self` to modify state.". Works as expected if moved to the State.

 - Python Version: 3.12
 - Reflex Version: 0.4.9
 - OS: WSL2
 - Browser (Optional): Chrome
 
**Additional context**
If it's not easy to implement, it would be good to get some compilation errors :) 

Related PR that allowed `rx.var` to work in mixins #2351

benedikt-bartscher:
Addresses issues with cached_vars reported in #3253

Closes #3253


masenf (approved):
Looks good; thanks @benedikt-bartscher for jumping in on these fixes so quickly

benedikt-bartscher:
You are welcome. I'll add more tests for cached vars and background tasks in another PR

PR #3255, opened 2024-05-08
copy background task marker

benedikt-bartscher:
Addresses issues with background tasks reported in https://github.com/reflex-dev/reflex/issues/3253


masenf (approved):
killer!

Issue #3262, opened 2024-05-09
[REF-2802] [0.5.0] `add_style` not respected if component is used in a foreach


masenf:
**Describe the bug**
The `rx.foreach` seems to have special handling with regard to its child components. In particular, the added `style` is not respected.

**To Reproduce**
```python
def test_add_style_foreach():
    class StyledComponent(Component):
        tag = "StyledComponent"
        ix: Var[int]

        def add_style(self):
            return Style({"color": "red"})

    page = rx.vstack(
        rx.foreach(Var.range(3), lambda i: StyledComponent.create(i))
    )
    page._add_style_recursive(Style())
    print(page.children[0].children[0])
    print(page.children[0])
    assert 'css={{"color": "red"}}' in str(page.children[0].children[0])
    assert 'css={{"color": "red"}}' in str(page.children[0])
```

**Expected behavior**
Default styles added with `add_style` method work, even in a `foreach`

**Specifics (please complete the following information):**
 - Python Version: 3.11.8
 - Reflex Version: 0.5.0a1
 - OS: macOS 14

<sub>[REF-2802](https://linear.app/reflex-dev/issue/REF-2802/[050]-add-style-not-respected-if-component-is-used-in-a-foreach)</sub>

masenf:
Also discovered through testing and code inspection component styles are not applied to components inside a foreach. Meaning this bug has been there since at least 0.3.x! How did we never catch it?

```python
import reflex as rx

def index():
    return rx.foreach(
        rx.Var.range(5),
        lambda i: rx.box(f"Item {i}")
    ), rx.box("Outside Foreach")

app = rx.App(style={rx.box: {"background_color": rx.color("tomato", 7)}})
app.add_page(index)
```

<img width="169" alt="image" src="https://github.com/reflex-dev/reflex/assets/1524005/08a95249-7d1e-4340-bb79-2da7db49d30d">

PR #3263, opened 2024-05-09
[REF-2802] Foreach should respect modifications to children

masenf:
Components are mutable, and there is logic that depends on walking through the component tree and making modifications to components along the way. These modifications _must_ be respected by foreach for consistency.

Modifications necessary to fix the bug #3262:

* Change the hash function in `_render` to get a hash over the render_fn's `__code__` object. This way we get a stable hash without having to call the render function with bogus values.
* Call the render function once during `create` and save the result as a child of the Foreach component (tree walks will modify this instance).
* Directly render the original (and possibly modified) child component instead of calling the render_fn again and creating a new component instance at render time.

Additional changes because they're nice:

* Deprecate passing `**props` to `rx.foreach`. No one should have been doing this anyway, because it just does not work in any reasonable way.
* Raise `ForeachVarError` when the iterable type is Any
* Raise `ForeachRenderError` when the render function does not take 1 or 2 args.
* Link to the foreach component docs when either of those errors are hit.
* Change the `iterable` arg in `create` to accept `Var[Iterable] | Iterable` for better typing support (and remove some type: ignore comments)
* Simplify `_render` and `render` methods -- remove unused and potentially confusing code.

### DIFF
--- a/reflex/components/core/foreach.py
+++ b/reflex/components/core/foreach.py
@@ -95,7 +95,15 @@ class Foreach(Component):
             props["index_var_name"] = params[1].name
         else:
             # Otherwise, use a deterministic index, based on the render function bytecode.
-            code_hash = hash(self.render_fn.__code__).to_bytes(8, signed=True).hex()
+            code_hash = (
+                hash(self.render_fn.__code__)
+                .to_bytes(
+                    length=8,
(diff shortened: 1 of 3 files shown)


martinxu9 (approved):
LGTM

PR #3264, opened 2024-05-09
Allow `Component.add_style` to return a regular dict

masenf:
It's more convenient to allow returning a regular dict without having to import and wrap the value in `rx.style.Style`.

If the dict contains any Var or encoded VarData f-strings, these will be picked up when the plain dicts are passed to Style.update().

Because Style.update already merges VarData, there is no reason to explicitly merge it again in this function; this change keeps the merging logic inside the Style class.

--------

I also noticed that the `Style.update` function is slightly broken:

If a Style class with _var_data is passed to `Style.update` along with kwargs, then the `_var_data` was lost in the double-splat dictionary expansion.

Instead, only apply the kwargs to an existing or new `Style` instance to retain `_var_data` and properly convert values.

Added a test case and fix.

### DIFF
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -801,20 +801,16 @@ class Component(BaseComponent, ABC):
             The style to add.
         """
         styles = []
-        vars = []
 
         # Walk the MRO to call all `add_style` methods.
         for base in self._iter_parent_classes_with_method("add_style"):
             s = base.add_style(self)  # type: ignore
             if s is not None:
                 styles.append(s)
-                vars.append(s._var_data)
 
         _style = Style()
         for s in reversed(styles):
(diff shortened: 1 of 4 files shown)


picklelo (approved):

reflex/components/component.py
```diff
@@ -782,7 +782,7 @@ def validate_children(children):
 
         return cls(children=children, **props)
 
-    def add_style(self) -> Style | None:
+    def add_style(self) -> Dict[str, Any] | None:
```
picklelo, line 785: nit: can use lowercase `dict` here

picklelo (dismissed):

picklelo (approved):

PR #3266, opened 2024-05-09
rx.accordion customizability overhaul

masenf:
* fix low contrast font in "classic" variant
* relative font sizing, so setting font_size on the root works now
* use CSS vars wherever possible to make downstream styling easier
* optional dividers between accordion items
* support `radius`
* support `duration` and `easing` for controlling the animation
* fix animation jank while keeping padding_y
* lookup `data-variant` via CSS instead of passing props in python
* fix "surface" variant to use `--accent-surface` as radix intended


martinxu9 (approved):

masenf:
The radius is still a bit off here; need to make some tweaks

martinxu9 (approved):

PR #3271, opened 2024-05-10
[REF-2774] Add ReflexError and subclasses, send in telemetry

martinxu9:
## Summary
- Add ReflexError as base type exception for all reflex specific errors.
- Add more fine grained exception subclasses for state, var, etc.
- Catch and send in telemetry certain reflex errors.
- Need the change in telemetry from #3270, merge after #3270.

### DIFF
--- a/reflex/utils/prerequisites.py
+++ b/reflex/utils/prerequisites.py
@@ -210,28 +210,35 @@ def get_app(reload: bool = False) -> ModuleType:
 
     Raises:
         RuntimeError: If the app name is not set in the config.
+        exceptions.ReflexError: Reflex specific errors.
     """
-    os.environ[constants.RELOAD_CONFIG] = str(reload)
-    config = get_config()
-    if not config.app_name:
-        raise RuntimeError(
-            "Cannot get the app module because `app_name` is not set in rxconfig! "
-            "If this error occurs in a reflex test case, ensure that `get_app` is mocked."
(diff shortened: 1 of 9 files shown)


reflex/utils/prerequisites.py
```diff
@@ -210,28 +210,35 @@ def get_app(reload: bool = False) -> ModuleType:
...
 
-        # Reload the app module.
-        importlib.reload(app)
+            # Reload the app module.
+            importlib.reload(app)
 
-    return app
+        return app
+    except exceptions.ReflexError as ex:
+        telemetry.send("error", context="backend", detail=str(ex))
```
masenf, line 240: Not sure what exactly the `context` is trying to capture, but i think `"compile"` might be more appropriate, because the errors that this will catch can be either frontend or backend. Actually i would say most compile errors are "frontend".

masenf (commented):
I think we need another big try/catch in the event processing loop so we can report runtime errors, if we want.

reflex/utils/prerequisites.py
martinxu9: Yeah, good point about the compile error. I'll fix it.

So the PR doesn't send all the errors. I'm getting a sense if we feel ok about this approach. I'll add those in a followup PR for the event loop.

masenf (approved):

PR #3276, opened 2024-05-10
[REF-2774] Send runtime error in telemetry

martinxu9:
## Summary
- Send runtime error in telemetry
- Remove logs from telemetry for frontend package download.
## Tests
image

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -1067,50 +1067,61 @@ async def process(
         headers: The client headers.
         client_ip: The client_ip.
 
+    Raise:
+        ReflexError: If a reflex specific error occurs during processing the event.
+
     Yields:
         The state updates after processing the event.
     """
-    # Add request data to the state.
-    router_data = event.router_data
-    router_data.update(
-        {
-            constants.RouteVar.QUERY: format.format_query_params(event.router_data),
-            constants.RouteVar.CLIENT_TOKEN: event.token,
-            constants.RouteVar.SESSION_ID: sid,
-            constants.RouteVar.HEADERS: headers,
-            constants.RouteVar.CLIENT_IP: client_ip,
-        }
-    )
-    # Get the state for the session exclusively.
-    async with app.state_manager.modify_state(event.substate_token) as state:
-        # re-assign only when the value is different
-        if state.router_data != router_data:
-            # assignment will recurse into substates and force recalculation of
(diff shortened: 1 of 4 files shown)


reflex/app.py
```diff
@@ -1067,50 +1067,61 @@ async def process(
...
+                async for update in state._process(event):
+                    # Postprocess the event.
+                    update = await app._postprocess(state, event, update)
+
+                    # Yield the update.
+                    yield update
+    except ReflexError as ex:
+        print("martin")
+        telemetry.send("error", context="backend", detail=str(ex))
+        raise
```
masenf, line 1124: would it make sense to wrap this in a decorator so we don't need the extra indentation and it would be reusable in other places where we need this functionality?

something like

```python
def exception_telemetry(exc_filter, context, **dec_kwargs):
    def decorator(f):
        @functools.wraps(f)
        def telemetry_wrapper(*args, **kwargs):
            try:
                f(*args, **kwargs)
            except exc_filter as ex:
                telemetry.send("error", context=context, detail=str(ex), **dec_kwargs)
                raise
        return telemetry_wrapper
    return decorator
```

I get that the invocation does look different in various places, but it would be nice to find a way to encapsulate most of this in one place rather than sprinkling throughout the codebase

masenf (commented):
stray `print("martin")`, but looks like it would work.

one suggestion to consider, but would require some re-working.

martinxu9:
> stray `print("martin")`, but looks like it would work.
> 
> 
> 
> one suggestion to consider, but would require some re-working.

Oh 🫢

reflex/app.py
```diff
@@ -1067,50 +1067,60 @@ async def process(
...
-            yield update
-
-        # Only process the event if there is no update.
-        else:
-            if app._process_background(state, event) is not None:
-                # `final=True` allows the frontend send more events immediately.
-                yield StateUpdate(final=True)
-                return
+    from reflex.utils import telemetry
+    from reflex.utils.exceptions import ReflexError
```
masenf, line 1077: i've seen this pattern a lot... is there a circular dep when trying to import reflex.utils at the top of the module?

masenf (approved):

reflex/app.py
martinxu9: Yeah, at least one of the files, import at the top causes circular dependency

PR #3312, opened 2024-05-16
add support for lifespan tasks

Lendemor:
Add support for running task during the whole lifespan of the app, rather than linked to any specific state.

```python
import asyncio
import reflex as rx

@rx.page()
def index():
    return rx.box(), rx.color_mode.button(position="top-right")

async def long_running_task(key: str = "A"):
    i = 1
    while True:
        await asyncio.sleep(5)
        print(f"{key}: Oh I woke up {i} times.... Back to sleep again")
        i += 1


async def longer_running_task():
    while True:
        await asyncio.sleep(10)
        print("I sleep more than the others.")

@contextlib.asynccontextmanager
async def my_custom_context():
    print("Startup process....")
    yield
    print("Shutdown process....")

app = rx.App()
app.register_lifespan_task(long_running_task)
app.register_lifespan_task(long_running_task, "B")
app.register_lifespan_task(longer_running_task)
app.register_lifespan_task(my_custom_context)
```


masenf (commented):
The tasks should have a chance to clean up nicely without being cancelled right?

Lendemor:
> The tasks should have a chance to clean up nicely without being cancelled right?

I think it's up to the task itself to handle being cancelled cleanly, by using `try/finally` or catching the `asyncio.CancelledError`

reflex/app.py
```diff
@@ -100,7 +100,47 @@ class OverlayFragment(Fragment):
...
+    @contextlib.asynccontextmanager
+    async def _run_lifespan_tasks(self, app: FastAPI):
+        running_tasks = []
+        try:
+            async with contextlib.AsyncExitStack() as stack:
+                for task in self.lifespan_tasks:
+                    if isinstance(task, asyncio.Task):
+                        running_tasks.append(task)
+                    else:
+                        _t = task()
```
masenf, line 118: should we pass the `app` argument through so the API signature is the same as FastAPI's `lifetime` function

masenf (commented):
Looks good.

I'd like to see at least one test in the integration tests that exercises this.

Maybe have a lifespan task and lifespan contextmanager set a module global and send it to the frontend via an `rx.var`

masenf (approved):

PR #3313, opened 2024-05-16
update README to use 0.5.0

Lendemor:
Update README.md


masenf (dismissed):
i'm not 100% sure we want to show the experimental features right on the readme.

@picklelo what do you think?

picklelo:
Agreed - discussed this with Thomas in our meeting today, let's remove the toast and keep the window alert for now

README.md
```diff
@@ -100,7 +100,7 @@ class State(rx.State):
     def get_image(self):
         """Get the image from the prompt."""
         if self.prompt == "":
-            return rx.window_alert("Prompt Empty")
+            return rx._x.toast.error("Prompt Empty")
```
picklelo, line 103: let's revert this

README.md
```diff
@@ -185,15 +186,15 @@ class State(rx.State):
 
 The state defines all the variables (called vars) in an app that can change and the functions that change them.
 
-Here the state is comprised of a `prompt` and `image_url`. There are also the booleans `processing` and `complete` to indicate when to show the circular progress and image.
+Here the state is comprised of a `prompt` and `image_url`. There are also the booleans `processing` and `complete` to indicate when to block the button and when to show the image.
```
picklelo, line 189: Maybe "disable the button" rather than block

picklelo (changes requested):

tgberkeley:
When this PR is merged in I can do a follow up PR to update the image so that it exactly matches the code

README.md
```diff
@@ -185,15 +186,15 @@ class State(rx.State):
...
+Here the state is comprised of a `prompt` and `image_url`. There are also the booleans `processing` and `complete` to indicate when to disable the button (during image generation) and when to show the resulting image.
 
 ### **Event Handlers**
 
 ```python
 def get_image(self):
     """Get the image from the prompt."""
     if self.prompt == "":
-        return rx.window_alert("Prompt Empty")
+        return rx._x.toast.error("Prompt Empty")
```
tgberkeley, line 197: You might want to change this back as well so that it uses the window alert instead of the experimental toast

Lendemor: done

picklelo (approved):

PR #3324, opened 2024-05-16
Update docker-example

masenf:
Improvements:

  * use `exec` when starting `reflex` to make it PID 1 in the container
  * include `redis-server` in the simple containers for better concurrency

Create 4 separate examples with separate README files for different use cases:

* simple-two-port: this is the same as the previous Dockerfile, except with redis
* simple-one-port: this is the same as app.Dockerfile, but with the Caddyfile pulled out and redis included
* production-compose: for deployment on a standalone VPS
* production-app-platform: for deployment on AWS, Azure, GCE, etc. This deployment expects the frontend, redis, and database to be managed separately depending on the platform.


IABeresnev (approved):

docker-example/simple-two-port/Dockerfile
```diff
@@ -1,5 +1,8 @@
 # This Dockerfile is used to deploy a simple single-container Reflex app instance.
-FROM python:3.11
+FROM python:3.12
```
Lendemor, line 2: Others seems to use 3.11, why 3.12 for this one ?

picklelo (approved):
Missed this one, we can get it in

PR #3342, opened 2024-05-19
Cache the result of loading pyproject.toml file

### LINKED ISSUE #3341: custom_components._get_package_config called numerous times
**Describe the bug**
The function `_get_package_config` is called several times under normal operation of the `reflex run publish` command. The `_get_package_config` makes an IO call to load the pyproject.toml file. While not a huge issue, this is unnecessary to reload the file multiple times. 

**To Reproduce**
 - Code/Link to Repo: https://github.com/reflex-dev/reflex/blob/main/reflex/custom_components/custom_components.py#L73
 - To recreate:
 - Add a debug statement in `_get_package_config`
 - Run `run reflex component publish -u saf -p asdf`
 - Notice how many times this function is called.

**Expected behavior**
Typically, you only want to load a configuration file once and use the results. Or alternatively cache the results for future reference.

**Screenshots**
n/a

 - Python Version: 3.12.2
 - Reflex Version: 0.5.0.post1
 - OS: macOS Sonoma 14.4.1
 - Browser (Optional): n/a
 
**Additional context**

ericwb:
This change adds the functools cache decorator to avoid repeated calls to _get_package_config, which results in IO hits to read a file.

Closes #3341

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

### DIFF
--- a/reflex/custom_components/custom_components.py
+++ b/reflex/custom_components/custom_components.py
@@ -8,6 +8,7 @@ import subprocess
 import sys
 from collections import namedtuple
 from contextlib import contextmanager
+from functools import lru_cache
 from pathlib import Path
 from typing import Optional, Tuple
 
@@ -70,6 +71,7 @@ def _create_package_config(module_name: str, package_name: str):
         )
 
 
+@lru_cache(maxsize=None)
 def _get_package_config(exit_on_fail: bool = True) -> dict:
     """Get the package configuration from the pyproject.toml file.
 


ericwb:
Oops, functools.cache is only available in Python 3.9

reflex/custom_components/custom_components.py
```diff
@@ -70,6 +71,7 @@ def _create_package_config(module_name: str, package_name: str):
         )
 
 
+@lru_cache(maxsize=None)
 def _get_package_config(exit_on_fail: bool = True) -> dict:
```
picklelo, line 75: In `_validate_project_info` we interactively ask for info, and may end up updating the `pyproject.toml`, making this caching invalid - since afterwards we try to get the toml in `_collect_details_for_gallery`.

I'm thinking it may be okay as is originally - the read operations should be pretty fast, and will ensure we're always reading the up to date config. What do you think?

ericwb: Possibly. Probably the best thing to do next is write up some unit tests for this module. I do feel a bit uncomfortable changing code that I only tested manually.

ericwb: The pyproject.toml file is redundantly loaded at least 3 times with no changes until the caching would be invalidated. This code could probably use some heavy refactoring, but I'll leave it for now.

PR #3346, opened 2024-05-20
Catch more errors in frontend/backend

picklelo:
Catch all exception types rather than only Reflex errors. Also add logging for hot reloads.

### DIFF
--- a/reflex/app_module_for_backend.py
+++ b/reflex/app_module_for_backend.py
@@ -4,12 +4,14 @@ Only the app attribute is explicitly exposed.
 from concurrent.futures import ThreadPoolExecutor
 
 from reflex import constants
+from reflex.utils import telemetry
 from reflex.utils.exec import is_prod_mode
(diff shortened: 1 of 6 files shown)


Alek99 (approved):

reflex/app_module_for_backend.py
```diff
@@ -4,12 +4,14 @@
...
 
 from reflex import constants
+from reflex.utils import telemetry
 from reflex.utils.exec import is_prod_mode
 from reflex.utils.prerequisites import get_app
 
 if "app" != constants.CompileVars.APP:
     raise AssertionError("unexpected variable name for 'app'")
 
+telemetry.send("reload")
```
masenf, line 14: i think we should call this a `"compile"` event rather than `"reload"` because this will get hit for run dev, run prod, hot reload, and export

masenf (changes requested):

masenf (approved):

PR #3351, opened 2024-05-21
Added config for number of gunicorn workers

### LINKED ISSUE #3347: Configurable Number of Gunicorn Workers
**Describe the bug**
Number of workers are right now determined based on CPU count if redis is enabled.
This might not always be desired for instance if reflex runs as part of a k8s deployment and `os.cpu_count()` returns the number of cores on the node, not the pod.

**To Reproduce**
https://github.com/reflex-dev/reflex/blob/656914edefa7658c76e3d350abe7e9223db313bb/reflex/utils/processes.py#L31-L48

**Expected behavior**
Being able to specify the number of workers through an env var.

**Screenshots**

 - Python Version:
 - Reflex Version: 0.5.0
 - OS: 
 - Browser (Optional):
 
**Additional context**

Snaipergelka:
### Type of change

Added opportunity for users to configure number of Gunicorn workers through an env var `NUMBER_GUNICORN_WORKERS`

Closes #3347

### DIFF
--- a/reflex/config.py
+++ b/reflex/config.py
@@ -213,6 +213,9 @@ class Config(Base):
     # The worker class used in production mode
     gunicorn_worker_class: str = "uvicorn.workers.UvicornH11Worker"
 
+    # Number of gunicorn workers from user
+    number_gunicorn_workers: Optional[int] = None
+
     # Attributes that were explicitly set by the user.
     _non_default_attributes: Set[str] = pydantic.PrivateAttr(set())
 
--- a/reflex/config.py
+++ b/reflex/config.py
@@ -214,7 +214,7 @@ class Config(Base):
     gunicorn_worker_class: str = "uvicorn.workers.UvicornH11Worker"
 
     # Number of gunicorn workers from user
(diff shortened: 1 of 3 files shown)


reflex/config.py
```diff
@@ -213,6 +213,9 @@ class Config:
     # The worker class used in production mode
     gunicorn_worker_class: str = "uvicorn.workers.UvicornH11Worker"
 
+    # Number of gunicorn workers from user
+    number_gunicorn_workers: Optional[int] = None
```
masenf, line 217: Can we call this gunicorn_workers to match their config key: https://docs.gunicorn.org/en/stable/settings.html#workers

That would align with our other setting gunicorn_worker_class

masenf (changes requested):
Awesome! Thanks for this contribution.

reflex/config.py
Snaipergelka: Ok, committed changes.

Lendemor (approved):

PR #3353, opened 2024-05-21
Use twine environment variables if set

### LINKED ISSUE #3343: Shouldn't allow password and tokens as CLI arguments
**Describe the bug**
While not strictly a security exploit, the publish CLI can be considered weak and vulnerable to an accidental exposure of a password and/token to the shell history, ps output, or logs.

This can lead to [CWE-214: Invocation of Process Using Visible Sensitive Information](https://cwe.mitre.org/data/definitions/214.html)

**To Reproduce**
 - Code/Link to Repo: https://github.com/reflex-dev/reflex/blob/main/reflex/custom_components/custom_components.py#L621
 - Run `reflex component publish -u saf -p asdf`
 - Type `history`, and notice the password is in the shell history

**Expected behavior**
To avoid this possible exposure, it's always advised to interactively prompt for the password (not echoing to stdout) or read it from an environment variable. Tokens are similarly vulnerable.

As written in the [Command Line Interface Guidelines](https://clig.dev/):

_Do not read secrets directly from flags. When a command accepts a secret, eg. via a --password argument, the argument value will leak the secret into ps output and potentially shell history. And, this sort of flag encourages the use of insecure environment variables for secrets._

_Consider accepting sensitive data only via files, e.g. with a --password-file argument, or via stdin. A --password-file argument allows a secret to be passed in discreetly, in a wide variety of contexts._

_(It’s possible to pass a file’s contents into an argument in Bash by using --password $(< password.txt). This approach has the same security issue of leaking the file’s contents into the output of ps. It’s best avoided.)_

**Screenshots**
n/a

 - Python Version: 3.12.2
 - Reflex Version: 0.5.0.post1
 - OS: macOS 14.4.1
 - Browser (Optional): n/a
 
**Additional context**
[Command Line Interface Guidelines](https://clig.dev/)

ericwb:
The reflex publish command utilizes twine under the covers to publish to PyPI. However, reflex will effectively ignore twine environment variables and expect token or username/password to be passed via the CLI.

This change adds TWINE_USERNAME and TWINE_PASSWORD as default values for username and password. This has two postive effects; it excludes passing secrets in a command line, and it just utilizes variables already potentially set as env vars.

This addresses the issue in #3343 so that a user can avoid secrets in the command line. It's not the most ideal solution, which would be to fetch secrets from a file, but it is much better than CLI arguments.

Note: if the user wants to pass a token securely, they can still set the TWINE_USERNAME to __token__ and TWINE_PASSWORD to the token (with the pypi- prefix)

Closes #3343

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

### DIFF
--- a/reflex/custom_components/custom_components.py
+++ b/reflex/custom_components/custom_components.py
@@ -616,15 +616,17 @@ def publish(
         help="The API token to use for authentication on python package repository. If token is provided, no username/password should be provided at the same time",
     ),
     username: Optional[str] = typer.Option(
-        None,
+        os.getenv("TWINE_USERNAME"),
         "-u",
         "--username",
+        show_default="TWINE_USERNAME environment variable value if set",
         help="The username to use for authentication on python package repository. Username and password must both be provided.",
     ),
     password: Optional[str] = typer.Option(
-        None,
+        os.getenv("TWINE_PASSWORD"),
         "-p",
         "--password",
+        show_default="TWINE_PASSWORD environment variable value if set",
         help="The password to use for authentication on python package repository. Username and password must both be provided.",
     ),
     build: bool = typer.Option(


ericwb:
Let me know if you'd prefer different environment variable names.

masenf (approved):
i think it makes good sense to use the same var names.

thanks for contributing

PR #3355, opened 2024-05-21
Radix Themes + Tailwind Harmony

masenf:
Improve compatibility with tailwind classes used on Radix Themes components. According to the guide
https://www.radix-ui.com/themes/docs/overview/styling#tailwind-base-styles

* Setup postcss-import and manually import Tailwind base styles via @import tailwindcss/base before Radix Themes styles.
* If tailwind is NOT used, then bring in the styles via JS CSS import, like before.

### DIFF
--- a/reflex/components/radix/themes/base.py
+++ b/reflex/components/radix/themes/base.py
@@ -6,6 +6,7 @@ from typing import Any, Dict, Literal
 
 from reflex.components import Component
 from reflex.components.tags import Tag
+from reflex.config import get_config
 from reflex.utils import imports
 from reflex.vars import Var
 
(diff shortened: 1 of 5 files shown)


reflex/components/radix/themes/base.py
```diff
@@ -209,12 +210,20 @@ def create(
         return super().create(*children, **props)
 
     def _get_imports(self) -> imports.ImportDict:
```
Lendemor, line 212: Isn't it the occasion to convert this to the `add_imports` API ?

Lendemor (changes requested):

Issue #3365, opened 2024-05-22
[REF-3189] about add_hooks: _get_added_hooks make _get_all_hooks's return wrong 


seewindcn:
**Describe the bug**
```
    def _get_added_hooks(self) -> dict[str, imports.ImportDict]:
         ....

    def _get_all_hooks(self) -> dict[str, None]:
        """Get the React hooks for this component and its children.

        Returns:
            The code that should appear just before returning the rendered component.
        """
        code = {}

        # Add the hook code for this component.
        hooks = self._get_hooks()
        if hooks is not None:
            code[hooks] = None

        code.update(self._get_added_hooks())
        ......
        return code
```
_get_added_hooks will return `dict[str, imports.ImportDict]`, it is be called by _get_all_hooks, and `code.update(self._get_added_hooks())`,  this make code's content wrong, code will be dict[str, ImportDict].

**To Reproduce**

**Expected behavior**

**Screenshots**

**Specifics (please complete the following information):**
 - Python Version:
 - Reflex Version:
 - OS: 
 - Browser (Optional):
 
**Additional context**

<sub>[REF-3189](https://linear.app/reflex-dev/issue/REF-3189/about-add-hooks-get-added-hooks-make-get-all-hookss-return-wrong)</sub>

PR #3366, opened 2024-05-22
fix _get_all_hooks (#3365)

seewindcn:
#3365: about add_hooks: _get_added_hooks make _get_all_hooks's return wrong


masenf (changes requested):
use `poetry run pre-commit run --all-files`

and out of curiosity, is there an actual error case in the generated code? or just the return value is not matching the return annotation?

seewindcn:
> use `poetry run pre-commit run --all-files`
> 
> and out of curiosity, is there an actual error case in the generated code? or just the return value is not matching the return annotation?
yes, on my custom components [reflex-antd](https://github.com/seewindcn/reflex-antd), use get_hooks to make a VarData object, and pydantic will raise an exception because return value is not matching the VarData.hooks annotation.

PR #3372, opened 2024-05-22
[REF-2915] ComponentState subclasses are always treated as mixin

### LINKED ISSUE #3368: [REF-2915] Breaking change 0.5.0 -> 0.5.1, inheritance in ComponentState
**Describe the bug**
Due to a more strict check in the parent classes (here: https://github.com/reflex-dev/reflex/blob/main/reflex/state.py#L758), it seems no longer possible to create a component state, and then have multiple implementations with specific behavior.

**To Reproduce**
The following works in 0.5.0, but not in 0.5.1
```python
import reflex as rx


class MyBaseComponentState(rx.ComponentState):
    a: str = "1"

    def on_increment(self):
        raise NotImplementedError()

    @classmethod
    def get_component(cls, **props) -> rx.Component:
        return rx.box(
            rx.text("hello " + cls.a, **props),
            rx.button("Increment", on_click=cls.on_increment),
        )


class MyComponent(MyBaseComponentState):

    def on_increment(self):
        self.a = str(int(self.a) + 1)


my_component = MyComponent.create


def index() -> rx.Component:
    return rx.box(
        my_component(),
    )


app = rx.App()
app.add_page(index)
```

**Expected behavior**
Expected behavior is that this throw no errors, like in 0.5.0.

 - Python Version: 3.11.6
 - Reflex Version: 0.5.1
 - OS: Mac
 - Browser (Optional): Chrome

<sub>[REF-2915](https://linear.app/reflex-dev/issue/REF-2915/breaking-change-050-051-inheritance-in-componentstate)</sub>

masenf:
Always treat ComponentState subclasses as mixin, and explicitly pass `mixin=False` in `.create` classmethod when intentionally creating a substate.

This allows a "base" ComponentState to have subclasses that also work as ComponentState themselves.

Fix #3368

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -757,7 +757,9 @@ class BaseState(Base, ABC, extra=pydantic.Extra.allow):
             for base in cls.__bases__
             if issubclass(base, BaseState) and base is not BaseState and not base._mixin
         ]
-        assert len(parent_states) < 2, "Only one parent state is allowed."
+        assert (
+            len(parent_states) < 2
+        ), f"Only one parent state is allowed {parent_states}."
         return parent_states[0] if len(parent_states) == 1 else None  # type: ignore
 
     @classmethod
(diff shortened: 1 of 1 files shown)


benedikt-bartscher:
Looks good to me, thanks @masenf for the fast fix :rocket:

Lendemor (approved):

PR #3375, opened 2024-05-22
feat: Optionally comparing fields in Var.contains, e.g. on rx.Base based types.

abulvenz:
I want to make the example app at the bottom work.

TLDR:
```python
    rx.foreach(
        State.available_items,
        lambda item: rx.menu.item(
            rx.checkbox(
                item.name,
                checked=State.selected_items.contains(item.key, "key"),  # The argument "key" is new.
                on_change=lambda _: State.select_item(item),
            )
        ),
    )
```
The `contains` function checks if a var is contained in the array by comparing a specific field.
This can also be helpful in other circumstances.

Other implementation ideas:
- Use a function such as [`_.isEqual`](https://underscorejs.org/docs/modules/isEqual.html)
- By default compare the field if `item.key` is passed (Maybe interfering with other use cases).
- Pass the entire JS-expression that should be used for comparison (Messy code).
- Use only keys for `selected_items`, that is not always possible (e.g. when using sqlalchemy-ORM objects.

Screenshot of the app:
image

```python

import reflex as rx
from rxconfig import config


class ComplexItem(rx.Base):
    name: str
    color: str
    key: str


class State(rx.State):
    selected_items: list[ComplexItem] = []

    available_items: list[ComplexItem] = [
        ComplexItem(name="Apple", color="red", key="apple"),
        ComplexItem(name="Banana", color="yellow", key="banana"),
        ComplexItem(name="Orange", color="orange", key="orange"),
        ComplexItem(name="Grape", color="purple", key="grape"),
        ComplexItem(name="Kiwi", color="green", key="kiwi"),
        ComplexItem(name="Blueberry", color="blue", key="blueberry"),
    ]

    def select_item(self, item: ComplexItem):
        if item in self.selected_items:
            self.selected_items.remove(item)
        else:
            self.selected_items.append(item)
            

def index() -> rx.Component:
    return rx.container(
        rx.menu.root(
            rx.menu.trigger(rx.button("Select Items")),
            rx.menu.content(
                rx.foreach(
                    State.available_items,
                    lambda item: rx.menu.item(
                        rx.checkbox(
                            item.name,
                            checked=State.selected_items.contains(item.key, "key"),
                            on_change=lambda _: State.select_item(item),
                        )
                    ),
                )
            ),
        ),
        rx.grid(
            rx.foreach(
            State.selected_items,
            lambda item: rx.card(
                rx.text(item.name, size="2"),
                rx.text(item.key, size="1"),
                style={"width": "200px", "color": item.color},
            ),
        ),
        columns="4",
        ),
    )


app = rx.App()
app.add_page(index)
```

### DIFF
--- a/reflex/vars.py
+++ b/reflex/vars.py
@@ -1353,7 +1353,7 @@ class Var:
             "'in' operator not supported for Var types, use Var.contains() instead."
         )
 
-    def contains(self, other: Any, field: Union[str, None] = None) -> Var:
+    def contains(self, other: Any, field: Union[Var, None] = None) -> Var:
         """Check if a var contains the object `other`.
 
         Args:
@@ -1395,11 +1395,12 @@ class Var:
                     f"'in <string>' requires string as left operand, not {other._var_type}"
                 )
 
-            _var_name = (
-                f"{self._var_name}.includes({other._var_full_name})"
-                if field is None
-                else f"{self._var_name}.some(e=>e.{field}==={other._var_full_name})"
-            )
+            _var_name = None
+            if field is None:
+                _var_name = f"{self._var_name}.includes({other._var_full_name})"
+            else:
+                field = Var.create_safe(field, _var_is_string=isinstance(field, str))
+                _var_name = f"{self._var_name}.some(e=>e[{field._var_name_unwrapped}]==={other._var_full_name})"
 
             return self._replace(
                 _var_name=_var_name,
--- a/reflex/vars.py
+++ b/reflex/vars.py
@@ -838,19 +838,19 @@ class Var:
                 if invoke_fn:
                     # invoke the function on left operand.
                     operation_name = (
-                        f"{left_operand_full_name}.{fn}({right_operand_full_name})"  # type: ignore
-                    )
+                        f"{left_operand_full_name}.{fn}({right_operand_full_name})"
+                    )  # type: ignore
                 else:
                     # pass the operands as arguments to the function.
                     operation_name = (
-                        f"{left_operand_full_name} {op} {right_operand_full_name}"  # type: ignore
-                    )
+                        f"{left_operand_full_name} {op} {right_operand_full_name}"
+                    )  # type: ignore
                     operation_name = f"{fn}({operation_name})"
             else:
                 # apply operator to operands (left operand <operator> right_operand)
                 operation_name = (
-                    f"{left_operand_full_name} {op} {right_operand_full_name}"  # type: ignore
-                )
+                    f"{left_operand_full_name} {op} {right_operand_full_name}"
+                )  # type: ignore
                 operation_name = format.wrap(operation_name, "(")
         else:
             # apply operator to left operand (<operator> left_operand)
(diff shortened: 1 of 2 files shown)


benedikt-bartscher (commented):
Awesome, looks good to me. Thanks, @abulvenz

Lendemor (dismissed):

reflex/vars.py
```diff
@@ -1393,8 +1394,15 @@ def contains(self, other: Any) -> Var:
                 raise VarTypeError(
                     f"'in <string>' requires string as left operand, not {other._var_type}"
                 )
+
+            _var_name = (
+                f"{self._var_name}.includes({other._var_full_name})"
+                if field is None
+                else f"{self._var_name}.some(e=>e.{field}==={other._var_full_name})"
```
masenf, line 1401: i think this should work if the field is specified as a state var, so maybe we can write it like

```
field = Var.create_safe(field, _var_is_string=isinstance(field, str))
```

and 

```python
f"{self._var_name}.some(e=>e[{field._var_full_name}]==={other._var_full_name})"
```

masenf (changes requested):
i haven't tested the suggested change, but i think field should be specifiable as another var, like the other var operations.

reflex/vars.py
Lendemor: I think the purpose is being able to write something like 
```python
rx.foreach(
    State.available_items,
    lambda item: rx.menu.item(
        rx.checkbox(
            item.name,
            checked=State.selected_items.contains(item.key, State.filter_key),
            on_change=lambda _: State.select_item(item),
        )
    ),
)
```
where you can change the value of `filter_key` dynamically

abulvenz: Thanks @masenf, I will use your corrected version, it works with both strings and vars.

One thing that is that `e.{field}` could also take nested values like `field="address.street"` which will not work with the `e[...]` syntax.

I can also support this case, when I use this abomination:
```python
_var_name = f"{self._var_name}.some(e=>{field._var_name_unwrapped}.split(\".\").reduce((acc,v)=>acc[v],e)==={other._var_full_name})"

```

```python
class ComplexKey(rx.Base):
    part1: str
    part2: int

class ComplexItem(rx.Base):
    name: str
    color: str
    key: ComplexKey
```
Then you could write:
```python
State.selected_items.contains(item.key.part1, "key.part1"),
```


But at some point it becomes ridiculously complex...

-------- 8< --------- 8< --------------

And ideas for sleepless nights :sleeping_bed: :desktop_computer: 
Please nobody take this seriously :smile: 
```python
import reflex.js as js
...
checked=State.selected_items.some(
    js.λ("e", 
        js.var("e")
            .subscript(State.filter_key)                        
            .equals(js.var(State.search_text)
        )
    )
),
...
assert js.λ("e", 
        js.var("e")
            .subscript(State.filter_key)                        
            .equals(State.search_text)
        )
    ).__str__() == "e=>e[state__state.filter_key]===state__state.search_text"

assert is_turing_complete(js)
```

masenf: 🏆

masenf (approved):

PR #3379, opened 2024-05-23
[REF-2879] Make client_state work without global refs (local only)

masenf:
For better memory efficiency and scenarios like `@rx.memo` where the component is only generated once and cannot use a unique name.

Add `set_value` as a method, to set a specific value on the frontend; for use with `on_click` and other scenarios where the value passed from the event trigger is not actually the value that one wants to set -- this only works with literal values or vars, it cannot use logic based on any values passed from the event trigger itself.


picklelo (approved):
Working well for me

PR #3380, opened 2024-05-23
[REF-2878] Map fontFamily to fontFamily and --default-font-family

masenf:
When setting the font_family prop for a component, also set the radix token `--default-font-family` so that child radix components will inherit the font.


picklelo (approved):
nice

onexay:
Thanks!

PR #3383, opened 2024-05-24
include resize and radius props to the text area component

carlosabadia:
Included the following props for the [text area component](https://reflex.dev/docs/library/forms/textarea/#textarea) from the [radix component](https://www.radix-ui.com/themes/docs/components/text-area)

- resize
- radius


masenf (dismissed):

picklelo:
@carlosabadia I think you need to run `poetry run python scripts/make_pyi.py reflex/components/radix/themes/components/text_area.pyi` then it should be good to go.

masenf (approved):

PR #3388, opened 2024-05-27
add mapping between client_token and socket id

### LINKED ISSUE #3387: Can I detect if a client is disconnected within a State's method?
The page of code shown below will continue the while loop even if the browser is reloaded and disconnected from the client. What are possible ways to detect the disconnection from the client and break the while loop?

```
import reflex as rx

class State(rx.State):
    @rx.background
    async def loop_function(self):
        while True:
            # I want to terminate this loop when disconnected from the client.
            foo()

@rx.page(on_load=State.loop_function)
def index():
    return rx.text("Hello")

app = rx.App()
```

Lendemor:
Keep an internal mapping between client token and sid.

Can be used by end user to terminate some logic/tasks if the websocket associated with a given state is disconnected.

Will also be helpful later for some internal use.

Can be used like this to terminate a background task for example : 
```python
class State(rx.State):
    @rx.background
    async def loop_function(self):
        while True:
            if self.router.session.client_token not in app.event_namespace.token_to_sid:
                break
            await asyncio.sleep(2)


@rx.page(on_load=State.loop_function)
def index():
    return rx.text("Hello")
```

Possible improvement : Provide a "cleaner" access to this mapping for end user? 

Close #3387

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -1302,6 +1302,12 @@ class EventNamespace(AsyncNamespace):
     # The application object.
     app: App
 
+    # Keep a mapping between socket ID and client token.
+    token_to_sid: dict[str, str] = {}
+
+    # Keep a mapping between client token and socket ID.
+    sid_to_token: dict[str, str] = {}
+
     def __init__(self, namespace: str, app: App):
         """Initialize the event namespace.
 
@@ -1327,7 +1333,9 @@ class EventNamespace(AsyncNamespace):
         Args:
             sid: The Socket.IO session id.
         """
-        pass
+        disconnect_token = self.sid_to_token.pop(sid, None)
+        if disconnect_token:
+            self.token_to_sid.pop(disconnect_token, None)
 
     async def emit_update(self, update: StateUpdate, sid: str) -> None:
         """Emit an update to the client.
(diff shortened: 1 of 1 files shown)


Lendemor:
> LGTM - is there a concrete use case for this yet? Or just in case we need it later?

Without this users can not terminate logic in background tasks, so I'd say yes.

but there will also be other potential future use too.

masenf (approved):

5quinque:
How can you use this method if you have your `app` initialized in one module, and your `State` class defined in another?
If I want the background task to run when the page loads, there isn't a way without causing a circular import?

For example, 

```python
from app_name.state import State

app = rx.App()

app.add_page(component=chat_page(), route="/", on_load=State.background_event)
```

```python
from app_name.app_name import app

class State(rx.State):
    @rx.event(background=True)
    async def background_event(self):
        while True:
            if self.router.session.client_token not in app.event_namespace.token_to_sid:
                ...
```

benedikt-bartscher:
@5quinque maybe you can use get_app inside the handler?

5quinque:
> @5quinque maybe you can use get_app inside the handler?

Didn't know about this method, but it works perfectly, thanks :D

PR #3389, opened 2024-05-27
add config knob for react_strict_mode

### LINKED ISSUE #3382: on_unmount will run when entering page
**Describe the bug**
on_unmount will run when entering page

**To Reproduce**
 - Code/Link to Repo:
my code like this:
 ```python
rx.heading(f"current value: {ControlState.value}",on_mount=ControlState.update_value,on_unmount=ControlState.stop_update_value),
```
i expect to run stop_update_value when i leave the page, but it run already when i come in page.

 - Python Version:3.10
 - Reflex Version:reflex==0.5.1
 - OS: linux
 - Browser (Optional): chrome

### PR DESCRIPTION
**Pull request by Lendemor**
Default to True.

Also cleanup unused code.

Close #3382

### CODE REVIEW

File: reflex/utils/prerequisites.py [hunk 1]
```diff
@@ -625,7 +612,7 @@ def _update_next_config(
     next_config = {
         "basePath": config.frontend_path or "",
         "compress": config.next_compression,
-        "reactStrictMode": True,
+        "reactStrictMode": config.react_strict_mode,
```
picklelo, line 615: Should we just expose a `NextConfig` object? We can add all the params there eventually. And let them do `config.next_config.react_strict_mode`

picklelo (approved):
Looks good, just an idea on how we can make the config more scalable.

reflex/utils/prerequisites.py
Lendemor: I guess we could.

PR #3397, opened 2024-05-29
[REF-2202] Implement event handlers for Plotly

masenf:
Add all plotly event handlers.

TODO:
  * figure out the correct signature and description for all of these
  * get reflex-web to parse event handler descriptions from comments (like props)


reflex/components/plotly/plotly.py
```diff
@@ -11,6 +13,76 @@
...
+    Args:
+        e0: The button click data.
+
+    Returns:
+        The menu, button, and active state.
+    """
+    return [e0.menu, e0.button, e0.active]
+
+
+def _passthrough_signature(e0: Var) -> List[Any]:
```
Lendemor, line 65: Wouldn't the `_passthrough_signature` be useful for more than just plotly?

I'm wondering if we should have a `signature.py` for all "signature presets" (maybe we can do that in later PR tho)

Lendemor (dismissed):

HongyuHansonYao (dismissed):

HongyuHansonYao (approved):

PR #3420, opened 2024-06-03
Feat: Add Session storage to store data on client storage

TG199:
### Type of change

### Description of change
Implemented SessionStorage class for managing session-related data in sessionStorage, enhancing project functionality


TG199:
Why are they failing 😥

reflex/__init__.py
```diff
@@ -301,7 +301,7 @@
         "var",
         "Cookie",
         "LocalStorage",
-        "ComponentState",
+        "SessionStorage" "ComponentState",
```
masenf, line 304:
```suggestion
        "SessionStorage", "ComponentState",
```
```suggestion
        "SessionStorage" "ComponentState",
```

reflex/state.py
```diff
@@ -2835,6 +2835,54 @@ def __new__(
...
+        """
+        if encoding or errors:
+            inst = super().__new__(cls, object, encoding or "utf-8", errors or "strict")
+        else:
+            inst = super().__new__(cls, object)
+        inst.name = name
+        inst.sync = sync
+        inst.session_id = session_id
+        inst.auth = auth
+        inst.timeout = timeout
```
masenf, line 2882: how do these get handled on the frontend / where did they come from?

masenf (changes requested):
there should be more line diff in `state.js`, basically `hydrateClientStorage` and `applyClientStorageDelta` should be updated to handle the session storage cases as well

reflex/.templates/web/utils/state.js
```diff
@@ -506,7 +518,18 @@ export const hydrateClientStorage = (client_storage) => {
       }
     }
   }
-  if (client_storage.cookies || client_storage.local_storage) {
+  if (client_storage.sessionStorage && typeof window != "undefined") {
```
masenf, line 521:
```suggestion
  if (client_storage.session_storage && typeof window != "undefined") {
```

I think the key is `session_storage`, here and in various places below,

masenf (changes requested):
how did you test this? I tried extending `integration/test_client_storage.py` to add some `SessionStorage` state vars and they didn't get saved, so something isn't quite working.

I'll push uo the tests that i wrote quickly so you can run them, but its not necessarily exhaustive.

TG199:
Finally!

masenf (approved):
Looks good!

.pre-commit-config.yaml
```diff
@@ -31,4 +31,4 @@ repos:
         always_run: true
         language: system
         description: 'Update pyi files as needed'
-        entry: python scripts/make_pyi.py
+        entry: python3 scripts/make_pyi.py
```
picklelo, line 34: do we need this change?

reflex/event.py
```diff
@@ -612,6 +612,34 @@ def remove_local_storage(key: str) -> EventSpec:
     )
 
 
+def clear_session_storage() -> EventSpec:
+    """Set a value in the session storage on the frontend.
```
picklelo, line 616: should be "clear all values", looks like we missed this in the local_storage also

reflex/event.py
```diff
@@ -612,6 +612,34 @@ def remove_local_storage(key: str) -> EventSpec:
...
+        EventSpec: An event to clear the session storage.
+    """
+    return server_side(
+        "_clear_session_storage",
+        get_fn_signature(clear_session_storage),
+    )
+
+
+def remove_session_storage(key: str) -> EventSpec:
+    """Set a value in the session storage on the frontend.
```
picklelo, line 628: Should be "remove a value"

picklelo (approved):
Some minor docstring issues that we can fix in a follow up

PR #3424, opened 2024-06-03
Set use resize handler default to true

HongyuHansonYao:


reflex/components/plotly/plotly.py
```diff
@@ -109,7 +109,7 @@ class Plotly(PlotlyLib):
     config: Var[Dict]
 
     # If true, the graph will resize when the window is resized.
-    use_resize_handler: Var[bool]
+    use_resize_handler: Var[bool] = Var.create(True)
```
picklelo, line 112: Need to use `Var.create_safe` to make the type checker happy

HongyuHansonYao: tyty

masenf (approved):

PR #3426, opened 2024-06-03
[REF-3006] Inline code rendered in rx.markdown has extra trailing space

### LINKED ISSUE #3425: Inline code rendered in `rx.markdown` has extra trailing space
**Describe the bug**
Any inline code in an `rx.markdown` component seems to have a trailing space

**To Reproduce**
```python
import reflex as rx


def index() -> rx.Component:
    return rx.container(
        rx.vstack(
            rx.text("prefix ", rx.code("normal code"), " suffix"),
            rx.markdown("prefix `markdown code` suffix"),
        ),
        rx.logo(),
    )


app = rx.App()
app.add_page(index)
```

**Expected behavior**
Both lines should look the same

**Screenshots**
image

 - Python Version: 3.11
 - Reflex Version: 0.5.3a

masenf:
Avoid replacing newlines with whitespaces... these show up in the rendered output 😱

Fix #3425

### DIFF
--- a/reflex/components/markdown/markdown.py
+++ b/reflex/components/markdown/markdown.py
@@ -229,7 +229,7 @@ class Markdown(Component):
         Returns:
             The formatted component.
         """
-        return str(self.get_component(tag, **props)).replace("\n", " ")
+        return str(self.get_component(tag, **props)).replace("\n", "")
 
     def format_component_map(self) -> dict[str, str]:
         """Format the component map for rendering.


picklelo (commented):
Ahh good catch!! That's been there for a while never new the reason. Will this mess up any other formatting though? Let me test it a little bit

picklelo (approved):
Looks good to me

PR #3429, opened 2024-06-03
Webdriver ArgOptions improvements

benedikt-bartscher:
- allow passing custom selenium ArgOptions to webdriver
- allow passing capabilities to auto-generated ArgOptions

### DIFF
--- a/reflex/testing.py
+++ b/reflex/testing.py
@@ -522,6 +522,7 @@ class AppHarness:
         driver_kwargs: dict[str, Any] | None = None,
         driver_options: ArgOptions | None = None,
         driver_option_args: List[str] | None = None,
+        driver_option_capabilities: dict[str, Any] | None = None,
     ) -> "WebDriver":
         """Get a selenium webdriver instance pointed at the app.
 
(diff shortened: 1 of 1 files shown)


reflex/testing.py
```diff
@@ -544,37 +548,43 @@ def frontend(
...
         want_headless = False
-        options: ArgOptions | None = None
         if os.environ.get("APP_HARNESS_HEADLESS"):
             want_headless = True
         if driver_clz is None:
             requested_driver = os.environ.get("APP_HARNESS_DRIVER", "Chrome")
             driver_clz = getattr(webdriver, requested_driver)
-            options = getattr(webdriver, f"{requested_driver}Options")()
+            if driver_options is None:
+                driver_options = getattr(webdriver, f"{requested_driver}Options")()
```
masenf, line 557: seems we can get rid of these now that each driver has a similar bit of code below.

masenf (commented):
i haven't used capabilities, what is their primary function? looking at the docs it seemed kind of user-agent-y, but curious what your use case is.

masenf (approved):

PR #3430, opened 2024-06-03
New pr for color and event triggers

HongyuHansonYao:


reflex/components/recharts/charts.py
```diff
@@ -522,6 +510,15 @@ def create(cls, *children, **props) -> Component:
...
         )
 
+    def get_event_triggers(self) -> dict[str, Union[Var, Any]]:
+        """Get the event triggers that pass the component's value to the handler.
+        Returns:
+            A dict mapping the event trigger to the var that is passed to the handler.
+        """
+        return {
+            EventTriggers.ON_ANIMATION_START: lambda: [],
+            EventTriggers.ON_ANIMATION_START: lambda: [],
```
Lendemor, line 520: typo with copy/paste? should be END I guess

HongyuHansonYao: Yes, thank you!

reflex/app.pyi
```diff
@@ -0,0 +1,426 @@
+"""Stub file for reflex/app.py"""
```
masenf, line 1: remove this file. yes the whole file

masenf (changes requested):
i'm not sure why there are so many new `.pyi` files... those shouldn't be included here.

i think we have a bug in the `pyi_generator.py` that tries to create them if any Component is imported and there is a change in the git index. i'll try to fix that. in the meantime, you should remove these from your commit.

only keep the pyi file changes in the `reflex/components` subtree.

masenf:
depends on #3227 for proper `rx.color` serialization for `fill` and `stroke` props.

PR #3433, opened 2024-06-04
[REF-3010]Circular Imports Fix

ElijahAhianyo:
Resolve circular imports between some `refex/utils` files which throw errors when importing some components from an interpreter than running reflex.
We probably got away with this previously because:

1. Everything was likely imported via other modules during an import, and Python typically resolved the look-up in the sys.modules.
2. During reflex runtime/compile time, some modules are imported earlier and thus no longer need to be resolved locally.

This PR also includes a tiny fix for missing `rx.el.tbody`

### DIFF
--- a/reflex/components/core/html.py
+++ b/reflex/components/core/html.py
@@ -9,7 +9,7 @@ class Html(Div):
     """Render the html.
 
     Returns:
-        The code to render the  html component.
+        The code to render the html component.
     """
 
     # The HTML to render.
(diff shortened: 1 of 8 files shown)


Lendemor (approved):

masenf (approved):
It would be nice to actually untangle all of our circular imports with some refactoring someday, but thanks for the quick fix to get 0.5.3 out the door!

PR #3441, opened 2024-06-04
fix: allow specifying websocket port

cdmistman:
A [Replit](https://replit.com) user reached out with pain getting a Reflex app set up on Replit. After some digging, this change seems to make the pain vanish

As a short explanation of why:
- all Replit projects are Docker containers
- we have reverse proxies set up to route traffic to repls given some URL, eg `https://f3ea4bab-7ba5-4eb2-9ad4-9ec4f3702158-00-1squ4svnaus5u.infra-staging.replit.dev/`
- we may also do port mapping to make the experience as pleasant as possible, eg 3000->80 for http ports
- because this is happening over the internet, we do also support auto tls promotion, which means that content being served from a project running in dev mode (eg `reflex dev`) will probably be accessed via `https://`
- all of this results in the next server being served over :443, but the web socket still being provided via `wss` on :8080 (note that we had to tell reflex's backend to bind to port :8080 due to a bug in our reverse proxying)

this change allows the next server's client to connect to the `wss` websocket on :8080 as expected

### DIFF
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -101,7 +101,6 @@ export const getBackendURL = (url_str) => {
       } else if (endpoint.protocol === "http:") {
         endpoint.protocol = "https:";
       }
-      endpoint.port = ""; // Assume websocket is on https port via load balancer.
     }
   }
   return endpoint;


masenf:
> all of this results in the next server being served over :443, but the web socket still being provided via wss on :8080

i'm unsure about this line; is replit terminating TLS for the websocket separately and mapping 8080 external -> 8000 internal?

blast-hardcheese:
> is replit terminating TLS for the websocket separately and mapping 8080 external -> 8000 internal?

In our case, we were doing

```python
config = rx.Config(
    ...
    backend_port=8080,
    api_url=f"wss://{replit_dev_domain}:8080"
)
```
which is where 8080 is coming from. We're doing `wss://...:8080` external -> `ws://...:8080` internal. It should be equivalent to 8000 external being mapped to 8000 internal, iiuc.

masenf:
Interesting. If the `api_url` is being set explicitly to something that is not one of the `SAME_DOMAIN_HOSTNAMES` then i would have expected the line changed in the patch to not even be executed. This endpoint rewriting logic should only be occuring if the `api_url` is `localhost`.

```
// These hostnames indicate that the backend and frontend are reachable via the same domain.
const SAME_DOMAIN_HOSTNAMES = ["localhost", "0.0.0.0", "::", "0:0:0:0:0:0:0:0"];
```

blast-hardcheese:
I may be mistaken then. @cdmistman, were you still using the port override, or did we back out the configuration change and just `8080 -> 8000`?

cdmistman:
ah sorry, it seems there's been some work since this project was vanilla!

give me a sec, i'm gonna revalidate this patch from a fresh reflex project...

cdmistman:
ah ok yes - @blast-hardcheese when we were walking through this a few mins ago we didn't get the same problem, but the patch from this PR isn't sufficient to fix the problems here, there's still another part missing

@masenf if you'd like to TAL, you can fork this Repl to a Replit account and click the run button, you'll see the websocket error pop up once everything's running https://replit.com/@coltonatreplit/CriminalEnchantingEfficiency

blast-hardcheese:
So, my full reproducing case is:

1. `reflex init`, choose the sidebar template
2. ✍️ Modify `rxconfig.py` to add `backend_port=8080,` to work around the aforementioned port binding bug
3. ✍️ Configure the VM to have `443` -> `3000` and `8080` -> `8080`
4. ❌ Observe that the UI is loadable at `https://${uuid}.replit.dev/` and I see attempts to connect to `wss://${uuid}.replit.dev/` as well (note missing port), the protocol switch is not handled by the UI server, and the `Cannot connect to server: timeout` popup (first screenshot below) is shown.
5. ✍️ Modify `.web/utils/state.js` to introduce @cdmistman 's change
6. ✅ Loading the same URL as before, `https://${uuid}.replit.dev/`, observe that `wss://...` now includes port `:8080` (screenshot two)

image
image

masenf:
@blast-hardcheese got it, thanks for the explanation.

It sounds like you're wanting to take advantage of the automatic hostname rewriting without also assuming the wss port will be the standard https port 443.

**Am I correct in my previous observation that setting `api_url="https://actual-hostname.repl.it:8080"` in the `rxconfig.py` does work around this issue without the patch being proposed here?**

I'm hesitant to accept the patch as-is because it would be a behavioral regression that would break common load balancer setups that we've been recommending over the past 6 months or so. **But I do want to find a solution** that minimizes the amount of configuration we expect end users to do and understand.

I'm currently leaning to a solution wherein (when connecting to the frontend over https)

* the `api_url` is set to `https://localhost:8080`
  * do not strip off the port
  * assume you want to connect to `wss://{frontend}:8080`
* the `api_url` is set to `http://localhost:8080` (the default)
  * strip off the port
  * assume you want to connect to `wss://{frontend}`

In this configuration, you would still have to set `api_url` yourself, but you could set it to a generic value and take advantage of the hostname rewriting.

blast-hardcheese:
I have confirmed that the following in `rxconfig.py` does work:

```python
import reflex as rx
import os

replit_dev_domain = os.getenv("REPLIT_DEV_DOMAIN")

config = rx.Config(
    app_name="reflex_test",
    backend_port=8080,
    api_url=f"https://{replit_dev_domain}:8080"
)
```

For context, what is the underlying rationale for dropping the port? You mention common load-balancer setups, are you referring to production environments multiplexing both servers on the same port, either via path matching or protocol matching?

One solution we discussed internally was to expose a websocket proxy from the frontend repo to the backend, possibly printing a helpful `WARN: Using a development websocket proxy! For improved performance, see [docs link](...)`.

The downside would be that you'd have a larger surface area to maintain, and it may obscure the error if the user doesn't notice that they're using the dev proxy in production, but maybe this is mitigated by misconfigurations just being slow instead of correlating with downtime for the user.

masenf:
> You mention common load-balancer setups, are you referring to production environments multiplexing both servers on the same port, either via path matching or protocol matching?

Yeah exactly. We recommend setting up either caddy or nginx to terminate TLS and have the webserver proxy the requests to the correct port. I would say it's less common to have TLS for the websocket terminated separately on an uncommon port.

> expose a websocket proxy from the frontend repo to the backend

I've played around with this a bit in the last couple of week, but the other direction: hitting the backend on `/` actually proxies to the frontend (or a static export of the same). It seems to work pretty well, but is currently still PoC and the performance is less than stellar.

--------------

Another option for this specific usage is to special case the `REPLIT_DEV_DOMAIN` like we do for github codespaces: https://github.com/reflex-dev/reflex/blob/6420fb12526bed1d76e94fdfc1c6fdabbb2c3638/reflex/config.py#L319-L329

Then we could get it working out of the box and even workaround the port 8000 / 8080 issue for users automatically.

blast-hardcheese:
> I've played around with this a bit in the last couple of week, but the other direction: hitting the backend on `/` actually proxies to the frontend (or a static export of the same). It seems to work pretty well, but is currently still PoC and the performance is less than stellar.

Yeah, I'd imagine that the frontend would involve sending quite a bit more content than would otherwise be transferred over the websocket port, but definitely quite a bit easier to maintain. 

> Another option for this specific usage is to special case the `REPLIT_DEV_DOMAIN` like we do for github codespaces:

That seems like a good idea. Good news is the fix for `:8000` is going out shortly, so that shouldn't be an issue. I'll open a PR to add support for `REPLIT_DEV_DOMAIN` for the time being, and wherever you decide to land on the 8000-over-3000 or 3000-over-8000 should be fine.

Thank you!

masenf:
> I'll open a PR to add support for `REPLIT_DEV_DOMAIN` for the time being

Aweomse! that should be a much less controversial change. Thanks for helping unpack this issue.

I've playing around with reflex on repl.it and it's quite nice to work with and feels faster than codespaces. I definitely could see myself using this for sharing repros and helping users debug their code.

cdmistman:
glad to hear you like it! also glad to hear a better solution was agreed upon. thank you @masenf for working with us on this problem!

blast-hardcheese:
> I have confirmed that the following in `rxconfig.py` does work:

Turns out I had forgotten to revert the manual changes to `state.js`. When I did so, the bug reappeared, so I am convinced that the change laid out in this PR is actually the correct one. To avoid impinging on your efforts to help folks productionize reflex, I gated it behind a suffix check, `hostname.endsWith('replit.dev')`, which is spiritually equivalent to the previously agreed upon workaround similar to what was done for codespaces.

Happy to remove that check once the non-prod-one-port solution has been found. Thank you!

PR #3442, opened 2024-06-04
[REF-2977] [REF-2982] Merge layout prop and automatic darkmode

masenf:
* Expose `template` prop in rx.plotly
* Set default `template` to a color_mode_cond that responds to dark mode
* Merge user-provided `layout` and `template` props into the serialized plot data


reflex/components/plotly/plotly.py
```diff
@@ -4,14 +4,18 @@
...
+from reflex.components.core.cond import color_mode_cond
 from reflex.event import EventHandler
 from reflex.vars import Var
 
 try:
-    from plotly.graph_objects import Figure
+    from plotly.graph_objects import Figure, layout
+
+    Template = layout.Template
 except ImportError:
```
Lendemor, line 16: Do we still need to handle this import error now that we have lazy loading for the imports?
We should just raise and tell them to install plotly package

Lendemor (changes requested):

Lendemor (approved):

reflex/components/plotly/plotly.py
```diff
@@ -89,7 +93,7 @@ class PlotlyLib(NoSSRComponent):
 
     library = "react-plotly.js@2.6.0"
 
-    lib_dependencies: List[str] = ["plotly.js@2.22.0"]
+    lib_dependencies: List[str] = ["plotly.js@2.22.0", "mergician@2.0.2"]
```
picklelo, line 98: I don't think this is needed since you specify the library in `add_imports`. Also maybe we should just combine this class and the bottom class together, since there aren't any other components in PlotlyLib

reflex/components/plotly/plotly.py
```diff
@@ -210,14 +228,62 @@ def add_custom_code(self) -> list[str]:
...
+        Returns:
+            The Plotly component.
+        """
+        from plotly.io import templates
+
+        responsive_template = color_mode_cond(
+            light=Var.create_safe(templates["plotly"]).to(dict),
+            dark=Var.create_safe(templates["plotly_dark"]).to(dict),
+        )
+        if isinstance(responsive_template, Var):
```
picklelo, line 250: When would this check fail?

reflex/components/plotly/plotly.py
```diff
@@ -210,14 +228,62 @@ def add_custom_code(self) -> list[str]:
...
+
     def _render(self):
         tag = super()._render()
         figure = self.data.to(dict)
-        if self.layout is None:
-            tag.remove_props("data", "layout")
+        merge_dicts = []  # Data will be merged and spread from these dict Vars
+        if self.layout is not None:
+            # Why is this not a literal dict? Great question... it didn't work
+            # reliably because of how _var_name_unwrapped strips the outer curly
```
picklelo, line 266: Is this the same issue from serializing rx.Base types in a list? Once we fix that maybe we'll be able to clean this up.

reflex/components/plotly/plotly.py
masenf: added a warning if users try

```python
>>> import reflex as rx
>>> rx.plotly
Warning: Plotly is not installed. Please run `pip install plotly`.
<bound method Plotly.create of <class 'reflex.components.plotly.plotly.Plotly'>>
```

it does still work though

reflex/components/plotly/plotly.py
masenf: it wouldn't it's just to please the type checker, since `rx.cond` can be either a Var or a Component

reflex/components/plotly/plotly.py
masenf: yep

reflex/components/plotly/plotly.py
picklelo: Ah got it - can also do `assert isinstance(...)` to please it

picklelo (approved):

Lendemor (approved):

PR #3443, opened 2024-06-04
state.js: set event_processing = false when websocket connects

### LINKED ISSUE #3404: state.js processEvent hang when eventSent then network disconnect
**Describe the bug**
when debug, I found serveral times, the web hang;
like this:
image

event_processing is true, and processEvent don't apply event_queue;

**To Reproduce**
maybe Steps to reproduce the behavior:
1. in processEvent, eventSent=true, 
2. backend process kill or network close;
3. reconnect;
4. processEvent hang;

**Expected behavior**
maybe we can set event_processing=true when on connect?

masenf:
In case an event was pending when the websocket went down, allow further events to be processed when it comes back up.

Fix #3404 

Various improvements to `integration/test_connection_banner.py` to catch this issue and fix overall reliability of the test.

### DIFF
--- a/integration/test_connection_banner.py
+++ b/integration/test_connection_banner.py
@@ -11,13 +11,27 @@ from reflex.testing import AppHarness, WebDriver
 
 def ConnectionBanner():
     """App with a connection banner."""
+    import asyncio
+
     import reflex as rx
 
     class State(rx.State):
         foo: int = 0
 
+        async def delay(self):
+            await asyncio.sleep(2)
+
     def index():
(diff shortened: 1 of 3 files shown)


masenf:
Had to tweak the PR to instead reset the flag as the socket disconnects. See https://github.com/reflex-dev/reflex/pull/3443/commits/6b789b63d0bc5b0e50faa7aba2242b68abc00a3a

picklelo (approved):

PR #3448, opened 2024-06-05
use add_imports everywhere

Lendemor:
Replace use of `_get_imports` with the new API `add_imports`.

Also tried to unify the typing for the message by using `ImportDict` and `ParsedImportDict` types.


reflex/compiler/utils.py
```diff
@@ -82,7 +83,7 @@ def validate_imports(import_dict: imports.ImportDict):
                 used_tags[import_name] = lib
 
 
-def compile_imports(import_dict: imports.ImportDict) -> list[dict]:
+def compile_imports(import_dict: imports.ParsedImportDict) -> list[dict]:
```
masenf, line 86:
```suggestion
def compile_imports(import_dict: ParsedImportDict) -> list[dict]:
```

reflex/vars.py
```diff
@@ -120,7 +125,7 @@ class VarData(Base):
     state: str = ""
 
     # Imports needed to render this var
-    imports: ImportDict = {}
+    imports: Union[ImportDict, ParsedImportDict] = {}
```
masenf, line 128: Should we just make this `imports: ParsedImportDict` and have the `imports` arg for `__init__` take the `Union[ImportDict, ParsedImportDict]`

masenf (commented):
Mostly looks right, i have yet to check it out and test it out locally though

Lendemor:
> Mostly looks right, i have yet to check it out and test it out locally though

Addressed your two comment, I also added unit tests for the new `parse_imports` function

reflex/components/component.py
```diff
@@ -1301,24 +1301,29 @@ def _get_imports(self) -> imports.ImportDict:
...
-                        for tag in _make_list(maybe_tags)
-                    ]
-                    for package, maybe_tags in clz.add_imports(self).items()
-                }
-            )
+            added_imports_ = clz.add_imports(self)
+            if not isinstance(added_imports_, list):
+                added_imports_ = [added_imports_]
+            for added_imports in added_imports_:
+                _added_import_dicts.append(parse_imports(added_imports))
```
masenf, line 1317: the naming here is a bit hard to read because `added_imports_` and `added_imports` are very similar looking

maybe you can use `list_of_import_dicts` instead of `added_imports_` and `import_dict` instead of `added_imports`?

reflex/vars.py
```diff
@@ -201,7 +219,10 @@ def dict(self) -> dict:
             "state": self.state,
             "interpolations": list(self.interpolations),
             "imports": {
-                lib: [import_var.dict() for import_var in import_vars]
+                lib: [
+                    import_var.dict() if isinstance(import_var, ImportVar) else {}
```
masenf, line 223: is this for the type checker? when would `import_var` not be an `ImportVar` here? what else could it be?

Lendemor: Will remove, it was needed in the middle of dev because the import_var could be a str, but now it's parsed properly ahead of time so it won't happens.

masenf (approved):

PR #3451, opened 2024-06-06
fix warning about props

Lendemor:
prevent the warning about props `type_` from popping up when using `rx.upload`


masenf (approved):
got it, this popped up when we switched from the chakra input to the base html input. good fix

PR #3453, opened 2024-06-07
update launch message

Lendemor:
small improvement for QoL when launching reflex app

### DIFF
--- a/reflex/utils/exec.py
+++ b/reflex/utils/exec.py
@@ -114,7 +114,7 @@ def run_process_and_launch_url(run_command: list[str], backend_present=True):
                         )
                         if backend_present:
                             console.print(
-                                f"Backend running at: [bold green]{get_config().api_url}[/bold green]"
+                                f"Backend running at: [bold green]http://0.0.0.0:{get_config().backend_port}[/bold green]"
                             )
                         first_run = False
                     else:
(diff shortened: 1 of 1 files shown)


reflex/utils/exec.py
```diff
@@ -108,7 +108,14 @@ def run_process_and_launch_url(run_command: list[str], backend_present=True):
                         url = match.group(1)
                         if get_config().frontend_path != "":
                             url = urljoin(url, get_config().frontend_path)
-                        console.print(f"App running at: [bold green]{url}")
+                        if backend_present:
+                            console.print(
```
picklelo, line 112: nit: maybe put the backend url on a newline. Also remove the ending `)` from the green in the url, or it appears to be part of the url

reflex/utils/exec.py
```diff
@@ -108,7 +108,14 @@ def run_process_and_launch_url(run_command: list[str], backend_present=True):
...
                         if get_config().frontend_path != "":
                             url = urljoin(url, get_config().frontend_path)
-                        console.print(f"App running at: [bold green]{url}")
+
+                        console.print(
+                            f"App running at: [bold green]{url}[/bold green]{' (Frontend-only mode)' if not backend_present else ''}"
+                        )
+                        if backend_present:
+                            console.print(
+                                f"Backend running at: [bold green]{get_config().api_url}[/bold green]"
```
masenf, line 117: This could be kind of confusing if the user puts in something like `API_URL=https://my-api.example.com`, but then misconfigures their domain or load balancer:

The message would say "Backend running at https://my-api.example.com", but it might not actually be running there... Maybe we should use `f"http://0.0.0.0:{get_config().backend_port}"`?

masenf (approved):

PR #3458, opened 2024-06-07
swap all use of get_event_triggers to rx.EventHandler props

Lendemor:
what the title says.

### DIFF
--- a/reflex/components/datadisplay/dataeditor.py
+++ b/reflex/components/datadisplay/dataeditor.py
@@ -2,11 +2,12 @@
 from __future__ import annotations
 
 from enum import Enum
-from typing import Any, Callable, Dict, List, Literal, Optional, Union
+from typing import Any, Dict, List, Literal, Optional, Union
 
 from reflex.base import Base
 from reflex.components.component import Component, NoSSRComponent
 from reflex.components.literals import LiteralRowMarker
+from reflex.event import EventHandler
 from reflex.utils import console, format, imports, types
 from reflex.utils.imports import ImportVar
 from reflex.utils.serializers import serializer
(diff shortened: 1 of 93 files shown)


reflex/components/datadisplay/dataeditor.py
```diff
@@ -223,35 +276,6 @@ def _get_imports(self):
             },
         )
 
-    def get_event_triggers(self) -> Dict[str, Callable]:
```
masenf, line 226: removing this has resulted in the addition of the base Component.get_event_triggers default triggers, is that appropriate for this component? or should we have a `get_event_triggers` here that returns an empty dict?

reflex/components/radix/themes/components/alert_dialog.py
```diff
@@ -48,18 +40,14 @@ class AlertDialogContent(elements.Div, RadixThemesComponent):
...
-            EventTriggers.ON_ESCAPE_KEY_DOWN: lambda e0: [e0],
-        }
+    # Fired when the dialog is opened.
+    on_open_auto_focus: EventHandler[lambda e0: [e0]]
+
+    # Fired when the dialog is closed.
+    on_close_auto_focus: EventHandler[lambda e0: [e0]]
+
+    # Fired when the escape key is pressed.
+    on_escape_key_down: EventHandler[lambda e0: [e0]]
```
masenf, line 50: i wonder if we should mention that the intended purpose of these event handlers is primarily to call `rx.prevent_default`

masenf (dismissed):

picklelo (dismissed):
Can we delete the `EventTriggers` class now?

Lendemor:
> Can we delete the `EventTriggers` class now?

I'm not sure yet, they are still used in the base Component I think.

I'm hesitant to convert this one to rx.EventHandler right now because I'm not sure it won't have side effect.

picklelo (approved):

reflex/components/datadisplay/dataeditor.py
Lendemor: Adding `get_event_triggers` that returns `{}` has caused events triggers to be completely removed from the docs.

PR #3485, opened 2024-06-12
export != zip

benedikt-bartscher:
`reflex export --no-zip` skips sitemap generation


Lendemor:
I think this is covered by PR #3462, can you check ?

PR #3491, opened 2024-06-13
notifying frontend about backend error looks better

Lendemor:
Use toast as the default for notifying about backend errors.

Can set `debug=True` in the config or the env var `REFLEX_DEBUG=True` to have a more helpful message in the frontend.

Will fallback to the old window_alert if the toast provider has been removed.

### DIFF
--- a/reflex/components/sonner/toast.py
+++ b/reflex/components/sonner/toast.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import Any, Literal, Optional, Union
+from typing import Any, ClassVar, Literal, Optional, Union
 
 from reflex.base import Base
 from reflex.components.component import Component, ComponentNamespace
@@ -211,6 +211,9 @@ class Toaster(Component):
     # Pauses toast timers when the page is hidden, e.g., when the tab is backgrounded, the browser is minimized, or the OS is locked.
     pause_when_page_is_hidden: Var[bool]
 
+    # Marked True when any Toast component is created.
+    is_used: ClassVar[bool] = False
+
     def add_hooks(self) -> list[Var | str]:
         """Add hooks for the toaster component.
 
@@ -231,7 +234,7 @@ class Toaster(Component):
         return [hook]
 
     @staticmethod
-    def send_toast(message: str, level: str | None = None, **props) -> EventSpec:
+    def send_toast(message: str = "", level: str | None = None, **props) -> EventSpec:
         """Send a toast message.
 
         Args:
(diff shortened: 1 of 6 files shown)


masenf (commented):
i like the direction here. will need a rebase after #3476 since the exception handling code has moved

benedikt-bartscher:
Just a heads up, the new exception handling was merged in #3572 and #3637

reflex/app.py
```diff
@@ -112,11 +112,29 @@ def default_backend_exception_handler(exception: Exception) -> EventSpec:
...
+
     error = traceback.format_exc()
 
     console.error(f"[Reflex Backend Exception]\n {error}\n")
 
-    return window_alert("An error occurred. See logs for details.")
+    error_message = (
+        [f"{type(exception).__name__}: {exception}.", "See logs for details."]
+        if get_config().debug is True
+        else ["Contact the website administrator."]
```
picklelo, line 124: Should we still mention that an error occured here

reflex/app.py
```diff
@@ -112,11 +112,29 @@ def default_backend_exception_handler(exception: Exception) -> EventSpec:
...
+    )
+    if Toaster.is_used:
```
picklelo, line 126: This doesn't seem to work for me out of the box? I have to do `rx.toast.provider()` within my app and then it shows up

reflex/config.py
```diff
@@ -219,6 +219,9 @@ class Config:
     # Number of gunicorn workers from user
     gunicorn_workers: Optional[int] = None
 
+    # If debug is true, more details will show when an error happens.
+    debug: bool = False
```
picklelo, line 223: Should we just always enable this in dev mode?

reflex/app.py
```diff
@@ -112,11 +112,29 @@ def default_backend_exception_handler(exception: Exception) -> EventSpec:
...
+    from reflex.components.sonner.toast import Toaster, toast
+
     error = traceback.format_exc()
 
     console.error(f"[Reflex Backend Exception]\n {error}\n")
 
-    return window_alert("An error occurred. See logs for details.")
+    error_message = (
+        [f"{type(exception).__name__}: {exception}.", "See logs for details."]
+        if get_config().debug is True
```
picklelo, line 123: don't need to compare `is True`

reflex/app.py
Lendemor: It is still mentioned as the title of the toast.

There does seems to be a bug where the title disappear for some reason.

reflex/app.py
Lendemor: Fixed this by making sure that `default_overlay_component` is always called before any page compilation.

reflex/components/sonner/toast.py
```diff
@@ -231,18 +234,27 @@ def add_hooks(self) -> list[Var | str]:
...
         Returns:
             The toast event.
         """
+        if not Toaster.is_used:
+            raise ValueError(
+                "Toaster component must be created before sending a toast. (use `rx.toast.provider()`)"
+            )
         toast_command = f"{toast_ref}.{level}" if level is not None else toast_ref
+        if message == "" and ("title" not in props or "description" not in props):
+            raise ValueError("Toast message or title or descrition must be provided.")
```
picklelo, line 257: typo: description

picklelo (approved):

PR #3497, opened 2024-06-14
upgrade to latest ruff

Lendemor:


masenf (dismissed):
looking into why the CI is buggin out

Lendemor:
Figured out the dependency review one, still have to understand why app-harness fail in CI even though it doesn't in local.

masenf (approved):
We decided that downpinning `fastapi < 0.111.0` (before it got extra deps), is the safest approach to solve our integration failure in `integration/test_event_chain.py::test_yield_state_update[click_yield_interim_value]` introduced by running with `uvloop` from fastapi 0.111.0.

[Private discussion regarding the pin](https://discord.com/channels/1029853095527727165/1082391228915646515/1252753274856276099) 🔒

PR #3510, opened 2024-06-17
pre-commit improvements

benedikt-bartscher:


Lendemor:
#3497 will switch around the order of pre-commit checks (pyi generation will happens before pyright checks)

You can base your change on that.

Issue #3487, opened 2024-06-13
Removing Rich


Alek99:
### Discussed in https://github.com/orgs/reflex-dev/discussions/3479

<div type='discussions-op-text'>

<sup>Originally posted by **itsmeadarsh2008** June 11, 2024</sup>
Rich is too heavy with bloatware in the dependency. Would it be a better idea to make a progress bar using `tqdm` only because using rich doesn't make any sense with web apps in the production? It will be a good idea if we replace it with them and keep it simple. You can convert it to an issue if you want. Here are some example sentences. (No Boxes, Just Logs And These Sentences)
1.  ⚡ Compiling ETA: `{seconds}`
2. 🚧 Stopping the server
3. 🏓 Frontend is playing ping-pong on `{port}`
4. 🖥 Backend is serving food on `{port}` 
5. Your server is ready to go 🚀</div>

masenf:
Rich is now used as a log sink instead of directly, so we're a step closer to removing it / making it optional.

We already have PoorProgress which replaces the rich progress bar we use in some parts of the app. The last outstanding piece is just deciding how we want to handle the default log sink. I would say make rich optional, and if it's there then use it; otherwise just output normal log lines.

PR #3514, opened 2024-06-18
Replace rich progress bar to tqdm progress bar

TG199:
### Type of change

### New Feature Submission:

### Changes To Core Features:

### Description
This commit replaces the Rich progress bar implementation with TQDM #3487. The following changes were made:
- **Replaced Rich Progress Bar**:
  - Updated the `_zip` function to use TQDM for progress bar handling.
  - TQDM is now responsible for displaying progress during zipping operations, providing a simpler and lighter alternative to Rich.

- **Preserved Pretty Printing with Rich**:
  - Retained the use of `rich.console` for pretty printing logs and outputs to maintain readability and formatting in the console.

- **Updated Dependencies**:
  - Added TQDM to `pyproject.toml` as a dependency.


picklelo (commented):
I think you need to `poetry lock` again for the tests

TG199:
Okay

masenf:
i pushed a fix for the merge conflict here... but i think there are some other issues we need to resolve here before we can take this change.

1. update the dependency review CI job to accept the MIT / APL-2.0 license... unfortunately these combo licenses need to all be added in the workflow
2. this adds a new dependency but doesn't actually remove the `rich` dependency as specified by #3487

i'm hesitant to take the new dep without fully removing the old dep

itsmeadarsh2008:
We can use the `termcolor` (wheel size - 12.7KB source and built distribution 7.7KB) instead but for prompts (that uses rich too) that ask for what kind of template are you building. Or maybe select your region for deployment. I found an alternative born as an inquirer on PyPI (14.5KB). The `tqdm` is about 169.4KB on the source and 78.3KB on the built distribution.

TG199:
It appears tqdm lacks necessary features such as colored text support and interactive prompts, so I left that part of the rich lib to handle it

itsmeadarsh2008:
> It appears tqdm lacks necessary features such as colored text support and interactive prompts, so I left that part of the rich lib to handle it

Use termcolor and inquirer python

TG199:
Okay

ericwb:
>     1. update the dependency review CI job to accept the MIT / APL-2.0 license... unfortunately these combo licenses need to all be added in the workflow
> 

```
  The following dependencies have incompatible licenses:
  poetry.lock » tqdm@4.66.4 – License: MIT AND MPL-2.0
  pyproject.toml » tqdm@4.66.4 – License: MIT AND MPL-2.0
  Error: Dependency review detected incompatible licenses.
```

The dependency review failure is with MPL-2.0 license not APL-2.0. Note, that MPL is a copyleft license, so the maintainers should consider whether this might affect stakeholders. Might not be for most, but some enterprises steer clear of anything copyleft.

Here's a nice FAQ on MPL:
https://www.mozilla.org/en-US/MPL/2.0/FAQ/

Also, the fix is simply, just add MPL-2.0 to the allow list.

adhami3310:

PR #3519, opened 2024-06-19
pass validation of valid_parent if inheriting from valid class

Lendemor:
Improve the cases where component can pass the validation check on parent/children.

## Setup
```python
import reflex as rx

from reflex.components.radix.themes.components.context_menu import (
    ContextMenuRoot,
    ContextMenuTrigger,
)

class ValidRoot(ContextMenuRoot): ...

class InvalidRoot(rx.Component): ...
```

## Pass the check
```python
@rx.page()
def index():
    return ValidRoot.create(ContextMenuTrigger.create("Click me!"))
```

## Trigger the error
```python
@rx.page()
def index():
    return InvalidRoot.create(ContextMenuTrigger.create("Click me!"))
```

### DIFF
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -1022,10 +1022,10 @@ class Component(BaseComponent, ABC):
                     f"The component `{comp_name}` only allows the components: {valid_child_list} as children. Got `{child_name}` instead."
                 )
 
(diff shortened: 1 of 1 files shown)


ElijahAhianyo (approved):
Nice

PR #3522, opened 2024-06-19
add bare sqlalchemy session, Closes #3512

### LINKED ISSUE #3512: add bare sqlalchemy rx.session
Add something like `rx.sqla_session` or `rx.session(bare=True)` to get a bare sqlalchemy session instead of sqlmodel

benedikt-bartscher:


Lendemor (approved):

reflex/__init__.py
```diff
@@ -304,7 +304,7 @@
         "window_alert",
     ],
     "middleware": ["middleware", "Middleware"],
-    "model": ["session", "Model"],
+    "model": ["session", "sqla_session", "Model"],
```
picklelo, line 307: I think this shouldn't be in the main `rx.`namespace as it's more of a special case. For people who need it, they can do the full import as `rx.model.sqla_session`

picklelo (changes requested):
Code looks good - just a request to remove this from the top-level namespace

reflex/__init__.py
benedikt-bartscher: Thanks for your review, done!

Lendemor (approved):

PR #3529, opened 2024-06-20
lru_cache on few functions and no_args_is_help by default

itsmeadarsh2008:
### All Submissions:

### DIFF
--- a/reflex/reflex.py
+++ b/reflex/reflex.py
@@ -7,6 +7,7 @@ import os
 import webbrowser
 from pathlib import Path
 from typing import List, Optional
+from functools import lru_cache
 
 import typer
 import typer.core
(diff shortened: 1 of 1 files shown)


masenf (commented):
what is the purpose of the `lru_cache` here? seems these typer cli command functions will really only get called once per process invocation and neither seems computationally expensive. i must be missing something?

itsmeadarsh2008:
> what is the purpose of the `lru_cache` here? seems these typer cli command functions will really only get called once per process invocation and neither seems computationally expensive. I must be missing something.

This should speed up the execution by running the cache. I'm not sure which functions do I need to lru_cache. (for you to know, I have done on the commands that are not important maybe). Maybe we could do caching in the run command? So whenever we have to open IDE to do a reflex run. It should execute the cache instead if the codebase remains the same (last edited). Talking about no_args_is_help, This would enable the automatic execution help command if no args is given. Editing is enabled.

PR #3540, opened 2024-06-21
[REF-3079] state.js: disconnect websocket for window "pagehide" event

### LINKED ISSUE #3478: State hangs after using back button after following a link.
**Describe the bug**
When clicking a link and navigating to a subdomain of the website, pressing back on the browser results in a webpage where the backend seems to not be connected. Hardcoded links on the page still work, but any events will not work like button presses until user refreshes page and then state works again.

**To Reproduce**
Navigate to nursereports.org. Open the login modal using the 'sign in' button in the upper right and then close it (remember that it works.) Click any of the links at the top of the page (students, staff, travelers, etc.) Now hit the back button. Clicking sign in or any other button on the page now doesn't work.

Dev console on navigating back states 'socket reconnected.' This issue has been ongoing since at least 0.4.x

 - Code/Link to Repo:
https://github.com/hjpr/nursereports

**Expected behavior**
Buttons work regardless of navigation events.

 - Python Version: 3.11
 - Reflex Version: 0.5.3
 - OS: Windows, Mac, Android
 - Browser (Optional): Chrome, Safari, Firefox
 
**Additional context**

masenf:
All major browser use a "bfcache" to freeze pages when navigating away from the domain, then they restore the page when going back.

However if the page uses websockets, these get kind of stuck unless you disconnect them before freezing (and have a mechanism for reconnecting, which we already do).

Ref: https://web.dev/articles/bfcache

Fix #3478

### DIFF
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -388,9 +388,17 @@ export const connect = async (
     }
   }
 
+  const pagehideHandler = (event) => {
+    if (event.persisted && socket.current?.connected) {
+      console.log("Disconnect backend before bfcache on navigation");
+      socket.current.disconnect();
+    }
+  }
+
   // Once the socket is open, hydrate the page.
   socket.current.on("connect", () => {
     setConnectErrors([]);
+    window.addEventListener("pagehide", pagehideHandler);
   });
 
   socket.current.on("connect_error", (error) => {
@@ -400,6 +408,7 @@ export const connect = async (
   // When the socket disconnects reset the event_processing flag
   socket.current.on("disconnect", () => {
     event_processing = false;
+    window.removeEventListener("pagehide", pagehideHandler);
   });
 
   // On each received message, queue the updates and events.


picklelo (approved):
interesting, didn't know about bfcache

masenf:
> didn't know about bfcache

Me either, but Chrome at least logs a message on the browser console with a link to the page above that explains it in pretty good detail, so that was a nice hint.

I wish i knew why it didn't work in dev mode though... maybe dev mode has an `unload` event listener?

> On desktop, Chrome and Firefox have chosen to make pages ineligible for bfcache if they add an unload listener, which is less risky but also disqualifies a lot of pages. Safari will attempt to cache some pages with an unload event listener, but to reduce potential breakage it won't run the unload event when a user is navigating away, which makes the event very unreliable. [[1]](https://web.dev/articles/bfcache#never-use-the-unload-event)

Lendemor (approved):

Lendemor (approved):

PR #3507, opened 2024-06-17
add allow_system prop to colormode iconbutton, and clean up logic

Lendemor:

PR #3542, opened 2024-06-21
fix hydration error

### LINKED ISSUE #3543: [REF-3165] Warning: Text content did not match. Server: "" Client: "system"
**Describe the bug**
When `chakra-ui-color-mode` does not match the system mode, nextjs throws a hydration error.

**To Reproduce**

```python
import reflex as rx


def index() -> rx.Component:
    return rx.text(rx.style.resolved_color_mode)


app = rx.App()
app.add_page(index)
```

**Expected behavior**
`resolved_color_mode` should not cause a hydration error. We should set it to a definitely value for server side rendering, and then update it on the client at runtime.

Setting default `resolved_color_mode` to the `resolvedTheme` outside of an effect means that the server and the client will render different a different initial DOM.

* Python Version: 3.11.8
* Reflex Version: main 958c4fa

<sub>From [SyncLinear.com](https://synclinear.com) | [REF-3165](https://linear.app/reflex-dev/issue/REF-3165/warning-text-content-did-not-match-server-client-system)</sub>

Lendemor:
fix the hydration that could occurs in some case following the PR #3507 

closes #3543

### DIFF
--- a/reflex/.templates/web/components/reflex/radix_themes_color_mode_provider.js
+++ b/reflex/.templates/web/components/reflex/radix_themes_color_mode_provider.js
@@ -5,7 +5,7 @@ import { ColorModeContext, defaultColorMode } from "/utils/context.js";
 export default function RadixThemesColorModeProvider({ children }) {
   const { theme, resolvedTheme, setTheme } = useTheme();
   const [rawColorMode, setRawColorMode] = useState(defaultColorMode);
-  const [resolvedColorMode, setResolvedColorMode] = useState(theme);
+  const [resolvedColorMode, setResolvedColorMode] = useState("dark");
 
   useEffect(() => {
     setRawColorMode(theme);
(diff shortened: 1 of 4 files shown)


reflex/.templates/web/components/reflex/radix_themes_color_mode_provider.js
```diff
@@ -5,7 +5,7 @@ import { ColorModeContext, defaultColorMode } from "/utils/context.js";
 export default function RadixThemesColorModeProvider({ children }) {
   const { theme, resolvedTheme, setTheme } = useTheme();
   const [rawColorMode, setRawColorMode] = useState(defaultColorMode);
-  const [resolvedColorMode, setResolvedColorMode] = useState(theme);
+  const [resolvedColorMode, setResolvedColorMode] = useState("dark");
```
picklelo, line 8: Is this right we want to default to dark?

picklelo (approved):

reflex/.templates/web/components/reflex/radix_themes_color_mode_provider.js
Lendemor: yes it's for the pre-rendering to be consistant and avoid hydration error.
It will not impact the actual render of the page in the browser.

PR #3547, opened 2024-06-24
German README.md 

unfortunatelyalex:
I've added a translation `docs/de` to include german, I've also added ` | [Deutsch](https://github.com/reflex-dev/reflex/blob/main/docs/de/README.md)` to display the availability of this translation in the other readme files

Please let me know if I have overlooked something or if there is a problem with my PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 
 ---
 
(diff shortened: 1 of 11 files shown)


Lendemor (approved):

Lendemor:
Thanks for the contribution.

PR #3548, opened 2024-06-24
[REF-2588]Clear color mode local storage for dev mode

### LINKED ISSUE #2047: Dev Mode Clear Dark Cookie
During dev mode sometimes when you are on another app and running dark mode stop that app and go to a new one the cookie for local host is cached, and your other app has dark mode. On every run dev, we should clear this cookie

ElijahAhianyo:
Sometimes an app's `appearance` value(`rx.App(theme=rx.theme(appearance="dark"))`) does not persist correctly in the local storage(as `chakra-ui-color-mode`). Eg, if you manipulate the local storage value and restart the app, the app's theme/color-mode is not respected. This is typically annoying in dev mode. 
We want to clear this local storage key(`chakra-ui-color-mode`) when the app is started or when the app is hot reloaded, in this case, the app's theme/appearance/color-mode will be persisted and respected.

We do this by keeping a timestamp of the last time the app was compiled and saving this value in the local storage. For cases where the local storage value of the timestamp is any other than what is generated in `context.js` at the time of compilation, we clear the local storage key(of `chakra-ui-color-mode`).

closes #2047 #3548


reflex/.templates/web/components/reflex/radix_themes_color_mode_provider.js
```diff
@@ -1,13 +1,31 @@
 import { useTheme } from "next-themes";
 import { useEffect, useState } from "react";
-import { ColorModeContext, defaultColorMode } from "/utils/context.js";
+import {
+  ColorModeContext,
+  defaultColorMode,
+  isDevMode,
+} from "/utils/context.js";
+import { lastCompiledTimeStamp } from "/utils/context.js";
```
masenf, line 8: i think this can go in the import list in the line above

masenf (approved):

PR #3566, opened 2024-06-25
[REF-3185][REF-3180]Dont escape backticks in JS string interpolation

ElijahAhianyo:
Javascript is not a happy person when you escape backticks in a string interpolation. Eg.

```js 
  const value = `Getting the value of ${ someValue[\`key\`]}`
``` 
when trying to format or escape strings with backticks, we want to exclude anything in `${}`.


masenf (approved):
sweet hack!

PR #3568, opened 2024-06-25
introduce customizable breakpoints

adhami3310:
### All Submissions:

- [ x Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/reflex-dev/reflex/pulls ) for the desired changed?

### Type of change

### New Feature Submission:

### DIFF
--- a/reflex/components/core/breakpoints.py
+++ b/reflex/components/core/breakpoints.py
@@ -1,9 +1,11 @@
 """Breakpoints utility."""
 
+from typing import Tuple
+
 breakpoints_values = ["30em", "48em", "62em", "80em", "96em"]
 
 
-def set_breakpoints(values: tuple[str, str, str, str, str]):
+def set_breakpoints(values: Tuple[str, str, str, str, str]):
     """Overwrite default breakpoint values.
 
     Args:
--- a/reflex/components/core/breakpoints.py
+++ b/reflex/components/core/breakpoints.py
@@ -1,6 +1,6 @@
 """Breakpoints utility."""
 
-breakpoints_values = ["520px", "768px", "1024px", "1280px", "1640px"]
+breakpoints_values = ["30em", "48em", "62em", "80em", "96em"]
 
 
 def set_breakpoints(values: tuple[str, str, str, str, str]):
--- a/reflex/components/core/breakpoints.py
+++ b/reflex/components/core/breakpoints.py
@@ -1,0 +1,60 @@
+from typing import Self
+
+
+breakpoints_values = ["520px", "768px", "1024px", "1280px", "1640px"]
+
+
+def set_breakpoints(values: tuple[str, str, str, str, str]):
+    breakpoints_values.clear()
+    breakpoints_values.extend(values)
+
+
+class Breakpoints(dict):
+    """A responsive styling helper."""
+
+    def __init__(self, mapping: dict):
+        super().__init__(mapping)
+
+    @classmethod
+    def create(
+        cls,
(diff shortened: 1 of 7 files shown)


adhami3310:
It's possible to be more strict in typings and validation, let me know if any is needed.

reflex/style.py
```diff
@@ -5,6 +5,7 @@
 from typing import Any, Literal, Tuple, Type
 
 from reflex import constants
+from reflex.components.core.breakpoints import Breakpoints, breakpoints_values
```
picklelo, line 8: I'm wondering if we can define the breakpoint values within the `Style` class itself (rather than having it be a global). That way users can set it in their global style dict:

```python
app = rx.App(style={"breakpoints": [...]})
```

Since the `convert` method is only used in the Style class, it can pass its breakpoints to it then.

reflex/components/core/breakpoints.py
```diff
@@ -0,0 +1,74 @@
...
+class Breakpoints(dict):
+    """A responsive styling helper."""
+
+    def __init__(self, mapping: dict):
+        """Instantiate breakpoints mapping.
+
+        Args:
+            mapping: a dictionary where breakpoints are keys and CSS values are values
+        """
+        super().__init__(mapping)
```
masenf, line 27: is this function necessary? or is it mostly for the docstring?

reflex/components/core/breakpoints.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @classmethod
+    def create(
+        cls,
+        custom: Optional[dict] = None,
+        initial=None,
+        xs=None,
```
picklelo, line 34: I think we should keep this as is, but one thing that's a bit confusing is the breakpoints we size as sm, md, lg - whereas the radix core components do sizing as 1, 2, 3. I prefer the former as it's more descriptive, but there is an inconsistency.

masenf (dismissed):

reflex/components/core/breakpoints.py
adhami3310: It's used in style.py once. (and in, that file, but that can be replaced)

masenf: i just meant since it's inherited from the base class `dict`, it should still work even if its not overridden here.

masenf:
in testing, i actually noticed that `rx.breakpoints` is not recognized in vscode, i think for whatever reason the `reflex/__init__.pyi` did not get regenerated.

i tried running `poetry run python scripts/make_pyi.py reflex/__init__.py` myself, but i don't actually see a diff in `reflex/__init__.pyi` 🤔 seems like there is something wrong with the pyi generator, i'll look into it

picklelo (changes requested):
Let's try to move the parameter into the `Style` class and I think this is good.

reflex/components/core/breakpoints.py
picklelo: Radix also uses sm md lg for their responsive props, so it is consistent with them.

reflex/components/core/breakpoints.py
adhami3310: you are correct, will fix this then

reflex/style.py
adhami3310: @picklelo  I'm trying to implement this, but it seems that `style=` is processed after processing the `rx.breakpoints`. This means the modification happens after it already generates the CSS for the element.

picklelo (approved):
Looks good to me!

PR #3573, opened 2024-06-26
add computed backend vars

benedikt-bartscher:


Lendemor (approved):

masenf (commented):
for consistency with the normal backend vars would it make sense to treat any @rx.var where the fget starts with an underscore as backend and not allow users to mark a non-underscore var as backend?

oops i accidentally merged your later PRs that introduced conflicts in this one

benedikt-bartscher:
@masenf no worries, I will fix the conflicts. To be honest, I do not really like the magic underscore backend vars. But I can implement it. Actually there is some code which did this before and I did not clean it up properly.

benedikt-bartscher:
I will clean this one up once #3587 is merged, I would appreciate a fast merge of #3587

masenf:
> I would appreciate a fast merge of https://github.com/reflex-dev/reflex/pull/3587

For sure. i think our github actions is kind of backed up today, but i'll merge it as soon as we see the check.

I'm trying to cut 0.5.6a1 in the next 2 hours, and i'd like these pieces to be included. There won't be a Monday release on Jul 1, but it will be good to integrate the variety of changes we have and test more with some of our more exotic apps.

benedikt-bartscher:
Yeah, noticed that as well... Thanks again for your fast reviews. I will try to get this one ready for 0.5.6a1.
As I am writing this: ci for #3587 has passed! :heavy_check_mark:

masenf:
> I do not really like the magic underscore backend vars

To this point, we're cooking up a Var refactor that (in an ideal world) should allow for a plain class to be used as a state class provided that all of the Vars are declared as `rx.Var` / `rx.BackendVar` and all event handlers are decorated with `@rx.EventHandler`.

We will still keep the `rx.State` around for compatibility, but if users want to opt-out of the "magic" of the current `rx.State` implementation, I want a clean way for them to do so.

benedikt-bartscher:
> To this point, we're cooking up a Var refactor that (in an ideal world) should allow for a plain class to be used as a state class provided that all of the Vars are declared as `rx.Var` / `rx.BackendVar` and all event handlers are decorated with `@rx.EventHandler`.

I really like the explicit decorator-based approach. It would simplify the codebase and improve typing by a lot. I think this will actually avoid some confusion for reflex newcomers. I've heard a lot of feedback where people are confused about the "magic".
This would also improve compatibility with existing code, other libs and `rx.State`, caused by naming constraints. With decorator only, I could finally write PEP8 "compliant" code again, where `_` should mean internal use.

benedikt-bartscher:
> We will still keep the `rx.State` around for compatibility, but if users want to opt-out of the "magic" of the current `rx.State` implementation, I want a clean way for them to do so.

How about using ast to add the missing decorators to existing code automatically? This would allow us to move faster and drop some dirty legacy code

masenf (approved):

PR #3578, opened 2024-06-27
classvars should not be backend vars

benedikt-bartscher:
Fixes issue where `_mixin` is propagated as backend var with state mixins and ComponentState


masenf (approved):
thanks for the nice fix!

PR #3583, opened 2024-06-27
add responsive props to radix components

adhami3310:
### All Submissions:

### Type of change

### New Feature Submission:


reflex/components/core/breakpoints.py
```diff
@@ -15,19 +16,41 @@ def set_breakpoints(values: Tuple[str, str, str, str, str]):
...
         cls,
-        custom: Optional[dict] = None,
-        initial=None,
-        xs=None,
-        sm=None,
-        md=None,
-        lg=None,
-        xl=None,
+        custom: Optional[dict[K, V]] = None,
+        initial: Optional[V] = None,
```
picklelo, line 48: nit: You can do `V | None` instead of `Optional` for type annotations. Only within the pydantic class definitions do we need to keep it using the old syntax

reflex/utils/types.py
```diff
@@ -351,6 +355,33 @@ def _issubclass(cls: GenericType, cls_check: GenericType) -> bool:
     if isinstance(cls_base, tuple):
         return False
 
+    # Check that fields of breakpoints match the expected values.
+    if issubclass(cls_base, Breakpoints) and instance is not None:
+
+        def satisfies(cls_check: GenericType):
```
picklelo, line 361: If possible we should pull this function out to the top level - otherwise it will get redefined every time we call `issubclass` and may have performance impact

reflex/components/core/breakpoints.py
adhami3310: The union operator I believe is only compatible with Python 3.10, but we have compatibility with 3.8 so I'm not sure we can do that?

masenf: it works, but you have to ensure the first non-docstring line of the module is `from __future__ import annotations`. This enables the new typing features in 3.7+

adhami3310:
One behavior we're missing is using list operator to define responsive props for radix components. Is this something we're interested in having? @picklelo

reflex/components/core/breakpoints.py
benedikt-bartscher: Should we enable this automatically? I am using this in all my reflex projects:

```toml
[tool.ruff.lint.isort]
required-imports = ["from __future__ import annotations"]
```

adhami3310:
In the interest of easing merge requests, we can merge this and add list syntax radix props in another PR.

reflex/components/radix/themes/components/avatar.pyi
```diff
@@ -20,17 +22,29 @@ class Avatar(RadixThemesComponent):
...
+                            str, Literal["1", "2", "3", "4", "5", "6", "7", "8", "9"]
+                        ],
+                    ]
+                ],
+                Union[
+                    Literal["1", "2", "3", "4", "5", "6", "7", "8", "9"],
+                    reflex.components.core.breakpoints.Breakpoints[
+                        str, Literal["1", "2", "3", "4", "5", "6", "7", "8", "9"]
+                    ],
+                ],
```
masenf, line 42: not clear why this is generating a separate internal `Union`, since there is already an "outer" `Union` wouldn't the following code work the same:

```suggestion
                Literal["1", "2", "3", "4", "5", "6", "7", "8", "9"],
                reflex.components.core.breakpoints.Breakpoints[
                    str, Literal["1", "2", "3", "4", "5", "6", "7", "8", "9"]
                ],
```

masenf (commented):
does this _actually_ work if the `Breakpoints` is inside a `Var`? for example in state:

```python
class S(rx.State):
    bps: rx.breakpoints.__self__ = rx.breakpoints(
        initial="1",
        xs="2",
        sm="2",
        md="3",
        lg="4",
        xl="5",
    )

def index():
    return rx.vstack(
        "hello world",
        "foo",
        "bar",
        "baz",
        spacing=S.bps,
    )
```

when i try it like that it kind of blows up

```console
  File "/Users/masen/code/reflex-dev/reflex/reflex/components/component.py", line 811, in create
  File "/Users/masen/code/reflex-dev/reflex/reflex/components/component.py", line 444, in __init__
    and not types._issubclass(passed_type, expected_type, value)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masen/code/reflex-dev/reflex/reflex/utils/types.py", line 392, in _issubclass
    return _breakpoints_satisfies_typing(cls_check, instance)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masen/code/reflex-dev/reflex/reflex/utils/types.py", line 352, in _breakpoints_satisfies_typing
    return any(
           ^^^^
  File "/Users/masen/code/reflex-dev/reflex/reflex/utils/types.py", line 353, in <genexpr>
    _breakpoints_satisfies_typing(type_to_check, instance)
  File "/Users/masen/code/reflex-dev/reflex/reflex/utils/types.py", line 346, in _breakpoints_satisfies_typing
    for value in instance.values():
                 ^^^^^^^^^^^^^^^
  File "/Users/masen/code/reflex-dev/reflex/reflex/vars.py", line 819, in __getattr__
    raise VarAttributeError(
reflex.utils.exceptions.VarAttributeError: The State var `state__s.bps` has no attribute 'values' or may have been annotated wrongly.
```

i think, at least initially, we need to ensure that the `Breakpoints` is handled at compile time (i.e. cannot be in a dynamic frontend Var), or we need to ignore the type checking if the `Breakpoints` is in a `Var`.

I tend to lean toward the former, because we definitely cannot support `Var[Breakpoint]` for the CSS props as it's written now, so we shouldn't introduce an inconsistency between the two.

My other question is about `Var` within an `rx.breakpoints` dict, the type checking seems to get bungled up on this, but since the style props have no type checking, it actually _does_ seem to work if the value comes from a `Var`

reflex/components/radix/themes/components/avatar.pyi
adhami3310: That's interesting, I'm not sure how it generated this piece of code. I will take a look into it.

adhami3310:
> does this _actually_ work if the `Breakpoints` is inside a `Var`?

Good catch! I think the issue I ran through was doing a union simply broke pydantic and I wasn't sure how, so that was the only thing that typechecked fine. I will see if it can be fixed more towards the right way.

adhami3310:
It actually seems fine, we can handle breakpoints in var after fixing an issue with typing. But we do lose validation when used in a state.

masenf (approved):
will take this for now, but i have a followup PR to clean up some of the generated .pyi files coming

PR #3585, opened 2024-06-27
order type annotations in pyi_generator

benedikt-bartscher:


masenf (approved):
thanks for the fix. so confused about why this is hitting us now though... not sure what actually changed to cause these pyi mismatches. any idea?

benedikt-bartscher:
You're welcome. No, tbh, I do not know why this was hitting us so randomly. However, I observed this issue for quite a while now. It got a bit better with my differential pyi generator PR.

PR #3591, opened 2024-06-29
pyi_generator cleanup: ruff, remove fully qualified reflex. names

masenf:
* Remove black dependency; format and fix generated files using `ruff check` and `ruff format` now
  * all unused imports are removed and duplicate imports are collapsed and ordered correctly, which reduces file size and line count of shipped pyi files
* Replace special-case global import logic with creation of ast nodes from `DEFAULT_IMPORTS`
* Remove fully-qualified `reflex.` names in pyi files. The generator now raises an error when trying to resolve a type hint that is not in the namespace of the pyi file, and asks the user to add it to `DEFAULT_IMPORTS`
* Replace `function` with `Callable` in the event trigger annotation, since `function` is not actually a valid builtin name

pyi_generator changes only https://github.com/reflex-dev/reflex/pull/3591/files/780a36925bd3e6c29af0d5d6279f73d973f45a6c


Alek99 (approved):
nice, glad we can switch fully to this

PR #3592, opened 2024-06-29
[REF-3076] Do not purge .web/pages in dev mode

### LINKED ISSUE #3416: [REF-3076] 404 with recomplilation when working on routes
When working with pages that are routes auto reloading leads to 404 errors that cannot be recovered without restarting reflex.

**To Reproduce**
I have an event hander in the state for the below page. If i change an invalid variable name to a correct one i can force throwing an attribute error. Switching between valid and invalid 2-3’times throws a 404 and reloading doesnt work. Restarting reflex does

This is my route:
`@rx.page(route="/entity/[entity_id]/")
def entity_editor():`

**Expected behavior**
Editing pages should result in reloading a working reflex app

**Screenshots**

 - Python Version: 3.12
 - Reflex Version: 0.52
 - OS: osx 13.3.1
 - Browser (Optional): firefox
 
**Additional context**
Generally happens often with all sorts of editing, but reporducible with the above

<sub>[REF-3076](https://linear.app/reflex-dev/issue/REF-3076/404-with-recomplilation-when-working-on-routes)</sub>

masenf:
Avoid race condition in hot reload that occurs when pages are removed before they are recreated.

Fix #3416

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -14,6 +14,7 @@ import os
 import platform
 import sys
 from datetime import datetime
+from pathlib import Path
 from typing import (
     Any,
     AsyncIterator,
@@ -1055,9 +1056,6 @@ class App(LifespanMixin, Base):
 
         progress.advance(task)
 
-        # Empty the .web pages directory.
-        compiler.purge_web_pages_dir()
-
         progress.advance(task)
         progress.stop()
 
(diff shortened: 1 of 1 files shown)


benedikt-bartscher:
@masenf shouldn't this be REF-3076 instead of REF-3067?

picklelo (approved):
nice fix

Issue #1529, opened 2023-08-05
Consider tailwindcss/typography as default for rx.markdown and rx.html components


masenf:
**Describe the bug**
Rendering raw HTML can have unexpected results because tailwind and in some cases chakra kills the standard formatting of headings and other common tags. The markdown situation is a little better, because reflex automatically maps common tags to reflex components which are styled appropriately.

A general solution to this problem is to use tailwindcss/typography plugin via `class_name="prose"`, which brings back some standard formatting rules. We should consider this as the default for the `rx.html` component.

**To Reproduce**
```python
import reflex as rx

class State(rx.State):
    pass


def index() -> rx.Component:
    return rx.html("""
        <h1>Foo</h1>
        <h2>Bar</h2>
        <p>This text is <em>italic</em> and this is <strong>bold</strong></p>""",
        class_name="prose")


app = rx.App()
app.add_page(index)
app.compile()
```

<img width="473" alt="image" src="https://github.com/reflex-dev/reflex/assets/1524005/026ecc7b-b7b9-4e12-8b62-2ac534c1bd44">

With typography added (`rxconfig.py`):

```python
import reflex as rx

class ReprotypographyConfig(rx.Config):
    pass

config = ReprotypographyConfig(
    app_name="repro_typography",
    tailwind={
        "plugins": ["@tailwindcss/typography"],
    },
    frontend_packages=["@tailwindcss/typography"],
)
```

It looks much nicer

<img width="473" alt="image" src="https://github.com/reflex-dev/reflex/assets/1524005/70b0d89b-64d7-4246-998a-28001bde3ea8">

Issue #613, opened 2023-02-27
pc.html does not pay attention to tags


hirscr:
**Describe the bug**
header tags like `<h1></h1>` have no effect on the text between the tabs

**To Reproduce**
you can just look at your own docs on website
 - Code/Link to Repo:](https://pynecone.io/docs/library/other/html)

**Expected behavior**
header tags should produce bolded and englarged text

**Screenshots**

** Specifics (please complete the following information):**
 - Python Version: 3.9
 - Pynecone Version: latest
 - OS:  mac
 - Browser (Optional): safari, chrome
 
**Additional context**

Kaushal-Dhungel:
Hi, Can I work on this issue?

Alek99:
> Hi, Can I work on this issue?

Yes you can work on this issue. Thanks!

masenf:
The problem here is that reflex has a CSS reset which removes all distinctions between headings. So unless your site is using tailwind typography or setting your own CSS rules for the headings, they are going to appear visually identical.

kylelee1027:
Hello, can I work on this?

picklelo:
This is still an issue, but not sure what the solution is. As mentioned above the css is getting reset within the Reflex app, so not sure how to get the original one back.

masenf:
This has been attempted a few times in the past. See https://github.com/reflex-dev/reflex/issues/1529

In that issue i have an example code which restores "nice looking" default styles for headers and other components via the tailwind/typography package and the `prose` class.

If you're planning to fix this issue, then i would suggest a similar route as described in that bug.

kylelee1027:
I thought that maybe adding the typography plugin as a default value for tailwind in the reflex config would work but it didn't :( or maybe I just implemented it wrong?

masenf:
You also need to add the `class_name="prose"` to the `rx.html` as a default. I think a recent PR just added `rx-Html` as another default class name, so you should be able to piggy back on that.

PR #3593, opened 2024-06-29
Use tailwind typography plugin by default

minimav:
This PR addresses #1529 to make the HTML component display tags correctly by default via the Tailwind typography plugin as suggested on the issue itself and using the `prose` class as mentioned [here](https://github.com/reflex-dev/reflex/issues/613#issuecomment-2080271098).

Example of this working:

Screenshot 2024-06-29 at 15 28 19

via this:

Screenshot 2024-06-29 at 15 28 37

-------

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:


Lendemor:
I don't have an issue about the code itself, I'm just not sure if we should have the `tailwindcss/typography` as a default for every project, rather than enabling it when rx.html is used.

masenf:
@Lendemor is there a perceived downside to bringing in the typography plugin for every project? Slightly large download size perhaps, but i think in deploy/prod mode the extra classes will be pruned out if `prose` is not actually used anywhere.

Lendemor:
> @Lendemor is there a perceived downside to bringing in the typography plugin for every project? Slightly large download size perhaps, but i think in deploy/prod mode the extra classes will be pruned out if `prose` is not actually used anywhere.

Probably nothing big, I just remembered some users disabling tailwind in their projects (we had it included as default before, then we made it optional at some point) because it was conflicting with some others stuff they were using I think.

If we enable it forcefully again, it might break these users apps? I'm not sure how many users this is tho, so maybe it's not a big deal.

masenf:
> If we enable it forcefully again

If user sets `tailwind=None` in their config, then they will not get the plugins either, so i think it's safe

adhami3310 (approved):
Sorry for sleeping on it 😅

PR #3597, opened 2024-06-30
do not get_config in global scope

### LINKED ISSUE #3596: Importing module containing State classes in rxconfig.py breaks reflex
Introduced in https://github.com/reflex-dev/reflex/pull/3523. Caused by `config = get_config()` call in global context.

benedikt-bartscher:
Closes #3596


masenf (approved):
Thanks for the fix

PR #3600, opened 2024-06-30
dedupe config overrides

### LINKED ISSUE #3599: Dedupe env config override logs
```log
Info: Overriding config value telemetry_enabled with env var TELEMETRY_ENABLED=false
Info: Overriding config value next_compression with env var NEXT_COMPRESSION=false
Info: Overriding config value telemetry_enabled with env var TELEMETRY_ENABLED=false
Info: Overriding config value next_compression with env var NEXT_COMPRESSION=false
──────────────────────────────────────────────────────────────────────────── Starting Reflex App ─────────────────────────────────────────────────────────────────────────────
[20:06:07] Compiling: ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 13/13 0:00:00
──────────────────────────────────────────────────────────────────────────────── App Running ─────────────────────────────────────────────────────────────────────────────────
App running at: http://localhost:3000
Backend running at: http://0.0.0.0:8000
Info: Overriding config value telemetry_enabled with env var TELEMETRY_ENABLED=false
Info: Overriding config value next_compression with env var NEXT_COMPRESSION=false
```

benedikt-bartscher:
Closes #3599

### DIFF
--- a/reflex/config.py
+++ b/reflex/config.py
@@ -277,7 +277,8 @@ class Config(Base):
             if env_var is not None:
                 if key.upper() != "DB_URL":
                     console.info(
-                        f"Overriding config value {key} with env var {key.upper()}={env_var}"
+                        f"Overriding config value {key} with env var {key.upper()}={env_var}",
+                        dedupe=True,
                     )
 
                 # Convert the env var to the expected type.
--- a/reflex/utils/console.py
+++ b/reflex/utils/console.py
@@ -17,6 +17,9 @@ _LOG_LEVEL = LogLevel.INFO
 # Deprecated features who's warning has been printed.
 _EMITTED_DEPRECATION_WARNINGS = set()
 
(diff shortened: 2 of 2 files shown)


Lendemor:
I was thinking the message for config override would be more pertinent at the DEBUG level rather than INFO, what do you think?

benedikt-bartscher:
I actually like that this is INFO, and with this PR it's not that disturbing, because it's only shown once on startup. But if you and your team like DEBUG more, i can adjust it for you.

masenf (approved):
i think INFO is fine... if not we can adjust that in a future PR, since it was already at INFO currently

Lendemor:
Hmm, true, with dedupe, info will be less spammy yeah.

Guess it's fine like this 👍

Lendemor (approved):

PR #3607, opened 2024-07-02
add immutable var class

adhami3310:
> The `ImmutableVar` is a dataclass that inherits from `Var`, very similar to the current `BaseVar`, except without any setter or default value logic and with `frozen=True` so that it is immutable.

### DIFF
--- a/reflex/experimental/__init__.py
+++ b/reflex/experimental/__init__.py
@@ -8,6 +8,7 @@ from reflex.components.sonner.toast import toast as toast
 
 from ..utils.console import warn
 from . import hooks as hooks
+from . import vars as vars
 from .assets import asset as asset
 from .client_state import ClientStateVar as ClientStateVar
 from .layout import layout as layout
--- a/reflex/experimental/__init__.py
+++ b/reflex/experimental/__init__.py
@@ -42,6 +42,7 @@ _x = ExperimentalNamespace(
     asset=asset,
     client_state=ClientStateVar.create,
     hooks=hooks,
+    vars=vars,
     layout=layout,
     progress=progress,
     PropsBase=PropsBase,
--- a/reflex/experimental/vars/__init__.py
+++ b/reflex/experimental/vars/__init__.py
@@ -1,0 +1,1 @@
+"""Experimental Immutable-Based Var System."""
--- a/reflex/experimental/vars/__init__.py
+++ b/reflex/experimental/vars/__init__.py
@@ -1,1 +1,3 @@
 """Experimental Immutable-Based Var System."""
+
+from .base import ImmutableVar as ImmutableVar
--- a/reflex/experimental/vars/base.py
+++ b/reflex/experimental/vars/base.py
@@ -2,9 +2,8 @@
 
 import dataclasses
 import sys
-from typing import Any, Literal, Optional, Self, Type, Union, get_args, get_origin
+from typing import Any, Optional, Self, Type
 
-from reflex.utils import types
 from reflex.vars import Var, VarData
 
 
--- a/reflex/experimental/vars/base.py
+++ b/reflex/experimental/vars/base.py
@@ -2,7 +2,7 @@
 
 import dataclasses
 import sys
-from typing import Any, Literal, Optional, Type, Union, get_args, get_origin
+from typing import Any, Literal, Optional, Self, Type, Union, get_args, get_origin
 
 from reflex.utils import types
 from reflex.vars import Var, VarData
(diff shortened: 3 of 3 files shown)


reflex/experimental/vars/base.py
```diff
@@ -0,0 +1,73 @@
...
+    # Whether the var is a string literal.
+    _var_is_string: bool = dataclasses.field(default=False)
+
+    # _var_full_name should be prefixed with _var_state
+    _var_full_name_needs_state_prefix: bool = dataclasses.field(default=False)
+
+    # Extra metadata associated with the Var
+    _var_data: Optional[VarData] = dataclasses.field(default=None)
+
+    def get_default_value(self) -> Any:
```
masenf, line 32: i think we can leave this function out here, as this functionality is only used in State classes (and IMO is a bit of a kludge that we should avoid bringing into the new API)

masenf (changes requested):
needs an overridden `_replace` method that returns `ImmutableVar` instances.

it actually might make sense to override `create` as well to something like this

```python
@classmethod
def create(self, ...):
    return super().create(...)._replace()
```

That way we can create `ImmutableVar` instances in the same way as `Var` instances until more special-case behavior is added.

reflex/experimental/vars/base.py
adhami3310: should be resolved!

reflex/experimental/vars/base.py
```diff
@@ -0,0 +1,135 @@
...
+    _var_type: Type = dataclasses.field(default=Any)
+
+    # Whether this is a local javascript variable.
+    _var_is_local: bool = dataclasses.field(default=False)
+
+    # Whether the var is a string literal.
+    _var_is_string: bool = dataclasses.field(default=False)
+
+    # _var_full_name should be prefixed with _var_state
+    _var_full_name_needs_state_prefix: bool = dataclasses.field(default=False)
```
masenf, line 33: Lets remove these fields from `ImmutableVar`, and implement them as `@property` functions that return `False`. This will let us work with the existing `Var` API without bringing along the API kludge from the old system.

Similarly in `create` we can accept these as kwargs, but raise `TypeError` if they are passed in.

In `_replace`, you can check something like `if kwargs.pop("_var_is_local", self._var_is_local) != self._var_is_local: raise TypeError("passing _var_is_local is deprecated")` for each of the removed fields.

With the new Var system, the formatting of the Var should not depend on these kinds of flags, but instead each `ImmutableVar` subclass will know how to format itself. The basic `ImmutableVar` should pretty much just use `_var_name` as its `__str__`. Any type of value which needs to be directly translated to JS code can use the base `ImmutableVar` and it should render similarly to a `BaseVar` with `_var_is_local=True` and `_var_is_string=False`.

masenf (changes requested):

reflex/experimental/vars/base.py
```diff
@@ -0,0 +1,182 @@
...
+            merge_var_data: VarData to merge into the existing VarData.
+            **kwargs: Var fields to update.
+
+        Returns:
+            A new ImmutableVar with the updated fields overwriting the corresponding fields in this Var.
+
+        Raises:
+            TypeError: If _var_is_local, _var_is_string, or _var_full_name_needs_state_prefix is not None.
+        """
+        if kwargs.get("_var_is_local", None) is not None:
```
masenf, line 71: Lets get the default as `False` and raise the error if it's _not_ `False`. Some of the existing legacy var operations will call into `_replace` and if they don't pass conflicting values for these deprecated fields, then they should still work.

That will allow us to zero in on the legacy operations that are toggling these flags to control rendering, but keep most of the operations, which don't toggle these flags, to keep working and be tested with the new `ImmutableVar`

reflex/experimental/vars/base.py
```diff
@@ -0,0 +1,182 @@
...
+
+        field_values = dict(
+            _var_name=kwargs.pop("_var_name", self._var_name),
+            _var_type=kwargs.pop("_var_type", self._var_type),
+            _var_is_local=kwargs.pop("_var_is_local", self._var_is_local),
+            _var_is_string=kwargs.pop("_var_is_string", self._var_is_string),
+            _var_full_name_needs_state_prefix=kwargs.pop(
+                "_var_full_name_needs_state_prefix",
+                self._var_full_name_needs_state_prefix,
+            ),
```
masenf, line 94: these should be removed now, since passing them to the constructor would always raise a TypeError

reflex/experimental/vars/base.py
```diff
@@ -0,0 +1,182 @@
...
+            console.deprecate(
+                feature_name=f"Creating a Var ({value}) from a string without specifying _var_is_string",
+                reason=(
+                    "Specify _var_is_string=False to create a Var that is not a string literal. "
+                    "In the future, creating a Var from a string will be treated as a string literal "
+                    "by default."
+                ),
+                deprecation_version="0.5.4",
+                removal_version="0.6.0",
+            )
```
masenf, line 176: this logic can be removed since we don't allow passing `_var_is_string` anymore

masenf (changes requested):

Lendemor (dismissed):

reflex/experimental/__init__.py
```diff
@@ -42,6 +42,7 @@ def toast(self):
     asset=asset,
     client_state=ClientStateVar.create,
     hooks=hooks,
+    vars=vars,
```
masenf, line 45: need to add `from . import vars as vars`, otherwise this name is the builtin `vars`

```python
>>> rx._x.vars
<built-in function vars>
```

reflex/experimental/vars/__init__.py
```diff
@@ -0,0 +1 @@
+"""Experimental Immutable-Based Var System."""
```
masenf, line 1: we should also bring `ImmutableVar` into the namespace here via

```
from .base import ImmutableVar as ImmutableVar
```

ultimately we want to be able to write something like `rx._x.ImmutableVar.create(...)`

masenf (changes requested):
we can hold off on adding tests for now, because most of the API for `ImmutableVar` is coming from `Var` itself and won't represent the final usage of the API.

but in a near-future PR, we will want to make sure these changes are coming in with near-full test coverage, because it's much easier to add tests incrementally, rather than at the end.

masenf (approved):

PR #3613, opened 2024-07-03
fix sqla python_type issues, add tests

benedikt-bartscher:
This contains a fix for custom sqlalchemy types without `python_type` implemented. Since `get_attribute_access_type` is already pretty big, I will probably write a followup PR to split it into multiple functions.


masenf (approved):

abulvenz:
:heart_decoration:

PR #3617, opened 2024-07-03
[REF-3225] implement __format__ for immutable vars

adhami3310:
> For new-style immutable vars, the format string mechanism will use a tagged marker that can be used to reconstruct the original Var from the formatted string.


reflex/experimental/vars/base.py
```diff
@@ -28,6 +29,10 @@ class ImmutableVar(Var):
     # Extra metadata associated with the Var
     _var_data: Optional[VarData] = dataclasses.field(default=None)
 
+    def __post_init__(self):
+        """Post-initialization."""
+        _global_vars[hash(self)] = self
```
masenf, line 34: my initial thought was to store the reference only in `__format__`, to avoid keeping unused references to Vars that never actually get formatted. Saving refs to all Vars could cause memory use issues where Vars go out of scope, but cannot be garbage collected because they are in this global map.

tests/test_var.py
```diff
@@ -835,6 +839,30 @@ def test_state_with_initial_computed_var(
...
+    )
+
+    var_with_data = var_without_data._replace(merge_var_data=original_var_data)
+
+    f_string = f"foo{var_with_data}bar"
+
+    assert REFLEX_VAR_OPENING_TAG in f_string
+    assert REFLEX_VAR_CLOSING_TAG in f_string
+
+    result_var_data = Var.create_safe(f"foo{var_with_data}bar")._var_data
```
masenf, line 859:
```suggestion
    result_var_data = Var.create_safe(f_string)._var_data
```

reflex/vars.pyi
```diff
@@ -35,6 +35,9 @@ USED_VARIABLES: Incomplete
 
 def get_unique_variable_name() -> str: ...
 def _encode_var(value: Var) -> str: ...
+
+_global_vars: Dict[int, Var] = {}
```
masenf, line 39:
```suggestion
_global_vars: Dict[int, Var]
```

assignments can and should be left off for `.pyi` files

masenf (commented):
code looks good, going to test it now

masenf (changes requested):
Currently the `ImmutableVar` cannot be created from an f-string with vars inside of it

```python
import reflex as rx


class S(rx.State):
    foo: str = "foo"


# Create an new style Var with data.
iv = rx._x.vars.ImmutableVar.create("foo", _var_data=S.foo._var_data)

# Create an f-string using the new style Var.
f_s = f"{iv}bar"

# Create an old style Var from the f-string.
n_v = rx.Var.create(f_s)
print(n_v, repr(n_v))

# Create a new style Var from the f-string.
n_iv = rx._x.vars.ImmutableVar.create(f_s)
print(n_iv, repr(n_iv))
```

It seems there is a bit of code in `rx.Var.__post_init__` which is trying to assign to `self._var_name` and `self._var_data`, but this does not work for immutable instances.

Instead we can change the code to call `__init__` again with the updated values, and this should be _fine_ in nearly every case. It sort of breaks immutability, but if we consider `__post_init__` to be just an extension of `__init__`, then it makes sense.

```diff
diff --git a/reflex/vars.py b/reflex/vars.py
index ad7c223a..f55aa025 100644
--- a/reflex/vars.py
+++ b/reflex/vars.py
@@ -502,8 +502,10 @@ class Var:
         # Decode any inline Var markup and apply it to the instance
         _var_data, _var_name = _decode_var(self._var_name)
         if _var_data:
-            self._var_name = _var_name
-            self._var_data = VarData.merge(self._var_data, _var_data)
+            self.__init__(
+                _var_name=_var_name,
+                _var_data=VarData.merge(self._var_data, _var_data),
+            )
 
     def _replace(self, merge_var_data=None, **kwargs: Any) -> BaseVar:
         """Make a copy of this Var with updated fields.
```

tests/test_var.py
```diff
@@ -835,6 +839,30 @@ def test_state_with_initial_computed_var(
...
+    )
+
+    var_with_data = var_without_data._replace(merge_var_data=original_var_data)
+
+    f_string = f"foo{var_with_data}bar"
+
+    assert REFLEX_VAR_OPENING_TAG in f_string
+    assert REFLEX_VAR_CLOSING_TAG in f_string
+
+    result_var_data = Var.create_safe(f_string)._var_data
```
masenf, line 859: can we also roundtrip with `ImmutableVar.create_safe(f_string)` as well, to make sure the new `__post_init__` stuff works

masenf (approved):

PR #3624, opened 2024-07-04
rx._x.asset improvements

### LINKED ISSUE #3623: Multiple rx._x.asset issues
- Assets should not be checked and copied in backend only mode. Even if you do not put them in globals, they will for example be executed if you use them in a `@rx.page` which will be loaded. Even in backend only mode.
- Local/Normal asset support - Assets which are in the local `assets/` directory should not be copied to `assets/external`. But we can still validate that the asset exists and return the serving path.

@abulvenz

benedikt-bartscher:
Closes #3623


abulvenz:
@benedikt-bartscher I've had a look at the changes and slept one night on it :)
We have two different use cases that both deal with the asset folder:
- The file an URL is created for is already in the asset folder. In this case the function prepends the path with `"/assets/"` and the presence of the file is checked.
- The external asset use case where the file is located relative to the including `.py` file and copied only when `not ENV_BACKEND_ONLY` and the resulting file is returned as link.

On my option the automatic detection can be removed and then immediately raise an exception when `shared=True`, but the file is not located relative to the caller. A file name collision is very unlikely especially because `assets/external` is on `.gitignore`
Otherwise I think the implementation is cool.

One question: Is it wanted, that missing file presence of local assets should raise when `ENV_BACKEND_ONLY`? I thought that might be the case where no assets folder is present at all?

reflex/experimental/assets.py
```diff
@@ -1,57 +1,90 @@
...
 
-    dst_file = asset_folder / relative_filename
+        dst_file = asset_folder / path
 
-    if not dst_file.exists():
-        dst_file.symlink_to(src_file)
+        if not dst_file.exists() and (
+            not dst_file.is_symlink() or dst_file.resolve() != src_file_shared.resolve()
+        ):
+            dst_file.symlink_to(src_file_shared)
```
masenf, line 88: this makes me a bit nervous because symlinks are not generally enabled or available to unprivileged users on windows.

can we have a fallback path that just copies the files on windows or if this operation fails

reflex/experimental/assets.py
```diff
@@ -1,57 +1,90 @@
...
 from typing import Optional
 
 from reflex import constants
 
 
-def asset(relative_filename: str, subfolder: Optional[str] = None) -> str:
-    """Add an asset to the app.
+def asset(
+    path: str,
+    shared: bool = False,
```
masenf, line 13: the `shared` argument is a little bit confusing to me... is `shared=False` for referencing files that are already in the app's `assets/` directory?

i suppose this is to align the API to work with source-relative assets as well as assets that are already in the correct place? But setting the default to `False` seems to break how existing callers are already using this function to reference source-relative assets.

reflex/experimental/assets.py
benedikt-bartscher: this PR did not introduce the symlinks, they were there before. i can address it in another PR

reflex/experimental/assets.py
benedikt-bartscher: `shared=True` ist for librarys which want to expose the assets to other python modules/reflex apps. Shared mode expects the files to be relative to the python source file where `rx._x.asset` is called.

`shared=False` is for "normal"/"local" assets in reflex. It just returns the serving path and checks if the file really exits.

reflex/experimental/assets.py
```diff
@@ -1,57 +1,90 @@
...
     Raises:
         FileNotFoundError: If the file does not exist.
+        ValueError: If subfolder is provided for local assets.
 
     Returns:
-        The relative URL to the copied asset.
+        The relative URL to the asset.
     """
+    assets = constants.Dirs.APP_ASSETS
+    backend_only = os.environ.get(constants.ENV_BACKEND_ONLY)
```
maximvlah, line 51: @benedikt-bartscher hey Bene, I found a minor bug - here you will get either a string or a None which means when you are checking for 

```python

if not backend_only:
   ...
 ```
if the env var is "False", it will be evaluated as True

so I suggest doing sth like:

```python

backend_only = os.environ.get(constants.ENV_BACKEND_ONLY, "false").lower() == "true"

```

maximvlah (changes requested):

reflex/experimental/assets.py
abulvenz: @maximvlah Thank you for analyzing it down to the root.

Lendemor:
Marked this as draft until re-synced with main.

Lendemor (approved):

masenf (approved):
@benedikt-bartscher i updated this PR to move `asset` out of the experimental namespace with the new API (`shared=False` default). The existing `rx._x.asset` has been modified to call the new API with the previous default to avoid breaking any code.

The old `_x` name will be removed in a few weeks when 0.7.0 is released.

benedikt-bartscher:
@masenf that's awesome, thank you very much. I will rebase our fork and report if everything still works

PR #3636, opened 2024-07-09
feat: Adding an event to go back just as the user would.

abulvenz:
```python
# This will do the same as if the user presses the back button.
rx.button("Go back", on_click=rx.event.back())

```

### DIFF
--- a/reflex/event.py
+++ b/reflex/event.py
@@ -515,7 +515,7 @@ def back() -> EventSpec:
     Returns:
         An event to go back one page.
     """
-    return server_side("_back", get_fn_signature(back))
+    return call_script("window.history.back()")
 
 
 def window_alert(message: str | Var[str]) -> EventSpec:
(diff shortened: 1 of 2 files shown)


benedikt-bartscher (approved):

reflex/event.py
```diff
@@ -509,6 +509,15 @@ def console_log(message: str | Var[str]) -> EventSpec:
     return server_side("_console", get_fn_signature(console_log), message=message)
 
 
+def back() -> EventSpec:
+    """Do a history.back on the browser.
+
+    Returns:
+        An event to go back one page.
+    """
+    return server_side("_back", get_fn_signature(back))
```
masenf, line 518: do we need to handle this as a special case in `state.js`? seems like it would also work to have

```python
return call_script("window.history.back()")
```

masenf (approved):

reflex/event.py
abulvenz: Thanks, such an improvement. But now it looks trivial :smile:

masenf: thats good. trivial is easier to maintain 😸 

in the mid term, we're planning to refactor `state.js` to break it up into smaller parts and move more logic to the python side so there are fewer "moving parts"

PR #3637, opened 2024-07-09
Fix the error boundary

maximvlah:
- Fixed the error where the default `ErrorBoundary` component would always be rendered even if we pass a custom one. See https://github.com/reflex-dev/reflex/blob/9e1789a6c2c297066d4ed77be02f097ac3d7ab0b/reflex/app.py#L754
- Replaced handcoded `FrontendEventExceptionState` name across multiple files with a single constant in the repo to avoid mistakes (based on https://github.com/reflex-dev/reflex/pull/3476#discussion_r1647915633)
- Also it makes it more compatible with https://github.com/reflex-dev/reflex/pull/3214


maximvlah:
@masenf @Lendemor I have found few minor bugs in https://github.com/reflex-dev/reflex/pull/3572 and fixed them here, please have a look

masenf (approved):

PR #3654, opened 2024-07-10
feat: Added two more options to setvar.

abulvenz:
Either provide a dot-separated string to access
fields on objects or use a var to describe which
value to set.

This enables you to write something like this:
```python
import reflex as rx

class Object(rx.Base):
    name: str = "1234"

class State(rx.State):
    message: str = "Default"
    object: Object = Object()

def index() -> rx.Component:
    return rx.container(
        rx.text(f"Object name: ", State.object.name),
        rx.input(value=State.object.name, on_change=State.setvar(State.object.name)),  # new syntax
        rx.input(value=State.object.name, on_change=State.setvar("object.name")),  #new syntax
        rx.text(f"Message: ", State.message),
        rx.input(value=State.message, on_change=State.setvar(State.message)),  # new syntax
        rx.input(value=State.message, on_change=State.setvar("message")),
    )

app = rx.App()
app.add_page(index)
```

It might need some polishing with regard to nonexisting fields on objects, which is not checked for now. 
But I want to share a first version to get feedback.

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -295,13 +295,11 @@ class EventHandlerSetVar(EventHandler):
         from reflex.utils.exceptions import EventHandlerValueError
 
         if args:
-            var_name: Union[str, BaseVar] = args[0]
-            if not isinstance(args[0], str) and not isinstance(var_name, BaseVar):
+            var_name: str = args[0]
+            if not isinstance(args[0], str):
                 raise EventHandlerValueError(
                     f"Var name must be passed as a string, got {args[0]!r}"
                 )
-            if isinstance(var_name, BaseVar):
-                var_name = var_name._var_name
             fracs = var_name.split(".")
             # Check that the requested Var setter exists on the State at compile time.
             if (
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -260,6 +260,19 @@ class EventHandlerSetVar(EventHandler):
             var_name: The name of the variable to set.
             value: The value to set the variable to.
         """
+        var_name_fracs = var_name.split(".")
+        if len(var_name_fracs) > 1:
+            if isinstance(getattr(self, var_name_fracs[0]), Base):
+                obj = initial_obj = getattr(self, var_name_fracs[0])
+                for frac in var_name_fracs[1:-1]:
+                    obj = getattr(obj, frac)
+                setattr(obj, var_name_fracs[-1], value)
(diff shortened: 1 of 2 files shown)


masenf:
> use a var to describe which value to set

when i initially implemented `setvar`, this behavior was included, but a point was raised that it could introduce an incongruity and confusion.

consider the following code

```python
rx.input(value=State.message, on_change=State.setvar(State.message))
```

it's not clear if i'm intending to set `State.messages` to the changed value, or if i'm trying to do indirection and set a var that is named the _value of_ `State.message` to the changed value.

benedikt-bartscher:
I still like "the `State.a.set` thing" https://github.com/reflex-dev/reflex/pull/3163#issuecomment-2083439794

masenf:
> we need descriptor based vars and event handlers for this to pass type checkers

This is actually on the horizon. We're working on the Var refactor now and descriptor-based state Vars are part of that project.

abulvenz:
OK, I see. Sorry that I didn't do the history digging :+1: 

Thanks!

masenf:
to be clear, i still think `State.setvar("object.name")` is sweet! we should keep that.

abulvenz:
> to be clear, i still think `State.setvar("object.name")` is sweet! we should keep that.

Ah, sorry, tiny misunderstanding :smile: , so we are still in the game :game_die: . With your suggested change we can still write something like the following, which can help when refactoring, but typing is a bit :broken_heart:, but of course we can use a little `# pyright: ignore[my-typing]`.
```python
  rx.input(
      value=State.object.name, on_change=State.setvar(State.object.name._var_name)
  )
```
@benedikt-bartscher Do you think this could still help then with our wild internal meta-programming dreams?

Or to help the confused API-users, we rename the function from `setvar` to `apply_event_to` or `apply_value_to` or `apply_to` and read this:
`rx.input(value=State.message, on_click=State.apply_to(State.message))`

Lendemor:
Marking this as draft until resynced with main.
Some of the recent addition for typing event handler might unblock this PR.

adhami3310:

abulvenz:
> closing this due to inactivity, feels free to resurrect it when appropriate 👍

Thanks for closing it for me. FMPOV it lost its charm when making it string-based again which is unstable under refactoring and not the best option when working with IDEs.

PR #3669, opened 2024-07-12
[REF-3228] implement LiteralStringVar and format/retrieval mechanism

adhami3310:
> A LiteralStringVar is slightly more complex in that it may be created from a string that originated from a python f-string and have other Var references embedded within it.
>
> If the original _var_value does contain embedded Vars, then the constructor for such a Var would actually return a ConcatVarOperation over all literal strings and embedded Var references.


reflex/experimental/vars/base.py
```diff
@@ -234,3 +237,119 @@ class ArrayVar(ImmutableVar):
...
+                                value[end : (end + string_length)], _var_data=var_data
+                            )
+                        )
+                        value = value[(end + string_length) :]
+
+                offset += end - start
+
+            strings_and_vals.append(LiteralStringVar.create(value))
+
+            var_name = "+".join([element._var_name for element in strings_and_vals])
```
masenf, line 330:
```suggestion
            var_name = "+".join([str(element) for element in strings_and_vals])
```

It also feels like this logic should be in a cached property of the `ConcatVarOperation`

reflex/experimental/vars/base.py
```diff
@@ -234,3 +237,119 @@ class ArrayVar(ImmutableVar):
...
+
+            return ConcatVarOperation(
+                _var_value=strings_and_vals,
+                _var_type=str,
+                _var_name=var_name,
+                _var_data=_var_data,
+            )
```
masenf, line 337: i think this should move into `ConcatVarOperation.create` and it should only take the `strings_and_vals` (and maybe `_var_data`) as an argument.

reflex/experimental/vars/base.py
```diff
@@ -234,3 +237,119 @@ class ArrayVar(ImmutableVar):
...
+
+@dataclasses.dataclass(
+    eq=False,
+    frozen=True,
+    **{"slots": True} if sys.version_info >= (3, 10) else {},
+)
+class ConcatVarOperation(StringVar):
+    """Representing a concatenation of literal string vars."""
+
+    _var_value: Optional[list[Var]] = dataclasses.field(default=None)
```
masenf, line 355: The base `Var` should get a new property called `_get_all_var_data() -> VarData` that, by default, just returns the `_var_data` field.

This concat operation needs its `_get_all_var_data` to be a cached_property that returns a single `VarData` instance that is merged from the operation's `_var_data` and `_get_all_var_data` property of each Var in `_var_value`.

reflex/experimental/vars/base.py
```diff
@@ -234,3 +238,163 @@ class ArrayVar(ImmutableVar):
...
+
+        Args:
+            _var_value: The list of vars to concatenate.
+            _var_data: Additional hooks and imports associated with the Var.
+        """
+        super(ConcatVarOperation, self).__init__(
+            _var_name="", _var_data=_var_data, _var_type=str
+        )
+        object.__setattr__(self, "_var_value", _var_value)
+        object.__setattr__(self, "_var_name", self._cached_var_name)
```
masenf, line 362: this feels kind of messy, but i dont know of a cleaner way to accomplish it. (tried some stuff locally, but nothing really worked)

adhami3310: Yea cached properties replacing existing fields are very iffy.

reflex/experimental/vars/base.py
```diff
@@ -59,12 +70,25 @@ def _var_full_name_needs_state_prefix(self) -> bool:
...
+                ImmutableVarData.merge(self._var_data, _var_data),
             )
 
+    def __hash__(self) -> int:
+        """Define a hash function for the var.
+
+        Returns:
+            The hash of the var.
+        """
+        return hash((self._var_name, self._var_type, self._var_data))
```
masenf, line 87: do we really need this defined here? my concern is that some subclass adds additional fields, but then doesn't override `__hash__` and thus would compare the same as another Var, even though it really was different.

maybe that's not an issue though. i know `__hash__` is a bit weird, in that it is used for `set`/`dict` operations, but then they use equality for collision detection

reflex/vars.py
```diff
@@ -231,6 +238,188 @@ def dict(self) -> dict:
...
+            == (
+                other.hooks
+                if isinstance(other, ImmutableVarData)
+                else tuple(other.hooks.keys())
+            )
+            and imports.collapse_imports(self.imports)
+            == imports.collapse_imports(other.imports)
+        )
+
+    def dict(self) -> dict:
```
masenf, line 348: i don't think we need this method, since `ImmutableVarData` shouldn't ever be serialized with the new global f-string lookup table

reflex/vars.py
```diff
@@ -231,6 +238,188 @@ def dict(self) -> dict:
...
+            "state": self.state,
+            "imports": {
+                lib: [import_var.dict() for import_var in import_vars]
+                for lib, import_vars in self.imports
+            },
+            "hooks": {k: None for k in self.hooks},
+        }
+
+
+def _decode_var_immutable(value: str) -> tuple[ImmutableVarData | None, str]:
```
masenf, line 364: who calls this function? it seems like the functionality is part of `LiteralStringVar.create` now and the old-style vars don't need the ImmutableVarData

reflex/experimental/vars/base.py
adhami3310: If it's not defined here we would have to define it elsewhere anyways, so I don't particularly see a way out of it.

Also `__eq__` is generally made to use the weird system of comparison to serialize it for the front end instead of actually comparing the two and returning a bool. It feels like this could cause a limitation between true equality and front-end equality.

reflex/vars.py
adhami3310: It's currently being called in `__post_init_` of `ImmutableVar`. If it's desired we can push the behavior of creating `ConcatVarOperation` into there so not use this.

masenf (approved):
i think this is good to go

PR #3673, opened 2024-07-15
[REF-3197] Browser init workflow

masenf:


picklelo (commented):
Do we also want to add this as option 0 within the interactive reflex init?

picklelo (approved):
working nicely for me

PR #3677, opened 2024-07-16
update init prompt to use new templates from reflex-dev/templates

Lendemor:


reflex/utils/prerequisites.py
```diff
@@ -1310,39 +1310,63 @@ def migrate_to_reflex():
...
-        }
-    except httpx.HTTPError as ex:
-        console.info(f"Failed to fetch app templates: {ex}")
+        releases = response.json()
+        for release in releases:
+            if release["tag_name"] == f"v{tag}":
+                return release
+        return None
+
+    release = get_release_by_tag(version)
```
picklelo, line 1335: I think we should try / except around the httpx calls like before. If they don't have network access, we should still let them init with the blank template.

reflex/utils/prerequisites.py
```diff
@@ -1310,39 +1310,63 @@ def migrate_to_reflex():
...
+
+    assets = release.get("assets", [])
+    asset = next((a for a in assets if a["name"] == "templates.json"), None)
+    if asset is None:
+        print(f"Templates metadata not found for version {version}")
         return {}
+    else:
+        templates_url = asset["browser_download_url"]
 
+    templates_data = httpx.get(templates_url, follow_redirects=True).json()["templates"]
```
picklelo, line 1348: Same here for the try/except

picklelo (commented):
Looks good - just a small comment on maintaining the blank template working if httpx fails

picklelo (approved):

PR #3680, opened 2024-07-16
Benchmarking Updates

Alek99:
Reworking how we benchmark and store our benchmarking results from the github CI to posthog and refactor the code.


masenf (approved):
there's some extra trailing whitespace added to some of the yml files, but overall looks good to me.

we might need to make some tweaks after we start getting some data in, but i think it can go in, considering it doesn't affect the operation of the framework itself.

PR #3684, opened 2024-07-18
[REF-3273]Add SVG circle, polygon and rect components

ElijahAhianyo:
```python
import reflex as rx

def index():
    return rx.vstack(
        rx.el.svg(
            rx.el.svg.rect(
                x=10,
                y=10,
                width=30,
                height=30,
                stroke="black",
                fill="transparent",
                stroke_width=5
            ),
            rx.el.svg.circle(
                cx=25,
                cy=75,
                r=20,
                stroke="red",
                fill="transparent",
                stroke_width=5
            ),
            rx.el.svg.polygon(
                points="50 160 55 180 70 180 60 190 65 205 50 195 35 205 40 190 30 180 45 180",
                stroke="green",
                fill="transparent",
                stroke_width=5
            ),
            width=250,
            height=250,
            xmlns="http://www.w3.org/2000/svg",
        )
    )


app = rx.App()
app.add_page(index)

```

attributes/props in the example above such as `stroke`, `stroke_width`, `fill` are not explicitly defined as props because these are/can typically be added as CSS props (https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Fills_and_Strokes#using_css)


picklelo (dismissed):
Looks good once the CI is passing

picklelo (approved):

PR #3687, opened 2024-07-18
[REF-3227] implement more literal vars

adhami3310:
implements number, boolean, object, and array

### DIFF
--- a/reflex/experimental/vars/base.py
+++ b/reflex/experimental/vars/base.py
@@ -515,6 +515,9 @@ class LiteralVar(ImmutableVar):
                 value.dict(), _var_type=type(value), _var_data=_var_data
             )
 
+        if isinstance(value, str):
+            return LiteralStringVar.create(value, _var_data=_var_data)
+
         constructor = type_mapping.get(type(value))
 
         if constructor is None:
@@ -942,7 +945,6 @@ class LiteralArrayVar(LiteralVar):
 
 
 type_mapping = {
-    str: LiteralStringVar,
     int: LiteralNumberVar,
     float: LiteralNumberVar,
     bool: LiteralBooleanVar,
--- a/reflex/experimental/vars/base.py
+++ b/reflex/experimental/vars/base.py
@@ -10,6 +10,7 @@ from functools import cached_property
 from typing import Any, Optional, Tuple, Type
 
 from reflex import constants
+from reflex.base import Base
 from reflex.constants.base import REFLEX_VAR_CLOSING_TAG, REFLEX_VAR_OPENING_TAG
 from reflex.utils import serializers, types
 from reflex.utils.exceptions import VarTypeError
@@ -284,7 +285,7 @@ class LiteralVar(ImmutableVar):
         cls,
         value: Any,
         _var_data: VarData | None = None,
-    ) -> Var | None:
+    ) -> LiteralVar:
         """Create a var from a value.
 
         Args:
@@ -294,22 +295,20 @@ class LiteralVar(ImmutableVar):
         Returns:
             The var.
         """
-        if value is None:
-            return None
-
-        if isinstance(value, Var):
+        if isinstance(value, LiteralVar):
             return value
 
-        type_mapping = {
-            str: LiteralStringVar,
-            int: LiteralNumberVar,
-            float: LiteralNumberVar,
-            bool: LiteralBooleanVar,
-            dict: LiteralObjectVar,
-            list: LiteralArrayVar,
-        }
+        if isinstance(value, Base):
+            return LiteralObjectVar(
+                value.dict(), _var_type=type(value), _var_data=_var_data
+            )
 
-        return type_mapping[type(value)].create(value, _var_data=_var_data)
+        constructor = type_mapping.get(type(value))
+
+        if constructor is None:
+            raise TypeError(f"Unsupported type {type(value)} for LiteralVar.")
+
+        return constructor.create(value, _var_data=_var_data)
 
     def __post_init__(self):
         """Post-initialize the var."""
@@ -444,7 +443,7 @@ class ConcatVarOperation(StringVar):
         """
         if name == "_var_name":
             return self._cached_var_name
-        return getattr(super(), name)
+        return super(type(self), self).__getattr__(name)
 
     @cached_property
     def _cached_var_name(self) -> str:
@@ -578,6 +577,7 @@ class LiteralObjectVar(LiteralVar):
     def __init__(
         self,
         _var_value: dict[str, Var | Any],
+        _var_type: Type = dict,
         _var_data: VarData | None = None,
     ):
         """Initialize the object var.
@@ -588,7 +588,7 @@ class LiteralObjectVar(LiteralVar):
         """
         super(LiteralObjectVar, self).__init__(
             _var_name="",
-            _var_type=dict,
+            _var_type=_var_type,
             _var_data=ImmutableVarData.merge(_var_data),
         )
         object.__setattr__(
@@ -609,7 +609,7 @@ class LiteralObjectVar(LiteralVar):
         """
         if name == "_var_name":
             return self._cached_var_name
-        return getattr(super(), name)
+        return super(type(self), self).__getattr__(name)
 
     @cached_property
     def _cached_var_name(self) -> str:
@@ -641,6 +641,7 @@ class LiteralObjectVar(LiteralVar):
     def create(
         cls,
         value: dict[str, Var | Any],
+        _var_type: Type = dict,
         _var_data: VarData | None = None,
     ) -> LiteralObjectVar:
         """Create a var from an object value.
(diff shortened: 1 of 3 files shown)


reflex/experimental/vars/base.py
```diff
@@ -452,3 +465,241 @@ def create(
...
+
+@dataclasses.dataclass(
+    eq=False,
+    frozen=True,
+    **{"slots": True} if sys.version_info >= (3, 10) else {},
+)
+class LiteralObjectVar(LiteralVar):
+    """Base class for immutable literal object vars."""
+
+    _var_value: Tuple[Tuple[str, Var], ...] = dataclasses.field(default_factory=tuple)
```
masenf, line 544: should both key and value should be `Var` type to support indirection of the key?

if we do it this way, then the js literal would put square brackets around the key.

reflex/experimental/vars/base.py
```diff
@@ -452,3 +465,241 @@ def create(
...
+    def _cached_var_name(self) -> str:
+        """The name of the var.
+
+        Returns:
+            The name of the var.
+        """
+        return (
+            "{ "
+            + ", ".join(
+                [str(key) + ": " + str(value) for key, value in self._var_value]
```
masenf, line 588:
```suggestion
                [str(key) + ": " + str(value) for key, value in self._var_value]
```

Eventually we should be converting non-Var literal values into their appropriate literal value here so they get serialized correctly. For example, if I create a `LiteralObjectVar` from `{"a": {"b": "c"}}` there is no mechanism by which the nested dict would be converted into a `LiteralObjectVar`

reflex/experimental/vars/base.py
```diff
@@ -452,3 +465,241 @@ def create(
...
+        return getattr(super(), name)
+
+    @cached_property
+    def _cached_var_name(self) -> str:
+        """The name of the var.
+
+        Returns:
+            The name of the var.
+        """
+        return "[" + ", ".join([str(element) for element in self._var_value]) + "]"
```
masenf, line 674: Similarly, i think we should convert `element` into a `LiteralVar` if it is not already to better handle nested stuff

masenf (commented):
I think `LiteralVar.create` should have a kind of dispatch table based on the type, that would allow us to write `LiteralVar.create(["test", 43, [1, 2, 3], {"a": {"b": "c"}}])` and get back a `LiteralArrayVar`

reflex/experimental/vars/base.py
```diff
@@ -452,3 +465,241 @@ def create(
...
+
+        Args:
+            name: The name of the attribute.
+
+        Returns:
+            The attribute of the var.
+        """
+        if name == "_var_name":
+            return self._cached_var_name
+        return getattr(super(), name)
```
masenf, line 576: doesn't seem like this actually works. i had to use `return super(type(self), self).__getattr__(name)`

> slots: If true (the default is False), [__slots__](https://docs.python.org/3/reference/datamodel.html#object.__slots__) attribute will be generated and new class will be returned instead of the original one. If __slots__ is already defined in the class, then [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) is raised. Calling no-arg [super()](https://docs.python.org/3/library/functions.html#super) in dataclasses using slots=True will result in the following exception being raised: TypeError: super(type, obj): obj must be an instance or subtype of type. The two-arg [super()](https://docs.python.org/3/library/functions.html#super) is a valid workaround. See [gh-90562](https://github.com/python/cpython/issues/90562) for full details. [[1]](https://docs.python.org/3/library/dataclasses.html)

masenf (commented):
I made some local changes while testing that you may want to consider

```diff
diff --git a/reflex/experimental/vars/base.py b/reflex/experimental/vars/base.py
index fce1a958..35376281 100644
--- a/reflex/experimental/vars/base.py
+++ b/reflex/experimental/vars/base.py
@@ -10,6 +10,7 @@ from functools import cached_property
 from typing import Any, Optional, Tuple, Type
 
 from reflex import constants
+from reflex.base import Base
 from reflex.constants.base import REFLEX_VAR_CLOSING_TAG, REFLEX_VAR_OPENING_TAG
 from reflex.utils import serializers, types
 from reflex.utils.exceptions import VarTypeError
@@ -279,6 +280,55 @@ class FunctionVar(ImmutableVar):
 class LiteralVar(ImmutableVar):
     """Base class for immutable literal vars."""
 
+    @classmethod
+    def create(
+        cls,
+        value: Any,
+        _var_data: VarData | None = None,
+    ) -> LiteralVar:
+        """Create a var from a value.
+
+        Args:
+            value: The value to create the var from.
+            _var_data: Additional hooks and imports associated with the Var.
+
+        Returns:
+            The var.
+        """
+        # Eventually this needs to replace the @rx.serializer API to allow custom types to be converted to Var.
+        if isinstance(value, str):
+            return LiteralStringVar.create(value, _var_data=_var_data)
+        if isinstance(value, bool):
+            return LiteralBooleanVar.create(value, _var_data=_var_data)
+        if isinstance(value, (int, float)):
+            return LiteralNumberVar.create(value, _var_data=_var_data)
+        if isinstance(value, dict):
+            return LiteralObjectVar.create(value, _var_data=_var_data)
+        if isinstance(value, Base):
+            return LiteralObjectVar.create(value.dict(), _var_type=type(value), _var_data=_var_data)
+        if isinstance(value, (list, tuple, set)):
+            return LiteralArrayVar.create(value, _var_data=_var_data)
+        raise TypeError(f"Unsupported type {type(value)} for LiteralVar.")
+
+    @classmethod
+    def create_safe(
+        cls,
+        value: Any,
+        _var_data: VarData | None = None,
+    ) -> LiteralVar:
+        """Create a var from a value, asserting that it is not None.
+
+        Args:
+            value: The value to create the var from.
+            _var_data: Additional hooks and imports associated with the Var.
+
+        Returns:
+            The var.
+        """
+        var = cls.create(value, _var_data=_var_data)
+        assert var is not None
+        return var
+
     def __post_init__(self):
         """Post-initialize the var."""
 
@@ -412,7 +462,7 @@ class ConcatVarOperation(StringVar):
         """
         if name == "_var_name":
             return self._cached_var_name
-        return getattr(super(), name)
+        return super(type(self), self).__getattr__(name)
 
     @cached_property
     def _cached_var_name(self) -> str:
@@ -546,6 +596,7 @@ class LiteralObjectVar(LiteralVar):
     def __init__(
         self,
         _var_value: dict[str, Var],
+        _var_type: Type | None = None,
         _var_data: VarData | None = None,
     ):
         """Initialize the object var.
@@ -556,7 +607,7 @@ class LiteralObjectVar(LiteralVar):
         """
         super(LiteralObjectVar, self).__init__(
             _var_name="",
-            _var_type=dict,
+            _var_type=_var_type or dict,
             _var_data=ImmutableVarData.merge(_var_data),
         )
         object.__setattr__(self, "_var_value", tuple(_var_value.items()))
@@ -573,7 +624,7 @@ class LiteralObjectVar(LiteralVar):
         """
         if name == "_var_name":
             return self._cached_var_name
-        return getattr(super(), name)
+        return super(type(self), self).__getattr__(name)
 
     @cached_property
     def _cached_var_name(self) -> str:
@@ -585,7 +636,13 @@ class LiteralObjectVar(LiteralVar):
         return (
             "{ "
             + ", ".join(
-                [str(key) + ": " + str(value) for key, value in self._var_value]
+                [
+                    (
+                        str(key if isinstance(key, Var) else LiteralVar.create_safe(key)) + ": " +
+                        str(value if isinstance(value, Var) else LiteralVar.create_safe(value))
+                    )
+                    for key, value in self._var_value
+                ]
             )
             + " }"
         )
@@ -605,6 +662,7 @@ class LiteralObjectVar(LiteralVar):
     def create(
         cls,
         value: dict[str, Var],
+        _var_type: Type | None = None,
         _var_data: VarData | None = None,
     ) -> LiteralObjectVar:
         """Create a var from an object value.
@@ -618,6 +676,7 @@ class LiteralObjectVar(LiteralVar):
         """
         return LiteralObjectVar(
             _var_value=value,
+            _var_type=_var_type,
             _var_data=_var_data,
         )
 
@@ -662,7 +721,7 @@ class LiteralArrayVar(LiteralVar):
         """
         if name == "_var_name":
             return self._cached_var_name
-        return getattr(super(), name)
+        return super(type(self), self).__getattr__(name)
 
     @cached_property
     def _cached_var_name(self) -> str:
@@ -671,7 +730,7 @@ class LiteralArrayVar(LiteralVar):
         Returns:
             The name of the var.
         """
-        return "[" + ", ".join([str(element) for element in self._var_value]) + "]"
+        return "[" + ", ".join([str(element if isinstance(element, Var) else LiteralVar.create_safe(element)) for element in self._var_value]) + "]"
 
     @cached_property
     def _get_all_var_data(self) -> ImmutableVarData | None:
```

It's probably also time to split out the literal defs from `base.py` into `literals.py`

reflex/experimental/vars/base.py
adhami3310: I have moved this logic into init

masenf (commented):
Code looks good. At the very least though, we need test cases for each of the new types and the call operation and combinations there of (like passing a ConcatVarOperation to a FunctionStringVar)

masenf (approved):
awesome thanks

PR #3691, opened 2024-07-19
reflex init --ai fixups

masenf:
* Catch exceptions while polling to avoid exiting the loop
* Increase polling interval to 2s
* Allow `--template` to be a generation hash when `--ai` is also specified


picklelo (approved):
awesome!

PR #3698, opened 2024-07-23
[REF-3321] implement var operation decorator

adhami3310:
```python
@var_operation(output=NumberVar)
def add(a: NumberVar | int | float, b: NumberVar | int | float):
    return f"({a} + {b})"
```


reflex/experimental/vars/base.py
```diff
@@ -943,6 +2179,55 @@ def _get_all_var_data(self) -> ImmutableVarData | None:
...
+
+    Returns:
+        The decorator.
+    """
+
+    def decorator(func: Callable[P, str], output=output):
+        @functools.wraps(func)
+        def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
+            return output(
+                _var_name=func(*args, **kwargs),
```
masenf, line 2216: i think any literals would need to be converted into LiteralVar first before passing them to the function

masenf (commented):
still testing it out locally, but looks good so far

masenf:
Can you start breaking some of this out into separate modules

reflex/experimental/vars/base.py
adhami3310: It would make it impossible to implement something like this though:
```py
@var_operation(output=NumberVar)
def add(a: number_types, b: number_types) -> str:
    if isinstance(b, (int, float)) and b == 0:
        return str(a)
    if isinstance(a, (int, float)) and a == 0:
        return str(b)
    return f"({a} + {b})"
```

masenf: i think thats okay to not allow python logic in the var operation, since really it's designed to be a snippet of js code.

the danger in allowing python literals to flow through is i can pass some type that's not valid js when converted with simple f-string template, such as literal boolean (will convert to capital `True` or `False`) or complex numbers.

With Var operations, we must treat the vars as essentially opaque to the python side and transform it using valid operations that chain to correct js code. By converting literals to Var upfront, then we'll definitely catch any unserializable types at compile time, instead of generating invalid code.

masenf (approved):
awesome!

PR #3701, opened 2024-07-24
feat: minified state names

benedikt-bartscher:
followup of #3214

### DIFF
--- a/reflex/constants/compiler.py
+++ b/reflex/constants/compiler.py
@@ -66,18 +66,18 @@ class CompileVars(SimpleNamespace):
     ENV_MINIFY_STATES = "REFLEX_MINIFY_STATES"
     # Whether to minify states.
     MINIFY_STATES = os.environ.get(ENV_MINIFY_STATES, False)
+    # The name of the OnLoadInternal state.
+    ON_LOAD_INTERNAL_STATE = (
+        "l" if MINIFY_STATES else "reflex___state____on_load_internal_state"
+    )
     # The name of the internal on_load event.
-    ON_LOAD_INTERNAL = (
-        "l"
-        if MINIFY_STATES
-        else "reflex___state____on_load_internal_state.on_load_internal"
+    ON_LOAD_INTERNAL = f"{ON_LOAD_INTERNAL_STATE}.on_load_internal"
+    # The name of the UpdateVarsInternal state.
+    UPDATE_VARS_INTERNAL_STATE = (
+        "u" if MINIFY_STATES else "reflex___state____update_vars_internal_state"
     )
     # The name of the internal event to update generic state vars.
-    UPDATE_VARS_INTERNAL = (
-        "u"
-        if MINIFY_STATES
-        else ("reflex___state____update_vars_internal_state.update_vars_internal")
-    )
+    UPDATE_VARS_INTERNAL = f"{UPDATE_VARS_INTERNAL_STATE}.update_vars_internal"
     # The name of the frontend event exception state
     FRONTEND_EXCEPTION_STATE = (
         "e" if MINIFY_STATES else "reflex___state____frontend_event_exception_state"
@@ -87,9 +87,9 @@ class CompileVars(SimpleNamespace):
         f"reflex___state____state.{FRONTEND_EXCEPTION_STATE}"
     )
     INTERNAL_STATE_NAMES = {
-        ON_LOAD_INTERNAL,
-        UPDATE_VARS_INTERNAL,
-        FRONTEND_EXCEPTION_STATE_FULL,
+        ON_LOAD_INTERNAL_STATE,
+        UPDATE_VARS_INTERNAL_STATE,
+        FRONTEND_EXCEPTION_STATE,
     }
 
 
--- a/reflex/constants/compiler.py
+++ b/reflex/constants/compiler.py
@@ -6,7 +6,7 @@ from enum import Enum
 from types import SimpleNamespace
 
 from reflex.base import Base
-from reflex.constants import Dirs
+from reflex.constants import ENV_MODE_ENV_VAR, Dirs, Env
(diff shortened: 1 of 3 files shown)


reflex/constants/compiler.py
```diff
@@ -61,18 +62,35 @@ class CompileVars(SimpleNamespace):
     CONNECT_ERROR = "connectErrors"
     # The name of the function for converting a dict to an event.
     TO_EVENT = "Event"
+    # The env var to toggle minification of states.
+    ENV_MINIFY_STATES = "REFLEX_MINIFY_STATES"
+    # Whether to minify states.
+    MINIFY_STATES = os.environ.get(ENV_MINIFY_STATES, False)
```
masenf, line 68: i'm thinking we should minify state names by default in prod mode

masenf (commented):
this is great!

masenf (commented):
unfortunately trying to run it on reflex-web is giving a cryptic error

```python
  File "/Users/masenf/code/reflex-dev/reflex-web/pcweb/pcweb.py", line 8, in <module>
    from pcweb.pages import page404, routes
  File "/Users/masenf/code/reflex-dev/reflex-web/pcweb/pages/__init__.py", line 3, in <module>
    from .blog import blog_routes
  File "/Users/masenf/code/reflex-dev/reflex-web/pcweb/pages/blog/__init__.py", line 1, in <module>
    from .blog import *
  File "/Users/masenf/code/reflex-dev/reflex-web/pcweb/pages/blog/blog.py", line 2, in <module>
    from pcweb.templates.webpage import webpage
  File "/Users/masenf/code/reflex-dev/reflex-web/pcweb/templates/__init__.py", line 1, in <module>
    from .docpage import docpage
  File "/Users/masenf/code/reflex-dev/reflex-web/pcweb/templates/docpage/__init__.py", line 1, in <module>
    from .docpage import *
  File "/Users/masenf/code/reflex-dev/reflex-web/pcweb/templates/docpage/docpage.py", line 545, in <module>
    class RadixDocState(rx.State):
  File "pydantic/main.py", line 282, in pydantic.main.ModelMetaclass.__new__
  File "/Users/masenf/code/reflex-dev/reflex/reflex/state.py", line 563, in __init_subclass__
    raise StateValueError(
reflex.utils.exceptions.StateValueError: The substate class 'a' has been defined multiple times. Shadowing substate classes is not allowed.
```

reflex/constants/compiler.py
```diff
@@ -61,18 +62,35 @@ class CompileVars(SimpleNamespace):
...
+        "e" if MINIFY_STATES else "reflex___state____frontend_event_exception_state"
+    )
     # The full name of the frontend exception state
     FRONTEND_EXCEPTION_STATE_FULL = (
         f"reflex___state____state.{FRONTEND_EXCEPTION_STATE}"
     )
+    INTERNAL_STATE_NAMES = {
+        ON_LOAD_INTERNAL,
+        UPDATE_VARS_INTERNAL,
+        FRONTEND_EXCEPTION_STATE_FULL,
```
masenf, line 92: should be just `FRONTEND_EXCEPTION_STATE` so that `e` ends up in the exclusion list.

`FRONTEND_EXCEPTION_STATE_FULL` also can't really be defined here, because `rx.State` might get minified itself and not be `reflex___state____state`

reflex/constants/compiler.py
```diff
@@ -61,18 +62,35 @@ class CompileVars(SimpleNamespace):
...
     # The name of the function for converting a dict to an event.
     TO_EVENT = "Event"
+    # The env var to toggle minification of states.
+    ENV_MINIFY_STATES = "REFLEX_MINIFY_STATES"
+    # Whether to minify states.
+    MINIFY_STATES = os.environ.get(ENV_MINIFY_STATES, False)
     # The name of the internal on_load event.
-    ON_LOAD_INTERNAL = "reflex___state____on_load_internal_state.on_load_internal"
+    ON_LOAD_INTERNAL = (
+        "l"
```
masenf, line 71: seems like this would need to be `l.on_load_internal`, otherwise it doesn't reference the actual event handler that needs to be called.

reflex/state.py
```diff
@@ -461,6 +520,10 @@ def __init_subclass__(cls, mixin: bool = False, **kwargs):
         if mixin:
             return
 
+        # Generate a minified state name by converting state count to string
+        if not cls._state_name:
```
masenf, line 524: Seems like we need something like this

```python
if not cls._state_name or cls._state_name == cls.get_parent_state()._state_name:
```

Otherwise once we minify a parent state name `cls._state_name` seems to get the parent's minified name... maybe i'm missing something though.

With the suggested change, i'm actually able to get past the reflex-web error that i mentioned earlier.

masenf (commented):
okay, i found a few other problems that need to be addressed, but nothing that suggests this is the wrong approach.

reflex/constants/compiler.py
benedikt-bartscher: I agree. Should we also move this to `rx.Config.minify_states`?

reflex/state.py
benedikt-bartscher: Thanks for digging into this. You are absolutely right, and I like your suggested fix.

benedikt-bartscher: Sadly, your solution causes issues with some tests. In fact, just calling `cls.get_parent_state()` in the init_subclass hook causes some tests to fail.  
I adjusted it to `not cls._state_name or cls._state_name in all_state_names`.

reflex/constants/compiler.py
benedikt-bartscher: Thanks, I have now separated the state names from the event handler paths

benedikt-bartscher:
Superseded by #3728

PR #3702, opened 2024-07-24
use different registry when in china, fixes #3700

### LINKED ISSUE #3700: Can Chinese developers use pnpm or yarn for package management?
In China, the package management tools Bun and npm are too slow to download dependencies and are very unstable, sometimes taking several hours to start a project.

Debug: Running command: ['C:\\Users\\admxcy\\AppData\\Local\\reflex\\bun\\bin\\bun.exe', 'install']
Debug: Installing base frontend packages
Debug: bun install v1.1.10 (5102a944)
Debug: Resolving dependencies
Debug: Resolved, downloaded and extracted [10]
Debug: Saved lockfile
Debug:                                                                                                                                                                                  
Debug: + @radix-ui/react-dialog@1.1.1                                                                                                                                                   
Debug:                                                                                                                                                                                  
Debug: 33 packages installed [323.38s]                                                                                                                                                  
Debug: Running command: ['C:\\Users\\admxcy\\AppData\\Local\\reflex\\bun\\bin\\bun.exe', 'add', '-d', 'tailwindcss@3.3.2']                                                              
Debug: Installing tailwind                                                                                                                                                              
Debug: bun add v1.1.10 (5102a944)
Debug: Saved lockfile

adhami3310:


adhami3310:
@reigadegr can you give this a try?

reflex/utils/prerequisites.py
```diff
@@ -577,6 +578,15 @@ def initialize_package_json():
     code = _compile_package_json()
     output_path.write_text(code)
 
+    if _is_in_china() or True:
```
Lendemor:
```suggestion
    if _is_in_china():
```

Lendemor (changes requested):

Lendemor (dismissed):

reflex/utils/china.py
```diff
@@ -0,0 +1,27 @@
...
+        None: If an error occurred.
+    """
+    try:
+        response = requests.get(f"http://ip-api.com/json")
+        data = response.json()
+
+        if data["status"] == "success":
+            return data["country"] == "China"
+        else:
+            print(f"Error: {data['message']}")
```
masenf, line 22: we should use `rx.utils.console.info` to log this, and maybe include some more context, like "Failed geo-location lookup for your IP address"

reflex/utils/china.py
```diff
@@ -0,0 +1,27 @@
...
+
+def _is_in_china() -> Union[bool, None]:
+    """Check if the current IP address is in China.
+
+    Returns:
+        bool: True if the IP address is in China, False otherwise.
+        None: If an error occurred.
+    """
+    try:
+        response = requests.get(f"http://ip-api.com/json")
```
masenf, line 16: We should use `httpx` to make any requests.

According to `pyproject.toml` reflex does not directly depend on `requests`, so if it happens to be installed, then we're getting it as a transitive dependency and cannot rely on it being in the environment (for example, if the dependency that brings it in transitively switches to `httpx` or `urllib3` or something else)

masenf (commented):
i'm really hesitant to depend on some remote IP location service.

1. it's exposing our users to a third party every time they use reflex
2. at best it's a proxy for the information we actually want: whether the chinese npm mirror has a faster response time than the default mirror

I think this functionality should instead compare the response times of various mirrors and pick the one that is the fastest.

masenf (commented):
i definitely like this better, but determining the fastest mirror adds noticeable time now. does this only hit at init time?

Lendemor:
Maybe we should just check is the env var REFLEX_NPM_REGISTRY is set.
If it is, we use the value set in there, if not we keep using the default like now.

adhami3310:
It only adds a delay when initializing the web directory (once per init).

adhami3310:
> If it slow the process, maybe we could just check is an env var like REFLEX_NPM_REGISTRY is set. If it is, we use the value set in there, if not we keep using the default like now.

This could be a cool option but many might not find it and has to put up with worse experience without knowing how to fix it.

Alek99:
@FHU-yezi Would you be interested in testing this?

reigadegr:
@[adhami3310](https://github.com/adhami3310) 
Sorry，I am not able to compile the Python source code of reflex into an executable file.Could you teach me how to compile it, or send me the compiled file for testing?
I found a  method, which is to use the command 
```cmd
taskkill /F /IM bun.exe
```
 to let reflex call npm for installation, which is quite fast. The reason is that the 
```cmd
bun install
```
 command is very slow for me.

seewindcn:
> @FHU-yezi Would you be interested in testing this?
> 
> Let's wait to merge until someone can properly test and compare.

I had test "r.cnpmjs.org" by hand create .web\bunfig.toml; 
yes, it improved the package download experience;

picklelo (approved):

PR #3705, opened 2024-07-25
[REF-3328] Implement __getitem__ for ArrayVar

adhami3310:


reflex/experimental/vars/object.py
```diff
@@ -0,0 +1,439 @@
...
+        object.__delattr__(self, "_var_name")
+
+    @cached_property
+    def _cached_var_name(self) -> str:
+        """The name of the operation.
+
+        Returns:
+            The name of the operation.
+        """
+        return f"{self.value._var_name}[{self.key._var_name}]"
```
masenf, line 405: Feels like this should be 
```suggestion
        return f"{str(self.value)}[{str(self.key)}]"
```

Is there a reason it's not?

masenf (commented):
i still need to do more testing, but it doesn't look like the _var_type is properly passed for the `keys`, `values` and `entries` operations

reflex/experimental/vars/object.py
adhami3310: That's correct!

reflex/experimental/vars/object.py
```diff
@@ -88,18 +104,19 @@ class LiteralObjectVar(LiteralVar, ObjectVar):
...
 
         Args:
             _var_value: The value of the var.
+            _var_type: The type of the var.
             _var_data: Additional hooks and imports associated with the Var.
         """
         super(LiteralObjectVar, self).__init__(
             _var_name="",
-            _var_type=_var_type,
+            _var_type=type(_var_value) if _var_type is None else _var_type,
```
masenf, line 119: not sure if the `type` cast is needed here, because `type(dict[str, str])` returns `<class 'types.GenericAlias'>` and we almost certainly want the former so we can call `typing.get_args` on it

reflex/experimental/vars/object.py
```diff
@@ -109,6 +126,27 @@ def __init__(
...
 
+    def _key_type(self) -> Type:
+        """Get the type of the keys of the object.
+
+        Returns:
+            The type of the keys of the object.
+        """
+        print(self._var_type)
+        return (
+            self._var_type.__args__[0] if hasattr(self._var_type, "__args__") else Any
```
masenf, line 137: should be `typing.get_args` instead

masenf (commented):
partway through the new stuff

reflex/experimental/vars/sequence.py
```diff
@@ -1238,11 +1239,10 @@ def __init__(
             i: The index.
             _var_data: Additional hooks and imports associated with the Var.
         """
+        element_type = typing.get_args(a._var_type)
         super(ArrayItemOperation, self).__init__(
             _var_name="",
-            _var_type=(
-                a._var_type.__args__[0] if hasattr(a._var_type, "__args__") else Any
-            ),
+            _var_type=element_type[0] if element_type else Any,
```
masenf, line 1245: if the `i` is actually an int, then we should use `element_type[i % len(element_type)]` so a tuple, which has index-wise args can be indexed correctly

masenf (commented):
the following code needs to work if the annotation is there

```python
>>> ov2 = rx._x.vars.LiteralVar.create({"foo": "bar"})
>>> ov2["foo"].upper()
  File "<stdin>", line 1, in <module>
  File "/Users/masen/code/reflex-dev/reflex/reflex/vars.py", line 1726, in upper
    return self._replace(
           ^^^^^^^^^^^^^^
  File "/Users/masen/code/reflex-dev/reflex/reflex/experimental/vars/base.py", line 160, in _replace
    return type(self)(**field_values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: ObjectItemOperation.__init__() got an unexpected keyword argument '_var_name'
```

This seems to be hitting `upper` in the base Var, because the resulting `ObjectItemOperation` isn't a `StringVar`.

Similar but different, if we cast the value to `dict[str, str]`, then it still doesn't really work

```python
>>> ov = rx._x.vars.LiteralVar.create({"foo": "bar"}).to(rx._x.vars.ObjectVar, dict[str,str])
Warning: `rx._x` contains experimental features and might be removed at any time in the future .
>>> ov["foo"].lower()
  File "<stdin>", line 1, in <module>
  File "/Users/masen/code/reflex-dev/reflex/reflex/vars.py", line 1702, in lower
    raise VarTypeError(
reflex.utils.exceptions.VarTypeError: Cannot convert non-string var ({ ["foo"] : "bar" })["foo"] to lowercase.
>>> ov["foo"]
ObjectItemOperation(_var_name='({ ["foo"] : "bar" })["foo"]', _var_type=<class 'reflex.experimental.vars.base.ImmutableVar'>, _var_data=None, value=ToObjectOperation(_var_name='({ ["foo"] : "bar" })', _var_type=dict[str, str], _var_data=None, _original_var=LiteralObjectVar(_var_name='({ ["foo"] : "bar" })', _var_type=typing.Dict[str, str], _var_data=None, _var_value={'foo': 'bar'})), key=LiteralStringVar(_var_name='"foo"', _var_type=<class 'str'>, _var_data=None, _var_value='foo'))
```
in this case, the resulting `ObjectItemOperation` gets `_var_type=ImmutableVar` which doesn't seem quite right.

It seems like these item accessing operations probably need to also use `.to` in order to correctly cast the resulting value based on the annotation in the original var.

we also need test cases for these chain operations

reflex/experimental/vars/sequence.py
```diff
@@ -1232,6 +1278,20 @@ def unionize(*args: Type) -> Type:
     return Union[first, unionize(*rest)]
 
 
+def is_tuple_type(t: GenericType) -> bool:
```
masenf, line 1281: i think `rx.utils.types._issubclass` implements this already

reflex/experimental/vars/base.py
```diff
@@ -433,6 +473,16 @@ def create(
     def __post_init__(self):
         """Post-initialize the var."""
 
+    def json(self) -> str:
```
masenf, line 476: what is the `json()` method intended to be used for?

masenf (approved):

PR #3706, opened 2024-07-25
[REF-3203] Find a DOM event-like object in addEvents

masenf:
Do NOT assume that the first or last arg is the browser event that triggered the event handler.

Instead, pass all of the `args` through and find the first object which defines `preventDefault` and treat that as the event.

This fixes `stop_propagation` and `prevent_default` on recharts charts, which define their mouse event handler signatures as `(nextState, event)`. Before this fix, the generated code was passing `nextState` through as `_e` and since the event actions were not defined on that object, they had no effect.

This also fixes some other funny behavior where the `arg_def` was being passed through in parenthesis, which in JS evaluates to the last item in the list. So even if we defined some more complex `args_spec` for a particular trigger that correctly defined it's `_e` as the first argument, this would still be broken, because when multiple arguments were defined in an `arg_spec`, the _last_ one was implicitly being treated as the browser event.

### DIFF
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -647,7 +647,12 @@ export const useEventLoop = (
   const [connectErrors, setConnectErrors] = useState([]);
 
   // Function to add new events to the event queue.
-  const addEvents = (events, _e, event_actions) => {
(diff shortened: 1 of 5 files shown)


picklelo (approved):
nice

PR #3719, opened 2024-07-29
fix silly bug when style is set directly to breakpoints

adhami3310:
```py
def index() -> rx.Component:
    return rx.text(
        "Hello, World!",
        style=rx.breakpoints(
            md={"color": "blue"},
            lg={"color": "green"},
        ),
        color="red",
        font_weight="bold",
    )
```

### DIFF
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -469,6 +469,7 @@ class Component(BaseComponent, ABC):
 
         if isinstance(style, Breakpoints):
             style = {
+                # Assign the Breakpoints to the self-referential selector to avoid squashing down to a regular dict.
                 "&": style,
             }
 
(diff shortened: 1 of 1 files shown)


reflex/components/component.py
```diff
@@ -466,6 +467,11 @@ def __init__(self, *args, **kwargs):
             # Merge styles, the later ones overriding keys in the earlier ones.
             style = {k: v for style_dict in style for k, v in style_dict.items()}
 
+        if isinstance(style, Breakpoints):
+            style = {
+                "&": style,
```
masenf, line 472:
```suggestion
                # Assign the Breakpoints to the self-referential selector to avoid squashing down to a regular dict.
                "&": style,
```

A comment for the future might be nice

masenf (dismissed):

masenf (approved):

Issue #3720, opened 2024-07-30
[REF-3448] 404 not rendering


charlestondance:
**Describe the bug**
entering a random url give a 404 with text in the page
going to localhost:3000/404 updates the title but nothing is rendered in the page. Making a custom 404 has the same issue where no components are rendered in the page

**To Reproduce**
 - enter URL as above

**Expected behavior**
correct 404 page be returned at localhost:3000/404 

**Screenshots**

**Specifics (please complete the following information):**
 - Python Version:
 - Reflex Version:
 - OS: 
 - Browser (Optional):
 
**Additional context**

<sub>[REF-3448](https://linear.app/reflex-dev/issue/REF-3448/404-not-rendering)</sub>

masenf:
I'm able to reproduce the issue, even with our main website https://reflex.dev/404/ renders as a blank page.

Not sure what the problem is.

Is there a use case? Or are you just wondering for consistency sake why `/404` doesn't render? (I'm also curious)

Lendemor:
If only for consistency we should figure out what happens and fix it I think.

charlestondance:
No probs. My two use cases are:

1) Custom 404. You cant make a custom 404 now as no components will be rendered
2) I am using chakra.box as an iframe so i don't see the title change. All i can see in the iframe is blank space and i have no idea what is going on.

masenf:
> Custom 404

does it not work when you use `app.add_custom_404_page`? reflex.dev uses custom 404 and it seems to work fine (although accessing it directly via `/404` does not work)

charlestondance:
Custom or built in 404s do not render at /404. I am trying to handle and return /404 under certain cases that i check for. Automatically returning 404 on random routes doesnt help my use case.

charlestondance:
Problems with dynamic routes return the broken /404 automatically as well.

Manas1820:
@charlestondance can you share a small example snippet ? Just want to test if its fixed with the mentioned PR !

Manas1820:
```python3
"""Welcome to Reflex! This file outlines the steps to create a basic app."""

import reflex as rx

from rxconfig import config


class State(rx.State):
    """The app state."""

    @rx.var
    def test_id(self) -> str:
        return self.router.page.params.get("id", "No Test ID")


@rx.page(route="/test/[id]")
def test():
    """A page that updates based on the route."""
    return rx.container(
        rx.heading(f"Testing ID: {State.test_id}", size="9"),
[...]

PR #3723, opened 2024-07-31
Improve Client-Side Routing on Page Not Found

Manas1820:
## Describe your changes.
  
The issue is well described in #3720 and also fixes it.

In the ```useClientSideRouting``` function, the ```routeNotFound``` state was incorrectly set to ```false``` even when the current route was actually not found. To address this, I added a hard check to verify the pathname after attempting a redirection. If the pathname remains ```/404``` after the redirection attempt, the ```routeNotFound``` state is correctly set to ```true```. This ensures that the correct 404 page is displayed when the route is genuinely not found.

Example is mentioned in the same issue's [thread](https://github.com/reflex-dev/reflex/issues/3720#issuecomment-2259891448)

### Type of change

### New Feature Submission:

### Changes To Core Features:

### DIFF
--- a/reflex/.templates/web/utils/client_side_routing.js
+++ b/reflex/.templates/web/utils/client_side_routing.js
@@ -23,7 +23,12 @@ export const useClientSideRouting = () => {
       router.replace({
           pathname: window.location.pathname,
           query: window.location.search.slice(1),
-      })
+      }).then(()=>{
+          // Check if the current route is /404
+        if (router.pathname === '/404') {
+          setRouteNotFound(true); // Mark as an actual 404
+        }
+    })
       .catch((e) => {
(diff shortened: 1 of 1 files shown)


masenf (commented):
Thanks for your contribution. Seems to resolve the issue on my end.

Is there a reason why we check the `router.pathname` inside a chained function instead of _before_ we even attempt the `router.replace`?

Manas1820:
@masenf Just did it to be completely sure, thought that when router.replace is called, it might initiate a client-side navigation that potentially changes the current route. By checking router.pathname after router.replace completes, we can be completely positive that we have reached the final state.

Lendemor (approved):

PR #3725, opened 2024-07-31
Add more `rx.script` props

ElijahAhianyo:
This PR updates the props for `rx.script`.
ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script


masenf (commented):
i dont think these props are recognized by the next/script `<Script>` element... https://nextjs.org/docs/pages/api-reference/components/script

ElijahAhianyo:
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

I see, I might have been looking into the wrong `rx.script`, these props probably belong to `rx.el.script`, Ill go ahead and close this for now

PR #3726, opened 2024-07-31
fix initial_value for computed_var

### LINKED ISSUE #3722: [REF-3453] @rx.var(cache=True) does not get the correct initial value
**Describe the bug**
It seems like the `@rx.var(cache=True)` decorated function doesn't get run when computing the initial state. Changing to `@rx.cached_var` works fine.

**To Reproduce**

```python
import reflex as rx


class State(rx.State):
    @rx.var(cache=True)
    def my_list(self) -> list[int]:
        return [1, 2, 3]


def index() -> rx.Component:
    return rx.container(
        rx.foreach(
            State.my_list,
            lambda item: rx.box(f"Item {item}"),
        ),
        rx.logo(),
    )


app = rx.App()
app.add_page(index)
```

**Expected behavior**
`@rx.cached_var` and `@rx.var(cache=True)` should work the same.

**Screenshots**
image

 - Python Version: 3.11
 - Reflex Version: 0.5.8
 - OS: 
 - Browser (Optional):

<sub>[REF-3453](https://linear.app/reflex-dev/issue/REF-3453/rxvarcache=true-does-not-get-the-correct-initial-value)</sub>

benedikt-bartscher:
Closes #3722

### DIFF
--- a/reflex/vars.py
+++ b/reflex/vars.py
@@ -2482,7 +2482,7 @@ class ComputedVar(Var, property):
 
 def computed_var(
     fget: Callable[[BaseState], Any] | None = None,
-    initial_value: Any | None = None,
+    initial_value: Any | types.Unset = types.Unset(),
     cache: bool = False,
     deps: Optional[List[Union[str, Var]]] = None,
     auto_deps: bool = True,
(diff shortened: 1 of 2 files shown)


masenf (approved):
Thanks for the quick fix 😌

PR #3738, opened 2024-08-03
Use the new state name when setting `is_hydrated` to false

masenf:
This change ensures that `is_hydrated` is reset to `false` when navigating between pages to avoid the "flicker" that shows incorrect data during navigation events.

### DIFF
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -782,7 +782,7 @@ export const useEventLoop = (
   // Route after the initial page hydration.
   useEffect(() => {
     const change_start = () => {
-      const main_state_dispatch = dispatch["state"]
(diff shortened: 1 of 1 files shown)


benedikt-bartscher:
Nice catch, I think I missed this one because of the hidden templates folder.

masenf:
@benedikt-bartscher instead of templating state.js, i think we should just move the "main_state_name" to context.js.jinja2 and import it from state.js

The main reason being that the complexity in state.js is already quite high and moving it to a jinja2 file means we lose intellisense and lenting capability when editing that file.

picklelo (approved):

benedikt-bartscher:
@masenf sounds like a good approach, i will adjust it in #3728

Issue #2155, opened 2023-11-11
[REF-1626] Allow string fomatting in rx.html component


abdullahkhalids:
**Describe the bug**
Currently, the `rx.html` component `src` is not formatted.

**To Reproduce**
Compile this app

```
from rxconfig import config
import reflex as rx

class State(rx.State):
    """The app state."""

    myvar: str = "Blue"


def index() -> rx.Component:
    return rx.fragment(
	rx.heading(
            f"The number is {State.myvar}"
            ), #works as expected
        rx.html(
            f"The number is {State.myvar}"
            ), #doesn't work
    )

app = rx.App()
app.add_page(index)
app.compile()
```

**Expected behavior**
The html format string should be processed correctly.

**Screenshots**

![image](https://uploads.linear.app/9513bc1c-787d-499b-8c86-46952ab7013c/7a9a2074-267c-4f20-9ce8-af35104e748c/9ee098b5-6402-4206-aab7-15c99364d1b1?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzk1MTNiYzFjLTc4N2QtNDk5Yi04Yzg2LTQ2OTUyYWI3MDEzYy83YTlhMjA3NC0yNjdjLTRmMjAtOWNlOC1hZjM1MTA0ZTc0OGMvOWVlMDk4YjUtNjQwMi00MjA2LWFhYjctMTVjOTkzNjRkMWIxIiwiaWF0IjoxNzA0NzI4NjMzLCJleHAiOjE3MDQ4MTUwMzN9.9DUU0skoP2cG6xWUJliLpIacseMRCOxpEBhmz8VVcaE)

**Specifics (please complete the following information):**

* Python Version: 3.10.12
* Reflex Version: 0.3.2
* OS: Debian testing
* Browser (Optional):

<sub>From [SyncLinear.com](https://synclinear.com) | [REF-1626](https://linear.app/reflex-dev/issue/REF-1626/allow-string-fomatting-in-rxhtml-component)</sub>

masenf:
potentially fixed by #2149

ipcloudlive:
I am using multiline text like this
```
return rx.html(
"""
<h1> Hello {State.text1}</h1>
<h1> Bye {State.text2}</h1>
""")
```

masenf:
Thanks for the update, will investigate.

adhami3310:
This is fixed with the new immutable var system.

PR #3743, opened 2024-08-05
fully migrate vars into new system

adhami3310:
By the end of this PR, creating `Var` or `BaseVar` would be deprecated in favor of `ImmutableVar`.


masenf (commented):
review checkpoint

reflex/ivars/number.py
```diff
@@ -1065,9 +1022,7 @@ def _cached_var_name(self) -> str:
         Returns:
             The name of the var.
         """
-        first_value = self.a if isinstance(self.a, Var) else LiteralVar.create(self.a)
-        second_value = self.b if isinstance(self.b, Var) else LiteralVar.create(self.b)
-        return f"({str(first_value)} != {str(second_value)})"
+        return f"({str(self._lhs)} != {str(self._rhs)})"
```
masenf, line 1025:
```suggestion
        return f"({str(self._lhs)} !== {str(self._rhs)})"
```

reflex/ivars/number.py
```diff
@@ -1399,26 +1383,100 @@ class ToBooleanVarOperation(BooleanVar):
...
         )
-        object.__setattr__(self, "_original_value", _original_value)
+
+
+@dataclasses.dataclass(
+    eq=False,
+    frozen=True,
+    **{"slots": True} if sys.version_info >= (3, 10) else {},
+)
+class TernaryOperator(ImmutableVar):
```
masenf, line 1465: i think this should be in `base.py` or a separate module. doesn't seem like it fits with the other `number.py` types because it fundamentally is not a number but the type if either `_if_true` or `_if_false`

reflex/components/core/banner.py
```diff
@@ -19,50 +18,45 @@
...
-    _var_is_string=False,
+has_too_many_connection_errors: Var = ImmutableVar.create_safe(
+    value="(connectErrors.length >= 2)",
     _var_data=connect_error_var_data,
-).to(bool)
+).to(BooleanVar)
 
 
-class WebsocketTargetURL(Bare):
+class WebsocketTargetURL:
```
masenf, line 59: can probably be a subclass of ImmutableVar now

and `add_imports` can be removed in favor of the included VarData with the necessary imports

reflex/components/core/match.py
```diff
@@ -117,17 +120,12 @@ def _create_case_var_with_var_data(cls, case_element):
...
         """
-        _var_data = case_element._var_data if isinstance(case_element, Style) else None  # type: ignore
-        case_element = Var.create(
-            case_element,
-            _var_is_string=isinstance(case_element, (str, Color)),
-        )
-        if _var_data is not None:
-            case_element._var_data = VarData.merge(case_element._var_data, _var_data)  # type: ignore
+        _var_data = case_element._var_data if isinstance(case_element, Style) else None
+        case_element = LiteralVar.create(case_element, _var_data=_var_data)
```
masenf, line 124: what happens here when `case_element` is already a `Var`?

reflex/components/core/match.py
```diff
@@ -230,28 +228,22 @@ def _create_match_cond_var_or_component(
...
 
-        # match cases and default should all be Vars at this point.
-        # Retrieve var data of every var in the match cases and default.
-        var_data = [
-            *[el._var_data for case in match_cases for el in case],
-            default._var_data,  # type: ignore
-        ]
-
-        return match_cond_var._replace(
+        return ImmutableVar(
```
masenf, line 235: created https://linear.app/reflex-dev/issue/REF-3532/create-matchvar-to-handle-rxmatch to eventually replace `format_match` with an actual Var

reflex/components/datadisplay/code.py
```diff
@@ -505,16 +523,11 @@ def _render(self):
...
                 format.to_camel_case(f"{predicate}{qmark}{value.replace('`', '')}"),
-                _var_is_local=False,
-                _var_is_string=False,
-            )
-        ).remove_props("theme", "code")
-        if self.code is not None:
-            out.special_props.add(
-                Var.create_safe(f"children={str(self.code)}", _var_is_string=False)
             )
+        ).remove_props("theme", "code").add_props(children=self.code)
```
masenf, line 529: overall much nicer!

is is possible to get rid of the `format.to_camel_case(f"{predicate}{qmark}{value.replace('`', '')}")` and just use `self.theme` directly?

reflex/components/next/image.py
```diff
@@ -104,6 +105,6 @@ def check_prop_type(prop_name, prop_value):
 
         src = props.get("src", None)
         if src is not None and not isinstance(src, (Var)):
-            props["src"] = Var.create(value=src, _var_is_string=True)
+            props["src"] = LiteralVar.create(src)
```
masenf, line 108: another place where it seems like this would just be handled by `Component.__init__`

masenf (commented):
Review checkpoint 😁

reflex/ivars/base.py
```diff
@@ -0,0 +1,1736 @@
...
+            **kwargs,
+        )
+
+    return wrapper
+
+
+# Partial function of computed_var with cache=True
+cached_var = functools.partial(
+    immutable_computed_var, cache=True, _deprecated_cached_var=True
+)
```
masenf, line 1736: we can get rid of this and the `_deprecated_cached_var` parameter, since this change will go into 0.6.0 and `@rx.cached_var` is slated for removal in 0.6.0

reflex/ivars/object.py
```diff
@@ -22,18 +22,19 @@
...
-from reflex.experimental.vars.number import NumberVar
-from reflex.experimental.vars.sequence import ArrayVar, StringVar
-from reflex.utils.exceptions import VarAttributeError
-from reflex.utils.types import GenericType, get_attribute_access_type
-from reflex.vars import ImmutableVarData, Var, VarData
+from .number import BooleanVar, NumberVar
+from .sequence import ArrayVar, StringVar
 
-OBJECT_TYPE = TypeVar("OBJECT_TYPE")
+OBJECT_TYPE = TypeVar("OBJECT_TYPE", bound=Dict)
```
masenf, line 37: do we want `Dict` here or `Mapping`?

reflex/ivars/object.py
```diff
@@ -488,89 +468,144 @@ def _get_all_var_data(self) -> ImmutableVarData | None:
...
+
     @cached_property
     def _cached_var_name(self) -> str:
         """The name of the operation.
 
         Returns:
             The name of the operation.
         """
-        return f"Object.values({self.value._var_name})"
+        return f"Object.values({self._value._var_name})"
```
masenf, line 551:
```suggestion
        return f"Object.values({str(self._value)})"
```
might be better to rely on `__str__` like in ObjectKeysOperation

reflex/ivars/object.py
```diff
@@ -488,89 +468,144 @@ def _get_all_var_data(self) -> ImmutableVarData | None:
...
+
     @cached_property
     def _cached_var_name(self) -> str:
         """The name of the operation.
 
         Returns:
             The name of the operation.
         """
-        return f"Object.entries({self.value._var_name})"
+        return f"Object.entries({self._value._var_name})"
```
masenf, line 586: same here, prefer `str(self._value)` unless i'm missing something about why this needs to be different here

reflex/ivars/object.py
```diff
@@ -581,30 +616,103 @@ def _cached_var_name(self) -> str:
...
+    )
+
+    @cached_property
+    def _cached_var_name(self) -> str:
+        """The name of the operation.
+
+        Returns:
+            The name of the operation.
+        """
+        return f"Object.assign({self._lhs._var_name}, {self._rhs._var_name})"
```
masenf, line 633: seems unsafe to mutate `self._lhs`, is there a reason to not use the object spread syntax?

```
return f"{{...{str(self._lhs)}, ...{str(self._rhs)}}}
```

reflex/ivars/sequence.py
```diff
@@ -69,29 +71,29 @@ def __radd__(self, other: StringVar | str) -> ConcatVarOperation:
...
         Returns:
-            The string concatenation operation.
+            StringVar: The resulting sequence after multiplication.
         """
-        return ConcatVarOperation(*[self for _ in range(other)])
+        return (self.split() * other).join()
 
-    def __rmul__(self, other: int) -> ConcatVarOperation:
-        """Concatenate two strings.
+    def __rmul__(self, other: NumberVar | int) -> StringVar:
```
masenf, line 87:
```python
__rmul__ = __mul__
```

can we just alias it?

reflex/vars.py
```diff
@@ -857,6 +902,7 @@ def __getitem__(self, i: Any) -> Var:
         Raises:
             VarTypeError: If the var is not indexable.
         """
+        print(repr(self))
```
masenf, line 905: ❌

masenf (commented):
Review checkpoint

I'm almost through it all, just need to look through the test changes

tests/components/core/test_cond.py
```diff
@@ -57,15 +58,23 @@ def test_validate_cond(cond_state: BaseState):
 
     [true_value_text] = true_value["children"]
     assert true_value_text["name"] == "RadixThemesText"
-    assert true_value_text["children"][0]["contents"] == "{`cond is True`}"
+    assert true_value_text["children"][0]["contents"] in (
+        "{`cond is True`}",
+        '{"cond is True"}',
+        "{'cond is True'}",
```
masenf, line 64: is the test non-deterministic? seems like it would only ever be one of these right?

tests/components/core/test_html.py
```diff
@@ -15,8 +15,8 @@ def test_html_many_children():
 
 def test_html_create():
     html = Html.create("<p>Hello !</p>")
```
masenf, line 17: lets add another test case that passes an f-string to `Html.create` to hit https://github.com/reflex-dev/reflex/issues/2155

masenf (commented):
review complete. ping me when you're ready for re-review

reflex/ivars/number.py
adhami3310: the reason i put it there in the first places because the condition is of `BooleanVar` type which is in the same file

reflex/components/core/match.py
adhami3310: it gets returned as is

reflex/ivars/base.py
```diff
@@ -0,0 +1,1736 @@
...
+                except ValueError:
+                    console.debug(
+                        f"{type(state).__name__}.{self._var_name}: Failed conversion of {value} to '{self._var_type.__name__}'. Value not set.",
+                    )
+            else:
+                setattr(state, actual_name, value)
+
+        setter.__qualname__ = self.get_setter_name()
+
+        return setter
```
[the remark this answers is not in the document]
adhami3310: The issue lies in inheritance. We would need to subclass all Var variants or do something similarly ugly.

reflex/ivars/base.py
```diff
@@ -0,0 +1,1736 @@
...
+
+
+OUTPUT = TypeVar("OUTPUT", bound=ImmutableVar)
+
+
+class LiteralVar(ImmutableVar):
+    """Base class for immutable literal vars."""
+
+    @classmethod
+    def create(
```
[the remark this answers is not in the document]
adhami3310: We can simply call serialize on the value if the known types don't resolve, and then recall LiteralVar on that.

reflex/ivars/object.py
adhami3310: unsure, it's your call

masenf (approved):

PR #3749, opened 2024-08-06
fix import clash between connectionToaster and hooks.useState

### LINKED ISSUE #3604: [REF-3219] use hooks.useState will raise: ValueError: Can not compile, the tag useState is used multiple time from react and react
**Describe the bug**
use `hooks.useState`, will raise: ValueError: Can not compile, the tag useState is used multiple time from react and react

`hooks._compose_react_imports` function, create ImportVar object with `install=False`; and in banner.py, `ConnectionToaster.add_hooks` create ImportVar object with `install=True`; 

these will make `compler.validate_imports` raise ValueError.

<sub>[REF-3219](https://linear.app/reflex-dev/issue/REF-3219/use-hooksusestate-will-raise-valueerror-can-not-compile-the-tag)</sub>

Lendemor:
closes #3604

### DIFF
--- a/reflex/components/core/banner.py
+++ b/reflex/components/core/banner.py
@@ -141,7 +141,7 @@ useEffect(() => {{
             _var_is_string=False,
         )
         imports: ImportDict = {
-            "react": ["useEffect", ImportVar(tag="useState", install=False)],
+            "react": ["useEffect", "useState"],
             **target_url._get_imports(),  # type: ignore
         }
         hook._var_data = VarData.merge(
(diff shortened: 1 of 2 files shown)


picklelo (approved):

masenf (commented):
i think it would be better to change `hooks.useState` to not specify `install=False`, since `react` is never installed, it's less likely to clash with other places/3rd party packages that need `useState`, including `rx._x.client_state`

picklelo (approved):

PR #3753, opened 2024-08-07
removed posthog secret

itsmeadarsh2008:
### All Submissions:

### Type of change

**Dear maintainer(s), you have leaked a posthog secret in the source code. I have replaced it with os.getenv. Please edit my code if it's incorrect.**

### DIFF
--- a/benchmarks/utils.py
+++ b/benchmarks/utils.py
@@ -60,7 +60,7 @@ def send_data_to_posthog(event, properties):
         HTTPError: When there is an error sending data to PostHog.
     """
     event_data = {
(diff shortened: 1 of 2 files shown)


Lendemor:
This is a public key that is needed to send the data for the telemetry.

Your fix would make it not work out of the box.

Since we want telemetry to work on any basic install (unless turned off by user) then we need to have the key part of the installable package.

itsmeadarsh2008:
> This is an API key that is needed to send the data for the telemetry.
> 
> Your fix would make it not work out of the box.
> 
> Since we want telemetry to work on any basic install (unless turned off by user) then we need to have the key part of the installable package.

So, it's not dangerous to keep the secret public?

PR #3768, opened 2024-08-09
minor State cleanup

benedikt-bartscher:


reflex/state.py
```diff
@@ -1259,7 +1271,7 @@ async def _populate_parent_states(self, target_state_cls: Type[BaseState]):
 
         # Fetch all missing parent states and link them up to the common ancestor.
         parent_states_tuple = self._get_parent_states()
-        root_state = parent_states_tuple[-1][1]
+        root_state = self._get_root_state()
```
picklelo, line 1274: Since parent_states_tuple is already calculated, wouldn't this introduce additional unneeded computation?

benedikt-bartscher: yes, it does. i can change this line back

benedikt-bartscher: done

picklelo (approved):

PR #3769, opened 2024-08-09
wip: iter states

benedikt-bartscher:


benedikt-bartscher:
@masenf do you have any idea if there is a cool way to automatically detect if `modify_states` was called from an event handler to set `from_state=self` for deadlock prevention?

benedikt-bartscher:
> @masenf do you have any idea if there is a cool way to automatically detect if `modify_states` was called from an event handler to check for `from_state=self` for deadlock prevention?

We could also add a `BaseState.modify_states` partial-like method which sets `from_state=self ` and advise users to use `self.modify_states` in event handlers

reflex/utils/string.py
```diff
@@ -0,0 +1,21 @@
+"""String utility functions."""
+
+import sys
+
+
+def remove_prefix(text: str, prefix: str) -> str:
```
picklelo, line 6: nit: this can be moved to `format.py`

picklelo (changes requested):
Sorry for the delay here. I think the changes look good - we need to rebase this onto `main` and implement it for `StateManagerDisk` which is becoming the default in 0.7.0

benedikt-bartscher:
> Sorry for the delay here. I think the changes look good - we need to rebase this onto `main` and implement it for `StateManagerDisk` which is becoming the default in 0.7.0

No worries, I just implemented it for `StateManagerDisk`

picklelo (approved):
Code works well for me. I'll wait for @masenf 's final approval to merge. What was the use case for this?

benedikt-bartscher:
> Code works well for me. I'll wait for @masenf 's final approval to merge. What was the use case for this?

You can use it for cache invalidation, f.e. to have all blog pages as infinitely cached computed vars and invalidate/reload them globally if a page changes.

benedikt-bartscher:
Will rebase if there is intention to merge this

reflex/utils/prerequisites.py
```diff
@@ -279,6 +282,18 @@ def get_app(reload: bool = False) -> ModuleType:
         raise
 
 
+def get_app(reload: bool = False) -> App:
```
masenf, line 285: can we make this new function `get_app_instance` and leave `get_app` the way it was, to retain as much compatibility as possible

reflex/state.py
```diff
@@ -2812,6 +2864,24 @@ async def modify_state(self, token: str) -> AsyncIterator[BaseState]:
...
+        """Iterate over the state names.
+
+        Args:
+            substate_cls: The subclass of BaseState to filter by.
+
+        Yields:
+            The state names.
+        """
+        for path in self._iter_pkl_files():
+            token = path.stem
```
masenf, line 2880: unfortunately this wont work anymore, because we started hashing the token to avoid file path limits on windows.

however, we should be able to use the same implementation here as we do for `StateManagerMemory` as the on-disk pickles are only read when the backend starts after a hot reload

masenf (commented):
the main problem i forsee with this feature is that it works great in testing/dev mode, and then once it's deployed prod with redis, some clients are not going to get deltas sent when their state is modified, because they're connected to a different instance of the app than where this API is used.

it's not a new problem, we've had it for a long time. but currently there is no mechanism for an app instance to send a delta to a client that is connected to a different instance. @Lendemor had a WiP to handle this, but i think it's closed without merging.

Lendemor:
Marking this as draft until ready for review.

benedikt-bartscher:
> it's not a new problem, we've had it for a long time. but currently there is no mechanism for an app instance to send a delta to a client that is connected to a different instance. @Lendemor had a WiP to handle this, but i think it's closed without merging.

Thanks for the review. Yes, that's currently a limitation, but i think it can be implemented with redis pub/sub and some maps.
Do you mean this PR https://github.com/reflex-dev/reflex/pull/2819 ?

masenf:
Closing in favor of #5927 

https://github.com/reflex-dev/reflex/pull/5927/files#diff-f54b166b5fbb1cbd472d99aa3a943150590610a2f3fd2db8745f277ce443b9b1R70-R77

Using the new `TokenManager.enumerate_tokens` interface, along with `lost+found`, we can finally achieve the dream of editing states on arbitrary instances _and_ delivering updates to the proper frontend client

PR #3790, opened 2024-08-13
do not reload compilation if using local app in AppHarness

Lendemor:


reflex/testing.py
```diff
@@ -274,7 +274,9 @@ def _initialize_app(self):
             before_decorated_pages = reflex.app.DECORATED_PAGES[self.app_name].copy()
             # Ensure the AppHarness test does not skip State assignment due to running via pytest
             os.environ.pop(reflex.constants.PYTEST_CURRENT_TEST, None)
-            self.app_module = reflex.utils.prerequisites.get_compiled_app(reload=True)
+            self.app_module = reflex.utils.prerequisites.get_compiled_app(
+                reload=self.app_source is not None
```
masenf, line 278:
```suggestion
                # Do not reload the module for pre-existing apps (only apps generated from source)
                reload=self.app_source is not None
```

masenf (commented):
maybe a comment for future maintainers?

reflex/testing.py
Lendemor: sure 👍

masenf (approved):

PR #3812, opened 2024-08-20
delay page until _compile gets called

adhami3310:


reflex/app.py
```diff
@@ -515,58 +538,38 @@ def add_page(
...
-                else format.make_default_page_title(get_config().app_name, route)
-            ),
-            "image": image,
-            "meta": meta,
-        }
+        self.unevaluated_pages[route] = UnevaluatedPage(
+            component=component,
+            route=route,
+            title=title or constants.DefaultPage.TITLE,
+            description=description or constants.DefaultPage.DESCRIPTION,
```
masenf, line 550: prefer a ternary here, in case `title` is a Var

```suggestion
            title=title if title is not None else constants.DefaultPage.TITLE,
            description=description if description is not None else constants.DefaultPage.DESCRIPTION,
```

adhami3310:
We should keep this around for 0.6.3 because it has considerable changes and likely has bugs.

reflex/app.py
```diff
@@ -969,36 +931,70 @@ def get_compilation_time() -> str:
...
+
+        for route, page in self.unevaluated_pages.items():
+            if route in self.pages:
+                continue
+
+            ExecutorSafeFunctions.UNCOMPILED_PAGES[route] = page
+
+        ExecutorSafeFunctions.STATE = self.state
+
+        pages_results = []
```
masenf, line 947: seems unexpectedly duplicated with the block below

Lendemor (approved):

masenf (approved):
alright lets bring this in now for 0.6.5 so we can get some bake time on `main`

Lendemor:
After investigation, this PR seems to have caused #4384

PR #3824, opened 2024-08-22
Upper bound for reflex-chakra dependency

masenf:
We have to make a breaking change in reflex-chakra, and it will happen in 0.6.0, so ensure that reflex 0.5.10 never installs the breaking version of reflex-chakra.


picklelo (approved):
we can remove the dependency remove for now I think

PR #3828, opened 2024-08-24
add generate palette function in experimental

Lendemor:
Add the logic to generate radix custom color palette directly in Reflex.


pyproject.toml
```diff
@@ -62,6 +62,7 @@ twine = ">=4.0.0,<6.0"
 tomlkit = ">=0.12.4,<1.0"
 lazy_loader = ">=0.4"
 reflex-chakra = ">=0.1.1a1"
+coloraide = "^3.3.1"
```
picklelo, line 65: I don't want to include another dependency (trying to cut down the default package install size). I think we should make this an optional install for those who want to use the custom color feature. Or reimplement the functions we use, but that seems like a lot of work.

Lendemor: I'll look into that.

picklelo (approved):

masenf (commented):
maybe this could be added to the `_x` namespace via a property that delays the import. we definitely don't want to be importing this unless it will be used.

PR #3843, opened 2024-08-27
Forbid Computed var shadowing

benedikt-bartscher:
based on #3833


benedikt-bartscher:
will update and fix conflicts once https://github.com/reflex-dev/reflex/pull/3833 is merged

picklelo (approved):

PR #3845, opened 2024-08-27
Only serialize used rx.Base fields

abulvenz:
If you check the payloads when running the example below, you might notice that `value` is only serialized when you comment in line 43 where the `value` is displayed.
Fields are checked on class level, so always the union of all used fields will be serialized (Not per usage, which would be cool, but more sophisticated.)
```python

import reflex as rx
from reflex.base import SlimBase


def on_frontend_exception(exception: Exception) -> None:
    pass

app: rx.App = rx.App(frontend_exception_handler=on_frontend_exception)


class SmartObjectChild(SlimBase):
    name: str = "Smart Object Child"
    value: int = 0

class SmartObject(SlimBase):
    name: str = "Smart Object"
    value: int = 0
    child: SmartObjectChild = SmartObjectChild()
    children: list[SmartObjectChild] = [SmartObjectChild(), SmartObjectChild()]

    def increment(self):
        self.value += 1

    def decrement(self):
        self.value -= 1


class LSpace(rx.State):
    computed_var: str = "Computed var"
    some_list: list[SmartObject] = [SmartObject(), SmartObject()]

    smartie: SmartObject = SmartObject()

    def set_smartie_name(self, name: str) -> None:
        self.smartie.name = name


def index() -> rx.Component:
    return rx.box(
        rx.heading("Test smart object!"),
        rx.foreach(LSpace.some_list, lambda x: rx.box(rx.text(x.name))),
       # rx.text(LSpace.smartie.value),
        rx.text(LSpace.smartie.child.name),        #rx.text(LState.smartie.child.value),

        rx.foreach(LSpace.smartie.children, lambda x: rx.box(rx.text(x.name))),


        rx.button("Decrement", on_click=LSpace.set_smartie_name("Decremented")),
        rx.button("Increment", on_click=LSpace.set_smartie_name("Incremented")),
    )

app.add_page(index)

```


benedikt-bartscher:
@abulvenz I think we can rebase this and mark it as ready
should we make SlimBase the default Base?

adhami3310:
@abulvenz is this relevant after the removal of rx.Base

PR #3846, opened 2024-08-27
[REF-3633] Introduce a workaround for enterprise users who get stuck with httpx.get SSL

masenf:
Setting SSL_NO_VERIFY=1 will disable SSL verification during `reflex init`


Alek99 (approved):
lgmt

PR #3848, opened 2024-08-28
Exclude reflex-ai scratch from hot reload

picklelo:
Hardcode the scratch directory in the unicorn exclude.


masenf (commented):
i think if we can just put the scratch dir inside `.web` (per discord chat), then that's a cleaner solution with the side benefit of we're not really expected to clean it up, and `.web` is a generated directory and shouldn't be checked into version control.

PR #3853, opened 2024-08-28
[REF-3589] raise EventHandlerArgMismatch when event handler args don't match spec

masenf:
2 new error messages when the event trigger spec does not match the function definiton.

```
reflex.utils.exceptions.EventHandlerArgMismatch: The number of arguments accepted by State.no_arg (0) does not match the arguments passed by the event trigger: ['form_data']
See https://reflex.dev/docs/events/event-arguments/
```

```
reflex.utils.exceptions.EventFnArgMismatch: The number of arguments accepted by <function index.<locals>.<lambda> at 0x10c791ee0> (0) does not match the arguments passed by the event trigger: ['form_data']
See https://reflex.dev/docs/events/event-arguments/
```

### DIFF
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -527,15 +527,7 @@ class Component(BaseComponent, ABC):
             for v in value:
                 if isinstance(v, (EventHandler, EventSpec)):
                     # Call the event handler to get the event.
-                    try:
-                        event = call_event_handler(v, args_spec)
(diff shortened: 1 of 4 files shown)


masenf:
Need to strip one argument off the count when the event fn is a bound method / partial, etc.

picklelo (approved):

PR #3860, opened 2024-08-30
Use bun run when supported

picklelo:
Remove dependency on FNM + Node on supported operating systems as it seems bun can now run and export the apps well. For unsupported systems, we fallback to use NPM for everything.


picklelo:
Bun failed to export reflex-web, closing this for now.

PR #3868, opened 2024-09-02
Adding array to array pluck operation.

abulvenz:
Who doesn't know it, who doesn't like it?
https://underscorejs.org/#pluck

It can be quite useful to reduce state duplications, e.g. when you have selections of objects, you pluck a field and use array.contains to check in the selected items.

If there are any questions, please let me know!

```python
class Person(rx.Base):
  first_name: str
  last_name: str

class State:
  persons: list[Person] = []

.... rx.cond(State.persons.pluck("last_name").contains("Furer"), "There is a maintainer in the list", "... ...
```


masenf (dismissed):
nice! way to use the new modular var operations api 🎉 

cc @adhami3310 for input

adhami3310:
This looks fine! We can merge it and then I will add field guards with my open PR.

benedikt-bartscher:
@masenf we have had some migration challenges in our projects, but overall we really like the new var api. Thank you!

reflex/ivars/sequence.py
```diff
@@ -707,6 +707,19 @@ def contains(self, other: Any) -> BooleanVar:
...
+    def pluck(self, field: StringVar | str) -> ArrayVar:
+        """Pluck a field from the array.
+
+        Args:
+            field: The field to pluck from the array.
+
+        Returns:
+            The array pluck operation.
+        """
+        if field is None:
```
benedikt-bartscher, line 719: Either remove this or add `None` to the field type union

abulvenz: OK, I just removed it. It should anyway have resulted in an empty array for logical reasons.

abulvenz:
@masenf Thanks for approving. Usually I reset approvals by some tiny commits afterwards :100: Is there any other best practice or is this OK? What is most convenient for You as maintainers?

Issue #3745, opened 2024-08-05
Add Load Tests Using Locust


Alek99:
We are looking to add load tests to Reflex, to show how many connections we can handle and what this does cpu utilization. Will use this library https://github.com/locustio/locust. This will probably need some plugin like https://github.com/SvenskaSpel/locust-plugins/blob/master/examples/socketio_ex.py to handle websocket interactions since it's all pretty vanilla http out of the box.

Some things we want to test for: 
- idle websocket connections
- rps events handled (no state modification)
- rps state modifications with delta (probably router_data)
- rps (small) substate modifications

These probably need to be with and without redis (but that shouldn't affect how the tests are written)

tsaipraveen99:
Hi, I would love to work on this issue. Could you please assign it to me?

Alek99:
Yes will do!

Alek99:
> Hi, I would love to work on this issue. Could you please assign it to me?

Hi are you still working on this ticket another contributors is interested so please let me know, if not I will un assign and hand it over.

tsaipraveen99:
Hey Alex, you can assign this issue to other contributors. Thanks!

Alek99:
Sure I'll assign it to you, yes second approach sounds good!

masenf:
may reapproach down the road, but since any load testing is going to be app-specific, it makes more sense to invest in micro benchmarks via codspeed throughout the framework and potentially add load tests for driving some select sample apps.

PR #3870, opened 2024-09-02
[WIP] - Add load tests for reflex using locust

sksDonni:
Work in progress PR requesting comments  for #3745

### DIFF
--- a/integration/loadtesting/.gitignore
+++ b/integration/loadtesting/.gitignore
@@ -1,2 +1,5 @@
-./env
+*.db
+*.py[cod]
 .web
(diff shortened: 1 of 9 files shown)


Lendemor:
Marking this as Draft until the issue in CI are addressed : 
- pre-commit must be run on the file impacted by the PR

masenf:
I have another `masenf/locust` branch based on this PR and cleaned up a bit which I will be resubmitting. We don't need this much UI stuff for the load test with websockets as they never actually open the browser. We could be running backend-only and still get the same information.

PR #3875, opened 2024-09-03
Use old serializer system in LiteralVar

adhami3310:


reflex/ivars/number.py
```diff
@@ -20,13 +21,33 @@
...
+    """Raise an unsupported operand types error.
+
+    Args:
+        operator: The operator.
+        operands_types: The types of the operands.
+
+    Raises:
+        VarTypeError: The operand types are unsupported.
+    """
+    raise VarTypeError(
```
masenf, line 42: i think the proper thing to do in this scenario is to `return NotImplemented` [[1]](https://docs.python.org/3.6/library/constants.html#NotImplemented)

Not sure if that totally makes sense for us though, because ostensibly we're not expecting an incompatible type to have a better comparison checker than the Var itself, and raising the exception means we fail loud and fast (and record the exception type for telemetry).

Not suggesting a change here, just posing a point for discussion.

masenf (dismissed):

reflex/ivars/number.py
adhami3310: that seems valid, we would need to also remove `.operation` from `Var` for both to work. Then Python should do the heavy lifting for us.

picklelo (approved):
This PR fixes the ClientState issues and I can run flexgen prod with it. Should we consider adding an integration test for ClientState? Currently it doesn't work on `main` though all the tests are passing.

reflex/ivars/base.py
```diff
@@ -759,10 +758,123 @@ def _type(self) -> StringVar:
...
+        """
+        return self
+
+    def reverse(self):
+        """Reverse the var.
+
+        Raises:
+            TypeError: If the var does not support reverse.
+        """
+        raise TypeError("Cannot reverse non-list var var.")
```
masenf, line 801:
```suggestion
        raise TypeError("Cannot reverse non-list var.")
```

reflex/ivars/base.py
```diff
@@ -759,10 +758,123 @@ def _type(self) -> StringVar:
...
+        Returns:
+            The decoded value or the Var name.
+        """
+        try:
+            return json.loads(str(self))
+        except ValueError:
+            try:
+                return json.loads(self.json())
+            except (ValueError, NotImplementedError):
+                return str(self)
```
masenf, line 847: for `LiteralVar`, it seems like this should just return the original value if we know it, because the `str` operation does not necessarily return json

reflex/ivars/sequence.py
```diff
@@ -696,30 +929,176 @@ def range(
...
+
+        Returns:
+            The array greater than or equal operation.
+        """
+        if not isinstance(other, (ArrayVar, list, tuple)):
+            raise_unsupported_operand_types(">=", (type(self), type(other)))
+
+        return array_ge_operation(self, other)
+
+    def foreach(self, fn: Any):
```
masenf, line 1058: do we have any test cases for this?

reflex/ivars/sequence.py
```diff
@@ -758,13 +1139,13 @@ def _cached_var_name(self) -> str:
         )
 
     @cached_property_no_lock
-    def _cached_get_all_var_data(self) -> ImmutableVarData | None:
-        """Get all the VarData associated with the Var.
+    def _cached_get_all_var_data(self) -> VarData | None:
+        """Get all the VarData asVarDatae Var.
```
masenf, line 1143: seems like a typo

reflex/ivars/sequence.py
```diff
@@ -933,6 +1314,62 @@ def array_reverse_operation(
     )
 
 
+@var_operation
+def array_lt_operation(lhs: ArrayVar | list | tuple, rhs: ArrayVar | list | tuple):
```
masenf, line 1318: these `array_*_operation` functions seem really similar to the string flavored functions, are they different just due to the annotations? or is there some other difference i'm missing

reflex/ivars/sequence.py
```diff
@@ -1029,42 +1486,67 @@ def array_contains_operation(haystack: ArrayVar, needle: Any | Var):
     frozen=True,
     **{"slots": True} if sys.version_info >= (3, 10) else {},
 )
-class ToStringOperation(CachedVarOperation, StringVar):
+class ToStringOperation(StringVar):
```
masenf, line 1489: seems like we could use a new mixin for `ToXOperation` that removes the `__getattr__`, `__post_init__`, and `__hash__` boilerplate

reflex/vars.py
```diff
@@ -599,59 +310,37 @@ def create(
...
+                value,
+                _var_data=_var_data,
             )
 
-        return BaseVar(
-            _var_name=name,
-            _var_type=type_,
-            _var_is_local=_var_is_local,
-            _var_is_string=_var_is_string if _var_is_string is not None else False,
+        # If the value is a string, create a LiteralVar.
```
masenf, line 335: comment doesn't seem to describe what's happening

masenf (commented):
review checkpoint. haven't tested it locally yet, interested to see how you were able to clear out so much of reflex.utils.format and keep things working

reflex/components/core/banner.py
```diff
@@ -68,7 +68,7 @@ def create(cls) -> ImmutableVar:
         """
         return ImmutableVar(
             _var_name="getBackendURL(env.EVENT).href",
-            _var_data=ImmutableVarData(
+            _var_data=VarData(
```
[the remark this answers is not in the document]
adhami3310: ImmutableVarData doesn't exist anymore

reflex/ivars/sequence.py
adhami3310: no test exist for foreach, it's spec has been always loosely defined. I only implemented it because reflex-chakra did.

reflex/ivars/sequence.py
adhami3310: it's only annotations wise, we could in theory merge the two although the functions would kinda except anything (which could be ok)

masenf (approved):

PR #3877, opened 2024-09-04
add fragment to foreach

adhami3310:
```py
import reflex as rx

class State(rx.State):
    choices: list[int] = []

    def add_choice(self):
        self.choices.append(len(self.choices))

@rx.page(route="/")
def index():
    return rx.vstack(
        rx.button("Add choice", on_click=State.add_choice),
        rx.match(
            State.choices.length(),
            (0, rx.text("No choices")),
            (1, rx.text("One choice")),
            rx.foreach(State.choices, lambda choice: rx.text(choice)),
        ),
)

app = rx.App()
```

### DIFF
--- a/integration/test_var_operations.py
+++ b/integration/test_var_operations.py
@@ -588,6 +588,16 @@ def VarOperations():
                 int_var2=VarOperationState.int_var2,
                 id="memo_comp_nested",
             ),
+            # foreach in a match
+            rx.box(
+                rx.match(
+                    VarOperationState.list3.length(),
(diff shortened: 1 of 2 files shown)


masenf (approved):
is there a test we can add that hits this case so we don't accidentally regress it when refactoring away from jinja

adhami3310:
> is there a test we can add that hits this case

the one above works, it would be better in an integration test to just validate that the case compiles, should i add it to this PR?

picklelo (approved):

PR #3907, opened 2024-09-10
Remove Pydantic from some classes

adhami3310:


masenf (commented):
this sets us up to integrate both components and event handlers with the new var system. nice work.

i like the frozen instances everywhere, but there has to be a nicer way to init without seeing lots of `object.__setattr__` calls. does dataclass not handle that internally with `frozen=True`? could we call `super().__init__` to the generated method?

reflex/state.py
```diff
@@ -1795,8 +1845,13 @@ def dict(
...
             self._mark_dirty()
 
+        def dictify(value: Any):
+            if dataclasses.is_dataclass(value) and not isinstance(value, type):
+                return dataclasses.asdict(value)
+            return value
+
         base_vars = {
-            prop_name: self.get_value(getattr(self, prop_name))
+            prop_name: dictify(self.get_value(getattr(self, prop_name)))
```
benedikt-bartscher, line 1854: This change caused issues for us, because `dataclasses.asdict` raises exceptions when used with sqlalchemy mapped dataclasses. Do we even need this?
If we need it, can we add `and not has_serializer(type(value))` to the condition in `dictify` to fix it?

cc @adhami3310 @masenf

adhami3310: It should work without it as we have now logic covering dataclasses inside of serialize. I'll make a PR soon.

adhami3310: #4141

PR #3910, opened 2024-09-11
override dict in propsbase to use camelCase

adhami3310:


masenf (dismissed):

masenf:
need to fix these https://github.com/reflex-dev/reflex/blob/63bf1b8817202ef70ba7e024a5e6321f2e220e55/reflex/components/sonner/toast.py#L174-L181

since they're going to be camelCase now

Lendemor (approved):

masenf (approved):

PR #3911, opened 2024-09-11
fallback to ObjectVar in guess_type

benedikt-bartscher:
`ImmutableVar.__getattr__` does not allow attribute access and `guess_type` does not guess `ObjectVar` for arbitrary objects. In our codebases, this results in a lot `var.to(ObjectVar)`s  
this pr currently just adds a fallback. Maybe we can instead check if there is a serializer which serializes to dict?


tests/components/datadisplay/test_datatable.py
```diff
@@ -40,13 +40,13 @@ def test_validate_data_table(data_table_state: rx.State, expected):
...
-    expected = f"{state_name}.{expected}" if expected else state_name
+    var = data_table_state
+    if expected:
+        var = getattr(var, expected)
 
     assert data_table_dict["props"] == [
-        f"columns={{{expected}.columns}}",
-        f"data={{{expected}.data}}",
+        f"columns={{{var.columns._var_name}}}",
+        f"data={{{var.data._var_name}}}",
```
masenf, line 49:
```suggestion
        f"columns={{{str(var.columns)}}}",
        f"data={{{str(var.data)}}}",
```

This should work, because `_var_name` has gone away recently as an implementation detail.

masenf (commented):
@adhami3310 what do you think of this one

adhami3310:
I'm a bit concerned here on the general case, for example, a union of an int and str might be an ObjectVar now. It feels, impure, but maybe practicality trumps correctness.

To engage more, why were you creating so many ObjectVar? Can we find a better adjustment to the API to allow this behavior be simpler as an alternative?

benedikt-bartscher:
> To engage more, why were you creating so many ObjectVar? Can we find a better adjustment to the API to allow this behavior be simpler as an alternative?

Because we often write serializers for existing objects and just want to use them in the frontend without creating an additional rx.Base "proxy class".
We could probably add some kind of marker which marks all classes which should be ObjectVars. Otherwise maybe follow my original idea from the PR description and only fallback to ObjectVar if there is a serializer registerd which serializes to dict/json.

PR #3919, opened 2024-09-13
can run with granian by setting REFLEX_USE_GRANIAN var

Lendemor:
WIP: allow use of Granian instead of the uvicorn/gunicorn for running backend.


itsmeadarsh2008:
Thanks for this PR.

bertbarabas:
Is it worth trying this currently? I'd love to check if it helps the performance of the file upload since that problem seems to be in uvicorn/gunicorn.

reflex/utils/exec.py
```diff
@@ -184,25 +193,46 @@ def run_backend(
...
-        app=f"{app_module}.{constants.CompileVars.API}",
-        host=host,
-        port=port,
-        log_level=loglevel.value,
-        reload=True,
-        reload_dirs=[config.app_name],
-    )
+    if should_use_granian():
+        console.debug("Using Granian for backend")
+        try:
```
picklelo, line 205: We should move this block into its own function

reflex/utils/exec.py
```diff
@@ -225,41 +255,77 @@ def run_backend_prod(
...
-        ]
-        if constants.IS_WINDOWS
-        else [
-            *RUN_BACKEND_PROD,
-            "--bind",
-            f"{host}:{port}",
-            "--threads",
+
+    if should_use_granian():
+        try:
```
picklelo, line 261: Should we do this in a more modular way? Rather than having if statements here make it possible to runners from the outside.

reflex/utils/exec.py
Lendemor: I can do that.

reflex/utils/exec.py
Lendemor: For now I want people to be able to test it to get feedback on it, since it'll only be used if the flag is set.
We can make it more modular in the future if feedback is positive, to let people choose what they want.

Lendemor:
> Is it worth trying this currently? I'd love to check if it helps the performance of the file upload since that problem seems to be in uvicorn/gunicorn.

We would love to have more people trying it.
It is working, but testing in real condition rather than on basic app will let us know if anything need to be adjusted.

bertbarabas:
How would I try this out? Is it in vv0.6.0a2 or do I need to do more or perhaps wait for it?

Lendemor:
> How would I try this out? Is it in vv0.6.0a2 or do I need to do more or perhaps wait for it?

If you want to try it right now, you can clone the project and use this branch to run reflex instead of the installed package.

It is not part of `0.6.0a2`, and I'm not sure if it will be in `0.6.0`, need to discuss it with the team.
At the latest I think it'll be in `0.6.1`

masenf:
> How would I try this out?

```console
pip install git+https://github.com/reflex-dev/reflex@lendemor/use_granian granian
REFLEX_USE_GRANIAN=1 reflex run
```

> Is it in vv0.6.0a2 or do I need to do more or perhaps wait for it?

planned for `v0.6.1` in ~2 weeks

@Lendemor I tried this out and it seems to work in dev mode, but the websocket doesn't connect in prod mode and `/ping` gives "Internal Server Error"

Lendemor:
I'll look into it, prod mode was running fine before my last commit I think

bertbarabas:
Here are some timings of uploading a file of size: 522,134,318 bytes in my actual application in dev with browser and backend on the same laptop.
(I killed reflex and relaunched between tests)

REFLEX_USE_GRANIAN=0 reflex run
Run 1 took: 127 seconds
Run 2 took: 183 seconds

REFLEX_USE_GRANIAN=1 reflex run
Run 1 took: 63 seconds
Run 2 took: 104 seconds

Looks like Granian is faster although my laptop may not be the best place to measure speed, maybe the laptop gets slower as it gets hotter…

Lendemor:
> I'll look into it, prod mode was running fine before my last commit I think

Turned out the backend never actually worked, because I tested it with a stateless app.

It's working now.

itsmeadarsh2008:
> Here are some timings of uploading a file of size: 522,134,318 bytes in my actual application in dev with browser and backend on the same laptop. (I killed reflex and relaunched between tests)
> 
> REFLEX_USE_GRANIAN=0 reflex run Run 1 took: 127 seconds Run 2 took: 183 seconds
> 
> REFLEX_USE_GRANIAN=1 reflex run Run 1 took: 63 seconds Run 2 took: 104 seconds
> 
> Looks like Granian is faster although my laptop may not be the best place to measure speed, maybe the laptop gets slower as it gets hotter…

Try it in a codespace.

adhami3310 (approved):

PR #3932, opened 2024-09-16
fix template fetch during init

Lendemor:


reflex/utils/prerequisites.py
```diff
@@ -1280,11 +1280,14 @@ def get_release_by_tag(tag: str) -> dict | None:
...
-        for tp in templates_data
-        if not tp["hidden"] and tp["code_url"] is not None
-    }
+
+    filtered_templates = {}
+    for tp in templates_data:
+        if tp["hidden"] or tp["code_url"] is None:
+            continue
+        tp.pop("hidden")
+        filtered_templates[tp["name"]] = Template(**tp)
```
masenf, line 1289: i think this would be more robust if we only include the fields that the class recognizes

```suggestion
        known_fields = set(f.name for f in dataclasses.fields(Template))
        filtered_templates[tp["name"]] = Template(**{k: v for k, v in tp.items() if k in known_fields})
```

masenf (approved):

PR #3934, opened 2024-09-16
use serializer for state update and rework serializers

adhami3310:


reflex/utils/serializers.py
```diff
@@ -264,13 +263,11 @@ def serialize_list(value: Union[List, Tuple, Set]) -> str:
     Returns:
         The serialized list.
     """
-    from reflex.vars import LiteralArrayVar
-
-    return str(LiteralArrayVar.create(value))
+    return [serialize(item) for item in value]
```
masenf, line 266: do we really have to recursively call `serialize` ourselves here and below? i thought having `default=serialize` in `format.json_dumps` meant we could just return the values as they are

adhami3310: My understanding is that if `json.dumps` calls the `default` function should return something that can be serialized directly, without further calls to `default`. I'm unsure why it even calls it as the docs say:

> If specified, default should be a function that gets called for objects that can’t otherwise be serialized. It should return a JSON encodable version of the object or raise a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError). If not specified, [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) is raised.

masenf (approved):

PR #3943, opened 2024-09-17
add special handling for infinity and nan

adhami3310:


reflex/vars/number.py
```diff
@@ -1038,7 +1039,12 @@ def json(self) -> str:
 
         Returns:
             The JSON representation of the var.
+
+        Raises:
+            ValueError: The JSON representation is invalid.
```
Lendemor, line 1044: I think we want to use more specific exception classes from now on, so maybe something like `InvalidJSONReprError` or something would be better here ?

masenf (dismissed):
unit test here would be nice

masenf (approved):
tks

PR #3946, opened 2024-09-18
use is true for boolify

adhami3310:


masenf (approved):

masenf (commented):
adding this to the integration test_var_operations.py might be helpful if this type of change doesn't trigger a test change, then we must not have good enough coverage

adhami3310:
I believe you meant `test_vars.py` (the unit one) and not `test_var_operations.py` (the integration one, that only tests the output).

But your assumption is correct, there is no tests that match with `.bool()`.

PR #3949, opened 2024-09-18
add few test cases for bool

adhami3310:


tests/test_var.py
```diff
@@ -958,6 +958,21 @@ def test_all_number_operations():
     )
 
 
+@pytest.mark.parametrize(
+    "var,expected",
```
masenf, line 962:
```suggestion
    ("var", "expected"),
```

although csv string does work, the preferred way i think is to use a tuple of args

masenf (approved):

masenf (approved):

PR #3963, opened 2024-09-20
[ENG-3824]Make AI template use light mode

ElijahAhianyo:
until the web editor generates code that is compatible with our color mode switching, we need to force the AI template (`reflex init --ai`) to use light mode


Lendemor (approved):

masenf (approved):
working for me

PR #3974, opened 2024-09-23
[ENG-3833] handle object in is bool

adhami3310:


reflex/.templates/web/utils/state.js
```diff
@@ -832,7 +832,11 @@ export const useEventLoop = (
  * @returns True if the value is truthy, false otherwise.
  */
 export const isTrue = (val) => {
-  return Array.isArray(val) ? val.length > 0 : !!val;
+  return Array.isArray(val)
+    ? val.length > 0
+    : val == Object(val)
+    ? Object.keys(val).length > 0
+    : Boolean(val);
 };
```
masenf, line 840: the readability here suffers here a bit, but maybe i'm just unaccustomed to reading nested js ternary ops

i feel like this would make more sense to me

```javascript
if (Array.isArray(val)) {
    return val.length > 0;
}
if (val == Object(val)) {
    return Object.keys(val).length > 0;
}
return Boolean(val);
```

your choice, either way is equivalent

masenf (approved):

masenf (approved):

PR #3977, opened 2024-09-23
remove on_scroll because it doesn't work

### LINKED ISSUE #463: Split the on_scroll Base Event Trigger
Split on_scroll, into on_scroll_up and on_scroll_down. In it's current state it's not insanely useful knowing only when the user scrolls the page and not which direction they may be scrolling.

Lendemor:
close #463


masenf (commented):
not sure who is using this event, but we probably should deprecate it rather than remove it without notice

PR #3979, opened 2024-09-23
remove format_state and override behavior for bare

adhami3310:


masenf (commented):
seems we can now also remove `serialize_list` and `serialize_dict`

masenf (approved):
red diff ftw

PR #3981, opened 2024-09-24
reorganize all tests in a single top folder

### LINKED ISSUE #3905: Add option for Nodejs path like `BUN_PATH`
**Describe the bug**
Reflex tries to install the required node version by itself into data directory but this fails on systems like NixOS or Guix because its not allowed to run binaries directly

```
NixOS cannot run dynamically linked executables intended for generic
linux environments out of the box. For more information, see:
https://nix.dev/permalink/stub-ld
```

**To Reproduce**
- reflex init and choose any template
- reflex run

**Expected behavior**
Reflex runs without any errors just like in the documentation

 - Python Version: Python 3.12.5
 - Reflex Version: 0.5.10
 - OS: NixOS 22.04
 
**Additional context**
Related issue: https://github.com/reflex-dev/reflex/issues/1470

Lendemor:


masenf (commented):
this one does kind of scare me a little bit... what if we're not compatible with the latest node version? (idk we don't test it)

Lendemor:
Understandable, but at the same time, we already have that risk with windows users, so if we ever need to restrain a specific node version, we'll need to hot patch anyway to limit said version.
So making this behaviour the same for all systems does make sense to me ?

Also I think it's probably better to take that risk of a potential risk vs actually preventing some users from using Reflex at all ?

.github/workflows/check_node_latest.yml
```diff
@@ -0,0 +1,38 @@
...
+                run-poetry-install: true
+                create-venv-at-path: .venv
+            - uses: actions/setup-node@v4
+              with:
+                node-version: ${{ matrix.node-version }}
+            - run: |
+                poetry run uv pip install playwright pytest-playwright pyvirtualdisplay pillow
+                poetry run playwright install --with-deps
+            - run: |
+                poetry run pytest integration/test_node_version.py
```
masenf, line 35:
```suggestion
                poetry run pytest integration/
```

can we run _all_ of the integration tests with latest node to improve our coverage and catch more potential errors?

.github/workflows/integration_app_harness.yml
```diff
@@ -51,7 +51,7 @@ jobs:
           SCREENSHOT_DIR: /tmp/screenshots
           REDIS_URL: ${{ matrix.state_manager == 'redis' && 'redis://localhost:6379' || '' }}
         run: |
-          poetry run pytest integration
+          poetry run pytest integration --ignore=integration/test_node_version.py
```
masenf, line 54: we need a better way to skip this test, otherwise it's just going to fail all the time when running locally

.github/workflows/check_node_latest.yml
```diff
@@ -0,0 +1,38 @@
...
+            - uses: ./.github/actions/setup_build_env
+              with:
+                python-version: ${{ matrix.python-version }}
+                run-poetry-install: true
+                create-venv-at-path: .venv
+            - uses: actions/setup-node@v4
+              with:
+                node-version: ${{ matrix.node-version }}
+            - run: |
+                poetry run uv pip install playwright pytest-playwright pyvirtualdisplay pillow
```
masenf, line 32: we need to add pytest-playwright to dev dependencies, and probably remove playwright and playwright-pytest from here

```
  def test_node_version(node_version_app: AppHarness, page: Page):
E       fixture 'page' not found
>       available fixtures: anyio_backend, anyio_backend_name, anyio_backend_options, app_harness_env, benchmark, benchmark_weave, cache, capfd, capfdbinary, caplog, capsys, capsysbinary, class_mocker, cov, doctest_namespace, event_loop, mocker, module_mocker, monkeypatch, no_cover, node_version_app, package_mocker, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, session_mocker, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory, unused_tcp_port, unused_tcp_port_factory, unused_udp_port, unused_udp_port_factory, xvfb
>       use 'pytest --fixtures [testpath]' for help on them.
```

masenf (approved):
@picklelo we're eventually moving all tests into subfolders of `tests/` if you think that's a good idea

adhami3310 (approved):

PR #3996, opened 2024-09-25
Update markdown component map to use new rx.code_block.theme enum

masenf:


reflex/components/markdown/markdown.py
```diff
@@ -173,7 +173,7 @@ def add_imports(self) -> ImportDict | list[ImportDict]:
                 component(_MOCK_ARG)._get_all_imports()  # type: ignore
                 for component in self.component_map.values()
             ],
-            CodeBlock.create(theme="light")._get_imports(),  # type: ignore,
+            CodeBlock.create(theme=Theme.light)._get_imports(),  # type: ignore,
```
Lendemor, line 176: I think we can remove the `# type : ignore` now that we're not using raw string

Lendemor (approved):

PR #4001, opened 2024-09-25
disable prose by default for rx.html

adhami3310:


Lendemor (dismissed):

masenf (commented):
what is the motivation for this change?

adhami3310:
> what is the motivation for this change?

some people (carlos) rely on existing html code that looks very different now, so we're preserving existing behavior.

additionally, it reflects the behavior of doing `rx.el.h1` for example, which is unstyled by default.

masenf:
so this partly reverts the change brought in by #3593

but looking at that PR closer i see that it doesnt do anything with `rx.markdown` but the original bug it was supposed to be fixing specifically mentions markdown and html being affected (maybe markdown got fixed since then).

i guess what im getting at is why not just revert that entire PR? this new styled prop is confusingly named IMO and not particularly discoverable.

adhami3310:
> not just revert that entire PR? this new styled prop is confusingly named IMO and not particularly discoverable

good question! i pose that some might find it convenient to bring back that property, but also, they could do `class_name="prose"` to do the same behavior, making this essentially uesless so :shrug: .

masenf (approved):

PR #4007, opened 2024-09-26
[ENG-3476]Setting State Vars that are not defined should raise an error

ElijahAhianyo:
```python
import reflex as rx

class State(rx.State):
   value: int = 0
   def change_value(self):
       self.value = 10 # valid
       self.num = 5   # error
       self._num = 10 # error
       self.__num = 10 # valid
```

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -1282,10 +1282,9 @@ class BaseState(Base, ABC, extra=pydantic.Extra.allow):
             and name not in self.vars
             and name not in self.get_skip_vars()
         ):
-            available_vars = ", ".join(self.vars) or "None"
             raise AttributeError(
                 f"The state variable '{name}' has not been defined in '{type(self).__name__}'. "
-                f"All state variables must be declared before they can be set. Available vars: {available_vars}"
+                f"All state variables must be declared before they can be set."
             )
 
         # Set the attribute.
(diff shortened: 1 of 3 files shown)


reflex/state.py
```diff
@@ -1233,6 +1236,18 @@ def __setattr__(self, name: str, value: Any):
...
             return
 
+        if (
+            not name.startswith(
+                "_"
+            )  # TODO: skipping backend vars and vars with double leading underscores for now. They should be supported, however.
+            and name not in self.vars
+            and name not in self.get_skip_vars()
+        ):
+            raise AttributeError(
```
Lendemor, line 1246: I think we should raise a SetUndefinedStateVar error instead.

Lendemor (changes requested):

Lendemor (approved):

reflex/state.py
```diff
@@ -1277,6 +1281,17 @@ def __setattr__(self, name: str, value: Any):
             self._mark_dirty()
             return
 
+        if (
+            not name.startswith("__")
+            and not name.startswith(f"_{type(self).__name__}__")
+            and name not in self.vars
```
picklelo, line 1287: nit: we may want to put this check first since it's the most common. Then if they are doing things correctly, it will avoid the above two checks

picklelo (approved):

PR #4010, opened 2024-09-26
Handle bool cast for optional NumberVar

masenf:
If the _var_type is optional, then also check that the value is not None

### DIFF
--- a/reflex/vars/number.py
+++ b/reflex/vars/number.py
@@ -526,7 +526,7 @@ class NumberVar(Var[NUMBER_T]):
             The boolean value of the number.
         """
         if is_optional(self._var_type):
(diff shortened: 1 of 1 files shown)


reflex/vars/number.py
```diff
@@ -524,6 +525,8 @@ def bool(self):
         Returns:
             The boolean value of the number.
         """
+        if is_optional(self._var_type):
+            return boolify((self != 0) & (self != None))  # noqa: E711
```
adhami3310, line 529:
```suggestion
            return boolify((self != None) & (self != 0))  # noqa: E711
```

adhami3310 (changes requested):

adhami3310 (approved):

PR #4014, opened 2024-09-27
improve lifespan typecheck and debug

### LINKED ISSUE #4009: Lifespan task without @asynccontextmanager are failing silently.
Repro example:
```python
def baz_lifespan():  
    print("baz_lifespan")
    yield
    print("baz_lifespan done")

app.register_lifespan_task(task=baz_lifespan) # should throw error
```

Lendemor:
closes #4009

### DIFF
--- a/reflex/app_mixins/lifespan.py
+++ b/reflex/app_mixins/lifespan.py
@@ -38,7 +38,7 @@ class LifespanMixin(AppMixin):
                         _t = task()
                         if isinstance(_t, contextlib._AsyncGeneratorContextManager):
                             await stack.enter_async_context(_t)
-                            console.debug(run_msg.format(type="async context manager"))
+                            console.debug(run_msg.format(type="asynccontextmanager"))
                         elif isinstance(_t, Coroutine):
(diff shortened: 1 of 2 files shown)


reflex/app_mixins/lifespan.py
```diff
@@ -26,6 +27,7 @@ async def _run_lifespan_tasks(self, app: FastAPI):
         try:
             async with contextlib.AsyncExitStack() as stack:
                 for task in self.lifespan_tasks:
+                    console.debug(f"Running lifespan task: {task}")
```
masenf, line 30: for some reason i dont see this one showing up on my end... maybe the loglevel isn't propagated into the subprocess?

Lendemor: Seems so. Will investigate on that problem.

Lendemor: follow up: bug as been present since forever, and is unrelated to the changes in this PR, so will need a PR of its own to resolve the issue.

masenf (approved):

PR #4018, opened 2024-09-27
implement _evaluate in state

adhami3310:
Introduces experimental inline-way to define State components.

```py
"""This is a mockup example to demonstrate the usage of server_side function."""

import random

import reflex as rx


class State(rx.State):
    large_number: int = 100
    header_name: str = "Factorial Calculator"

    def randomize_number(self):
        self.large_number = random.randint(1, 900)


def factorial(n: int) -> int:
    """Return the factorial of n."""
    if n == 0:
        return 1
    return n * factorial(n - 1)


def is_prime(n: int) -> bool:
    """Return True if n is prime, False otherwise."""
    if n < 2:
        return False
    return all(n % i != 0 for i in range(2, n))


app = rx.App()


@app.add_page
def index():
    return rx.vstack(
        rx.text(
            f"factorial of {State.large_number} is ",
            State._evaluate(
                lambda state: str(factorial(state.large_number)),
            ),
            width="100%",
        ),
        rx.text(
            f"url-name of {State.header_name} is ",
            State._evaluate(lambda state: state.header_name.lower().replace(" ", "-")),
        ),
        rx.text(
            f"{State.large_number} is ",
            State._evaluate(
                lambda state: (
                    "prime" if is_prime(state.large_number) else "not prime"
                ),
            ),
        ),
        rx.button(
            "Randomize",
            on_click=lambda: State.randomize_number,
        ),
    )
```

### DIFF
--- a/reflex/vars/base.py
+++ b/reflex/vars/base.py
@@ -1559,8 +1559,9 @@ class ComputedVar(Var[RETURN_TYPE]):
         Raises:
             TypeError: If the computed var dependencies are not Var instances or var names.
         """
-        hints = get_type_hints(fget)
-        hint = hints.get("return", Any)
+        hint = kwargs.pop("return_type", None) or get_type_hints(fget).get(
+            "return", Any
+        )
 
         kwargs["_js_expr"] = kwargs.pop("_js_expr", fget.__name__)
         kwargs["_var_type"] = kwargs.pop("_var_type", hint)
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -24,7 +24,6 @@ from typing import (
     Dict,
     List,
     Optional,
-    Self,
     Sequence,
     Set,
     Tuple,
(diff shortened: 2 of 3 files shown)


reflex/vars/base.py
```diff
@@ -1559,8 +1559,9 @@ def __init__(
         Raises:
             TypeError: If the computed var dependencies are not Var instances or var names.
         """
-        hints = get_type_hints(fget)
-        hint = hints.get("return", Any)
+        hint = kwargs.pop("return_type", None) or get_type_hints(fget).get(
```
masenf, line 1562: should we put `return_type` in the actual `__init__` signature? seems a little hard to find/undocumented this way

adhami3310: it might be intentionally that way (using might + intentionally in something i did is a bit questionable, but alas), i did it because users should almost always pass the return type through the function return signature, sometimes in reflex code we don't have that option so it's more of an internal thing.

masenf (approved):

PR #4019, opened 2024-09-27
remove all runtime asserts

Lendemor:


reflex/testing.py
```diff
@@ -787,13 +791,13 @@ def poll_for_clients(self, timeout: TimeoutType = None) -> dict[str, BaseState]:
...
+            ValueError: when the state_manager is not a memory state manager
         """
         if self.app_instance is None:
             raise RuntimeError("App is not running.")
         state_manager = self.app_instance.state_manager
-        assert isinstance(
-            state_manager, (StateManagerMemory, StateManagerDisk)
-        ), "Only works with memory state manager"
+        if not isinstance(state_manager, (StateManagerMemory, StateManagerDisk)):
+            raise ValueError("Only works with memory state manager")
```
masenf, line 800:
```suggestion
            raise ValueError("Only works with memory or disk state manager")
```

masenf (dismissed):
normally id say we should raise trackable exceptions here, but since these were originally `assert`, maybe they're not intended to ever be hit?

Lendemor:
> normally id say we should raise trackable exceptions here, but since these were originally `assert`, maybe they're not intended to ever be hit?

I think so yes.
That said, if there is any that should have more specific name, I don't mind changing them, I was just not very inspired for renaming since there is quite a few of them.

masenf (approved):

PR #4041, opened 2024-10-02
let users pick state manager mode

Lendemor:
In `rx.Config` instance, you can now pass a value between `"disk", "memory" or "redis"` to `state_manager_mode`


adhami3310:
I prefer passing the StateManager subclass itself rather than an enum because that allows one to pass any subclass, including custom defined ones.

Lendemor:
> I prefer passing the StateManager subclass itself rather than an enum because that allows one to pass any subclass, including custom defined ones.

I would be fine with this, but how would it work for setting it via env var?

adhami3310:
> how would it work for setting it via env var

the developer can special case on env var they give (like if is_prod use this else use sth else), i don't think we should support direct env var to set this

Lendemor:
> > how would it work for setting it via env var
> 
> the developer can special case on env var they give (like if is_prod use this else use sth else), i don't think we should support direct env var to set this

If we define it in Config it'll automatically accept the related env var, I'm not sure we have a mechanism to exclude some config field.

adhami3310:
> 'll automatically accept the related env var

wild, didn't even know we had that feature

> not sure we have a mechanism to exclude some config field

how would it even accept an env field for a non-trivially-serializable object like this? would it skip it?

Lendemor:
As a note: 
Using StateManager classes directly in config doesn't work at the moment due to some circular dependencies.

picklelo (approved):
awesome

PR #4042, opened 2024-10-02
bundle chakra in window for CSR

adhami3310:


adhami3310:
Question for reviewers, this solution isn't very scalable, it would be better if there's a way to register what libraries are going to be in `window`

Lendemor:
> Question for reviewers, this solution isn't very scalable, it would be better if there's a way to register what libraries are going to be in `window`

How about providing an app method `register_var_component` to allow arbitrary CSR components rather than just the predefined ones?

adhami3310:
> providing an app method register_var_component

well, it would need to happen at the library level (because we don't expect people to register every single component they are going to use), but that's a bit abstracted (we don't really mention radix much, so how would you figure out you have to write `@radix-ui/themes`

Lendemor:
> > providing an app method register_var_component
> 
> well, it would need to happen at the library level (because we don't expect people to register every single component they are going to use), but that's a bit abstracted (we don't really mention radix much, so how would you figure out you have to write `@radix-ui/themes`

Well the core components could be "automatically" registered, while third party comp would have to be registered manually if they want them for dynamic rendering.

adhami3310:
Currently one can do:

```py
bundle_library(rc.button())
```

and that would bundle reflex chakra

masenf (approved):
so as i read it, it's still up to the developer to call `rx.dynamic.bundle_library` at module level (i.e. not dynamically), to have their libs included in the `window`?

adhami3310:
> s still up to the developer to call rx.dynamic.bundle_library (...) to have their libs included in the window

yes, otherwise reflex wouldn't bundle the library in window

PR #4043, opened 2024-10-02
set loglevel to info with hosting cli

adhami3310:


reflex/reflex.py
```diff
@@ -563,6 +567,10 @@ def deploy(
         _init(name=config.app_name, loglevel=loglevel)
     prerequisites.check_latest_package_version(constants.ReflexHostingCLI.MODULE_NAME)
 
+    subprocesses_loglevel = (
```
Lendemor, line 570: If we're gonna repeat this logic in four or more places, maybe make it a function ?

Lendemor (changes requested):
Everything else looks fine.

Lendemor (approved):

PR #4044, opened 2024-10-02
move router dataclasses in their own file

Lendemor:


adhami3310:
Can we pick a better name than istate :sob:

Lendemor:
> Can we pick a better name than istate 😭

Sure, if you've got a better suggestion I can change it 👍

adhami3310 (approved):
i can't think of something better than istate (should prob be state once the reorganizing is done)

PR #4045, opened 2024-10-03
Track the last reflex run time

ElijahAhianyo:
Just like title say, keep track of the time `reflex run` is ran

### DIFF
--- a/reflex/utils/prerequisites.py
+++ b/reflex/utils/prerequisites.py
@@ -132,6 +132,21 @@ def get_or_set_last_reflex_version_check_datetime():
     return last_version_check_datetime
 
 
+def set_last_reflex_run_time():
+    """Set the last Reflex run time.
+
+    Returns:
+        The last time the reflex app was started.
+    """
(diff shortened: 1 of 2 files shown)


reflex/utils/prerequisites.py
```diff
@@ -132,6 +132,21 @@ def get_or_set_last_reflex_version_check_datetime():
...
+    """Set the last Reflex run time.
+
+    Returns:
+        The last time the reflex app was started.
+    """
+    reflex_json_file = get_web_dir() / constants.Reflex.JSON
+    if not reflex_json_file.exists():
+        return
+
+    data = json.loads(reflex_json_file.read_text())
```
Lendemor, line 145: You don't need to read + update if you are using `update_json_file`

Either just pass the additional data to `update_json_file`, or just do `fp.write` after read + update.

Lendemor (changes requested):

Lendemor (approved):

PR #4053, opened 2024-10-03
remove var operation error

adhami3310:
unsure where to add it back, but its currently a negative more than a positive


Lendemor (approved):

masenf (approved):
hopefully our Var exceptions are good enough

PR #4055, opened 2024-10-04
[ENG-2287] Avoid fetching same state from redis multiple times

### LINKED ISSUE #2950: [REF-2287] Redis fetches the same state multiple times
It's possible for a single state to be fetched from redis multiple times.

```
import reflex as rx


class State(rx.State):
    pass


class S2(State):
    @rx.var
    def foo(self):
        return "foo"


class S3(S2):
    def handler(self):
        print("foo")


def index():
    return rx.button("Click", on_click=S3.handler)


app = rx.App()
app.add_page(index)
```

This will load S2 from redis twice

Work branch: `masenf/optimize-redis-get-state`

<sub>From [SyncLinear.com](https://synclinear.com) | [REF-2287](https://linear.app/reflex-dev/issue/REF-2287/redis-fetches-the-same-state-multiple-times)</sub>

masenf:
Fix #2950 

When fetching a state from redis which has computed vars, it might get fetched twice when populating substates. To avoid this, pass a list of already-fetched substates to `get_state` and apply these before performing `populate_substates`. Then `populate_substates` skips fetching substates which are already in the tree.

Consolidate logic between the "deserialize instance" and "create instance" paths -- previously we needed the other path, because we had to instantiate a new state class with a valid `parent_state`, but with the merge of https://github.com/reflex-dev/reflex/commit/319ff17f4c4ea42a2163d0e0f3408d7400571d61, the parent_state pointer can be added later, so the two paths were equivalent.

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -2985,45 +2985,29 @@ class StateManagerRedis(StateManager):
         if redis_state is not None:
             # Deserialize the substate.
             state = BaseState._deserialize(data=redis_state)
-
-            # Populate parent state if missing and requested.
-            if parent_state is None:
-                parent_state = await self._get_parent_state(token, state)
-            # Set up Bidirectional linkage between this state and its parent.
-            if parent_state is not None:
-                parent_state.substates[state.get_name()] = state
-                state.parent_state = parent_state
-            # Avoid fetching substates multiple times.
-            if cached_substates:
-                for substate in cached_substates:
-                    state.substates[substate.get_name()] = substate
-                    if substate.parent_state is None:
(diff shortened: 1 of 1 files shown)


adhami3310 (dismissed):

reflex/state.py
```diff
@@ -2974,39 +2985,29 @@ async def get_state(
         if redis_state is not None:
             # Deserialize the substate.
             state = BaseState._deserialize(data=redis_state)
```
masenf, line 2987: need to catch schema mismatch error

Lendemor (approved):

PR #4073, opened 2024-10-05
Include step attribute in input

Joodith:
Changed the base class of Text field from rx.el.Div to rx.el.Input and added the step attribute.Works in the case of  value and on_change events being given(i.e DebounceInput ).


reflex/components/radix/themes/components/text_field.py
```diff
@@ -68,6 +68,8 @@ class TextFieldRoot(elements.Div, RadixThemesComponent):
     # Specifies the type of input
     type: Var[str]
 
+    step: Var[int]
```
masenf, line 71: If this is being included for documentation purposes, please add a preceeding comment, which will be used to generate the docs for this component.

masenf (commented):
looking at the radix source code, this _is_ the correct inheritance structure, because any props that are not explicitly recognized by radix are passed directly to the wrapped `<input>` component.

masenf:
retrigger CI hopefully

masenf (approved):

PR #4085, opened 2024-10-07
fail safely when pickling

adhami3310:


masenf (approved):
looks good, but we can't just ignore redis pickling errors long term, that's going to create problems for people

adhami3310:
> but we can't just ignore redis pickling errors

I think it makes sense here because it's going to show up during development, so people wouldn't be surprised when it doesn't work on redis.

PR #4089, opened 2024-10-08
test_dynamic_routes: log on_loads and poll for 60 seconds on order

masenf:
Assert on `list(...order)` so the error message prints actual value instead of MutableProxy's repr.

Not sure if this fixes it...


picklelo (approved):

masenf:
i've ran it 3 times now with no flakes, i'll keep re-running and see if any shake out.

i never did get a repro on my local (my machine is too fast)

PR #4136, opened 2024-10-08
Remove wrong event handlers

adhami3310:
also deprecate default annotations for strings (which means there will be good amount of fixes to event handlers in the library soon)


reflex/components/radix/themes/components/text_area.py
```diff
@@ -83,19 +83,19 @@ class TextArea(RadixThemesComponent, elements.Textarea):
     wrap: Var[str]
 
     # Fired when the value of the textarea changes.
-    on_change: EventHandler[lambda e0: [e0.target.value]]
+    on_change: EventHandler[input_event]
```
masenf, line 86: if this class inherits from `elements.Textarea`, do we need these definitions here? are they for autodoc purposes?

masenf: Soon TextField will inherit from `elements.Input`, so similar question there.

masenf (approved):
Probably followup work, but the drawer content component accepts `on_pointer_down_outside` and `on_open_auto_focus` events.

Looks like the root also exposes `on_close`.

The docs for vaul are a bit out of date with the code; we probably should go through and clean these up.

adhami3310:
> the drawer content component accepts on_pointer_down_outside and on_open_auto_focus events

it technically does, but also it's not documented in vaul, so we kinda agreed to leave it out

reflex/components/radix/themes/components/text_area.py
adhami3310: removed unnecessary ones! should work fine as PYI file remained the same

masenf (approved):

PR #4139, opened 2024-10-09
only run macOS job on merge

Lendemor:
They are slow in being picked up by github runner and should be very similar to the ubuntu one.


masenf (approved):
kinda sucks we have to duplicate the job definition, but lets take it for now; we can always improve that later.

Lendemor:
> kinda sucks we have to duplicate the job definition, but lets take it for now; we can always improve that later.

Yep, I first tried with adding some conditions and not duplicating the jobs, but it still created the job so we still got the delay in runner pickup.

PR #4084, opened 2024-10-07
Fix 'enter_key_submit=True' on 'rx.text_area' by disabling debounce if set

TimChild:
### All Submissions:

- [x ] Have you followed the guidelines stated in [CONTRIBUTING.md](https://github.com/reflex-dev/reflex/blob/main/CONTRIBUTING.md) file?
- [x ] Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/reflex-dev/reflex/pulls ) for the desired changed?

### Type of change

Please delete options that are not relevant.

- [x ] Bug fix (non-breaking change which fixes an issue)

### New Feature Submission:

- [x ] Does your submission pass the tests? 
- [x ] Have you linted your code locally prior to submission?

### Changes To Core Features:

- [x ] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your core changes, as applicable? -- no
- [x ] Have you successfully ran tests with your changes locally?

### **After** these steps, you're ready to open a pull request.

    b. Describe your changes.
    
   Add additional constraint on whether to wrap rx.text_area in DebounceInput. Don't wrap if `enter_key_submit` is True.
   Without this, the enter_key_submit behaviour is ignored.
   
   Closes #4080
   
   Note: If there is a way to do this without having to remove the DebounceInput, that would be even better.

PR #4142, opened 2024-10-09
Fix 'enter_key_submit=True' on 'rx.text_area' by carrying custom_code on debounce

### LINKED ISSUE #4080: `enter_key_submit` doesn't work for `rx.text_area` if `value` is set (works for `rx.el.textarea`).
**Describe the bug**
Trying to set `enter_key_submit` on the `rx.text_area` has no effect **if** `value` is set. It works as expected on the `rx.el.textarea`

**To Reproduce**
```
app = rx.App(state=rx.State)


class State(rx.State):
    submitted: str = ""

    area_1: str = ""
    area_2: str = ""

    def handle_submit(self, form_data: dict):
        self.submitted = str(form_data)


def layout() -> rx.Component:
    return rx.vstack(
        rx.form(
            rx.text_area(
                enter_key_submit=True,
                name="works_without_value",
                placeholder="works with no value set",
            ),
            rx.text_area(
                value=State.area_1,
                on_change=State.set_area_1,
                enter_key_submit=True,
                name="not_working",
                placeholder="doesn't work",
            ),
            rx.el.textarea(
                value=State.area_2,
                on_change=State.set_area_2,
                enter_key_submit=True,
                name="works",
                placeholder="works",
            ),
        ),
        rx.text(State.submitted)
    )


app.add_page(layout, route="/")
```

**Expected behavior**
Expect same behaviour in all three cases

 - Python Version: 3.12
 - Reflex Version: 0.6.1
 - OS: WSL2
 - Browser (Optional): Chrome
 
**Additional context**
Not obvious to me why the behaviour changes when `value` is set only on the radix themed text_area and not the regular one.

masenf:
Alternative to #4084 

* Allow `debounce` to carry any custom code defined in the child component
* Move special behavior handling to `create` classmethod to allow carrying of added props when wrapped in debounce (which does not call `_render` on the child component).

I think this is preferable to the original patch, because it still allows the debounce to work, avoiding typing jank, and it also works with `auto_height` behavior.

Fix #4080

### DIFF
--- a/reflex/components/core/debounce.py
+++ b/reflex/components/core/debounce.py
@@ -118,7 +118,7 @@ class DebounceInput(Component):
                 _var_type=Type[Component],
                 _var_data=VarData(
                     imports=child._get_imports(),
-                    hooks=child._get_hooks_internal(),
+                    hooks=child._get_all_hooks(),
                 ),
             ),
         )
@@ -128,6 +128,10 @@ class DebounceInput(Component):
         component.event_triggers.update(child.event_triggers)
         component.children = child.children
         component._rename_props = child._rename_props
+        outer_get_all_custom_code = component._get_all_custom_code
+        component._get_all_custom_code = lambda: outer_get_all_custom_code().union(
+            child._get_all_custom_code()
+        )
         return component
 
     def _render(self):
--- a/reflex/components/el/elements/forms.py
+++ b/reflex/components/el/elements/forms.py
@@ -600,6 +600,42 @@ class Textarea(BaseHTML):
     # Fired when a key is released
     on_key_up: EventHandler[lambda e0: [e0.key]]
 
+    @classmethod
+    def create(cls, *children, **props):
+        """Create a textarea component.
+
+        Args:
+            *children: The children of the textarea.
+            **props: The properties of the textarea.
+
+        Returns:
+            The textarea component.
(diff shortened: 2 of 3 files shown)


TimChild:
@masenf, Works great in the test cases I was using for text areas! 
Also, definitely better to take this approach. I was doing more testing this morning and realising how janky things get once it's not running locally!

On a related note. How hard would it be to make the `DebounceInput` work for the `rx.slider` as well?
For the slider, `rx.set_value` doesn't work, so I think there is currently no way to be able to use and set slider values without having a slow `value/on_change` combo.

masenf:
> How hard would it be to make the DebounceInput work for the rx.slider as well?
For the slider, rx.set_value doesn't work, so I think there is currently no way to be able to use and set slider values without having a slow value/on_change combo.

Can you make a new github discussion for this question and i'll answer it over there with a code example

masenf:
@TimChild sorry i forgot, we're moving long form support like that over to our new forum

https://forum.reflex.dev/

This will help people searching on google find the answers 😁

TimChild:
@masenf Created a new topic there https://forum.reflex.dev/t/how-to-control-slider-without-lag-debounced-inputs/443

picklelo (approved):

PR #4145, opened 2024-10-10
add type hinting to events

adhami3310:


adhami3310:
there's still more `EventHandler[lambda ...]`, more will be coming soon

reflex/utils/pyi_generator.py
```diff
@@ -427,18 +427,59 @@ def _generate_component_create_functiondef(
     all_props = [arg[0].arg for arg in prop_kwargs]
     kwargs.extend(prop_kwargs)
 
+    def figure_out_return_type(annotation: Any):
+        print(annotation)
```
masenf, line 431: stray print
```suggestion
```

reflex/event.py
```diff
@@ -1020,13 +1055,15 @@ def parse_args_spec(arg_spec: ArgsSpec):
...
     annotations = get_type_hints(arg_spec)
 
-    return arg_spec(
-        *[
-            Var(f"_{l_arg}").to(
-                unwrap_var_annotation(resolve_annotation(annotations, l_arg))
-            )
-            for l_arg in spec.args
-        ]
+    return list(
```
masenf, line 1058: why does this return a `list` now? this is the only change i don't really understand.

adhami3310: because argspec is now typed as "sequence"

masenf (approved):

PR #4153, opened 2024-10-10
unionize base var fields types

adhami3310:
```py
class Foo(rx.Base):
    name: str = "foo"


class Bar(rx.Base):
    field: str = "bar"


class State(rx.State):
    @rx.var
    def something(self) -> Foo | Bar:
        return Bar()


print(get_attribute_access_type(Foo | Bar, "name"))

print(State.something.name._var_type)
```

### DIFF
--- a/reflex/utils/types.py
+++ b/reflex/utils/types.py
@@ -195,8 +195,6 @@ def unionize(*args: GenericType) -> Type:
         return Any
     if len(args) == 1:
         return args[0]
-    if sys.version_info >= (3, 11):
-        return Union[*args]  # type: ignore
     # We are bisecting the args list here to avoid hitting the recursion limit
     # In Python versions >= 3.11, we can simply do `return Union[*args]`
     midpoint = len(args) // 2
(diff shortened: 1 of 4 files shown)


reflex/utils/types.py
```diff
@@ -182,6 +182,26 @@ def is_generic_alias(cls: GenericType) -> bool:
...
+
+    Returns:
+        The unionized types.
+    """
+    if not args:
+        return Any
+    if len(args) == 1:
+        return args[0]
+    # We are bisecting the args list here to avoid hitting the recursion limit
+    # In Python versions >= 3.11, we can simply do `return Union[*args]`
```
benedikt-bartscher, line 199: Can we add a case for >=python3.11 here to avoid unneeded computation?

adhami3310: turns out it fails with syntax error, meaning we can't have it in the code till 3.11 :L

masenf (approved):

masenf (approved):

PR #4166, opened 2024-10-12
[ENG-759] [ENG-1104] patch `json.dumps` to handle `__wrapped__` objects

masenf:
Unwrap proxied objects if the default serializer doesn't work.

Is this a good idea? That remains to be seen... but it does work.

```python
import json

import reflex as rx


class State(rx.State):
    value: list[int] = [50]

    def foo(self):
        print(json.dumps(self.value))


def index() -> rx.Component:
    return rx.button("Foo", on_click=State.foo)


app = rx.App()
app.add_page(index)
```


benedikt-bartscher:
> Is this a good idea? That remains to be seen... but it does work.

Maybe we should rename it to `__reflex_wrapped__` if we decide to patch python internals

picklelo (approved):

picklelo (approved):

PR #4169, opened 2024-10-13
fix: async default setters break setvar

benedikt-bartscher:


abulvenz:
Yes, that's what bit us.

benedikt-bartscher:
We could also try to implement async setvar instead. Or disallow overwriting default setter names either completely or with async functions

masenf:
> We could also try to implement async setvar instead.

Could it be this simple?

```diff
diff --git a/reflex/state.py b/reflex/state.py
index 51fd740df..65d95c53f 100644
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -191,7 +191,7 @@ class EventHandlerSetVar(EventHandler):
         )
         object.__setattr__(self, "state_cls", state_cls)
 
-    def setvar(self, var_name: str, value: Any):
+    async def setvar(self, var_name: str, value: Any):
         """Set the state variable to the value of the event.
 
         Note: `self` here will be an instance of the state, not EventHandlerSetVar.
@@ -200,7 +200,10 @@ class EventHandlerSetVar(EventHandler):
             var_name: The name of the variable to set.
             value: The value to set the variable to.
         """
-        getattr(self, constants.SETTER_PREFIX + var_name)(value)
+        res = getattr(self, constants.SETTER_PREFIX + var_name)(value)
+        if hasattr(res, "__await__"):
+            return await res
+        return res
 
     def __call__(self, *args: Any) -> EventSpec:
         """Performs pre-checks and munging on the provided args that will become an EventSpec.
@@ -214,7 +217,6 @@ class EventHandlerSetVar(EventHandler):
         Raises:
             AttributeError: If the given Var name does not exist on the state.
             EventHandlerValueError: If the given Var name is not a str
-            NotImplementedError: If the setter for the given Var is async
         """
         from reflex.utils.exceptions import EventHandlerValueError
 
@@ -232,11 +234,6 @@ class EventHandlerSetVar(EventHandler):
                     f"Variable `{args[0]}` cannot be set on `{self.state_cls.get_full_name()}`"
                 )
 
-            if asyncio.iscoroutinefunction(handler.fn):
-                raise NotImplementedError(
-                    f"Setter for {args[0]} is async, which is not supported."
-                )
-
         return super().__call__(*args)
 
```

benedikt-bartscher:
Thanks, @masenf, yes it could be that simple, maybe with `inspect.isawaitable`?
I will give it a try

masenf:
> this will change the setvar api to async, which seems like a bad idea

I'm not sure i see the problem with this API being async... it's really only used to create frontend events that set vars.

1. if i'm already running in an event handler, then i would just set the vars directly
2. if i pass `MyState.setvar(x)` as an event trigger, then it works just the same whether its async or not

masenf (approved):
taking this for 0.6.4, we can figure out async setters later

PR #4175, opened 2024-10-14
make var system expandable

adhami3310:


reflex/vars/base.py
```diff
@@ -1017,9 +1143,131 @@ def json(self) -> str:
...
+            name: The name of the attribute.
+
+        Returns:
+            The attribute of the var.
+        """
+        from .object import ObjectVar
+
+        if isinstance(self, ObjectVar) and name != "_js_expr":
+            return ObjectVar.__getattr__(self, name)
+        return getattr(object.__getattribute__(self, "_original"), name)
```
masenf, line 1162: does this work here?

```suggestion
        return getattr(self._original, name)
```

i'm seeing this all over, what am i missing?

reflex/vars/base.py
```diff
@@ -1017,9 +1143,131 @@ def json(self) -> str:
...
+        possible_bases = [
+            base
+            for base in bases_normalized
+            if issubclass(base, Var) and base != LiteralVar
+        ]
+
+        if not issubclass(cls, LiteralVar):
+            raise TypeError(
+                f"LiteralVar subclass {cls} must be a subclass of LiteralVar."
+            )
```
masenf, line 1241: under what circumstances do we hit this?

masenf (commented):
this is cool. cleans up a lot of more complex code that i think will improve maintainability

reflex/vars/base.py
adhami3310: I don't quite remember, but I had some recursion errors without it that got complex and that was once simple way of solving it. I will try without the hack to see if it works now.

reflex/vars/base.py
adhami3310: it was a remainder of being inside a different function, not needed anymore

reflex/vars/sequence.py
```diff
@@ -1654,3 +1633,119 @@ def array_concat_operation(
...
+            (
+                self.__class__.__name__,
+                self._var_value.color,
+                self._var_value.alpha,
+                self._var_value.shade,
+            )
+        )
+
+    @cached_property_no_lock
+    def _cached_var_name(self) -> str:
```
masenf, line 1692: this function needs more test cases.

The following seems to be a regression in this patch.
```
rx.Var.create(rx.color("blue", rx.Var.create(4), True))
```

Creates code like `Object.prototype.toString(4)`, which evaluates as `'[object Object]'`

-------------------

This snippet is not necessarily a regression, since it's also broken on `main`, but theoretically it should work to have `alpha` specified as a `Var`, right?

```
rx.Var.create(rx.color("blue", 4, rx.Var.create(True)))
```

adhami3310: oops, i thought Object.prototype.toString would be the same as .toString on a value, turns out it's more complicated, i will look into that

masenf (approved):

PR #4176, opened 2024-10-15
Remove demo command

Alek99:


picklelo (approved):
We should remove these from reflex-web docs as well if we mention it there

adhami3310 (approved):

PR #4182, opened 2024-10-16
add type hinting to error boundary

adhami3310:


reflex/state.py
```diff
@@ -2094,14 +2095,16 @@ class State(BaseState):
...
+    @event
+    def handle_frontend_exception(self, stack: str, component_stack: str) -> None:
         """Handle frontend exceptions.
 
         If a frontend exception handler is provided, it will be called.
         Otherwise, the default frontend exception handler will be called.
 
         Args:
             stack: The stack trace of the exception.
+            component_stack: The stack trace of the component where the exception occurred.
```
masenf, line 2107: what does the `component_stack` get used for? it's not being passed to the registered exception handler

masenf (approved):

reflex/state.py
adhami3310: it's not being used here, but it's provided by the react library

PR #4183, opened 2024-10-16
[GTM-345]Define component props in class for doc discoverability

ElijahAhianyo:
Defines/Moves `create` method args where possible in components as props so reflex-web can discover them


Alek99:
Tag me when this pr is ready

reflex/components/datadisplay/code.py
```diff
@@ -411,6 +411,12 @@ class CodeBlock(Component):
     # Props passed down to the code tag.
     code_tag_props: Var[Dict[str, str]]
 
+    # Whether a copy button should appear.
```
picklelo, line 414: In this approach, do these extra props end up getting compiled in? If so, I feel we should have a way to distinguish between these reflex-defined props vs the ones that get compiled in at the end.

reflex/components/datadisplay/code.py
```diff
@@ -411,6 +411,12 @@ class CodeBlock(Component):
...
+    can_copy: Var[Optional[bool]] = Var.create(False)
```
masenf, line 415: The other problem here is now these are annotated and treated as `Var`... but in many cases, the code in the `create` is using python `if` statements and potential other operations that are not valid on `Var` types.

I'd be more okay defining these as regular model fields (i.e. without the `Var`), but then i don't think they'd be automatically picked up by the pyi generator and the doc generation

masenf (commented):
i dont think we can take this as is, because some of these `create` kwargs that are replaced by props are not valid as `Var`

reflex/components/datadisplay/code.py
ElijahAhianyo: I think for these cases we can add them to the excluded props so they don't get compiled:

```python
class Code(rx.Component):
  ...
  
  def _exclude_prop(self) -> list[str]:
      return ["code_tag_props"]

```

reflex/components/datadisplay/code.py
ElijahAhianyo: The regular model fields get picked up by the pyi generator and reflex-web. Some kwargs (that were moved) accept literals or vars. Maybe in this case it might be worth specifying the typing of these model fields to accept both literals and vars

reflex/components/radix/primitives/accordion.py
```diff
@@ -193,6 +193,11 @@ class AccordionItem(AccordionComponent):
     # When true, prevents the user from interacting with the item.
     disabled: Var[bool]
 
+    # The header of the accordion item.
+    header: Optional[Union[Component, str, Var[Union[Component, str]]]] = None
+    # The content of the accordion item.
+    content: Optional[Union[Component, str, Var[Union[Component, str]]]] = None
```
masenf, line 199:
```suggestion
    # The header of the accordion item.
    header: Var[Union[Component, str]] = None
    # The content of the accordion item.
    content: Var[Union[Component, str]] = None
```

I think this is okay for these. If a `Var` is acceptable, then we can rely on the pyi generator to splat out the inner types

reflex/components/radix/themes/color_mode.py
```diff
@@ -96,26 +96,31 @@ def _set_static_default(props, position, prop, default):
...
         Args:
-            position: The position of the icon button. Follow document flow if None.
-            allow_system: Allow picking the "system" value for the color mode.
             **props: The props to pass to the component.
 
         Returns:
             The button component.
         """
+        position = props.pop("position", None)
+        allow_system = props.pop("allow_system", None)
```
masenf, line 119:
```suggestion
        allow_system = props.pop("allow_system", False)
```

to match the previous signature

reflex/components/radix/themes/layout/list.py
```diff
@@ -42,29 +42,32 @@ class BaseList(Component):
...
 
     # The style of the list. Default to "none".
-    list_style_type: Var[
+    list_style_type: Optional[
         Union[LiteralListStyleTypeUnordered, LiteralListStyleTypeOrdered]
-    ]
+    ] = "none"
+
+    # A list of items to add to the list.
+    items: Optional[Union[Iterable, Var[Iterable]]] = None
```
masenf, line 50:
```suggestion
    items: Var[Iterable]
```

this one can just be `Var`, since it's handled in a ForEach in that case

reflex/components/radix/themes/layout/list.py
```diff
@@ -42,29 +42,32 @@ class BaseList(Component):
     tag = "ul"
 
     # The style of the list. Default to "none".
-    list_style_type: Var[
+    list_style_type: Optional[
```
masenf, line 45:
```suggestion
    list_style_type: Var[
```

This one can also be a `Var`, since the value just gets shoved into the style dict

reflex/components/radix/themes/layout/list.py
```diff
@@ -87,32 +90,37 @@ def add_style(self) -> dict[str, Any] | None:
...
+        return ["items", "list_style_type"]
+
 
 class UnorderedList(BaseList, Ul):
     """Display an unordered list."""
 
     tag = "ul"
 
+    # The style of the list.
+    list_style_type: LiteralListStyleTypeUnordered = "disc"
```
masenf, line 103: this can be removed and inherited from the parent; the default is already being grabbed in the `create` function

(same with below for `OrderedList`)

reflex/components/radix/primitives/accordion.py
```diff
@@ -194,9 +194,9 @@ class AccordionItem(AccordionComponent):
     disabled: Var[bool]
 
     # The header of the accordion item.
-    header: Optional[Union[Component, str, Var[Union[Component, str]]]] = None
+    header: Var[Union[Component, str]] = Var.create(None)
```
masenf, line 197: i think the default is `None`
```suggestion
    header: Var[Union[Component, str]]
```

masenf (approved):

masenf (dismissed):

masenf (approved):

PR #4189, opened 2024-10-16
use larger or equal for node version check

adhami3310:


adhami3310:
on another note, the place that checks this just assumes it does >=, and prints a confusing error message like "node version X must be at least X"

masenf (approved):
nice clean up of the logic

PR #4192, opened 2024-10-16
move all environment variables to the same place

### LINKED ISSUE #4172: Unify env vars
https://github.com/reflex-dev/reflex/pull/4168#discussion_r1799775136

> For some env vars, we're checking against ["true", "1", "yes"] rather than just "true".
> I think it'd be better to have an uniform method doing the check for boolean env var rather than duplicating logic everywhere.

by @Lendemor

Should we keep it simple, only allow "true"/"false" and raise a deprecation warning (and later an error) if something different is set?
Also let's move all env var constants to a SimpleNamespace class?

adhami3310:
Fixes #4172

### DIFF
--- a/reflex/config.py
+++ b/reflex/config.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import dataclasses
 import importlib
 import os
 import sys
@@ -10,6 +11,7 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional, Set, Union
 
 from reflex.utils.exceptions import ConfigError
+from reflex.utils.types import value_inside_optional
 
 try:
     import pydantic.v1 as pydantic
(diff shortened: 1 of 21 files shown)


Lendemor (commented):
Code is much cleaner with this PR, should also allow us to re-introduce envfile usage in a follow up PR 👍

Lendemor (approved):

reflex/config.py
```diff
@@ -131,6 +135,198 @@ def get_url(self) -> str:
...
+    Raises:
+        EnvironmentVarValueError: If the path does not exist.
+    """
+    path = Path(value)
+    if not path.exists():
+        raise EnvironmentVarValueError(f"Path does not exist: {path}")
+    return path
+
+
+def interpret_env_var_value(value: str, field: dataclasses.Field) -> Any:
```
benedikt-bartscher, line 218: this probably needs handling for Union as well, see https://github.com/reflex-dev/reflex/pull/4205

benedikt-bartscher (commented):
can we add this testcase? https://github.com/reflex-dev/reflex/pull/4205/files#diff-32c634afa2617be71ee9450a74c25db8394f195be5be42e3a6ab2dd5387a37edR32

AkshayAwate (approved):

reflex/config.py
adhami3310: (at least not more than just optional)

benedikt-bartscher: > we're not using union for our field types though

What about `bun_path`? https://github.com/reflex-dev/reflex/blob/main/reflex/config.py#L194

adhami3310: `bun_path` goes through a different code path, and it doesn't use `interpret_env_var_value` (maybe it should?), the issue here is we accept `str` but we're going to convert it to `Path` anyways, so it acts like a nice API but not true "type", i propose changing it to just `Path` instead of `Union` (which does kick the bucket of figuring out how to do true unions to a later day)

benedikt-bartscher: Oh yes, you are right. I haven't realized that your PR did not touch `update_from_env` at all. Maybe i should just base https://github.com/reflex-dev/reflex/pull/4205/files on your PR and utilize your `interpret_env_var_value` to streamline this type parsing logic.

> i propose changing it to just Path instead of Union

sounds good!

> (which does kick the bucket of figuring out how to do true unions to a later day)

let's raise an exception for unhandled unions to prevent someone introducing a bug by adding a union to env/config

adhami3310: that sounds alright by me, we can merge this for now

PR #4193, opened 2024-10-17
[ENG-3954] Treat ArrayVar.foreach index as int

masenf:
```python
import reflex as rx

l = rx.Var.create(["a", "b"])
l_foo = l.foreach(lambda item: item + "_foo")
```


adhami3310 (approved):
looks alr o7

PR #4194, opened 2024-10-17
computed var default to cache=True

Lendemor:
Set the default value of `cache` to `True` for `@rx.var`

Lead to better performance "out of the box" since any computed var will be cached, unless explicitly disabled.

Potentially breaking some apps behaviour, so we are considering delaying this PR to `0.7.0`.


adhami3310 (approved):

masenf (commented):
can we actually make the signature `cache: bool = None`

then in `ComputedVar.__init__`, we can check `if cache is None` and throw a deprecation warning, keeping the default as `False` until 0.7.0; then we remove the warning and replace the `None` default with `True`.

If users want to get rid of the warning, then they will need an explicit value for the `cache` parameter.

The main downside with this approach is that `@rx.var` decoration (with no parentheses) will raise the deprecation warning until 0.7.0 is released... maybe that's not ideal

masenf:
@Lendemor if we haven't been warning users that this behavior will change in 0.7.0, i think we should merge a patch for our next release that does as i suggested back in october and print a warning when the user does not provide an explicit boolean for the `cache` parameter. We need to warn users that this change is coming.

masenf (approved):

PR #4195, opened 2024-10-17
implement rx dynamic

adhami3310:
```py
import reflex as rx


class State(rx.State):
    """A simple state that stores the score."""

    score: int = 0

    @rx.event
    def increment(self):
        """Increment the score by 10."""
        self.score += 10


@rx.page("/")
@rx.dynamic
def index(state: State):
    """A simple counter that increments the score when clicked."""
    return rx.vstack(
        rx.text(
            f"Score: {state.score}",
            " which is a high score!" if state.score > 100 else "",
        ),
    )


app = rx.App()
```


reflex/state.py
```diff
@@ -2091,6 +2092,51 @@ class State(BaseState):
...
+    """Create a dynamically generated components from a state class.
+
+    Args:
+        func: The function to generate the component.
+
+    Returns:
+        The dynamically generated component.
+
+    Raises:
+        ValueError: If the function does not have exactly one parameter.
```
Lendemor, line 2108: Maybe use `DynamicViewInvalidSignature` or something here.

Lendemor (dismissed):

Lendemor (approved):

masenf (approved):

PR #4219, opened 2024-10-22
HOS-93: add support for .env file

Kastier1:
### All Submissions:

- [ ] Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/reflex-dev/reflex/pulls ) for the desired changed?

### Type of change

### New Feature Submission:

### Changes To Core Features:

    b. Describe your changes.


masenf (commented):
so in this case, if the `env_file` is specified, then it overrides values defined in the environment already? is there any "automatic" env file? or is it only loaded if the user explicitly sets one?

-----------

remove stray print, merge from `main`, relock dependencies

Kastier1:
There is no automatic env file. If they do no supply one we do nothing. If they do we use those values over what is in their env already.

reflex/config.py
```diff
@@ -436,6 +436,9 @@ class Config:
     # Attributes that were explicitly set by the user.
     _non_default_attributes: Set[str] = pydantic.PrivateAttr(set())
 
+    # file path to env file
```
masenf, line 439: This will get rendered in the generated docs, so might i suggest

```suggestion
    # Path to file containing key-values pairs to override in the environment.
```

masenf (approved):
looks good

masenf (approved):

Kastier1:
Is there any reason in particular to make it optional?

`python-dotenv` is a small package with no other deps 

We are looking into making this optional but was curious as to why you wanted this

benedikt-bartscher:
Just trying to keep the list short in general. You're right, it's a rather small package, but it still needs to be maintained, updated, license reviewed etc. And it could (for some stupid reason) introduce other dependencies in the future.
In general, i prefer to make dependencies optional if they are not tightly coupled with the code.

PR #4220, opened 2024-10-22
move builtins states to their own file

Lendemor:


adhami3310:
can you export the builtins at the `rx.state` level if possible? it would be good if existing code doesn't have to follow new path

Lendemor:
> can you export the builtins at the `rx.state` level if possible? it would be good if existing code doesn't have to follow new path

Do you means importing them inside `reflex/state.py` ?
I tried at first but ruff remove them for being unused.

adhami3310:
we ended up moving other functions from state into istate, the builtin states are kind of small in comparison so it feels this is rather not going to be much impactful in the organization benefit

PR #4222, opened 2024-10-22
make python-dotenv optional

Lendemor:


adhami3310 (approved):

benedikt-bartscher (approved):
Thanks, @Lendemor !

Lendemor:
You're welcome.
By the way, if there is a specific reason you needed it to be optional, could you leave it her as a note for the future?

benedikt-bartscher:
> You're welcome. By the way, if there is a specific reason you needed it to be optional, could you leave it her as a note for the future?

Of course, i explained it here: https://github.com/reflex-dev/reflex/pull/4219#issuecomment-2431582692

PR #4223, opened 2024-10-23
components as literal vars

adhami3310:


reflex/components/component.py
```diff
@@ -2340,3 +2349,209 @@ def create(cls, *children, **props) -> Component:
...
+
+    Returns:
+        An empty component.
+    """
+    from reflex.components.base.bare import Bare
+
+    return Bare.create("")
+
+
+def render_dict_to_var(tag: dict | Component | str, imported_names: set[str]) -> Var:
```
masenf, line 2369: it would be nice if we could move some of this special specific logic closer to the components that will actually depend on it

adhami3310: at the current stage there's no standardization of the dict format, so the checks are very tight to what we have right now, i can restructure the rendering logic in the future if needed

adhami3310:
should be merged with https://github.com/reflex-dev/reflex-web/pull/1047

adhami3310:
waiting on https://github.com/reflex-dev/reflex-web/pull/1050

masenf (commented):
can we pare this down and keep the nextjs15 and turbo changes in the other PR #4243

Lendemor (approved):

tests/units/components/forms/test_uploads.py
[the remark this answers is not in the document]
adhami3310: they are removed because it's annoyingly specific, since there are integration test cases I thought it's not worth the effort of having test cases that just repeat the logic of the code.

masenf (approved):

PR #4229, opened 2024-10-23
Support `locale` prop in `rx.moment`

masenf:


reflex/components/moment/moment.py
```diff
@@ -101,22 +105,15 @@ def add_imports(self) -> ImportDict:
...
-            *children: The children of the component.
-            **props: The properties of the component.
+            imports["moment-timezone"] = ""
 
-        Returns:
-            The Moment Component.
-        """
-        comp = super().create(*children, **props)
-        if "tz" in props:
-            comp.lib_dependencies.append("moment-timezone")
```
Lendemor, line 121: Is `moment-timezone` no longer needed?

masenf: It's handled in add_imports

masenf:
Removed the date datetime change, as that was screwing up reflex web and we already have another issue tracking it.

Lendemor (approved):

Issue #4147, opened 2024-10-10
SQLAlchemy AssociationProxy creator lambda vs pickle


benedikt-bartscher:
**Describe the bug**
In SQLAlchemy AssociationProxy it's a common pattern to set the creator to an lambda like this:
```python
class User(Base):
    ...

    # use Keyword(keyword=kw) on append() events
    keywords: AssociationProxy[List[str]] = association_proxy(
        "kw", "keyword", creator=lambda kw: Keyword(keyword=kw)
    )
```
(ref: https://docs.sqlalchemy.org/en/20/orm/extensions/associationproxy.html#creation-of-new-values)
Since reflex uses bare pickle again, this causes issues with Redis State Manager

**Expected behavior**
Raise a better exception by either improving the PicklingError exception or checking all SQLAlchemy models before compile

**Specifics (please complete the following information):**
 - Python Version: 3.12.7
 - Reflex Version: main branch
 - OS: arch

PR #4239, opened 2024-10-24
[ENG-3970] When normal pickle fails, try dill

masenf:
If dill is not installed, suggest that the user `pip install` it.

Fix https://github.com/reflex-dev/reflex/issues/4147


adhami3310:
Since we are adding dill to pyproject, is there any way that dill wouldn't be installed?

masenf:
> Since we are adding dill to pyproject, is there any way that dill wouldn't be installed?

it's a dev dependency, so it only gets installed when you do `poetry install`. if you just `pip install reflex`, then it won't be included

adhami3310 (approved):

PR #4243, opened 2024-10-25
upgrade to nextJS v15

Lendemor:


masenf:
need to bump minimum node version as well

> You are using Node.js 18.17.0. For Next.js, Node.js version >= v18.18.0 is required.

Lendemor:
Done.
Also added the turbo flag removed from  #4223

adhami3310:
turbo requires babel to be removed, which that PR would address when it gets merged

adhami3310 (approved):

PR #4247, opened 2024-10-25
fix stateful components on delayed evaluation

adhami3310:


masenf (approved):
Fixes StatefulComponent creation 😁

PR #4248, opened 2024-10-26
More env var cleanup

### LINKED ISSUE #4172: Unify env vars

benedikt-bartscher:
closes #4172


benedikt-bartscher:
todo: add a flag to disable `Path.exists` validation

reflex/app.py
```diff
@@ -505,7 +505,10 @@ def add_page(
         # Check if the route given is valid
         verify_route_validity(route)
 
-        if route in self.unevaluated_pages and os.getenv(constants.RELOAD_CONFIG):
+        # TODO: this was broken?
```
benedikt-bartscher, line 508: @reflex-dev is this intended behavior? I wanted to use `environment.RELOAD_CONFIG.get` here, but it breaks the tests. Take a look here: https://github.com/reflex-dev/reflex/pull/4255

benedikt-bartscher: This should work as well
```diff
diff --git a/reflex/app.py b/reflex/app.py
index c4bbf8bf..40942d10 100644
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -12,7 +12,6 @@ import inspect
 import io
 import json
 import multiprocessing
-import os
 import platform
 import sys
 import traceback
@@ -505,9 +504,9 @@ class App(MiddlewareMixin, LifespanMixin, Base):
         # Check if the route given is valid
         verify_route_validity(route)
 
-        # TODO: this was broken?
-        if route in self.unevaluated_pages and os.getenv(
-            environment.RELOAD_CONFIG.name
+        if (
+            route in self.unevaluated_pages
+            and environment.RELOAD_CONFIG.get is not None
         ):
             # when the app is reloaded(typically for app harness tests), we should maintain
             # the latest render function of a route.This applies typically to decorated pages
diff --git a/reflex/config.py b/reflex/config.py
index 7b910e3e..fcf29d5f 100644
--- a/reflex/config.py
+++ b/reflex/config.py
@@ -486,7 +486,7 @@ class EnvironmentVariables:
     REFLEX_FRONTEND_ONLY: EnvVar[bool] = env_var(False)
 
     # Reflex internal env to reload the config.
-    RELOAD_CONFIG: EnvVar[bool] = env_var(False, internal=True)
+    RELOAD_CONFIG: EnvVar[Optional[bool]] = env_var(None, internal=True)
 
     # If this env var is set to "yes", App.compile will be a no-op
     REFLEX_SKIP_COMPILE: EnvVar[bool] = env_var(False, internal=True)

```

reflex/utils/exec.py
```diff
@@ -489,35 +491,5 @@ def is_prod_mode() -> bool:
...
         True if the app is running in production mode or False if running in dev mode.
     """
-    current_mode = os.environ.get(
-        constants.ENV_MODE_ENV_VAR,
-        constants.Env.DEV.value,
-    )
-    return current_mode == constants.Env.PROD.value
-
-
-def is_frontend_only() -> bool:
```
masenf, line 499: are these functions moved somewhere else? or just removed?

if removed, I think it would be prudent to deprecate them and remove them completely in 0.7.0 instead of in a 0.6.x dot release

masenf (commented):
overall i like the descriptor implementation

i'm not too sure about `.get` being a property though, to me that kind of looks weird. did you consider having it as a normal method like `.set(...)`

@adhami3310 wondering if you have any input here as well

benedikt-bartscher:
> overall i like the descriptor implementation

Thanks, i wanted to try binding the descriptors to `rx.Config` as well in a followup.

> i'm not too sure about `.get` being a property though, to me that kind of looks weird. did you consider having it as a normal method like `.set(...)`

Yeah i was thinking about making it a normal metehod.. This would also allow us to add f.e. a `cache` arg to `get` as well.
If @adhami3310 agrees i will refactor it.

reflex/utils/exec.py
benedikt-bartscher: I just removed them because they just added another indirection which did nothing. If you prefer i can just readd them or deprecate them.

adhami3310:
>  If @adhami3310 agrees i will refactor it.

yea that would be great

benedikt-bartscher:
Refactor done :heavy_check_mark:

reflex/app.py
benedikt-bartscher: I will just assume that this was the correct behavior. I slightly improved the EnvVar api for such cases by adding `EnvVar.is_set`

masenf (approved):

PR #4255, opened 2024-10-28
showcase potential bug for #4248

benedikt-bartscher:


reflex/app.py
```diff
@@ -505,7 +505,10 @@ def add_page(
         # Check if the route given is valid
         verify_route_validity(route)
 
-        if route in self.unevaluated_pages and os.getenv(constants.RELOAD_CONFIG):
+        if route in self.unevaluated_pages and interpret_boolean_env(
```
Lendemor, line 508: Rather than importing `interpret_boolean_env` here, we should define `RELOAD_CONFIG` in `EnvironmentVariables` and use the `environment` that is already imported with `environment.RELOAD_CONFIG`

benedikt-bartscher: It is, but it doesnt work, because the current behavior does not correctly check for true boolean vars. The direct interpret_boolean_env is just to showcase it for you

PR #4274, opened 2024-10-30
use better typing for on_load

adhami3310:


masenf (approved):
looks like a nice improvement

PR #4276, opened 2024-10-30
Add option to scroll to bottom

picklelo:
Add an option on `rx.scroll_to` to scroll to the bottom of the component instead of the top.

### DIFF
--- a/reflex/event.py
+++ b/reflex/event.py
@@ -777,8 +777,15 @@ def scroll_to(elem_id: str, align_to_top: bool | Var[bool] = True) -> EventSpec:
     Returns:
         An EventSpec to scroll the page to the selected element.
     """
-    js_code = f"document.getElementById('{elem_id}').scrollIntoView({str(Var.create(align_to_top))});"
-    return call_script(js_code)
+    get_element_by_id = FunctionStringVar.create("document.getElementById")
+
+    return call_script(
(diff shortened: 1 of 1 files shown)


reflex/event.py
```diff
@@ -767,17 +767,17 @@ def set_focus(ref: str) -> EventSpec:
...
-        elem_id: the id of the element
+        elem_id: The id of the element to scroll to.
+        align_to_top: Whether to scroll to the top (True) or bottom (False) of the element.
 
     Returns:
         An EventSpec to scroll the page to the selected element.
     """
-    js_code = f"document.getElementById('{elem_id}').scrollIntoView();"
-
+    js_code = f"document.getElementById('{elem_id}').scrollIntoView({str(Var.create(align_to_top))});"
```
masenf, line 780: `str` casting this Var will cause it to lose any `VarData`, so if someone inlines `rx.scroll_to` with `align_to_top` as a state Var, then it might throw a frontend error.

if they call it from the backend, it's fine, because they will be passing a real `bool` from the state instance, rather than a Var

picklelo: Tested this out with @adhami3310 's change and it works on my app

masenf (commented):
Shenanigans!

adhami3310 (approved):

PR #4280, opened 2024-10-31
allow custom bunfig.toml file

### LINKED ISSUE #4225: bunfig.toml always generated in `reflex init`, and it's not easy to use AWS CodeArtiafct repository as NPM registry
**Describe the bug**
`bunfig.toml` is always generated by `reflex init` and the registry URL is set as `https://registry.npmjs.org` by default.
Custom NPM registry can be specified by the environment variable `NPM_CONFIG_REGISTRY`, but it's impossible to use AWS CodeArtifact repository because we have to specify registry URL with an **auth-token**.

**Expected behavior**
When the environment variable `NPM_CONFIG_REGISTRY` is not set, `install.registry` in `bunfig.toml` should not be set.
Or at least, any option to disable `install.registry` in `bunfig.toml` should be provided.

Bun follows `.npmrc` next to `bunfig.toml`, and we can easily setup `.npmrc` by using AWS CLI.

As an another way, if an auth-token can be specified by any environment variable, we can get one and set to that.
Reference: https://bun.sh/docs/runtime/bunfig#install-registry
But the auth-token expires in 12 hours, so we have to update it. That seems harder than the way simply to follow `.npmrc`.

**Additional context**
When the environment variable `NPM_CONFIG_REGISTRY` is not set, `reflex init` always try to access public registries to resolve the best one automatically. I think it's an unsuitable behavior for self-hosting environments without any internet connection.

Lendemor:
If users define a `bunfig.toml` file at the root of their reflex project, during the init step, it will be copied in `.web` folder and used instead.
Closes #4225


adhami3310:
considering that this would require creating a file that cannot be checked to git, can we instead use something like `NPM_CONFIG_TOKEN` flag?

adhami3310:
actually it seems from the docs that it's possible to make it reference an env variable, then it's prob fine

reflex/utils/prerequisites.py
```diff
@@ -611,16 +613,26 @@ def _compile_package_json():
...
+
+    if (custom_bunfig := Path(constants.Bun.CONFIG_PATH)).exists():
+        custom_bunfig_content = custom_bunfig.read_text()
+        if "registry =" in custom_bunfig_content:
+            bun_config_path.write_text(custom_bunfig_content)
+            return
+        else:
+            console.error(
+                "The custom bunfig.toml file is invalid. Using default value."
+            )
```
masenf, line 631: why do we require the user to specify `registry =` with that exact spacing?

if we're going to enforce that the custom bunfig has a registry then

1. we need to parse the toml and check the resulting data structure
2. the error message needs to tell the user why their custom bunfig is being rejected

adhami3310: On that point, I think it's a better idea to always copy that file if it exists and add console.info that it did that

n0r1z0: I have the same opinion with @adhami3310 .

In our use case, we will place an EMPTY bunfig if we have to place one to disable that Reflex automatically resolve the best NPM registry.

reference: https://github.com/reflex-dev/reflex/issues/4225

adhami3310 (approved):

adhami3310 (approved):

PR #4282, opened 2024-10-31
allow for event handlers to ignore args

adhami3310:


masenf (approved):
looks good aside from the large amount of commented out test code; should this just be removed to avoid confusion?

adhami3310:
> should this just be removed to avoid confusion?

i added them to valid test cases instead

reflex/event.py
```diff
@@ -1058,8 +1058,8 @@ def get_hydrate_event(state) -> str:
 
 
 def call_event_handler(
-    event_handler: EventHandler | EventSpec,
-    arg_spec: ArgsSpec | Sequence[ArgsSpec],
+    event_callback: EventHandler | EventSpec,
+    event_spec: ArgsSpec | Sequence[ArgsSpec],
```
Lendemor, line 1062: Won't it be confusing to have 
`event_callback` of type `EventSpec`
but then
`event_spec` of type `ArgsSpec` ?

Tho I'm not sure what we could use instead, so maybe it's fine.

Lendemor (dismissed):

reflex/event.py
adhami3310: I would love to change existing terminology but that's the best i could do while working in the system :/

Lendemor (approved):

PR #4287, opened 2024-11-01
add metainfo to keyevent

adhami3310:


masenf (approved):
i think the `isort` formatting change probably should have come in a separate PR to make this easier to review the actual change, but looks good.

PR #4288, opened 2024-11-01
add noop event

adhami3310:


Lendemor (approved):
Looks fine to me.

The EventNamespace is starting to be quite big, maybe we should consider splitting some part into sub namespaces, in a follow up PR maybe?

adhami3310:
> we should consider splitting some part into sub namespaces

that would be a breaking change, so i'm hesitant to do it any time soon :(

PR #4301, opened 2024-11-04
handle none case in state setattr

adhami3310:
Fix:
```
DeprecationWarning: mismatched-type-assignment has been deprecated in version 
0.6.5 Tried to assign value None of type <class 'NoneType'> to field 
State.parent_state of type <class 'reflex.state.BaseState'>. This might lead to 
unexpected behavior. It will be completely removed in 0.7.0
```


masenf (approved):
gracias

PR #4307, opened 2024-11-05
implement redis key prefix for StateManagerRedis

### LINKED ISSUE #4303: Redis key prefix
Add env var to set a redis key prefix
Allows reflex deployments to share a redis server

benedikt-bartscher:
Closes #4303


benedikt-bartscher:
I think this is not worth the overhead, you can most likely use one redis server anyway because of uuid4?

adhami3310:
i think the point with uuid4 makes sense, there shouldn't be an issue on that regard

PR #4311, opened 2024-11-06
stop ignoring some lint rules

Lendemor:


tests/integration/tests_playwright/test_router_data.py
adhami3310, line 1: why is this file being added?

masenf (approved):
i'm actually surprised we had so many useless f-strings here

Lendemor:
> i'm actually surprised we had so many useless f-strings here

Yeah, I wonder why we ignored that rule in the first place, doesn't seems like it would change any behaviour in the code 🤔

PR #4329, opened 2024-11-07
redesign error boundary screen

adhami3310:


masenf (commented):
this looks much better than the crap we had before, nice.

i think visually it might be better if it were in the top third rather than vertically centered on the page, but i'm not splitting hairs as i'm not a designer.

you should add the screenshots to the PR description and have @Alek99 / @carlosabadia take a look at it

adhami3310:
image
image
image
image

adhami3310:
@Alek99 @carlosabadia let me know if improvements can be made!

Note this has to be implemented in pure HTML, as we don't get Radix in the fallback place.

adhami3310:
@masenf it doesn't seem we're getting feedback on this now, so we can merge it and redesign it later if needed

carlosabadia:
didn't see this, sorry. Looks good to me!

masenf (approved):

Issue #4264, opened 2024-10-29
Cast dict as model type if annotated


masenf:
**Describe the bug**
Event handlers always seem to get a `dict` even when the user expects to get a particular type.

**To Reproduce**
```python
import reflex as rx


class Amigo(rx.Base):
    nombre: str
    anos: int

class State(rx.State):
    amigos: list[Amigo] = [
        Amigo(nombre="Juan", anos=20),
        Amigo(nombre="Pedro", anos=30),
        Amigo(nombre="Maria", anos=40),
        Amigo(nombre="Jose", anos=50),
    ]

    amigo_actual: Amigo | None

    def set_amigo_actual(self, amigo: Amigo) -> None:
        print(repr(amigo))
        self.amigo_actual = amigo


def hacer_amigo(a: Amigo) -> rx.Component:
    return rx.button(a.nombre, on_click=State.set_amigo_actual(a))


def index() -> rx.Component:
    return rx.container(
        rx.color_mode.button(position="top-right"),
        rx.vstack(
            rx.heading("Welcome to Reflex!", size="9"),
            rx.text(f"Amigo Actual: {State.amigo_actual.nombre}"),
            rx.foreach(
                State.amigos,
                hacer_amigo,
            )
        ),
        rx.logo(),
    )


app = rx.App()
app.add_page(index)
```

**Expected behavior**
The argument received by the `set_amigo_actual` event handler should be an `Amigo`, not a `dict`.

**Specifics (please complete the following information):**
 - Python Version: 3.12
 - Reflex Version: 0.6.4a3
 - OS: macOS / Any
 
**Additional context**
From this forum thread https://forum.reflex.dev/t/understanding-state-types/488

However this is a long-standing problem in Reflex in that arguments passed to event handlers are always the bare un-JSON'd type, rather than the preferred model type.

There is some provision for casting number-type arguments to actual int/float before passing them through to the event handler, but this logic does not apply for dict types that are annotated with another, specific type.

benedikt-bartscher:
I think we need a flexible way to register deserialization/unmarshalling functions like reflex already does for serialization

PR #4331, opened 2024-11-08
convert event return types to type hints

adhami3310:


adhami3310:
```py
import dataclasses
import reflex as rx


@dataclasses.dataclass
class Guy:
    name: str


class State(rx.State):
    """State for the TestApp app."""

    @rx.event
    def on_event(self, payload: Guy):
        print(f"Hello {payload.name}!")


app = rx.App()


@app.add_page
def index():
    return rx.vstack(
        rx.button(
            on_click=State.on_event(Guy("Alice")),
        ),
    )
```

adhami3310:
Fixes #4264

adhami3310:
@benedikt-bartscher how would you imagine hooking into this system? maybe

```py
@deserializer
def deserialize_datetime(value: str) -> datetime:
    ...
```

masenf (approved):
pretty neat

adhami3310:
In case we want to improve the system, we can do so in later PRs o7

PR #4338, opened 2024-11-08
[ENG-3953] Support pydantic BaseModel (v1 and v2) as state var

masenf:
Provide serializers and mutable proxy tracking for pydantic models directly.

### DIFF
--- a/reflex/utils/serializers.py
+++ b/reflex/utils/serializers.py
@@ -24,6 +24,9 @@ from typing import (
     overload,
 )
 
+from pydantic import BaseModel as BaseModelV2
+from pydantic.v1 import BaseModel as BaseModelV1
+
 from reflex.base import Base
 from reflex.constants.colors import Color, format_color
 from reflex.utils import types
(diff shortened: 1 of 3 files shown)


reflex/utils/serializers.py
```diff
@@ -266,6 +269,32 @@ def serialize_base(value: Base) -> dict:
...
+def serialize_base_model_v2(model: BaseModelV2) -> dict:
+    """Serialize a pydantic v2 BaseModel instance.
+
+    Args:
+        model: The BaseModel to serialize.
+
+    Returns:
+        The serialized BaseModel.
+    """
+    return model.model_dump()
```
adhami3310, line 295:
```suggestion
if BaseModelV2 is not BaseModelV1:
    @serializer(to=dict)
    def serialize_base_model_v2(model: BaseModelV2) -> dict:
        """Serialize a pydantic v2 BaseModel instance.
    
        Args:
            model: The BaseModel to serialize.
    
        Returns:
            The serialized BaseModel.
        """
        return model.model_dump()
```

masenf: tks

masenf:
also need to make the import of `pydantic.v1` conditional, because 1.10.2 doesn't have that namespace

adhami3310 (approved):

PR #4348, opened 2024-11-10
Allow bound method as event handler

benedikt-bartscher:


adhami3310:
Also, in a case like this, tests would be appreciated to highlight if the existing code for this has regressed

benedikt-bartscher:
@adhami3310 I added a test which is similar to our regression. We are using smth like this for quite a while now and it works fine. However, this is not really a documented/supported feature of reflex. Hence, this PR could be seen as a feature, not a regression bug fix

adhami3310 (approved):

adhami3310:
Since it's similar to the lambda feature, I think this one makes sense. Although it would have been cooler to unite it with that other path.

PR #4368, opened 2024-11-12
Downgrade syntax highlighter to fix `wrapLongLines` issue

ElijahAhianyo:
ReactSyntaxHighligter `v15.16.1` breaks the functionality of the `wrap_long_line` prop. Downgrading to `15.16.0` works.

Related upstream PR: https://github.com/react-syntax-highlighter/react-syntax-highlighter/pull/487

### DIFF
--- a/reflex/components/datadisplay/code.py
+++ b/reflex/components/datadisplay/code.py
@@ -382,7 +382,7 @@ for theme_name in dir(Theme):
 class CodeBlock(Component, MarkdownComponentMap):
     """A code block."""
 
-    library = "react-syntax-highlighter@15.6.1"
+    library = "react-syntax-highlighter@15.6.0"
 
     tag = "PrismAsyncLight"
 


adhami3310:
How does ReactSyntaxHighligter `v15.16.1` break it? i looked at the PR and it doesn't seem doing anything bad?

ElijahAhianyo:
> How does ReactSyntaxHighligter `v15.16.1` break it? i looked at the PR and it doesn't seem doing anything bad?

That's what's not obvious to me. We don't, by default, set any css props, which should work the way it is. The referenced PR only swaps the order of args, so maybe a chain of reaction somewhere? I haven't dug into the reactsyntaxhighlighter codebase to to see what other parts may have been affected. Either way, Im not sure if this will require a fix from us than them.

Lendemor (approved):

PR #4372, opened 2024-11-13
add typing to function vars

adhami3310:


reflex/vars/function.py
```diff
@@ -118,9 +273,15 @@ def create(
         Returns:
             The function call var.
         """
+        function_return_type = (
+            func._var_type.__args__[1]
```
masenf, line 277: this could give a weird `IndexError` if someone tries to pass an incorrect _var_type for their `FunctionVar`

masenf (approved):

reflex/vars/function.py
adhami3310: Yea will enforce that it's `ReflexCallable` in following PRs o7

PR #4377, opened 2024-11-13
[GTM-836]Rework Init workflow

ElijahAhianyo:
Changing the `reflex init` workflow to below:

`reflex init` now gives 3 options:

```shell
Get started with a template:
(0) blank (https://blank-template.reflex.run) - A blank Reflex app.
(1) ai - Generate a template using AI [Experimental]
(2) choose templates - Choose an existing template.
$ Which template would you like to use? (0): 
```

- Blank template will initialize the app using the default template.
- The ai option redirects the user to flexgen where the app is initialized based on the generation. This option is the same as running `reflex init --ai`
- The third option now requests for a template name which you can find when you navigate to https://reflex.dev/tempates. The template name is on the template card which you can copy. When an invalid template is used, The user is prompted to enter a valid one

```shell
Visit https://reflex.dev/templates for the complete list of templates.
$ Enter a valid template name: invalid-template-name
Invalid template name. Please try again.
Visit https://reflex.dev/templates for the complete list of templates.
$ Enter a valid template name: 
```

You should also be able to pass in the template's name directly using the template flag
```shell
$ reflex init --template sales 
```


picklelo (commented):
Flow is working pretty well for me.

I'm wondering if we shouldn't show the urls for flexgen / templates and instead if they press (1) we do the ai workflow (this is already done). and if they press (2) then we open the templates page for them.

ElijahAhianyo:
> Flow is working pretty well for me.
> 
> I'm wondering if we shouldn't show the urls for flexgen / templates and instead if they press (1) we do the ai workflow (this is already done). and if they press (2) then we open the templates page for them.

redirecting users to the templates page was my initial intuition, however I felt redirecting users to the templates immediately wont be so obvious what they need to do on the page. It wont be an issue if they've done it multiple times, but for the first time, it could introduce some friction as opposed to explicitly stating they need to go on the page where they can get the names. Aside that concern,  it shouldn't be an issue to redirect them

picklelo (approved):
works well for me (we should merge the reflex-web PR along with this)

masenf (approved):
it would be cool if the success message for `ai` runs shows the generation hash, right now it says

```console
(VENV-dev311) masen@asmbp21 repro-init-flow % reflex init
──────────────────────────────────────────────────────── Initializing repro_init_flow ────────────────────────────────────────────────────────
[16:11:47] Initializing the web directory.                                                                                      console.py:104

Get started with a template:
(0) blank (https://blank-template.reflex.run) - A blank Reflex app.
(1) ai - Generate a template using AI [Experimental]
(2) choose templates - Choose an existing template.
Which template would you like to use? (0): 1
Info: Complete the workflow in the browser to continue.
[16:15:23] Initializing the app directory.                                                                                      console.py:104
Success: Initialized repro_init_flow using the blank template
```

PR #4380, opened 2024-11-13
ObjectVar __getitem__ typing issues

benedikt-bartscher:
@adhami3310 do you have any idea on how to fix the typing test?

`reflex/tests/units/vars/test_object.py:112:5 - error: "__getitem__" method not defined on type "ObjectItemOperation" (reportGeneralTypeIssues)`

### DIFF
--- a/tests/units/vars/test_object.py
+++ b/tests/units/vars/test_object.py
@@ -26,10 +26,17 @@ def serialize_bare(obj: Bare) -> dict:
     return {"quantity": obj.quantity}
 
 
+class Tag(rx.Base):
+    """A Tag for testing."""
+
+    pass
+
+
 class Base(rx.Base):
     """A reflex base class with a single attribute."""
 
     quantity: int = 0
+    collection: list[Tag] = []
 
 
 class ObjectState(rx.State):
@@ -100,3 +107,6 @@ def test_typing() -> None:
     # Base
     var = ObjectState.base
     _ = assert_type(var, ObjectVar[Base])
+
+    # Collection
+    ObjectState.base.collection[0]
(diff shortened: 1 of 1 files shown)


adhami3310:
`ObjectItemOperation` is essentially just a `Var` with no typing. You can generally call `.to(< the type it should be>` on it.

adhami3310:
I think in the example above it could lose the typing ? I can take a look into that.

benedikt-bartscher:
> I think in the example above it could lose the typing ? I can take a look into that.

Reflex always supported accessing vars in that way, so I think the general goal should be to add full typing support for this.
However, we could lose the typing as a mitigation. But I would not put too much effort into that, we can use `# pyright ignore` annotations until reflex supports it.

adhami3310:
by lose typing i meant runtime typing, reflex should maintain runtime typing as long as possible to limit the need for `.to`, but static typing is a different question, and generally difficult to get right with object vars

adhami3310:
i believe before, all vars supported `[]`, but right now we're moving away from that, the issue in this case is that static typing doesn't pick up on that field is an object var, as far as i know, it seems impossible to provide static typing that would not lose the context here without returning Any

benedikt-bartscher:
Sounds like you are hitting some of python's typing limits? :/

I have just added another test (https://github.com/reflex-dev/reflex/pull/4380/commits/13845143f8d5d571700cb6125e78adf2b8b10193) which is quite interesting. `dict[str, str]` works but `dict[str, bool]` does not. This sounds like it should be fixable.

benedikt-bartscher:
@adhami3310 any news regarding the objectvar `dict[str, bool]` thing? Shall I open an issue for this?

PR #4389, opened 2024-11-18
Unify `is_external` prop in `rx.redirect` and `rx.link`

ElijahAhianyo:
Deprecate `external` prop in `rx.redirect` in favor of  `is_external`


Lendemor (approved):

reflex/event.py
```diff
@@ -704,24 +704,35 @@ def fn():
...
         An event to redirect to the path.
     """
+    if external is not None:
+        console.deprecate(
+            "The `external` prop in `rx.redirect`",
+            "use `is_external` instead.",
+            "0.6.6",
+            "0.7.0",
+        )
+        is_external = external
```
masenf, line 729: hmm, if both `external` and `is_external` are specified, it seems maybe wrong if the deprecated arg takes over the value of the non-deprecated arg.

the code for sure gets uglier if both are `None` by default, but that way you'd be able to tell them apart. i'm on the fence about whether that's an improvement, but i wanted to mention it.

reflex/event.py
```diff
@@ -704,24 +704,35 @@ def fn():
 
 def redirect(
     path: str | Var[str],
-    external: Optional[bool] = False,
+    external: Optional[bool] = None,
+    is_external: Optional[bool] = False,
     replace: Optional[bool] = False,
```
masenf, line 709: if these have not `None` defaults, the annotation should just be `bool`

masenf (dismissed):

reflex/event.py
ElijahAhianyo: How about doing something like this which ensures that `is_external` always takes precedence:

```python
def redirect(
    path: str | Var[str],
    external: Optional[bool] = None,
    is_external: Optional[bool] = None,
    replace: bool = False,
) -> EventSpec:
    """Redirect to a new path.

    Args:
        path: The path to redirect to.
        external: Whether to open in new tab or not.
        is_external: Whether to open in new tab or not.
        replace: If True, the current page will not create a new history entry.

    Returns:
        An event to redirect to the path.
    """

    if external is not None:
        console.deprecate(
            "The `external` prop in `rx.redirect`",
            "use `is_external` instead.",
            "0.6.6",
            "0.7.0",
        )

    if external is None and is_external is None:
        is_external = False

    if external is not None and is_external is None:
        is_external = external

    return server_side(
        "_redirect",
        get_fn_signature(redirect),
        path=path,
        external=is_external,
        replace=replace,
    )

```

adhami3310:
```py
is_external = (False if external is None else external) if is_external is None else is_external
```

ElijahAhianyo: Not a huge fan of this one-liner but that works too

adhami3310: yea i did it for fun more than practicality (i do like having the logic laid out without multiple branches though)

reflex/event.py
```diff
@@ -704,24 +704,41 @@ def fn():
 
 def redirect(
     path: str | Var[str],
-    external: Optional[bool] = False,
-    replace: Optional[bool] = False,
+    external: Optional[bool] = None,
+    is_external: Optional[bool] = None,
+    replace: bool = False,
```
masenf, line 709:
```suggestion
    is_external: Optional[bool] = None,
    replace: bool = False,
    external: Optional[bool] = None,
```

Okay i could nitpick this PR to death i guess? everytime i look at it i see something else to call out; hear me out:

if we move the deprecated parameter to the end of the argument list this has the advantage that previous invocations of this function that were being called with position arguments continue to work fine.

consider

```python
rx.redirect("https://reflex.dev", True, True)
```

Before this PR the arguments would be interpreted as external=True and replace=True; although with the patch as it's currently written the behavior of the function would change, and now it's interpreted as replace=False because of the inserted argument.

If you take the suggestion to rename `external` to `is_external` and move the compat `external` argument _after_ `replace`, then nobody (foolishly) using positional arguments will be broken and there is no behavior change due to the signature changing.

For maximum API stability, a function's semantic signature should _never_ change without good reason.

ElijahAhianyo: yeah that makes a lot of sense

reflex/event.py
```diff
@@ -704,24 +704,41 @@ def fn():
 
 def redirect(
     path: str | Var[str],
```
adhami3310, line 706:
```suggestion

@overload
def redirect(
    path: str | Var[str],
    is_external: Optional[bool] = None,
    replace: bool = False,
) -> EventSpec: ...

@overload
@typing_extensions.deprecated("`external` is deprecatedm use `is_external` instead") 
def redirect(
    path: str | Var[str],
    is_external: Optional[bool] = None,
    replace: bool = False,
    external: Optional[bool] = None,
) -> EventSpec: ...

def redirect(
    path: str | Var[str],
```
nitpick beyond comprehension: this adds deprecation to the static checker

masenf (approved):

PR #4404, opened 2024-11-20
New Event Action: temporal

masenf:
When an event/event spec is marked as "temporal", it will not be queued unless the backend is up. This can be used to prevent periodic events (like from `rx.moment`) from queueing up while the backend is down, and then stampeding when the backend comes up and the queue is drained. It can be used to avoid processing many periodic events at once when the app is only expecting to process such an event every so often.

### DIFF
--- a/reflex/event.py
+++ b/reflex/event.py
@@ -181,6 +181,18 @@ class EventActionsMixin:
             event_actions={"debounce": delay_ms, **self.event_actions},
         )
 
+    @property
+    def temporal(self):
+        """Do not queue the event if the backend is down.
+
+        Returns:
+            New EventHandler-like with temporal set to True.
+        """
+        return dataclasses.replace(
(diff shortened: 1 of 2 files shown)


reflex/event.py
```diff
@@ -181,6 +181,18 @@ def debounce(self, delay_ms: int):
             event_actions={"debounce": delay_ms, **self.event_actions},
         )
 
+    @property
```
Lendemor, line 184: `throttle` and `debounce` don't have the `@property` decorator but are defined similarly, is it just those two that don't need it?

masenf: yeah this one is more along the lines of `prevent_default` and `stop_propagation`

Lendemor (approved):

adhami3310 (approved):

PR #4409, opened 2024-11-21
For non-cached computed vars, skip dep computation

masenf:


benedikt-bartscher:
@masenf this seems legit, only needs some test adjustments?

masenf:
merged with #4711

PR #4419, opened 2024-11-22
test fast yielding in background task

benedikt-bartscher:


benedikt-bartscher:
@masenf any idea why fast yielding background tasks seem to lock up after some state updates?

benedikt-bartscher:
@masenf i accidentally pushed unfinished changes and forgot to mention that the test needs to be run with redis

benedikt-bartscher:
Another idea to improve this: one subscription per token/lock_key instead of lock_id. Reuse the subscription and release/garbage collect it after a delay.

reflex/state.py
```diff
@@ -3475,6 +3502,19 @@ async def _wait_lock(self, lock_key: bytes, lock_id: bytes) -> None:
                         break
                 state_is_locked = await self._try_get_lock(lock_key, lock_id)
 
+    @override
+    async def disconnect(self, token: str):
```
benedikt-bartscher, line 3506: This is probably wrong. Clients can reconnect. Need to check if there are any unexpired redis tokens

masenf: i had a draft comment on my in progress review calling out this function, but i suppose your testing found it first. we need to keep the in memory states around. we do have an internal bug to add an expiry for memory state manager, but it just hasn't been high priority because we assume no one is using memory state manager in production

benedikt-bartscher: Maybe one asyncio Lock is enough, we wont need to garbage collection then

benedikt-bartscher:
@masenf is it true that the current state locking does not guarantee fifo? Imo the longest waiting task should get the lock next

reflex/state.py
benedikt-bartscher: Note to myself: subscribe to redis expire events to clean up locks, may be easier with #4459

benedikt-bartscher:
I have a local draft where only one pubsub instance is created per reflex worker. All lock releases are distributed to the waiting tasks via asyncio locks

masenf:
Closing in favor of #5932, which was inspired by this PR and also incorporates this concept:

> Another idea to improve this even further: one shared subscription per token and worker (if I am not mistaken, we basically need to lock a token to a worker on redis, further locking can be done with asyncio). Reuse the subscription and release/garbage collect it after a delay.

@benedikt-bartscher if you have some cycles to test on #5932 i'd appreciate any input you have. I'm working on some test cases for it, but since this isn't my main project at the moment, it might have to wait until next weekend.

PR #4430, opened 2024-11-24
WIP: State compression

benedikt-bartscher:


pyproject.toml
benedikt-bartscher, line 1: Just realized my toml formatter kicked in, shall I revert this?

masenf: this is fine, thanks for calling it out though 😀

reflex/state.py
```diff
@@ -2218,10 +2225,14 @@ def _deserialize(
             ValueError: If both data and fp are provided, or neither are provided.
             StateSchemaMismatchError: If the state schema does not match the expected schema.
         """
+        if environment.REFLEX_COMPRESS_STATE.get():
+            from blosc2 import decompress
+
+            data = decompress(data)
```
benedikt-bartscher, line 2231: this is broken for the fp case

reflex/state.py
```diff
@@ -2195,8 +2195,15 @@ def _serialize(self) -> bytes:
...
                 error += f"Dill was also unable to pickle the state: {ex}"
             console.warn(error)
+
+        if environment.REFLEX_COMPRESS_STATE.get():
+            from blosc2 import compress
+
+            payload = compress(payload)
+
         if environment.REFLEX_PERF_MODE.get() != PerformanceMode.OFF:
             self._check_state_size(len(payload))
```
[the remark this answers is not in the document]
benedikt-bartscher: Changed to consistent behavior, we now always check the uncompressed size

benedikt-bartscher:
While this can help a lot with big states, I think frontend payload compression could have an even bigger impact. Especially with slow connections, f.e. on mobile. Pickle payloads already benefit from referential serialization, json/ws does nothing similar. In fact, currently, the ws payloads are not compressed at all. Socketio only compresses in long-polling mode, which is only used for the initial connection and incompatible with gunicorn (I disabled it in https://github.com/reflex-dev/reflex/pull/4441)

masenf:
> In fact, currently, the ws payloads are not compressed at all.

how did you determine this? i'm looking at a packet capture of the websocket traffic and almost every message seems compressed

### Frame

image

### Decompressed payload

image

benedikt-bartscher:
@masenf i was just brainstorming in that comment and only read some docs. I did not test it. Just confirmed your wireshark capture, seems to be compressed by default. Interesting, i wonder where this happens, maybe i'll need to scroll trough some code to understand it.

benedikt-bartscher:
It is just not documented very well. python-socketio uses python-engineio which uses simple-websocket, which has compression enabled with hardcoded default settings since https://github.com/miguelgrinberg/simple-websocket/commit/9277e67140a456bd34e09146732d4bdca0c6db12

masenf:
In the JS version of socket.io, `perMessageDeflate` is now disabled by default: https://socket.io/docs/v4/server-options/#permessagedeflate

> This extension is known to add a significant overhead in terms of performance and memory consumption, so we suggest to only enable it if it is really needed.

benedikt-bartscher:
Very intersting, thanks for sharing. Maybe some (optimized) reflex apps could even benefit from disabling that compression. However with python-socketio reflex has no way to control this. Maybe we can adress that in the future. However we are getting a bit off-topic in this PR which is about pickle compression. Maybe i'll create an Issue to further discuss our performance impact and ideas for websocket compression.

masenf:

PR #4432, opened 2024-11-25
WIP - Shared redis subscribe

benedikt-bartscher:


masenf:
The spirit of this is also fixed by #5932, which optimizes the redis lock path:

* 1 pubsub per app instance
* create asyncio.Event to signal waiters to redis/lock lock expiration
* cached in-memory states when instance is holding redis lock with deferred write

PR #4433, opened 2024-11-25
Remove invitation code logic from reflex logoutv2

ElijahAhianyo:


reflex/reflex.py
```diff
@@ -629,12 +647,13 @@ def deployv2(
     vmtype: Optional[str] = typer.Option(
         None,
         "--vmtype",
-        help="Vm type id. Run `reflex apps vmtypes` to get options.",
+        help="Vm type id. Run reflex apps vmtypes list to get options.",
```
masenf, line 650: should leave the backticks
```suggestion
        help="Vm type id. Run `reflex apps vmtypes list` to get options.",
```

masenf (commented):
why are the version checks being removed? i thought @Kastier1 just added those

Kastier1:
Yeah please add back the version check!

ElijahAhianyo:
> why are the version checks being removed? i thought @Kastier1 just added those

wow, that's on me for not checking the PR after pushing the changes, this should only be a one-liner. Fixing this

Kastier1 (changes requested):
this change is not required

we are remove invitation codes for thew logout v2 not the v1

masenf (dismissed):

Kastier1 (approved):

Issue #4445, opened 2024-11-27
rx.[el.]input broken with null/None values


benedikt-bartscher:
**Describe the bug**
rx.input and rx.el.input do not update values if they are None/null

**To Reproduce**
```python
import reflex as rx
import random

class State(rx.State):
    simple: rx.Field[str | None] = rx.field(None)

    @rx.event
    def clear(self) -> None:
        self.simple = None

    @rx.event
    def state_reset(self) -> None:
        self.reset()

    @rx.event
    def set_random(self) -> None:
        self.simple = random.choice(["A", "B", "C"])


def index() -> rx.Component:
    return rx.container(
        rx.button("Reset", on_click=State.state_reset),
        rx.button("Clear", on_click=State.clear),
        rx.button("Random", on_click=State.set_random),
        rx.text("Simple", State.simple),
        rx.text("rx.input"),
        rx.input(value=State.simple),
        rx.text("rx.el.input"),
        rx.el.input(value=State.simple),
    )

app = rx.App()
app.add_page(index)
```

**Expected behavior**
Input clears on buttons reset and clear

**Specifics (please complete the following information):**
 - Python Version: 3.13
 - Reflex Version: main
 - OS: arch

benedikt-bartscher:
adding `.to(str)` to the input values doesn't change anything

benedikt-bartscher:
This also occurs with dict values and rx.Base models. Not only str base vars are affected

PR #4448, opened 2024-11-27
test #4445

benedikt-bartscher:
test for #4445


benedikt-bartscher:
Waiting for https://github.com/reflex-dev/reflex/pull/4528 to add more test cases without `rx.Field`

PR #4450, opened 2024-11-27
[ENG-4149] require login to deploy named templates

masenf:
* require `reflex loginv2` authentication and valid auth token to deploy a reflex-owned template
* github templates can be deployed without logging in


picklelo (approved):
working well for me

PR #4451, opened 2024-11-28
bump ruff to 0.8 and relock poetry

Lendemor:


adhami3310:
can we get rid of darglint yet?

Lendemor:
> can we get rid of darglint yet?

I think we need these rules to be out of preview : https://docs.astral.sh/ruff/rules/#pydoclint-doc

PR #4456, opened 2024-11-29
simplify redis code, less redis calls

benedikt-bartscher:


benedikt-bartscher:
@masenf was there a specific reason to use redis.exists and not directly try_get_lock?

reflex/state.py
```diff
@@ -3467,20 +3458,16 @@ async def _wait_lock(self, lock_key: bytes, lock_id: bytes) -> None:
...
-                    if message["data"] in self._redis_keyspace_lock_release_events:
-                        break
-                state_is_locked = await self._try_get_lock(lock_key, lock_id)
+            # wait for the lock to be released
+            while True:
+                # fast path
+                if await self._try_get_lock(lock_key, lock_id):
+                    return
+                # wait for lock events
+                _ = await pubsub.get_message(
```
benedikt-bartscher, line 3467: we could add a small wrapper around get_message which filters for _redis_keyspace_lock_release_events and calls get_message again with the time left

benedikt-bartscher: done in https://github.com/reflex-dev/reflex/pull/4456/commits/368a2ce55fd7286ecfbb8657c86bf821a1330ac5

masenf (approved):
looks good

PR #4459, opened 2024-12-01
redis hash for StateManagerRedis

benedikt-bartscher:
Drastically reduces the amount of redis calls for apps with many states
Keeps all states of a session in one redis hashmap instead of separate redis keys

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -3532,7 +3532,7 @@ class StateManagerRedis(StateManager):
             client_token: The name of the hash.
             redis_hashset: The keys and values to set.
         """
-        pipe = self.redis.pipeline()
+        pipe = self.redis.pipeline(transaction=False)
         pipe.hset(name=client_token, mapping=redis_hashset)
         if self._hexpire_not_supported:
             pipe.expire(client_token, self.token_expiration)


benedikt-bartscher:
ci failure is the typical dynamic routes flakyness

masenf:
@benedikt-bartscher i really like this one 👍

benedikt-bartscher:
@masenf thanks for the feedback and compability improvements you have pushed. I really appreciate it. We are already using this since some weeks and did not experience any issues. Will update our fork to the latest changes and keep you posted if everything still works.

benedikt-bartscher:
No issues so far, thanks again @masenf!
Going to fix the conflicts quickly

benedikt-bartscher:
`FAILED tests/units/test_state.py::test_state_manager_lock_warning_threshold_contend - AssertionError: assert 1 == 7`
the new test seems to fail on my machine with REDIS_URL set

masenf:
> the new test seems to fail on my machine with REDIS_URL set

pushed a fix; we only need to expect one warning message now, since we're only writing to redis once per event :sunglasses:

masenf:
@benedikt-bartscher have you profiled this? i notice that it's making fewer redis calls, but the state updates on reflex-web when bumping the counter on `/docs/getting-started/introduction` are taking ~0.04s whereas with `main` they only take ~0.02s

Not clear why it's taking twice as long to get/set the state with the new approach.

benedikt-bartscher:
> @benedikt-bartscher have you profiled this? i notice that it's making fewer redis calls, but the state updates on reflex-web when bumping the counter on `/docs/getting-started/introduction` are taking ~0.04s whereas with `main` they only take ~0.02s
> 
> Not clear why it's taking twice as long to get/set the state with the new approach.

@masenf i have not profiled it yet. Only checked the amount of redis calls. I wonder if the impact is from the actual redis calls or the computation of which states need to be fetched.

If this is due to redis hash, we can migrate back to normal redis keys using Pipelines to reduce the amount of requests. 

Will take a look on this over the weekend.

benedikt-bartscher:
@masenf I did some basic benchmarking with a redis server running on localhost (which is a bit unfair). Most state manager operations with this branch took longer than on main. I pushed the benchmark if you want to try or investigate.

benedikt-bartscher:
> If this is due to redis hash, we can migrate back to normal redis keys using Pipelines to reduce the amount of requests.

It is not.

reflex/state.py
```diff
@@ -1884,26 +1955,40 @@ def _dirty_computed_vars(
...
 
+    @classmethod
+    def _recursive_potentially_dirty_substates(cls) -> set[str]:
+        """Recursively determine substates which could be affected by dirty vars in this state.
+
+        Returns:
+            Set of full state names that may need to be fetched to recalc computed vars.
+        """
+        fetch_substates = cls._potentially_dirty_substates()
+        for substate_cls in cls.get_substates():
```
benedikt-bartscher, line 1986: this is not optimal, we only need to loop over the potentially dirty substates

masenf:

Will be moving to a different approach in the new state system.

PR #4463, opened 2024-12-02
rx.upload must include _var_data from props

masenf:
str-casting the dropzone arguments removed any VarData they depended on, like the state context.

update test_upload to include passing a prop from a state var.

update test_upload to include an upload handler that results in a large delta, plus a fix for handling multi-chunk deltas appropriately.

### DIFF
--- a/reflex/components/core/upload.py
+++ b/reflex/components/core/upload.py
@@ -293,13 +293,15 @@ class Upload(MemoizationLeaf):
             format.to_camel_case(key): value for key, value in upload_props.items()
         }
 
-        use_dropzone_arguments = {
-            "onDrop": event_var,
-            **upload_props,
-        }
+        use_dropzone_arguments = Var.create(
+            {
+                "onDrop": event_var,
+                **upload_props,
+            }
+        )
 
         left_side = f"const {{getRootProps: {root_props_unique_name}, getInputProps: {input_props_unique_name}}} "
-        right_side = f"useDropzone({str(Var.create(use_dropzone_arguments))})"
+        right_side = f"useDropzone({str(use_dropzone_arguments)})"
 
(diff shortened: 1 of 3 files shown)


reflex/components/core/upload.py
```diff
@@ -293,20 +293,23 @@ def create(cls, *children, **props) -> Component:
...
+        use_dropzone_arguments = Var.create(
+            {
+                "onDrop": event_var,
+                **upload_props,
+            }
+        )
 
         left_side = f"const {{getRootProps: {root_props_unique_name}, getInputProps: {input_props_unique_name}}} "
-        right_side = f"useDropzone({str(Var.create(use_dropzone_arguments))})"
+        right_side = f"useDropzone({str(use_dropzone_arguments)})"
```
Lendemor, line 304: I think f-string have an implicit call to str() by default, no ?

Lendemor (approved):

reflex/components/core/upload.py
benedikt-bartscher: iirc `__format__` takes precedence over `__str__`

masenf: > We can use the f-string flag `!s`

Interesting, i didn't realize that would bypass the normal `__format__` logic, but makes sense.

I think it's a bit cleaner looking, however I'm hesitant to make the change in this PR because we have the `f"{str(x)}"` pattern spread throughout the codebase already. If we're going to use `!s`, then we should have a separate PR that does it everywhere.

Lendemor: I mentioned this because I actually had a local PR where I'm doing this specific changes, let's go with this and I'll do the change on my PR.

PR #4469, opened 2024-12-03
enable PERF rules

Lendemor:
PERF are mostly micro-optimization, the impact should be negligible.
But it's better to have them than not.


ElijahAhianyo (dismissed):

reflex/compiler/utils.py
```diff
@@ -123,8 +123,7 @@ def compile_imports(import_dict: ParsedImportDict) -> list[dict]:
                 raise ValueError("No default field allowed for empty library.")
             if rest is None or len(rest) == 0:
                 raise ValueError("No fields to import.")
-            for module in sorted(rest):
-                import_dicts.append(get_import_dict(module))
+            import_dicts.extend([get_import_dict(module) for module in sorted(rest)])
```
masenf, line 126: if we're micro-optimizing, then you can ditch the list comprehension and just use a generator expression 
```suggestion
            import_dicts.extend(get_import_dict(module) for module in sorted(rest))
```

Lendemor: don't worry there is a ruff rules for those I'll add later anyway 🤣

masenf (approved):

PR #4470, opened 2024-12-03
[ENG-2157] [Refix] Allow `rx.download` to resolve `rx.get_upload_url`

### LINKED ISSUE #2812: [REF-2157] rx.download doesn't work with rx.get_upload_url from a backend event handler
**Describe the bug**
I want to write a file to a path obtained via `rx.get_upload_dir()` and then cause the user to download it via `rx.download`, but it doesn't work.

**To Reproduce**
```python
import reflex as rx

class State(rx.State):
    button_text: str = "Download"
    def do_download(self):
        filename = "foo.txt"
        (rx.get_upload_dir() / filename).write_text("test")
        return rx.download(rx.get_upload_url(filename))

@rx.page()
def index():
    return rx.button(State.button_text, on_click=State.do_download)

app = rx.App()
```

**Expected behavior**
Download the file that we write in the event handler

Instead, exception

```console
  File "/Users/masenf/code/reflex-dev/reflex/reflex/state.py", line 1439, in _process_event
    events = fn(**payload)
             ^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/repro-download-from-upload-dir/repro_download_from_upload_dir/repro_download_from_upload_dir.py", line 8, in do_download
    return rx.download(rx.get_upload_url(filename))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/reflex/reflex/event.py", line 589, in download
    raise ValueError("The URL argument should start with a /")
ValueError: The URL argument should start with a /
```

 - Python Version: 3.11.4
 - Reflex Version: 0.4.4a1
 - OS: macOS
 - Browser (Optional):
 
**Additional context**
[Add any other context about the problem here.](https://discord.com/channels/1029853095527727165/1061874061250150441/1215679066170982472)

<sub>[REF-2157](https://linear.app/reflex-dev/issue/REF-2157/rxdownload-doesnt-work-with-rxget-upload-url-from-a-backend-event)</sub>

masenf:
Update the special case in a way that works with the new Var system and its unstoppable determination to concatenate strings instead of using template string

This has been broken since 0.6.0, so a regression test was added to avoid future inadvertant breakage.

Fix #2812 (again)

## Break it?

```python
import reflex as rx


class State(rx.State):
    x: str

    def do_download(self):
        return rx.download(rx.get_upload_url(self.x))


def index() -> rx.Component:
    return rx.container(
        rx.vstack(
            rx.heading("Welcome to XSSflex!", size="9"),
            rx.form(
                rx.input(
                    rx.input.slot(
                        rx.icon_button(
                            "x",
                            size="1",
                            type="button",
                            on_click=State.set_x(""),
                            variant="ghost",
                        ),
                    ),
                    placeholder="A filename to download?",
                    value=State.x,
                    on_change=State.set_x,
                ),
                rx.button("Download"),
                on_submit=State.do_download(),
            ),
        ),
        rx.logo(),
    )


app = rx.App()
app.add_page(index)
```


Lendemor (approved):

masenf:
actually i just realized we can't merge this as is; since it's maybe creating a security problem for users that take whatever filename is given in the upload and potentially introduce script injection

masenf:
actually, maybe it's okay... @adhami3310 any idea about the script injection with this?

i tried the following 

```python
return rx.download(rx.get_upload_url('`${window.alert("foo")}`'))
```
```python
return rx.download(rx.get_upload_url('"; window.alert("foo"); "'))
```

and both of those had proper escapes around the danger characters

adhami3310 (approved):

PR #4473, opened 2024-12-04
use own cached templates release

adhami3310:
https://templates-releases.reflex.dev/ simply does the same request as https://api.github.com/repos/reflex-dev/templates/releases but doesn't have rate limiting, if the request fails in: https://templates-releases.reflex.dev/ it simply redirects to https://api.github.com/repos/reflex-dev/templates/releases in case something happens anyways, we have a try except here to use https://api.github.com/repos/reflex-dev/templates/releases as a fallback


masenf (commented):
what is running here https://templates-releases.reflex.dev/?

adhami3310:
```js
const GITHUB_RELEASES = "https://api.github.com/repos/reflex-dev/templates/releases";

export default {
  async fetch(request, env, ctx) {
    const cache = await caches.default;

    const cacheResponse = await cache.match(GITHUB_RELEASES);
    if (cacheResponse !== undefined) {
      return cacheResponse;
    }

    const githubResponse = await fetch(GITHUB_RELEASES, {
      method: 'GET',
      headers: request.headers
    });

    const response = new Response(githubResponse.body, githubResponse);

    if (response.ok) {
      response.headers.append("Cache-Control", "s-maxage=" + (60 * 30).toString());
      await cache.put(GITHUB_RELEASES, response.clone());
      return response;
    }
    
    return Response.redirect(GITHUB_RELEASES, 301);
  },
};
```

adhami3310:
@masenf do we want to bring this in? the pro would be that we don't run into rate limiting for unauthenticated API requests, the disadvantage is that we might have an hour of delay and another component to maintain, it would also be helpful in case the API goes down but the releases are still accessible somehow

masenf:
> do we want to bring this in?

not for 0.6.8

i like the idea, but before bringing it in i would want to see the service code in a repo somewhere that (hopefully) automatically deploys on push to `main` (at the very least we need deploy instructions that anyone can follow). Down the road it might make sense to cache the actual template archives as well, or maybe even push them to one of our r2/s3 buckets and not rely on github.

Issue #4262, opened 2024-10-29
Wrapping some extra components inside of the Context Menu Component


tgberkeley:
For the context menu we have not wrapped [RadioGroup](https://www.radix-ui.com/#radiogroup) [RadioItem](https://www.radix-ui.com/#radioitem) , [CheckboxItem](https://www.radix-ui.com/#checkboxitem), group  and label components. 

We wrap the context menu here in the code: https://github.com/reflex-dev/reflex/blob/main/reflex/components/radix/themes/components/context_menu.py and we want to add these other sub-components mentioned above with all their props and event handlers.

LucasOcampos:
Hey @tgberkeley! I would love to contribute to Reflex, the link that the linear bot posted is not opening up for me though.  Is this task still available?

Lendemor:
The link posted by the linear bot doesn't have additional info, it's just there for tracking in our private linear.
Since the task is still open, it's available yes.

LucasOcampos:
Awesome, if you could set me as assignee for this task please, I'd appreciate it!

KarinaNi:
Hi @tgberkeley  and @Lendemor if possible I would love to be assigned to this issue. I have got the project locally working and have an idea of how to wrap these components. If this is available, please assign it to me.

Lendemor:
Updated the initial post with a checklist for the missing components.
@LucasOcampos @vydpnguyen @KarinaNi feel free to pick up any of the components in there and submit them.

The PR for the checkbox seems to works, we just need the other now 👍

KarinaNi:
Hi,
I can do radiogroup and radio item.

vydpnguyen:
Hi @KarinaNi, how are the RadioGroup and RadioItem going for you? I've finished the CheckBoxItem part.

KarinaNi:
Hi, I shall try to finish it by today. I added all items so ill just commit the radio ones so give me a bit of time so i finish that

KarinaNi:
update: finished making the changes in code. Trying to get all success in the tests 
```
poetry run ruff check .
poetry run pyright reflex tests
find reflex tests -name "*.py" -not -path reflex/reflex.py | xargs poetry run darglint
```

Lendemor:
> update: finished making the changes in code. Trying to get all success in the tests
> 
> ```
> poetry run ruff check .
> poetry run pyright reflex tests
> find reflex tests -name "*.py" -not -path reflex/reflex.py | xargs poetry run darglint
[...]

PR #4479, opened 2024-12-04
Wrap Checkbox component in Context Menu (partial fix for #4262)

vydpnguyen:
### All Submissions:

### Type of change

### New Feature Submission:

    b. Describe your changes.
In `context_menu.py`, I added a new `ContextMenuCheckbox` class and initialized `checkbox`. Users can now use the checkbox like this:

checkbox-context-menu

This pull request only fixes part of issue #4262.

### DIFF
--- a/reflex/components/radix/themes/components/context_menu.py
+++ b/reflex/components/radix/themes/components/context_menu.py
@@ -2,14 +2,13 @@
 
 from typing import Dict, List, Literal, Union
 
-from reflex.components.component import ComponentNamespace
+from reflex.components.component import Component, ComponentNamespace
 from reflex.components.core.breakpoints import Responsive
 from reflex.event import EventHandler, no_args_event_spec, passthrough_event_spec
 from reflex.vars.base import LiteralVar, Var
-from reflex.components.component import Component
-from .checkbox import HighLevelCheckbox
 
 from ..base import LiteralAccentColor, RadixThemesComponent
+from .checkbox import HighLevelCheckbox
 
 LiteralDirType = Literal["ltr", "rtl"]
 
@@ -29,6 +28,7 @@ LiteralStickyType = Literal[
 LiteralCheckboxSize = Literal["1", "2", "3"]
 
 LiteralCheckboxVariant = Literal["classic", "surface", "soft"]
+
 
 class ContextMenuRoot(RadixThemesComponent):
     """Menu representing a set of actions, displayed at the origin of a pointer right-click or long-press."""
@@ -236,11 +236,12 @@ class ContextMenuSeparator(RadixThemesComponent):
 
     tag = "ContextMenu.Separator"
 
+
 class ContextMenuCheckbox(RadixThemesComponent):
     """The component that contains the checkbox."""
-    
+
     tag = "ContextMenu.Checkbox"
-    
+
     def create(cls, text: Var[str] = LiteralVar.create(""), **props) -> Component:
         """Create a checkbox with a label.
 
@@ -252,6 +253,7 @@ class ContextMenuCheckbox(RadixThemesComponent):
             The checkbox component with a label.
         """
         return HighLevelCheckbox.create(cls, **props)
+
 
 class ContextMenu(ComponentNamespace):
     """Menu representing a set of actions, diplayed at the origin of a pointer right-click or long-press."""
@@ -265,5 +267,6 @@ class ContextMenu(ComponentNamespace):
     item = staticmethod(ContextMenuItem.create)
     separator = staticmethod(ContextMenuSeparator.create)
     checkbox = staticmethod(ContextMenuCheckbox.create)
+
 
 context_menu = ContextMenu()
--- a/reflex/components/radix/themes/components/context_menu.py
+++ b/reflex/components/radix/themes/components/context_menu.py
@@ -231,6 +231,24 @@ class ContextMenuSeparator(RadixThemesComponent):
 
     tag = "ContextMenu.Separator"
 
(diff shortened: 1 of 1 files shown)


reflex/components/radix/themes/components/context_menu.py
```diff
@@ -232,8 +233,26 @@ class ContextMenuSeparator(RadixThemesComponent):
...
+
+        Returns:
+            The checkbox component with a label.
+        """
+        return HighLevelCheckbox.create(cls, **props)
+
+
 class ContextMenu(ComponentNamespace):
-    """Menu representing a set of actions, displayed at the origin of a pointer right-click or long-press."""
+    """Menu representing a set of actions, diplayed at the origin of a pointer right-click or long-press."""
```
Lendemor, line 255: This comment was correct in the first place, no need to remove the `s`.

reflex/components/radix/themes/components/context_menu.py
```diff
@@ -232,8 +233,26 @@ class ContextMenuSeparator(RadixThemesComponent):
     tag = "ContextMenu.Separator"
 
 
+class ContextMenuCheckbox(RadixThemesComponent):
```
Lendemor, line 236: I think if this inherit from HighLevelCheckbox, we won't need to redefined the `create` method, and the docs will have all the props of the checkbox when we generate them.

reflex/components/radix/themes/components/context_menu.py
```diff
@@ -232,8 +233,26 @@ class ContextMenuSeparator(RadixThemesComponent):
...
+    """The component that contains the checkbox."""
+
+    tag = "ContextMenu.Checkbox"
+
+    def create(cls, text: Var[str] = LiteralVar.create(""), **props) -> Component:
```
Lendemor, line 241: The text arg seems unused inside the method.

Lendemor (changes requested):

reflex/components/radix/themes/components/context_menu.py
vydpnguyen: I tried leaving it out, but Reflex doesn't recognize the `on_change` event trigger even though it was already defined in `checkbox.py`. I also saw how the `separator` was used in `context_menu.py`, but didn't see its imports. Could you give me some insights?

Lendemor: If you define it as `class ContextMenuCheckbox(RadixThemesComponent)`:
it means -> `on_change` is not known because you haven't defined it and `RadixThemesComponent` doesn't define it either

If you write as `class ContextMenuCheckbox(HighLevelCheckbox)`
it means -> `on_change` should be known because it is defined in `HighLevelCheckbox`

vydpnguyen: Thank you!

reflex/components/radix/themes/components/context_menu.py
```diff
@@ -232,6 +233,12 @@ class ContextMenuSeparator(RadixThemesComponent):
     tag = "ContextMenu.Separator"
 
 
+class ContextMenuCheckbox(HighLevelCheckbox):
+    """The component that contains the checkbox."""
+
+    tag = "ContextMenu.Checkbox"
+
```
Lendemor, line 240: I think here we want the props `color` and `shortcut` as defined in the docs of radix UI.

Lendemor (changes requested):

reflex/components/radix/themes/components/context_menu.py
Lendemor: Did some testing on my side, and the `shortcut` doesn't seems to work, so let's just go for the `color_scheme` for now.

reflex/components/radix/themes/components/context_menu.py
```diff
@@ -8,6 +8,7 @@
 from reflex.vars.base import Var
 
 from ..base import LiteralAccentColor, RadixThemesComponent
+from .checkbox import HighLevelCheckbox
```
Lendemor, line 11:
```suggestion
from .checkbox import Checkbox
```

reflex/components/radix/themes/components/context_menu.py
```diff
@@ -232,6 +233,12 @@ class ContextMenuSeparator(RadixThemesComponent):
     tag = "ContextMenu.Separator"
 
 
+class ContextMenuCheckbox(HighLevelCheckbox):
```
Lendemor, line 236:
```suggestion
class ContextMenuCheckbox(Checkbox):
```

reflex/components/radix/themes/components/context_menu.py
```diff
@@ -232,6 +233,12 @@ class ContextMenuSeparator(RadixThemesComponent):
...
+    """The component that contains the checkbox."""
+
+    tag = "ContextMenu.Checkbox"
```
Lendemor, line 239:
```suggestion
    tag = "ContextMenu.CheckboxItem"
```

reflex/components/radix/themes/components/context_menu.py
Lendemor, line 240:
```suggestion
    # Text to render as shortcut.
    shortcut: Var[str]
```

Lendemor (changes requested):
After running it again locally, I have some more changes to suggest.

You should also run this : `py scripts/make_pyi.py reflex/components/radix/themes/components/context_menu.py` before doing the commit.

Run `pre-commit install` if you want the command above to check automatically.

Lendemor (approved):

PR #4482, opened 2024-12-04
fix: multiple mismatched-type-assignment fixes

benedikt-bartscher:
mismatched-type-assignment check did not strip rx.Field annotations and added a second optional for annotations which were already optional


benedikt-bartscher:
@adhami3310 iirc you implemented `rx.Field`, could you maybe review this for me?

reflex/state.py
```diff
@@ -278,6 +279,22 @@ def __call__(self, *args: Any) -> EventSpec:
     from pydantic.v1.fields import ModelField
 
 
+def unwrap_field_type(type_: Type) -> Type:
```
Lendemor, line 282: Does this need to be a public method?
if not, maybe prefix it with `_`

benedikt-bartscher: No it does not, i migrated it :heavy_check_mark:

masenf (approved):

PR #4483, opened 2024-12-04
feat: add state_explicit_vars config option

benedikt-bartscher:


masenf:

In my State refactor work, only fields annotated with `rx.Field` will become base vars. In the old `rx.State` it will continue to function as it does today.

Issue #4487, opened 2024-12-05
Do not mark  dependent vars if value did not change


benedikt-bartscher:

adhami3310:
one thing we can do here is handle the simple cases of `int`, `str`, and other primitive types. It seems doing this in general is difficult.

PR #4488, opened 2024-12-05
wip #4487

benedikt-bartscher:


masenf (commented):
one potentially issue i see with this approach is it makes it harder to manually mark a var as dirty, for example if internal change tracking is not working for a certain kind of object, a hackaround was always to assign it to itself to signal to the framework that it should be included in the delta. i know @adhami3310 uses this trick quite a bit.

benedikt-bartscher:
@masenf I knew this comment was coming :laughing:  

However, I have 2 arguments against it:
- `self.var = self.var` is not really descriptive and needs a comment to explain what it's doing. `self.dirty_vars.add("var"); self._mark_dirty()` is pretty descriptive and doesn't need further explanation
- Currently, there is no way for users to stop reflex from marking all vars which depend on an unchanged computed var as dirty, even if the computed var did not change (`_mark_dirty_computed_vars`). Maybe I can adjust this PR to only change the computed var behavior?

benedikt-bartscher:
I just refactored one small repo to always check if the var has changed before assignment. Now I have a diff of 85 lines with repetitive 
```python
if self.var != new_var:
  self.var = new_var`
```

adhami3310:
haven't looked into this too deeply but `self.var = self.var` is only used to let reflex know about a difference that internal mutability proxy didn't cache

benedikt-bartscher:
current implementation is not optimal, i have a better one coming
BTW, this PR is now only about unchanged computed vars. I will split #4487 into multiple PRs

benedikt-bartscher:
need to recheck order of computed var computation

benedikt-bartscher:
> need to recheck order of computed var computation

It changed due to priming of the caches. I disabled it for now in https://github.com/reflex-dev/reflex/pull/4488/commits/cc6edf164a1ac31124c6b9ab07ecb25361af4730

This broke some of our apps without failing reflex tests. Maybe we can add a test for it (order and count of computed var execution)

benedikt-bartscher:
@masenf can you test if your model recursion issue also occurs with a computed var in this PR?

masenf:

With the State refactoring, dirty var status will be stored in an "EventContext" rather than in the class itself. This should make it easier to handle checking whether a variable is actually dirty and should be recalculated. Since the underlying code is going through quite a bit of change, will be closing this one for now, but will incorporate the ideas to reduce delta size where possible.

PR #4491, opened 2024-12-06
fix router var special case

benedikt-bartscher:
@masenf and I fixed some var dependency issues a while ago. Seems like this case is only needed for the root/main state


benedikt-bartscher:
seems like this case is only needed with redis

PR #4496, opened 2024-12-06
test dynamic route flakiness (can't reproduce locally)

benedikt-bartscher:


benedikt-bartscher:
Temporal mitigation until https://github.com/reflex-dev/reflex/pull/4446 is ready
CI ran 2 times without failure

Lendemor (approved):
Thanks, #4446 is blocked until playwright solve the issue on redirect upstream I think.

So this will do until then 👍

PR #4500, opened 2024-12-07
Minor pickle improvement 2

benedikt-bartscher:
based on #4499

### DIFF
--- a/tests/units/test_state.py
+++ b/tests/units/test_state.py
@@ -2513,10 +2513,7 @@ def test_mutable_copy(mutable_state: MutableTestState, copy_func: Callable):
         assert getattr(ms_copy, attr) is not getattr(mutable_state, attr)
     ms_copy.custom.array.append(42)
     assert "custom" in ms_copy.dirty_vars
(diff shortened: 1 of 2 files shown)


tests/units/test_state.py
```diff
@@ -2513,10 +2513,7 @@ def test_mutable_copy(mutable_state: MutableTestState, copy_func: Callable):
         assert getattr(ms_copy, attr) is not getattr(mutable_state, attr)
     ms_copy.custom.array.append(42)
     assert "custom" in ms_copy.dirty_vars
-    if copy_func is copy.copy:
```
benedikt-bartscher, line 2516: i can implement `__copy__` for `BaseState` if you want to keep this behavior

masenf: i dont think anything is depending on this behavior, so we're fine to simplify the test logic -- famous last words 🫣

masenf:

We actually _do_ want to persist these values in case the state is serialized to redis, but a delta wasn't sent before hand

PR #4505, opened 2024-12-09
deprecate add_custom_404_page

Lendemor:


reflex/app.py
```diff
@@ -633,10 +649,14 @@ def add_custom_404_page(
             on_load: The event handler(s) that will be called each time the page load.
             meta: The metadata of the page.
         """
-        if component is None:
-            component = Default404Page.create()
+        console.deprecate(
+            feature_name="App.add_custom_404_page",
+            reason="Use App.add_page(component, route=rx.constants.Page404.SLUG) instead.",
```
masenf, line 654:
```suggestion
            reason=f"Use app.add_page(component, route={rx.constants.Page404.SLUG!r}) instead.",
```

i dont think we should have the deprecation message show this kind of weird undocumented constant. the value is always going to be `/404`, so lets just show that

reflex/app.py
```diff
@@ -633,10 +649,14 @@ def add_custom_404_page(
             on_load: The event handler(s) that will be called each time the page load.
             meta: The metadata of the page.
         """
-        if component is None:
-            component = Default404Page.create()
+        console.deprecate(
+            feature_name="App.add_custom_404_page",
+            reason=f"Use app.add_page(component, route=/{constants.Page404.SLUG!r}) instead.",
```
masenf, line 654: this puts the slash outside the quotes

```
"Use app.add_page(component, route=/'404') instead."
```

either leave the leading slash off (it works just the same), or do the quoting yourself (without the `!r`)

```suggestion
            reason=f"Use app.add_page(component, route='/{constants.Page404.SLUG}') instead.",
```

masenf (approved):

masenf (approved):

PR #4506, opened 2024-12-09
add __repr__ method to MutableProxy

benedikt-bartscher:
Needed for debugging, otherwise if you print a mutable value inside an event handler you will get output which is not useful at all.


Lendemor (approved):
Seems fine, can you edit the description with a short comment on usecase where it's needed?

benedikt-bartscher:
@Lendemor it thought it is pretty obvious, done

PR #4507, opened 2024-12-10
[ENG-4100]reduce lock expiration time to 1s

ElijahAhianyo:
Reduce the Redis lock expiration time to 1s so users are guaranteed to get the lock expiration error when they run processes that block the main thread.

### DIFF
--- a/reflex/constants/config.py
+++ b/reflex/constants/config.py
@@ -26,7 +26,7 @@ class Expiration(SimpleNamespace):
     # Token expiration time in seconds
     TOKEN = 60 * 60
     # Maximum time in milliseconds that a state can be locked for exclusive access.
-    LOCK = 10000
+    LOCK = 1000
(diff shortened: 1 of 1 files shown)


benedikt-bartscher:
this seems to have introduce ci flakiness

`FAILED tests/units/test_state.py::test_state_manager_contend[redis] - reflex.utils.exceptions.LockExpiredError: Lock expired for token 2e09a593-fbff-4c74-ba16-9c15738747ea_tests___units___test_state____test_state.tests___units___test_state____child_state2 while processing. Consider increasing `app.state_manager.lock_expiration` (currently 1000) or use `@rx.event(background=True)` decorator for long-running tasks.
`

https://github.com/reflex-dev/reflex/actions/runs/12261807689/job/34209663455?pr=4441

maybe we should increase the lock time for CI?

PR #4518, opened 2024-12-10
add deps and position field in VarData

Lendemor:
When rendering hooks, some use case are not covered with the current hooks API.

This PR add two fields, `deps` and `position` to VarData that allow some controls on how to render hooks, both in relation to the `useCallback` generated by StatefulComponent.

Problem 1:
- if you are writing a signature spec for an event, and this one depend on a hook you wrote in `add_hooks`, when wrapped in useCallback, your event handler will not update its data if the `hook_name` is not added to the dependency array.
**`Solution: `** Passing the name of the hook in the new `deps` props.

Problem 2:
- When adding a hook, you might either want:
    - to reference your hook inside an event handler (as a `deps` for example), so the hook need to be defined BEFORE.
    - to reference a memoized event handler inside your hook, so the hook need to be defined AFTER. (see clipboard.py)
**`Solution: `** Inside a hook, pass either `HooksPosition.PRE_TRIGGER` or `HooksPosition.POST_TRIGGER` to the `position` props.


reflex/components/component.py
```diff
@@ -1532,17 +1535,15 @@ def _get_added_hooks(self) -> dict[str, ImportDict]:
...
 
         def extract_var_hooks(hook: Var):
-            _imports = {}
             var_data = VarData.merge(hook._get_all_var_data())
             if var_data is not None:
                 for sub_hook in var_data.hooks:
-                    code[sub_hook] = {}
-                if var_data.imports:
-                    _imports = var_data.imports
+                    code[sub_hook] = None
```
masenf, line 1541: how does this pass pyright when the return value is annotated as `dict[str, VarData]`?

maybe this should be `dict[str, VarData | None]` and then on like 1555 you can set `code[hook] = None` as well

reflex/vars/base.py
```diff
@@ -184,12 +194,29 @@ def merge(*all: VarData | None) -> VarData | None:
...
-        if state or _imports or hooks or field_name:
+        positions = list(
+            {
+                var_data.position
+                for var_data in all_var_datas
+                if var_data.position is not None
+            }
+        )
+        if positions:
+            if len(positions) > 1:
```
masenf, line 205: wondering if this should be `if len(set(positions)) > 1`, since you're attempting to check for unique positions being merged

reflex/.templates/jinja/web/pages/stateful_component.js.jinja2
```diff
@@ -5,11 +5,15 @@ export function {{tag_name}} () {
...
+  {% for hook, data in component._get_all_hooks().items() if not data.position or data.position == positions.PRE_TRIGGER %}
+  {{ hook }}
+  {% endfor %}
+
   {% for hook in memo_trigger_hooks %}
   {{ hook }}
   {% endfor %}
 
-  {% for hook in component._get_all_hooks() %}
+  {% for hook,data in component._get_all_hooks().items() if data.position and data.position == positions.POST_TRIGGER %}
```
masenf, line 16:
```suggestion
  {% for hook, data in component._get_all_hooks().items() if data.position and data.position == positions.POST_TRIGGER %}
```

format nit

reflex/components/component.py
Lendemor: Maybe because that line is inside an internal function `extract_var_hooks`, it might trick pyright ? not sure on that one.
Made the change anyway.

reflex/components/component.py
```diff
@@ -2168,6 +2169,7 @@ def _render_stateful_code(
             tag_name=tag_name,
             memo_trigger_hooks=memo_trigger_hooks,
             component=component,
+            positions=constants.Hooks.HookPosition,
```
masenf, line 2172: this is kind of hacky. constants should be added to the global `consts` namespace in `reflex/compiler/template.py`

masenf (dismissed):
i'm overall not crazy about packing more stuff in VarData because these fields do seem to have a pretty niche use in the framework... but i don't have an alternate suggestion that's better, so i think we can be pragmatic and take this.

Lendemor:
> i'm overall not crazy about packing more stuff in VarData because these fields do seem to have a pretty niche use in the framework... but i don't have an alternate suggestion that's better, so i think we can be pragmatic and take this.

I sort of agree on this, that's why at first I tried to pass the position to a HookVar with special logic, but in the end passing them through VarData made code much cleaner and easier to read.

Also even if they are niche, they won't impact the majority of people who don't need them so it should be fine.

reflex/compiler/templates.py
```diff
@@ -45,6 +45,7 @@ def __init__(self) -> None:
             "on_load_internal": constants.CompileVars.ON_LOAD_INTERNAL,
             "update_vars_internal": constants.CompileVars.UPDATE_VARS_INTERNAL,
             "frontend_exception_state": constants.CompileVars.FRONTEND_EXCEPTION_STATE_FULL,
+            "hook_positions": constants.Hooks.HookPosition,
```
masenf, line 48: seems like for consistency, this should be 
```suggestion
            "hook_position": constants.Hooks.HookPosition,
```
that way it matches the enum class plurality

masenf (approved):

PR #4519, opened 2024-12-11
allow to declare deps in event signature for memoized event triggers

Lendemor:


masenf (commented):
what is the use case here?

can you add some kind of test (probably integration) that would fail without this feature?

Lendemor:
> what is the use case here?
> 
> can you add some kind of test (probably integration) that would fail without this feature?

I'll merge this with #4518 since I have an example that need both.

PR #4521, opened 2024-12-11
[ENG-4153]Use empty string for None values in `rx.input` and `rx.el.input`

ElijahAhianyo:
React expects empty strings instead of `null` values for the `value` prop of the input component. putting this in a cond allows the input field to be updated when the `value` is `None`

### DIFF
--- a/reflex/components/el/elements/forms.py
+++ b/reflex/components/el/elements/forms.py
@@ -7,7 +7,6 @@ from typing import Any, Dict, Iterator, Set, Tuple, Union
 
 from jinja2 import Environment
 
-from reflex.components.core.cond import cond
 from reflex.components.el.element import Element
 from reflex.components.tags.tag import Tag
 from reflex.constants import Dirs, EventTriggers
@@ -400,7 +399,7 @@ class Input(BaseHTML):
 
         # React expects an empty string(instead of null) for uncontrolled inputs.
         if value is not None:
-            props["value"] = cond(value, value, "")
+            props["value"] = Var(_js_expr=f"{value} ?? ''", _var_type=str)
         return super().create(*children, **props)
 
 
--- a/reflex/components/el/elements/forms.py
+++ b/reflex/components/el/elements/forms.py
@@ -400,7 +400,7 @@ class Input(BaseHTML):
 
         value = props.get("value")
 
-        # React expects an empty string(instead of null) for uncontrolled inputs.
+        # React expects an empty string(instead of null) for controlled inputs.
         if value is not None and is_optional(
             (value_var := Var.create(value))._var_type
         ):
--- a/reflex/components/el/elements/forms.py
+++ b/reflex/components/el/elements/forms.py
@@ -395,11 +395,18 @@ class Input(BaseHTML):
         Returns:
             The component.
         """
+        from reflex.vars.number import ternary_operation
+
         value = props.get("value")
 
         # React expects an empty string(instead of null) for uncontrolled inputs.
         if value is not None:
-            props["value"] = Var(_js_expr=f"{value} ?? ''", _var_type=str)
+            props["value"] = ternary_operation(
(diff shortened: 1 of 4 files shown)


benedikt-bartscher:
Closes https://github.com/reflex-dev/reflex/issues/4445  

Thanks, @ElijahAhianyo!

I would love to have an integration test for the use case mentioned in my issue. I can add it if you want

reflex/components/el/elements/forms.py
```diff
@@ -384,6 +384,31 @@ class Input(BaseHTML):
...
+        Returns:
+            The component.
+        """
+        from reflex.vars.number import ternary_operation
+
+        value = props.get("value")
+
+        # React expects an empty string(instead of null) for uncontrolled inputs.
+        if value is not None:
+            props["value"] = ternary_operation(
```
masenf, line 404: It seems like adding a ternary to every `value` prop on every text input is a bit excessive. maybe we can check `isoptional(value._var_type)` and apply the ternary in that case.

This does potentially create an issue where the Var is not typed optional, but still ends up with a `None` value, however the fix for that is to improve the annotation of the Var.

I know it's probably negligible and maybe not worth optimizing for, but perhaps we should profile text field re-render time with and without a ternary in the `value`?

benedikt-bartscher: > It seems like adding a ternary to every value prop on every text input is a bit excessive. maybe we can check isoptional(value._var_type) and apply the ternary in that case.

This would be great and keep the generated code simpler and faster for values which will never be None

> This does potentially create an issue where the Var is not typed optional, but still ends up with a None value, however the fix for that is to improve the annotation of the Var.

I agree, we should not bother too much about wrongly annotated vars.

> I know it's probably negligible and maybe not worth optimizing for, but perhaps we should profile text field re-render time with and without a ternary in the value?

While profiling is a great idea, I think it's a bit overkill if we only do this for optional values. However, I do not want to stop anyone from doing that :D

ElijahAhianyo:
> Closes #4445
> 
> Thanks, @ElijahAhianyo!
> 
> I would love to have an integration test for the use case mentioned in my issue. I can add it if you want

yeah sure, feel free to add to this :), I'm currently looking into the other comments

ElijahAhianyo:
For now, the code below should work:

```python
import reflex as rx
import random


class State(rx.State):
    name: rx.Field[str | None] = "Hello world"


def index() -> rx.Component:
    return rx.container(
        rx.button("Clear", on_click=lambda: State.set_name(None)),
        rx.text("rx.input"),
        rx.input(value=State.name, on_change=State.set_name),
        rx.text("rx.el.input"),
        rx.el.input(value=State.name),
    )


app = rx.App()
app.add_page(index)
```

However, we'll need this https://github.com/reflex-dev/reflex/pull/4528 to work without `rx.Field`:

```python

class State(rx.State):
    name: str | None = "Hello world"
  # name: Optional[str] = "Hello world"
    
```

reflex/components/el/elements/forms.py
```diff
@@ -384,6 +385,33 @@ class Input(BaseHTML):
...
+            **props: The properties of the component.
+
+        Returns:
+            The component.
+        """
+        from reflex.vars.number import ternary_operation
+
+        value = props.get("value")
+
+        # React expects an empty string(instead of null) for uncontrolled inputs.
```
masenf, line 403: nit: aren't these controlled inputs when a value is passed

masenf (approved):

reflex/components/el/elements/forms.py
ElijahAhianyo: ah yeah, good catch

masenf (approved):

benedikt-bartscher:
Integration test is here: https://github.com/reflex-dev/reflex/pull/4448

PR #4536, opened 2024-12-13
enable C4 rule

Lendemor:


masenf (approved):
i always forget that `()` is empty tuple. i think i get tripped up because `(42)` is definitely not a 1-tuple.

personally i like the `dict(...)` format when i'm writing kwargs because they _look_ more like kwargs passed to a function call, but i'm not going to die on that hill. lets lint and standardize on literals

Issue #4481, opened 2024-12-04
Show password option for password input


dentroai:
**Describe the Custom Component**
there should be an inbuilt option of rx.input with type=password to display some eye which shows the password in plaintext when clicked on.

E.g.:

![image](https://github.com/user-attachments/assets/d401a3f4-c341-454c-9a35-9005c06ac30a)

then:

![image](https://github.com/user-attachments/assets/672cc5b1-5fb2-4a69-b887-b2856df3699d)

ANIR1604:
I want to work on this feature, if it's ok

Lendemor:
Sure, i assigned you to it.
Feel free to reach out to us here or on the discord if you want/need to discuss on it.

Rustix1604:
Can I work on this, plz assign it to me.

masenf:
old bug, lots of people are going to have different types of solutions for this, but here's a pointer for future wanderers

```python
def revealable_password(**props):
    revealed = rx._x.client_state(default=False, global_ref=False)
    return rx.input(
        rx.input.slot(
            rx.cond(
                revealed.value,
                rx.icon("eye_off", on_click=revealed.set_value(False)),
                rx.icon("eye", on_click=revealed.set_value(True)),
            ),
            side="right",
            cursor="pointer",
        ),
        type=rx.cond(revealed.value, "text", "password"),
        **props,
    )
```

likely not bringing this into the framework until we're maintaining our own component library

PR #4541, opened 2024-12-14
Feature : Show Password Update

Rustix69:
Fixes : #4481


Rustix69:
@Lendemor Sir this approach right?

PR #4549, opened 2024-12-17
use position in vardata to mark internal hooks

### LINKED ISSUE #4516: Conflict between _get_all_hooks and _get_all_hooks_internal
**Describe the bug**
Hook code snippets that are returned by both `_get_all_hooks` and `_get_all_hooks_internal` are not de-duplicated, which can lead to rendering problems when `const` declarations are re-defined.

**To Reproduce**
```python
import reflex as rx


class State(rx.State):
    """The app state."""
    v: str = "v"


class Foo(rx.Fragment):
    bar: rx.Var[str]

    def add_hooks(self) -> list[str]:
        return [
            rx.Var(f"console.log({self.bar})")
        ]

    def render(self):
        return {}


def index() -> rx.Component:
    return Foo.create(bar=State.v)


app = rx.App()
app.add_page(index)
```

**Expected behavior**
All hooks code should always be de-duplicated.

 - Python Version: 3.12
 - Reflex Version: 0.6.7.dev1
 - OS: macOS
 - Browser (Optional): chrome
 
**Additional context**
Care should be taken in regard to hook ordering (internal should always come first) and avoiding extra recursive calls (don't call `_get_all_hooks_internal` inside `_get_all_hooks` to avoid performance implications)

Lendemor:
Fixes #4516

~~Keeping as draft to figure a better way to collect hooks.~~

### DIFF
--- a/reflex/.templates/jinja/web/pages/_app.js.jinja2
+++ b/reflex/.templates/jinja/web/pages/_app.js.jinja2
@@ -1,4 +1,5 @@
 {% extends "web/pages/base_page.js.jinja2" %}
+{% from "web/pages/macros.js.jinja2" import renderHooks %}
 
 {% block early_imports %}
 import '$/styles/styles.css'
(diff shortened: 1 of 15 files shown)


masenf (approved):
pretty cool tbh

masenf (approved):

PR #4556, opened 2024-12-18
[ENG-4134]Allow specifying custom app module in rxconfig

ElijahAhianyo:
```python 

import sys
import reflex as rx
from pathlib import Path


sys.path.insert(0, str(Path("/path/to/app/foo/")))

config = rx.Config(
    app_name="my_project",
    app_module_path="foo.foo"
)
```

Assuming you have a reflex app `foo` with App instance living in `foo/foo/foo.py`

Example app: https://github.com/reflex-dev/reflex-examples/pull/291

### DIFF
--- a/reflex/config.py
+++ b/reflex/config.py
@@ -688,11 +688,11 @@ class Config(Base):
     # Maximum expiration lock time for redis state manager
     redis_lock_expiration: int = constants.Expiration.LOCK
 
-    # Token expiration time for redis state manager
-    redis_token_expiration: int = constants.Expiration.TOKEN
-
     # Maximum lock time before warning for redis state manager.
     redis_lock_warning_threshold: int = constants.Expiration.LOCK_WARNING_THRESHOLD
+
+    # Token expiration time for redis state manager
+    redis_token_expiration: int = constants.Expiration.TOKEN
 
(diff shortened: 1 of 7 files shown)


reflex/config.py
```diff
@@ -726,13 +727,34 @@ def __init__(self, *args, **kwargs):
...
+
     @property
     def module(self) -> str:
         """Get the module name of the app.
 
         Returns:
             The module name.
         """
+        if self.app_module:
+            return f"{Path(self.app_module.__file__).parent.name}.{Path(self.app_module.__file__).stem}"
```
ElijahAhianyo, line 757: wondering if this is what we should be doing. Or should we return the absolute path here? Im trying to think of where doing it like this could pose a real problem

ElijahAhianyo:
integration tests need https://github.com/reflex-dev/reflex-examples/pull/291 to pass

masenf:
@ElijahAhianyo i opened a PR into your branch #4644 for your review

ElijahAhianyo:
@masenf I think we need https://github.com/reflex-dev/reflex-examples/pull/291 merged to pass the CI and then it's ready to go

masenf (approved):

PR #4586, opened 2025-01-03
removing deprecated features for 0.7.0 and removing py3.9 support

Lendemor:


reflex/event.py
```diff
@@ -1420,19 +1345,15 @@ def resolve_annotation(annotations: dict[str, Any], arg_name: str):
...
     if annotation is None:
-        console.deprecate(
-            feature_name="Unannotated event handler arguments",
-            reason="Provide type annotations for event handler arguments.",
-            deprecation_version="0.6.3",
-            removal_version="0.7.0",
-        )
-        # Allow arbitrary attribute access two levels deep until removed.
-        return Dict[str, dict]
+        raise VarAnnotationError(arg_name, annotation)
```
adhami3310, line 1356: we can maintain the warning here but not cause the program to halt, what do you think @masenf ?

reflex/vars/base.py
```diff
@@ -2076,14 +2049,11 @@ def __get__(self, instance: BaseState | None, owner):
             value = getattr(instance, self._cache_attr)
 
         if not _isinstance(value, self._var_type):
-            console.deprecate(
-                "mismatched-computed-var-return",
-                f"Computed var {type(instance).__name__}.{self._js_expr} returned value of type {type(value)}, "
-                f"expected {self._var_type}. This might cause unexpected behavior.",
-                "0.6.5",
-                "0.7.0",
+            raise MismatchedComputedVarReturn(
```
adhami3310, line 2052: what if instead we do `console.error` here?

reflex/state.py
```diff
@@ -1335,12 +1337,10 @@ def __setattr__(self, name: str, value: Any):
             if field.allow_none and not is_optional(field_type):
                 field_type = Union[field_type, None]
             if not _isinstance(value, field_type):
-                console.deprecate(
-                    "mismatched-type-assignment",
-                    f"Tried to assign value {value} of type {type(value)} to field {type(self).__name__}.{name} of type {field_type}."
-                    " This might lead to unexpected behavior.",
-                    "0.6.5",
-                    "0.7.0",
+                raise MismatchedArgumentTypeError(
```
adhami3310, line 1340: same here, i think it would be better to not cause the program to fail completely but instead print warning/error level console log

.github/workflows/benchmarks.yml
```diff
@@ -81,24 +81,18 @@ jobs:
       matrix:
         # Show OS combos first in GUI
         os: [ubuntu-latest, windows-latest, macos-latest]
```
adhami3310, line 83: we happened to modify those versions since this PR, so this one reverts the change

reflex/event.py
```diff
@@ -813,61 +786,26 @@ def fn():
...
-        if is_external is None
-        else is_external
-    )
-
     return server_side(
         "_redirect",
         get_fn_signature(redirect),
         path=path,
-        external=is_external,
+        external=False if is_external is None else is_external,
```
masenf, line 808: suggest changing the signature to `is_external: bool = False` and passing this as `external=is_external` now that the compatibility logic is no longer needed.

(i guess sometime we should clean up the frontend handler arg name too 🤷)

reflex/utils/exceptions.py
```diff
@@ -95,11 +117,11 @@ class MatchTypeError(ReflexError, TypeError):
     """Raised when the return types of match cases are different."""
 
 
-class EventHandlerArgTypeMismatch(ReflexError, TypeError):
+class EventHandlerArgTypeMismatchError(ReflexError, TypeError):
```
masenf, line 120: seems this breaks any code which might have cared about this exception.

if we want to change it to add `Error` to the name, then we should still provide the old name as a subclass or alias of the new name and properly deprecate it.

that said, its a big release, so maybe it's okay

reflex/vars/base.py
```diff
@@ -583,12 +563,6 @@ def create(
         if not isinstance(value, str):
             return LiteralVar.create(value, _var_data=_var_data)
```
masenf, line 564: we should remove this too

reflex/__init__.py
```diff
@@ -367,19 +366,4 @@
 
 
 def __getattr__(name):
```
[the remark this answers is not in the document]
Lendemor: Removing this blow up all the CI. Gonna keep it for now

reflex/vars/base.py
masenf: the same exact line is just below. with the intervening fallback stuff taken out, there's no reason to create the LiteralVar early for `str` types

reflex/utils/exceptions.py
Lendemor: Will people actually make code based on this? It seems like reflex apps wouldn't run properly anyway if this is raised ?

reflex/utils/console.py
```diff
@@ -49,22 +49,7 @@ def set_log_level(log_level: LogLevel):
 
     Args:
         log_level: The log level to set.
-
-    Raises:
-        ValueError: If the log level is invalid.
     """
-    if not isinstance(log_level, LogLevel):
```
masenf, line 56: we should remove the auto-cast, but can we still raise the exception here, just to avoid the footgun slipping back in

masenf:
needs conflict resolution and address 1 comment, then i think we're good to merge 🙌

PR #4592, opened 2025-01-06
integration_tests_wsl: pin poetry==1.8.3

masenf:
poetry 2.0.0 seems to break on all platforms when doing `poetry run reflex`


masenf:
we'll just make it compatible instead

PR #4594, opened 2025-01-06
integration_app_harness.yml: cache playwright install

masenf:


masenf:
doesn't noticably improve startup time in CI

PR #4597, opened 2025-01-07
improve client state

adhami3310:


adhami3310:
this removes the need for the bare var in client state

masenf (approved):
interesting solution

PR #4602, opened 2025-01-08
make object var handle all mapping instead of just dict

adhami3310:


adhami3310:
this fixes few issues highlighted in #4380

masenf (approved):

PR #4607, opened 2025-01-08
fix duplicate tab issue

### LINKED ISSUE #4605: Duplicated tabs interfere with each other during background events
**Describe the bug**
If a new tab is opened by duplicating the first, and a background task is started in both tabs, the first (by order of event trigger) tab stops seeing updates.

**To Reproduce**

Here's the code that I show in the video clip below.
```python
import asyncio
import reflex as rx


class State(rx.State):
    state_value: str = "Initial value"

    @rx.event(background=True)
    async def do_something(self):
        async with self:
            self.state_value = "Initial change"
        await asyncio.sleep(2)
        async with self:
            self.state_value = "Final change"


@rx.page(route="/duplicate_tabs_issue", title="Duplicate tabs issue")
def index() -> rx.Component:
    return rx.container(
        rx.vstack(
            rx.markdown("Minimal example to demonstrate issues with duplicated tab in Chrome."),
            rx.markdown("`state.router.session.client_token` is not unique per tab when tab."),
            rx.text(f"state.router.session.client_token: {rx.State.router.session.client_token}"),
            rx.markdown("However, `state.router.session.session_id` does remain unique per tab."),
            rx.text(f"state.router.session.session_id: {rx.State.router.session.session_id}"),
            rx.divider(),
            rx.text("Somehow the tabs interfere with each other during background tasks."),
            rx.text(
                "Expect to see 'Initial change' followed by 'Final change' after 2 seconds. But that doesn't happen if the button is clicked in the other tab before the first event finishes."
            ),
            rx.button("Do something", on_click=State.do_something),
            rx.text(f"State.state_value: {State.state_value}"),
        )
    )

```

**Expected behavior**
Expect that the two tabs behave independently.

**Screenshots**
Short video clip to show duplication of tabs, and that the background events are interrupted in the first tab by running the event in the second tab.

https://github.com/user-attachments/assets/2bcc96bc-ca95-4601-bbba-11246a8ba422

In fact, it seems that the second tab get's the update that was intended for first tab (seen in the video by the `final_state` being set too quickly in the second tab).

 - Python Version: 3.12
 - Reflex Version: 0.6.5 -> 0.6.7
 - OS: WSL2
 - Browser (Optional): Chrome and Edge (not tried others)
 
**Additional context**

You can see that the `session_id` is unique per tab, but the `client_token` is not. I'm not sure if that is expected, but seems like it could be relevant.
Note, If the two tabs are opened separately (and don't share a `client_token` or `session_id`) then they do both behave independently.

Lendemor:
fix #4605


reflex/app.py
```diff
@@ -1528,14 +1529,18 @@ def __init__(self, namespace: str, app: App):
...
 
         Args:
             sid: The Socket.IO session id.
             environ: The request information, including HTTP headers.
         """
-        pass
+        query_string = environ.get("QUERY_STRING")
+        query_params = dict(
+            qc.split("=") for qc in query_string.split("&") if "=" in qc
+        )
```
masenf, line 1542: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.parse_qs

reflex/app.py
```diff
@@ -1610,3 +1612,17 @@ async def on_ping(self, sid):
...
+
+    async def link_token_to_sid(self, sid, token):
+        """Link a token to a session id.
+
+        Args:
+            sid: The Socket.IO session id.
+            token: The client token.
+        """
+        if token in self.sid_to_token.values() and sid != self.token_to_sid.get(token):
+            token = uuid.uuid4().hex
```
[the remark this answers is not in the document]
Lendemor: On that topic, I was considering always generating the token inside the on_connect, do you think it would be an acceptable approach?

reflex/utils/token_manager.py
```diff
@@ -0,0 +1,217 @@
...
+
+        Args:
+            token: The client token.
+            sid: The Socket.IO session ID.
+
+        Returns:
+            New token if duplicate detected and new token generated, None otherwise.
+        """
+        # Check if token is already mapped to a different SID (duplicate tab)
+        if token in self.sid_to_token.values() and sid != self.token_to_sid.get(token):
```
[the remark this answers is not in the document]
adhami3310: @Lendemor that looks reasonable

reflex/app.py
```diff
@@ -1966,17 +1962,45 @@ def __init__(self, namespace: str, app: App):
             app: The application object.
         """
         super().__init__(namespace)
-        self.token_to_sid = {}
-        self.sid_to_token = {}
         self.app = app
 
-    def on_connect(self, sid: str, environ: dict):
+        # Use TokenManager for distributed duplicate tab prevention
+        self._token_manager = TokenManager.create()
```
masenf, line 1968: consider putting the `TokenManager` on `App` as a real API into this data.

down the road we could even expose helper iterators on this new `TokenManager` to make cancelling background tasks easier, like `session_is_connected` and `token_is_connected` which would save the passed sid/token and `yield True` as long as these values were still being tracked. 

<details>
<summary>
Like so
</summary>

```python
@rx.event(background=True)
async def thing_checker(self):
    for _task_active in app.token_manager.session_is_connected(
        sid=self.router_data.session.session_id
    ):
        async with self:
            if await self._check_thing():
                await self._update_thing()
        await asyncio.sleep(CHECK_INTERVAL)
```

</details>

lot of possibilities for an asyncio.Event API for getting notified when a session or client disconnects, which people have asked for

reflex/app.py
```diff
@@ -1966,17 +1962,45 @@ def __init__(self, namespace: str, app: App):
...
+
+    @property
+    def token_to_sid(self) -> dict[str, str]:
+        """Get token to SID mapping for backward compatibility.
+
+        Returns:
+            The token to SID mapping dict.
+        """
+        # For backward compatibility, expose the underlying dict
+        return self._token_manager.token_to_sid
```
masenf, line 1978: should raise a deprecation here and below

masenf (commented):
functionality works great in my testing; a few comments from an API continuity perspective, since people actually do use this functionality and need it for background tasks and perhaps other cleanup steps

hr-alebel:
> Revisiting this PR, is there ever a case where we wouldn't get `on_disconnect` to unlink the sid/token, resulting in a refresh/reconnect leading to loss of token and state for a client?

Sry to ressurrect this PR. 

Yes there is a case where a client reconnects without firing "on_disconnect". Specifically when the Backend Process crashes. If the app is used by multiple users, one causing a crash (or restart by external controls, see containers in K8s/docker) will make the tokes/sessions stored in Redis persistent but still linked, meaning after the container restarts it causes a token duplicate, which in turn forces to create a new token (without hydration mind you) causing an inconsitent state between UI and Backend breaking the app.

Tested by:
1) having an app with for example selectors and a button to get the state from Backend
2) select a certain combination that differs from the default state
3) check state in Redis -> selection is persisted
4) restart the container running the backend
5) when the backend is back observe that in the Websocket Messages there is a new_token command
6) press button to get state -> default state from Backend is shown

This issue is reproducible 100% on any Version after 0.8.5 (last version before this PR got merged)

benedikt-bartscher:
@hr-alebel may I suggest opening an issue for this which references the old issue/PR?

PR #4623, opened 2025-01-11
improve type support for .get_state

adhami3310:
```py
from typing import assert_type

import reflex as rx


class AnotherState(rx.State): ...


class State(rx.State):
    async def hi_mom(self):
        x = await self.get_state(AnotherState)
        assert_type(x, AnotherState)
```


adhami3310:
@masenf is it fine we error here? i think it makes sense but it's a bit harsh

masenf (approved):

PR #4630, opened 2025-01-13
bump nextJS to v15

Lendemor:


adhami3310:
for reviewers, the CI barely runs any macos tests, so it would be great if you can verify this works on your macos machine on reflex-web and other repos of similar size

reflex/constants/installer.py
```diff
@@ -170,7 +170,7 @@ class PackageJson(SimpleNamespace):
     class Commands(SimpleNamespace):
         """The commands to define in package.json."""
 
-        DEV = "next dev"
+        DEV = "next dev --turbopack"
```
adhami3310, line 173: should we anticipate disaster and make this into an environment variable? (it's a bit annoying ik to have constants based on envs)

Lendemor: Something like `USE_TURBOPACK` or `REFLEX_USE_TURBOPACK`, which would default to true but can be set to False to disable it ?

adhami3310: that's what on my mind, it would be great if it was in `EnvironmentVariables` so we don't have to do much.

Alternatively, we can have REFLEX_NEXT_FLAGS that would be a string with a default value of `--turbopack`, but i prefer the boolean over that.

Alek99:
Testing this branch with the sandbox to see if it improves anything

reflex/utils/prerequisites.py
```diff
@@ -610,10 +610,14 @@ def initialize_web_directory():
...
 
+def _turbopack_flag() -> str:
+    return " --turbopack" if environment.REFLEX_USE_TURBOPACK.get() else ""
+
+
 def _compile_package_json():
     return templates.PACKAGE_JSON.render(
         scripts={
-            "dev": constants.PackageJson.Commands.DEV,
+            "dev": constants.PackageJson.Commands.DEV + _turbopack_flag(),
```
masenf, line 620: one thing i'm hesitant about here is setting/resetting the environment variable is not sufficient to rewrite the `package.json`, that only happens during init

adhami3310: i have looked into it, there's no way to do it with config files, so package.json is the only way (unless you want to make package.json read from environment as well, but that seems cursed)

Lendemor: I don't think it's unreasonable to expect users to re-init when changing how they want to run the frontend dev server, because it's not a flag you'll change frequently.

adhami3310: @masenf do we have alternatives for this or shall we merge this as is?

The only thing possible to do here is to log last value of this flag and if it changes we regenerate package.json

masenf (approved):

PR #4631, opened 2025-01-13
Show file and line number in deprecation warnings

masenf:
Make it easier to track down where the deprecation warnings are coming from.

### DIFF
--- a/reflex/utils/console.py
+++ b/reflex/utils/console.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import inspect
+import shutil
 from pathlib import Path
 from types import FrameType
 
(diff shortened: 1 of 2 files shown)


adhami3310 (approved):

reflex/utils/console.py
```diff
@@ -188,6 +192,20 @@ def warn(msg: str, dedupe: bool = False, **kwargs):
...
 
 
+def _get_first_non_framework_frame() -> FrameType | None:
+    import reflex as rx
+
+    framework_root = Path(rx.__file__).parent.resolve()
+    frame = inspect.currentframe()
+    while frame := frame and frame.f_back:
+        frame_path = Path(inspect.getfile(frame)).resolve()
+        if frame_path.name == "typing_extensions.py":
```
adhami3310, line 202: is there anything special about `typing_extensions`? can we make that behavior generic over such examples?

masenf: latest diff uses a more generic and extensible method for excluding packages

adhami3310 (approved):

PR #4635, opened 2025-01-14
[ENG-4356]Remove chakra completely from core

ElijahAhianyo:
Remove chakra shim completely.  Users should install `reflex-chakra` to continue usage


Lendemor:
#4586 already remove that

ElijahAhianyo:
> #4586 already remove that

RIght, Ill close this and reassign that ticket to you then

Issue #4628, opened 2025-01-13
Support rx.cond with Literal Values for tag in rx.icon


sorousch9:
**Describe the bug**
Currently, using rx.cond with literal string values for the tag argument in rx.icon raises issues with safety or functionality. This limits the ability to dynamically assign icons based on a conditional state in a clear and concise manner.

**To Reproduce**

```python
rx.icon(tag=rx.cond(MyState.bool, "arrow-up-wide-narrow", "arrow-down-narrow-wide"))
```

**Expected behavior**
It should be possible to use rx.cond with literal values for the tag argument safely and without requiring workarounds. The expected functionality would dynamically resolve the tag value based on the conditional expression.

**Specifics (please complete the following information):**
 - Python Version: 3.13.1
 - Reflex Version:0.6.7
 - OS: WSL2

 
**Additional context**
A potential workaround for this issue is to wrap the rx.icon calls within the rx.cond as follows:
```python
rx.cond(
    MyState.bool,
    rx.icon(tag="arrow-up-wide-narrow"),
    rx.icon(tag="arrow-down-narrow-wide")
)

```
This workaround avoids the direct use of literal values with rx.cond, but a built-in solution would streamline development and improve code clarity.

Lendemor:
I don't think this particular feature would work for `rx.icon` because the name need to be known during compilation for the import to work.

I'm tempted to mark this one as closed once we merge the dynamic Icon PR, unless we have other use case that would require this outside of `rx.icon` ?

PR #4636, opened 2025-01-14
allow dynamic icons name

Lendemor:
Latest lucide version allow use of DynamicIcon.

Need to be documented with a warning about slowing down apps, only to be used when really needed and prefer static names when possible.

### DIFF
--- a/reflex/components/lucide/icon.py
+++ b/reflex/components/lucide/icon.py
@@ -42,7 +42,6 @@ class Icon(LucideIconComponent):
         if children:
             if len(children) == 1 and isinstance(children[0], str):
                 props["tag"] = children[0]
-                children = []
             else:
                 raise AttributeError(
                     f"Passing multiple children to Icon component is not allowed: remove positional arguments {children[1:]} to fix"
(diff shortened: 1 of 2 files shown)


reflex/components/lucide/icon.py
```diff
@@ -47,6 +48,10 @@ def create(cls, *children, **props) -> Component:
         if "tag" not in props:
             raise AttributeError("Missing 'tag' keyword-argument for Icon")
 
+        if isinstance(props["tag"], Var):
+            icon_name = props.pop("tag")
+            return DynamicIcon.create(name=icon_name, **props)
+
```
adhami3310, line 54: can we special case LiteralVar?

benedikt-bartscher: That's what we wanted to archive in https://github.com/reflex-dev/reflex/issues/4628.

Still awesome to see full dynamic support!

adhami3310: i don't think we can get it to work to remain static with rx.cond sadly :( that might require vardata-level stuff

reflex/components/lucide/icon.py
```diff
@@ -47,6 +50,18 @@ def create(cls, *children, **props) -> Component:
...
 
+        if isinstance(props["tag"], LiteralVar):
+            if isinstance(props["tag"], LiteralStringVar):
+                props["tag"] = props["tag"]._var_value
+            else:
+                raise TypeError("Icon name must be a string")
+
+        if isinstance(props["tag"], Var):
+            icon_name: Var = props.pop("tag")
+            if icon_name._var_type is not str:
```
adhami3310, line 61: do we want to fail here? there's also cases where str is optional, or something weird

reflex/components/lucide/icon.py
```diff
@@ -32,39 +34,61 @@ def create(cls, *children, **props) -> Component:
...
+    tag = "DynamicIcon"
+
+    name: Var[str]
+
+    def _get_imports(self):
+        _imports = super()._get_imports()
+        if self.library:
+            _imports.pop(self.library)
+        _imports["lucide-react/dynamic"] = [ImportVar("DynamicIcon", install=False)]
+        return _imports
```
masenf, line 91: probably best to just override `Component.import_var` property here instead

Lendemor: Won't work, the name of the package import need to be different, `Component.import_var` doesn't allow that.

`_imports[self.library] = {self.import_var}` this is the line we want to change.

adhami3310 (approved):

PR #4640, opened 2025-01-15
Remove token check in reflex deploy

ElijahAhianyo:
This logic has been moved upstream to reflex-hosting-cli (via https://github.com/reflex-dev/reflex-hosting-cli/pull/102)

### DIFF
--- a/reflex/reflex.py
+++ b/reflex/reflex.py
@@ -503,13 +503,6 @@ def deploy(
     # Set the log level.
     console.set_log_level(loglevel)
 
-    if not token:
-        # make sure user is logged in.
-        if interactive:
-            hosting_cli.login()
-        else:
-            raise SystemExit("Token is required for non-interactive mode.")
-
     # Only check requirements if interactive.
     # There is user interaction for requirements update.
     if interactive:


Lendemor:
Going to mark this as draft until the PR for the hosting-cli is merged.

Also maybe we want the `poetry lock` update in the same PR, for the requirements to the next version of hosting-cli.

Lendemor (approved):

ElijahAhianyo:
> Going to mark this as draft until the PR for the hosting-cli is merged.
> 
> Also maybe we want the `poetry lock` update in the same PR, for the requirements to the next version of hosting-cli.

The requirements for that is already bounded (`reflex-hosting-cli = ">=0.1.29,<2.0"`), shouldn't that be automatically picked  up if the CLI gets released?

Lendemor:
> > Going to mark this as draft until the PR for the hosting-cli is merged.
> > Also maybe we want the `poetry lock` update in the same PR, for the requirements to the next version of hosting-cli.
> 
> The requirements for that is already bound (`reflex-hosting-cli = ">=0.1.29,<2.0"`), shouldn't that be automatically picked up in the next reflex release if the CLI gets released?

Next version won't work with <0.1.32, so we need to update the lower bound anyway.

masenf:
if the lower bound isn't raised, then upgrades of reflex won't install the new reflex-hosting-cli

that said, it's kind of a wash anyway because we have a secondary runtime check for latest version of hosting-cli and tell people to upgrade manually then

masenf:
i think it's better to bump the lower bound if this PR depends on things in a later version of `reflex-hosting-cli`

masenf (approved):

PR #4642, opened 2025-01-15
rename private fields with leading underscore in App

Lendemor:


reflex/app.py
```diff
@@ -251,36 +251,36 @@ class App(MiddlewareMixin, LifespanMixin):
...
-    load_events: Dict[str, List[IndividualEventType[[], Any]]] = dataclasses.field(
+    # Mapping from a route to event handlers to trigger when the page loads.
+    _load_events: Dict[str, List[IndividualEventType[[], Any]]] = dataclasses.field(
         default_factory=dict
     )
 
-    # Admin dashboard to view and manage the database. PRIVATE.
-    admin_dash: Optional[AdminDash] = None
+    # Admin dashboard to view and manage the database.
+    _admin_dash: Optional[AdminDash] = None
```
masenf, line 277: not sure if it's actually documented, but we should keep this one as it was because initializing the app with `_admin_dash` is weird and would break existing apps using this interface.

masenf (commented):
we should add read-only properties for `.api` and `.event_namespace`, since those are public APIs:

* `.api` is used for adding backend routes
* `.event_namespace` is used for accessing token_to_sid and sid_to_token mappings

These are private in the sense that they should not be set by the user when initializing the app, but their values are actually publicly documented.

reflex/app.py
```diff
@@ -251,36 +251,36 @@ class App(MiddlewareMixin, LifespanMixin):
...
+    # Mapping from a route to event handlers to trigger when the page loads.
+    _load_events: Dict[str, List[IndividualEventType[[], Any]]] = dataclasses.field(
         default_factory=dict
     )
 
-    # Admin dashboard to view and manage the database. PRIVATE.
+    # Admin dashboard to view and manage the database.
     admin_dash: Optional[AdminDash] = None
 
     # The async server name space. PRIVATE.
```
adhami3310, line 279:
```suggestion
    # The async server name space.
```

adhami3310 (approved):

PR #4644, opened 2025-01-15
CR Suggestions for #4556

masenf:


ElijahAhianyo:
what would the usage of this look like now? Would it look something like this?

foo/rxconfig.py
```python
import sys
from pathlib import Path

import reflex as rx

bar_app_dir = Path("/Users/eli/Documents/work/Pynecone-io/pynecone-forks/reflex/demos/oct_04_24/bar")

if str(bar_app_dir) not in sys.path:
    sys.path.insert(0, str(bar_app_dir))

config = rx.Config(
    app_name="foo",
    app_module_import="bar.bar",
)
```

I get the error below when I run  this:

```
  File "/Users/eli/Documents/work/Pynecone-io/pynecone-forks/reflex/reflex/utils/prerequisites.py", line 303, in get_app
    module = config.module
             ^^^^^^^^^^^^^
  File "/Users/eli/Documents/work/Pynecone-io/pynecone-forks/reflex/reflex/config.py", line 753, in module
    if self.app_module is not None:
       ^^^^^^^^^^^^^^^
  File "/Users/eli/Documents/work/Pynecone-io/pynecone-forks/reflex/reflex/config.py", line 741, in app_module
    importlib.import_module(self.app_module_import)
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1206, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1178, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1128, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1206, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1178, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1142, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'bar'

```

I put a breakpoint right before we import the module in `app_module` and it looks like the path might have been cleared somehow in the sys path before we got to this point

reflex/utils/exec.py
```diff
@@ -248,8 +248,17 @@ def get_reload_dirs() -> list[str]:
     """
     config = get_config()
     reload_dirs = [config.app_name]
-    if app_module_path := config.app_module_path:
-        reload_dirs.append(str(Path(app_module_path).resolve().parent.parent))
+    if config.app_module is not None and config.app_module.__file__:
```
ElijahAhianyo, line 251: I initially based my assumption on the current behavior of how we expect reflex apps to be structured based on the  `Config.module` property(`return ".".join([self.app_name, self.app_name])`). This actually assumes that the entrypoint (or where the App instance lives) will only be one level deep, So if I modify my file structure from `foo/foo/foo.py` to `foo/foo/bar/foo.py` this will currently fail. With this new approach, I'm wondering if we need to have a similar behavior where we provide users with a knob to change the default behavior, but I'm also wondering if adding another knob will make it confusing for users since we have the `app_module_import` knob now

masenf:
i think you may have uncovered a bug where due to how we `get_config` where anything added to sys.path during rxconfig import gets blown away.

```python
def get_config(reload: bool = False) -> Config:
    """Get the app config.

    Args:
        reload: Re-import the rxconfig module from disk

    Returns:
        The app config.
    """
    cached_rxconfig = sys.modules.get(constants.Config.MODULE, None)
    if cached_rxconfig is not None:
        if reload:
            # Remove any cached module when `reload` is requested.
            del sys.modules[constants.Config.MODULE]
        else:
            return cached_rxconfig.config

    with _config_lock:
        sys_path = sys.path.copy()
        sys.path.clear()
        sys.path.append(str(Path.cwd()))
        try:
            # Try to import the module with only the current directory in the path.
            return _get_config()
        except Exception:
            # If the module import fails, try to import with the original sys.path.
            sys.path.extend(sys_path)
            return _get_config()
        finally:
            # Restore the original sys.path.
            sys.path.clear()
            sys.path.extend(sys_path)
```

This logic basically reverts any `sys.path` changes that occur during `import rxconfig`.

I think we need to change the logic to also apply any new entries added to `sys.path` during the load.

reflex/utils/exec.py
masenf: i don't think we necessarily need to do anything special here. we just want to watch the top of the package hierarchy that the `app_import_module` belongs to. In fact the same logic would work just fine with the "default" structure as well.

ElijahAhianyo:
> my latest push should fix that sys.path bug, so your sample code above should be working on [9d2b4fa](https://github.com/reflex-dev/reflex/commit/9d2b4fae60ab9116b9aae57b429a874ea284b864)

Yeah this works for me, If you want, I can go ahead and merge this and update the integration test as well

PR #4665, opened 2025-01-20
[WIP]`reflex rename`

ElijahAhianyo:
proof of concept for renaming a reflex app. Not ready for review

### DIFF
--- a/reflex/reflex.py
+++ b/reflex/reflex.py
@@ -551,6 +551,22 @@ def deploy(
     )
 
 
+@cli.command()
+def rename(
+    new_name: str = typer.Argument(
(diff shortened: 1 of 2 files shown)


ElijahAhianyo:
This PR got semi-abandoned(not completely abandoned). It modifies the ASTs directly(which I admit might be the safer approach) but ASTs don't preserve whitespace or comments. astor preserves formatting only to small extent. so the code structure is guaranteed to change. 

Another approach might be to combine ASTs with the [tokenize](https://docs.python.org/3/library/tokenize.html) module?

masenf:
took the other one

PR #4666, opened 2025-01-20
enable N rules for naming conventions

Lendemor:


reflex/event.py
```diff
@@ -557,10 +560,10 @@ class JavasciptKeyboardEvent:
...
 
     key: str = ""
-    altKey: bool = False
-    ctrlKey: bool = False
-    metaKey: bool = False
-    shiftKey: bool = False
+    alt_key: bool = False
+    ctrl_key: bool = False
+    meta_key: bool = False
+    shift_key: bool = False
```
masenf, line 566: idk that we can just change these without doing some kind of mapping. these are the names given by the browser API

reflex/components/next/image.pyi
```diff
@@ -30,7 +32,7 @@ class Image(NextComponent):
         loading: Optional[
             Union[Literal["eager", "lazy"], Var[Literal["eager", "lazy"]]]
         ] = None,
-        blurDataURL: Optional[Union[Var[str], str]] = None,
+        blur_data_url: Optional[Union[Var[str], str]] = None,
```
masenf, line 35: this is an API breaking change, i wonder if we should map the old name in `create` and throw a deprecation message

reflex/event.py
adhami3310: we do the mapping! this is actually just to give the input API, the output API is `KeyInputInfo` which does use Python naming convention

reflex/utils/exec.py
```diff
@@ -306,7 +306,7 @@ def run_granian_backend(host, port, loglevel: LogLevel):
             interface=Interfaces.ASGI,
             log_level=LogLevels(loglevel.value),
             reload=True,
-            reload_paths=get_reload_dirs(),
+            reload_paths=[Path(rp) for rp in get_reload_dirs()],
```
adhami3310, line 309: this is not needed, get_reload_dirs is alr a list of paths

Lendemor: ah yeah it wasn't when I wrote that, need to change it back.

adhami3310 (approved):

PR #4667, opened 2025-01-20
Bump frontend and backend deps

masenf:
TBD


reflex/constants/installer.py
```diff
@@ -178,21 +178,21 @@ class Commands(SimpleNamespace):
...
         "json5": "2.2.3",
-        "next": "15.1.4",
+        "next": "15.1.5",
         "next-sitemap": "4.2.3",
-        "next-themes": "0.4.3",
-        "react": "18.3.1",
-        "react-dom": "18.3.1",
-        "react-focus-lock": "2.13.2",
+        "next-themes": "0.4.4",
+        "react": "18.3.0",
```
Lendemor, line 188: I see you reverted the change to react19, but why are we regressing to older version that initially ?

masenf: it was just a mistake

Lendemor (dismissed):

Lendemor (dismissed):

Lendemor (approved):

PR #4680, opened 2025-01-23
use older version of python for windows simple benchmarks

adhami3310:


adhami3310:
that should get post-merge tests to pass

Lendemor (approved):

PR #4683, opened 2025-01-23
return websocket protocol when asked

adhami3310:


masenf (approved):
Do you see the `env.TEST_MODE` hacks continuing to be useful?

My inclination is to remove the whole mechanism to avoid cases in the future where we special case our integration tests and end up actually hiding issues.

adhami3310:
> Do you see the `env.TEST_MODE` hacks continuing to be useful?

good point, i'm not sure, i could perceive it to be useful considering we haven't switched off to pyright completely, so some tests might need it

PR #4698, opened 2025-01-27
fix subprotocol for granian

adhami3310:


masenf (commented):
still doesn't work for uvicorn in dev mode accessing on chrome

gi0baro:
FYI, the ASGI protocol specifies you should just use the `subprotocol` message field, and avoid to have `sec-websocket-protocol` in headers. ref: https://asgi.readthedocs.io/en/latest/specs/www.html#accept-send-event

adhami3310:
> FYI, the ASGI protocol specifies you should just use the `subprotocol` message field

Thanks for the link, we were struggling with this a bit. The issue is that Granian/Uviocorn/Gunicorn behave slightly differently in different systems. We couldn't figure out why and how to reproduce it reliably so we opted out for using the header if it's there and subprotocol if it's there.

masenf (approved):

PR #4699, opened 2025-01-27
enable PGH, bump pyright and fix  all #type: ignore

Lendemor:


pyrightconfig.json
adhami3310, line 1: i believe we can have that in pyproject instead

reflex/utils/prerequisites.py
```diff
@@ -210,8 +210,9 @@ def get_bun_version() -> version.Version | None:
         return None
     try:
         # Run the bun -v command and capture the output
-        result = processes.new_process([str(bun_path), "-v"], run=True)
-        return version.parse(str(result.stdout))
+        result = processes.new_process([str(get_config().bun_path), "-v"], run=True)
+        result = processes.new_process([str(get_config().bun_path), "-v"], run=True)
```
adhami3310, line 214: why are we running this twice

tests/integration/test_var_operations.py
```diff
@@ -91,35 +89,35 @@ def index():
                 id="int_or_int",
             ),
             rx.text(
-                (VarOperationState.int_var1 > VarOperationState.int_var2).to_string(),
+                (VarOperationState.int_var1 > VarOperationState.int_var2).to_string(),  # pyright: ignore [reportAttributeAccessIssue]
```
adhami3310, line 92: i feel for those one it's better to wrap the fields with rx.Field

reflex/utils/prerequisites.py
Lendemor: Oops messed up when solving merge conflicts.

pyrightconfig.json
Lendemor: Sure will try

adhami3310 (approved):

PR #4706, opened 2025-01-28
merging two style instance should give a style instance

Lendemor:
`style1 | style2` currently give an object of type `dict`.

After fix, merging style together or dict with style will give a Style object

### DIFF
--- a/reflex/style.py
+++ b/reflex/style.py
@@ -301,7 +301,8 @@ class Style(dict):
         Returns:
             The combined style.
         """
-        return Style(super().__or__(self, other))  # pyright: ignore [reportGeneralTypeIssues, reportCallIssue]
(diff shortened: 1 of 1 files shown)


reflex/style.py
```diff
@@ -292,6 +292,17 @@ def __setitem__(self, key: str, value: Any):
...
+    def __or__(self, other: Style | dict) -> Style:
+        """Combine two styles.
+
+        Args:
+            other: The other style to combine.
+
+        Returns:
+            The combined style.
+        """
+        return Style(super().__or__(self, other))  # pyright: ignore [reportGeneralTypeIssues, reportCallIssue]
```
adhami3310, line 304: do we want to worry about var data here?

adhami3310 (approved):

PR #4729, opened 2025-02-01
better computed var static deps

benedikt-bartscher:


reflex/vars/base.py
```diff
@@ -2052,39 +2052,72 @@ def __init__(
...
             self,
             "_static_deps",
-            _static_deps,
+            self._calculate_static_deps(deps),
         )
         object.__setattr__(self, "_auto_deps", auto_deps)
 
         object.__setattr__(self, "_fget", fget)
 
+    def _calculate_static_deps(
```
adhami3310, line 2064: it seems this method should be a static method

benedikt-bartscher: Currently, `_calculate_static_deps` makes use of self, but if needed I can refactor that.
The main reason for this PR is to keep an interface to easily add static deps to a computed var without copying the new static dep dict calculation into downstream projects.

masenf (approved):

PR #4734, opened 2025-02-03
standarize filename from upload

### LINKED ISSUE #4562: Uploaded file.filename on safari and firefox has a leading slash
**Describe the bug**
The upload example code from the docs doesn't work on safari 😔

**To Reproduce**
```python

import reflex as rx

from rxconfig import config


class State(rx.State):
    """The app state."""

    # The images to show.
    img: list[str]

    @rx.event
    async def handle_upload(
        self, files: list[rx.UploadFile]
    ):
        """Handle the upload of file(s).

        Args:
            files: The uploaded files.
        """
        for file in files:
            outfile = rx.get_upload_dir() / file.filename

            # Save the file.
            with outfile.open("wb") as file_object:

            # Update the img var.


color = "rgb(107,99,246)"


def index():
    """The main view."""
    return rx.vstack(
        rx.upload(
            rx.vstack(
                rx.button(
                    "Select File",
                    color=color,
                    bg="white",
                    border=f"1px solid {color}",
                ),
                rx.text(
                    "Drag and drop files here or click to select files"
                ),
            ),
            id="upload1",
            border=f"1px dotted {color}",
            padding="5em",
        ),
        rx.hstack(
            rx.foreach(
                rx.selected_files("upload1"), rx.text
            )
        ),
        rx.button(
            "Upload",
            on_click=State.handle_upload(
                rx.upload_files(upload_id="upload1")
            ),
        ),
        rx.button(
            "Clear",
            on_click=rx.clear_selected_files("upload1"),
        ),
        rx.foreach(
            State.img,
            lambda img: rx.image(
                src=rx.get_upload_url(img)
            ),
        ),
        padding="5em",
    )


app = rx.App()
app.add_page(index)
```

**Expected behavior**
Dragging a file to upload should work the same on Chrome and Safari (and Firefox).

What actually happens

* Chrome gets the filename as `./image.png`
* Safari and Firefox get the filename as `/image.png`

The sample code from the website uses `rx.get_upload_dir() / file.filename` (arguably bad practice), but because safari and firefox pass an "absolute" path, the left part gets killed so we try to open `/image.png`, which hopefully the reflex app does not have permission to do 😅.

 - Python Version: 3.12
 - Reflex Version: 0.6.8.dev1

adhami3310:
fixes #4562

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -1559,7 +1559,7 @@ def upload(app: App):
             file_copies.append(
                 UploadFile(
                     file=content_copy,
-                    filename=file.filename,
+                    filename=Path(file.filename).name if file.filename else None,
                     size=file.size,
                     headers=file.headers,
                 )
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -22,6 +22,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     AsyncIterator,
+    BinaryIO,
     Callable,
     Coroutine,
     Dict,
@@ -36,12 +37,15 @@ from typing import (
     get_type_hints,
 )
 
-from fastapi import FastAPI, HTTPException, Request, UploadFile
+from fastapi import FastAPI, HTTPException, Request
+from fastapi import UploadFile as FastAPIUploadFile
 from fastapi.middleware import cors
 from fastapi.responses import JSONResponse, StreamingResponse
 from fastapi.staticfiles import StaticFiles
 from rich.progress import MofNCompleteColumn, Progress, TimeElapsedColumn
 from socketio import ASGIApp, AsyncNamespace, AsyncServer
+from starlette.datastructures import Headers
+from starlette.datastructures import UploadFile as StarletteUploadFile
 from starlette_admin.contrib.sqla.admin import Admin
 from starlette_admin.contrib.sqla.view import ModelView
 
@@ -190,6 +194,53 @@ class OverlayFragment(Fragment):
     """Alias for Fragment, used to wrap the overlay_component."""
 
     pass
+
+
+@dataclasses.dataclass(frozen=True)
+class UploadFile(StarletteUploadFile):
+    """A file uploaded to the server.
+
+    Args:
+        file: The standard Python file object (non-async).
+        filename: The original file name.
+        size: The size of the file in bytes.
+        headers: The headers of the request.
+    """
+
+    file: BinaryIO
+
+    path: Optional[Path] = dataclasses.field(default=None)
+
+    deprecated_filename: Optional[str] = dataclasses.field(default=None)
+
+    size: Optional[int] = dataclasses.field(default=None)
+
+    headers: Headers = dataclasses.field(default_factory=Headers)
+
+    @property
+    def name(self) -> Optional[str]:
+        """Get the name of the uploaded file.
+
+        Returns:
+            The name of the uploaded file.
+        """
+        if self.path:
+            return self.path.name
+
+    @property
+    def filename(self) -> Optional[str]:
+        """Get the filename of the uploaded file.
+
+        Returns:
+            The filename of the uploaded file.
+        """
+        console.deprecate(
+            feature_name="UploadFile.filename",
+            reason="Use UploadFile.name instead.",
+            deprecation_version="0.7.1",
+            removal_version="0.8.0",
+        )
+        return self.deprecated_filename
 
 
 @dataclasses.dataclass(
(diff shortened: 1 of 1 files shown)


reflex/app.py
```diff
@@ -1559,7 +1559,7 @@ async def upload_file(request: Request, files: List[UploadFile]):
             file_copies.append(
                 UploadFile(
                     file=content_copy,
-                    filename=file.filename,
+                    filename=Path(file.filename).name if file.filename else None,
```
masenf, line 1562: This strips off the path information.

```python
>>> Path("./foo/bar").name
'bar'
```

This functionality is important for recreating the tree when user drops a directory onto the dropzone component

adhami3310: that's interesting, but also, isn't `filename` supposed to be turly just the filename?

masenf: yeah, probably should expose a `.path` and `.filename` separately... but that would be an API-breaking change at this point. we should have a deprecation window and maybe drop it in 0.8

adhami3310: i'm in favor of leaving `.filename` as is, but introducing `.name` that should have the previous logic, (and maybe `.path` too). We can make `.name` a property when called warns the user about the deprecation.

adhami3310: For that we need to introduce our own UploadFile (maybe we can inherit from fastapi UploadFile though)

reflex/app.py
```diff
@@ -1559,7 +1610,8 @@ async def upload_file(request: Request, files: List[UploadFile]):
             file_copies.append(
                 UploadFile(
                     file=content_copy,
-                    filename=file.filename,
+                    path=Path(file.filename) if file.filename else None,
```
masenf, line 1613: can we `lstrip` a leading slash from the filename?

```suggestion
                    path=Path(file.filename.lstrip("/")) if file.filename else None,
```

This would give us good compatibility with all major browsers.

Currently we get the following

## Directory Upload

For directory upload, the result is the same, with a weird absolute path that can't be joined to the upload dir easily.

### Chrome

```python
>>> print([f.path for f in files])
[PosixPath('/app/blocks/tables/__init__.py'), PosixPath('/app/blocks/tables/table.py'), PosixPath('/app/blocks/__init__.py'), PosixPath('/app/rxconfig.py'), PosixPath('/app/app/models.py'), PosixPath('/app/app/__init__.py'), PosixPath('/app/app/components/__init__.py'), PosixPath('/app/app/components/dashboard/form_overview.py'), PosixPath('/app/app/components/dashboard/user_stats.py'), PosixPath('/app/app/components/dashboard/layout.py'), PosixPath('/app/app/components/dashboard/response_stats.py'), PosixPath('/app/app/components/dashboard/field_analytics.py'), PosixPath('/app/app/states/base_state.py'), PosixPath('/app/app/states/__init__.py'), PosixPath('/app/app/states/dashboard_state.py'), PosixPath('/app/app/app.py'), PosixPath('/app/app/pages/__init__.py'), PosixPath('/app/app/pages/dashboard.py'), PosixPath('/app/requirements.txt'), PosixPath('/app/.gitignore'), PosixPath('/app/assets/favicon.ico')]
```

### Safari

```python
>>> print([f.path for f in files])
[PosixPath('/app/blocks/tables/__init__.py'), PosixPath('/app/blocks/tables/table.py'), PosixPath('/app/blocks/__init__.py'), PosixPath('/app/rxconfig.py'), PosixPath('/app/app/models.py'), PosixPath('/app/app/__init__.py'), PosixPath('/app/app/components/__init__.py'), PosixPath('/app/app/components/dashboard/form_overview.py'), PosixPath('/app/app/components/dashboard/user_stats.py'), PosixPath('/app/app/components/dashboard/layout.py'), PosixPath('/app/app/components/dashboard/response_stats.py'), PosixPath('/app/app/components/dashboard/field_analytics.py'), PosixPath('/app/app/states/base_state.py'), PosixPath('/app/app/states/__init__.py'), PosixPath('/app/app/states/dashboard_state.py'), PosixPath('/app/app/app.py'), PosixPath('/app/app/pages/__init__.py'), PosixPath('/app/app/pages/dashboard.py'), PosixPath('/app/requirements.txt'), PosixPath('/app/assets/favicon.ico')]
```

## Single File

For single file, the behavior is different, chrome adds a leading `./`, which causes the resulting path to be nice and relative:

### Chrome

```python
>>> file.path
PosixPath('out-0-2.png')
```

### Safari

In safari, we seed the leading slash making the filename an absolute path.

```python
>>> file.path
PosixPath('/out-0-2.png')
```

That leading slash makes it easy to screw up

```python
>>> rx.get_upload_dir() / file.path
PosixPath('/out-0-2.png')
```

😢 

---------------------

I contend that the proper thing to do here is _always_ use a relative path for `path` field. We can leave the deprecated filename field as whatever the browser sends us.

PR #4737, opened 2025-02-03
workaround macOS EMFILE issue during reflex-web production build

masenf:
try to check and raise the limit as a first shot


masenf:
Issue seems to have disappeared 🫥

PR #4740, opened 2025-02-03
remove base_state from event types

adhami3310:


adhami3310:
(technically a breaking change, but a very easy one to abide by, simply removing, `BASE_STATE` and the extra brackets work, also: `EventType[[], BASE_STATE]` bcomes `EventType[()]`)

masenf (approved):
@adhami3310 do you think we should bring this for 0.7.0? i haven't updated my third party components yet, but some of them would need some tweaks to work with this change. Probably better to introduce the breakage at once here, or do you think there might be other implications if we bring this in?

PR #4749, opened 2025-02-04
Remove app_module_for_backend

masenf:
* Replace app_module_for_backend with ASGI factory on `rx.App.__call__`
* Consolidate module spec for uvicorn, gunicorn, and granian


adhami3310:
it seems --backend-only is failing on uvicorn

adhami3310 (approved):

PR #4751, opened 2025-02-05
improve foreach behavior with strings

adhami3310:


masenf (commented):
can you add a test for the string case in `test_var_operations.py`

Lendemor (dismissed):

Lendemor (approved):

Issue #3699, opened 2024-07-23
'module' object is not callable


reigadegr:
Using command: reflex run --loglevel debug
return this message:
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\admxcy\.conda\envs\py11\Scripts\reflex.exe\__main__.py", line 7, in <module>
  File "C:\Users\admxcy\.conda\envs\py11\Lib\site-packages\typer\main.py", line 326, in __call__
    raise e
  File "C:\Users\admxcy\.conda\envs\py11\Lib\site-packages\typer\main.py", line 309, in __call__
    return get_command(self)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\admxcy\.conda\envs\py11\Lib\site-packages\click\core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\admxcy\.conda\envs\py11\Lib\site-packages\typer\core.py", line 723, in main
    return _main(
           ^^^^^^
  File "C:\Users\admxcy\.conda\envs\py11\Lib\site-packages\typer\core.py", line 193, in _main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "C:\Users\admxcy\.conda\envs\py11\Lib\site-packages\click\core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\admxcy\.conda\envs\py11\Lib\site-packages\click\core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\admxcy\.conda\envs\py11\Lib\site-packages\click\core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 File "C:\Users\admxcy\.conda\envs\py11\Lib\site-packages\typer\main.py", line 692, in wrapper
    return callback(**use_params)
           ^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #4754, opened 2025-02-05
improve into component conversion

adhami3310:
makes #3699 easier to catch

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -2459,6 +2459,8 @@ class ComponentState(State, mixin=True):
         Returns:
             A new instance of the Component with an independent copy of the State.
         """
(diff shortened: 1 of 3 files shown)


reflex/state.py
```diff
@@ -2471,6 +2473,7 @@ def create(cls, *children, **props) -> "Component":
         setattr(reflex.istate.dynamic, state_cls_name, component_state)
         component = component_state.get_component(*children, **props)
         component.State = component_state
+        component = into_component(component)
```
masenf, line 2476: i think we want to invert the order of these lines; the `.State` should always be attached to the thing that's returned from `.create(...)`

masenf (approved):

PR #4758, opened 2025-02-05
move benchmarks and add some more

Lendemor:


tests/benchmarks/fixtures.py
```diff
@@ -0,0 +1,231 @@
...
+        rx.box(
+            rx.heading("Complicated Page", size="1"),
+            rx.text(LOREM_IPSUM),
+        ),
+    )
+
+
+@pytest.fixture
+def complicated_page():
+    return _complicated_page()
```
masenf, line 231: what about

```python
@pytest.fixture(params=[_complicated_page, _simple_page], ids=["_complicated_page", "_simple_page"])
def evaluated_page(request):
    return request.param()
```

Then we should be able to run the compile benchmarks against a variety of pages without further changing the structure of the fixture access

tests/benchmarks/test_evaluate.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+from .fixtures import _complicated_page
+
+
+@pytest.mark.benchmark
+def test_evaluate_page():
+    _complicated_page()
```
masenf, line 8: i also think we should parametrize this one pre-emptively. even though we only have one case now, if we parametrize it, then it's very easy to add additional pages to evaluate and the naming structure of the test case wouldn't change, the new ones would just be seen as new benchmarks.

pyproject.toml
```diff
@@ -102,6 +102,7 @@ lint.pydocstyle.convention = "google"
 "*/blank.py" = ["I001"]
 
 [tool.pytest.ini_options]
+filterwarnings = "ignore::RuntimeWarning"
```
masenf, line 105: What is this intended to ignore? Are the warnings unfixable? This seems like a heavy hammer to apply to all test cases in the repo, especially because RuntimeWarning have led us to discover real test bugs in the past.

Lendemor: I pushed a more specific ignore instead.

adhami3310 (approved):

PR #4760, opened 2025-02-05
Make custom component command an optional extra

masenf:
Reduce package scope and CLI entry point for normal installs.

Requires `pip install reflex[custom_components]` when publishing a component.


adhami3310:
we already made most of it option by now

PR #4762, opened 2025-02-06
make computed var generic over mapping

adhami3310:
```py
import reflex as rx


class State(rx.State):
    @rx.var
    def foo(self) -> int:
        return 42

    @rx.var(cache=True)
    def bar(self) -> dict:
        return {"foo": 42}


def fancy_number(n: rx.vars.NumberVar) -> rx.Component:
    return rx.text(n)


def fancy_object(o: rx.vars.ObjectVar) -> rx.Component:
    return rx.text(o.to_string())


def index() -> rx.Component:
    return rx.fragment(
        fancy_number(State.foo),
        fancy_object(State.bar),
    )


app = rx.App()
app.add_page(index)
```

### DIFF
--- a/reflex/vars/base.py
+++ b/reflex/vars/base.py
@@ -2219,10 +2219,10 @@ class ComputedVar(Var[RETURN_TYPE]):
 
     @overload
     def __get__(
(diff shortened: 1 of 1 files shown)


masenf (approved):
works like a charm, thank you

PR #4764, opened 2025-02-06
add stateful benchmarks

Lendemor:


adhami3310:
It seems the tests are very small for this, do we want to make it slightly larger? could be more deeply nested or such

adhami3310 (approved):

PR #4769, opened 2025-02-07
raise error when passing a str(var)

adhami3310:


reflex/components/base/bare.py
```diff
@@ -7,9 +7,44 @@
...
+from reflex.utils import console
+from reflex.utils.decorator import once
 from reflex.utils.imports import ParsedImportDict
 from reflex.vars import BooleanVar, ObjectVar, Var
 from reflex.vars.base import VarData
+from reflex.vars.sequence import LiteralStringVar
+
+
+@once
+def performace_mode():
```
Lendemor, line 20:
```suggestion
def performance_mode():
```

reflex/utils/decorator.py
```diff
@@ -0,0 +1,25 @@
+"""Decorator utilities."""
+
+from typing import Callable, TypeVar
+
+T = TypeVar("T")
+
+
+def once(f: Callable[[], T]) -> Callable[[], T]:
```
masenf, line 8: could we use the new global var cache for this so we have some control over evicting the values?

adhami3310: we could, but i think we should keep this one away from global var cache until we improve that one (also unsure if we want to mix var with not vars)

tests/units/test_var.py
```diff
@@ -1893,3 +1894,27 @@ def test_var_data_hooks():
...
     assert var_data == VarData(hooks=["what", "whot", "whott"])
+
+
+def test_str_var_in_components(mocker):
+    class StateWithVar(rx.State):
+        field: int = 1
+
+    mocker.patch(
+        "reflex.components.base.bare.get_performance_mode",
+        lambda: PerformanceMode.RAISE,
```
Lendemor, line 1905: We're using `return_value` in other place where we patch function, so we should keep things consistent I think.
```suggestion
        return_value=PerformanceMode.RAISE,
```

tests/units/test_var.py
```diff
@@ -1893,3 +1894,27 @@ def test_var_data_hooks():
...
+    )
+
+    with pytest.raises(ValueError):
+        rx.vstack(
+            str(StateWithVar.field),
+        )
+
+    mocker.patch(
+        "reflex.components.base.bare.get_performance_mode",
+        lambda: PerformanceMode.OFF,
```
Lendemor, line 1915:
```suggestion
        return_value=PerformanceMode.OFF,
```

Lendemor (changes requested):

Lendemor (approved):

PR #4773, opened 2025-02-07
fix port handling

Lendemor:


reflex/utils/processes.py
```diff
@@ -116,7 +116,7 @@ def change_port(port: int, _type: str) -> int:
     return new_port
 
 
-def handle_port(service_name: str, port: int, default_port: int) -> int:
+def handle_port(service_name: str, port: int, autoswitch: bool) -> int:
```
masenf, line 119: i think calling this `auto_increment` might be a bit nicer

reflex/utils/processes.py
```diff
@@ -126,21 +126,21 @@ def handle_port(service_name: str, port: int, default_port: int) -> int:
...
-        else:
-            console.error(f"{service_name.capitalize()} port: {port} is already in use")
-            raise typer.Exit()
-    return port
+    if not is_process_on_port(port):
+        return port
+    if autoswitch:
+        return change_port(port, service_name)
+    else:
+        console.error(f"{service_name.capitalize()} port: {port} is already in use")
```
masenf, line 142: if we could get the pid from `is_process_on_port`, that might improve this error message

reflex/config.py
```diff
@@ -544,6 +544,12 @@ class EnvironmentVariables:
     # Whether to run the frontend only. Exclusive with REFLEX_BACKEND_ONLY.
     REFLEX_FRONTEND_ONLY: EnvVar[bool] = env_var(False)
 
+    # The port to run the frontend on.
+    REFLEX_FRONTEND_PORT: EnvVar[int] = env_var(None)
+
+    # The port to run the backend on.
+    REFLEX_BACKEND_PORT: EnvVar[int] = env_var(None)
```
adhami3310, line 551: these should be typed as optional

adhami3310 (approved):

PR #4778, opened 2025-02-08
Add `awaitable_attrs` to rx.Model

masenf:
This works like sqlalchemy's AsyncAttrs mixin, providing an async-context for accessing fields and relationships when using async sessions.


masenf:

PR #4788, opened 2025-02-10
[ENG-4713] Cache pages which add states when evaluating

masenf:
allow the backend to come up without having to re-evaluate all of the pages:

1. cache Upload.is_used
2. cache any route which results in new state classes being defined

Still waiting on reflex-web / flexdown support to ensure frontend/backend states are aligned.


adhami3310:
since we're making another directory here inside of .web, should we start the process towards `.reflex`?

masenf:
> since we're making another directory here inside of `.web`, should we start the process towards `.reflex`?

i dont think we need to get caught up in that yet. this new directory currently only contains compile-time artifacts, so we won't run into the reloading issue because this directory wont be modified at runtime.

but yes ultimately we will need to reorganize at some point.

the other thing that i realized is missing from this PR is that we're not including this new `.web/backend` directory in the backend zip, so it won't really help on the hosting service as is, unless those files get included 🤔

Lendemor (approved):

PR #4790, opened 2025-02-10
Add auto scroll

adhami3310:
```py
import reflex as rx


class State(rx.State):
    messages: rx.Field[list[str]] = rx.field([])

    @rx.event
    def add_message(self):
        self.messages.append(f"Message {len(self.messages) + 1}")


def index():
    return rx.vstack(
        rx.auto_scroll(
            rx.foreach(State.messages, lambda message: rx.text(message)),
            height="300px",
            width="300px",
            border="1px solid red",
        ),
        rx.button("Add Message", on_click=State.add_message),
        padding="200px",
    )


app = rx.App()
app.add_page(index)
```


adhami3310:
we should have this for 0.7.1, it makes making a chat app interface much easier

masenf (commented):
does it work inside an rx.cond where it wont end up getting memoized?

adhami3310:
> does it work inside an rx.cond where it wont end up getting memoized?

it generally works, even without the memoization thing, but that helps specifically to reduce calls to useEffect (which you might have noticed, would run on every re-render)

tgberkeley (approved):

PR #4810, opened 2025-02-12
treat hyphen as underscore in keys of styles

adhami3310:


masenf (commented):
what case is being fixed here?

adhami3310:
> what case is being fixed here?

the console logs of any reflex app is bonkers at the moment, because we used to only convert `snake_case` into `camelCase`, but the sticky had `kebab-case` as its styling, this unifies the conversion with what `emotion` expects

masenf (approved):

PR #4814, opened 2025-02-13
change error connecting to backend when backend is cold started

adhami3310:
Adds `DOES_BACKEND_COLD_START`


masenf (commented):
where does this get used? is it for hosting?

adhami3310:
yes, in theory anyone can set the flag if their choice of hosting involves a cold start

masenf:
If i set this flag though and the backend has crashed or can't be contacted due to incorrect API_URL or some other problem, won't the "Backend is starting" message be somewhat misleading?

adhami3310:
> won't the "Backend is starting" message be somewhat misleading?

yes, for that i'm tagging @Kastier1

adhami3310:
@Kastier1 decided on adding `BACKEND_COLD_START_TIMEOUT`

masenf (approved):
At some point in the complexity, we should probably just implement it in js 😅.

adhami3310:
> we should probably just implement it in js

i think at some point yea, once the plugin system is implemented, all of this stuff should go downstream

reflex/config.py
```diff
@@ -713,6 +713,12 @@ class EnvironmentVariables:
     # Paths to exclude from the hot reload. Takes precedence over include paths. Separated by a colon.
     REFLEX_HOT_RELOAD_EXCLUDE_PATHS: EnvVar[List[Path]] = env_var([])
 
+    # Enables different behavior for when the backend would do a cold start if it was inactive.
+    DOES_BACKEND_COLD_START: EnvVar[bool] = env_var(False)
```
masenf, line 717: Is there a reason these don't have the `REFLEX_` prefix?

adhami3310: yea we should add that

masenf (approved):

PR #4819, opened 2025-02-13
auto hide badge for pro+ users for cloud deployments

Lendemor:
If `show_built_with_reflex` is not specified, it will default to False for pro+ users.
Using that config knob is still reserved for pro+ users, and value will be set to True for free users.

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -108,7 +108,7 @@ from reflex.utils import (
     prerequisites,
     types,
 )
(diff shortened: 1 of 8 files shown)


masenf (dismissed):

masenf (commented):
code looks fine, test needs to be adjusted to set the env var instead of the config knob when simulating a deploy

masenf (approved):

PR #4820, opened 2025-02-13
cache get_type_hints for environment

adhami3310:
get_type_hints is fairly expensive calculation done on a large class, this should make this basically one-time cost thing

### DIFF
--- a/reflex/config.py
+++ b/reflex/config.py
@@ -10,6 +10,7 @@ import os
 import sys
 import threading
 import urllib.parse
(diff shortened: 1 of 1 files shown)


masenf (approved):
should we use a get_type_hints cache in other places?

adhami3310:
> should we use a get_type_hints cache in other places?

eventually yes, but i'm hesitant to do it now, maybe in a follow up PR, once we have Var codspeed tests

PR #4827, opened 2025-02-14
components deserve to be first class props

adhami3310:
alternative to #4823

### DIFF
--- a/reflex/components/base/bare.py
+++ b/reflex/components/base/bare.py
@@ -2,9 +2,9 @@
 
 from __future__ import annotations
 
-from typing import Any, Iterator
+from typing import Any, Iterator, Sequence
 
-from reflex.components.component import Component, ComponentStyle
+from reflex.components.component import BaseComponent, Component, ComponentStyle
 from reflex.components.tags import Tag
 from reflex.components.tags.tagless import Tagless
 from reflex.config import PerformanceMode, environment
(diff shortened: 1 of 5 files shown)


adhami3310:
here's a nice demo (that i stole most of it from masen):

```py
import reflex as rx
from reflex.event import no_args_event_spec


class State(rx.State):
    foo: int = 42

    @rx.event
    def increment(self):
        self.foo += 1


class MyComponent(rx.Fragment):
    def add_custom_code(self) -> list[str]:
        return ["const foo = 'bologna';"]


@rx.memo
def my_memo_has_a_first_name(
    and_its_oscar: rx.Component, on_click: rx.EventHandler[no_args_event_spec]
) -> rx.Component:
    return rx.fragment(
        rx.heading("Hello, Oscar!"),
        rx.button("Click me", on_click=on_click),
        and_its_oscar,
    )


def index() -> rx.Component:
    return my_memo_has_a_first_name(
        and_its_oscar=MyComponent.create(
            rx.text(f"{rx.Var('foo')} is {State.foo} years old.", id="innertxt")
        ),
        on_click=State.increment,
    )


def bare_literal_component_var():
    return rx.fragment(rx.Var.create(rx.text(f"He is {State.foo} years old.")))


app = rx.App()
app.add_page(index)
app.add_page(bare_literal_component_var)
```

masenf (approved):
sweeet!

PR #4837, opened 2025-02-18
expose pages as public api in App()

Lendemor:


masenf (commented):
why do we want this?

* it might not contain all of the pages, depending on when the page is compiled
* it will be empty on the backend, because we evaluate but don't save compiled pages to avoid memory leaks

Lendemor:
> why do we want this?
> 
> * it might not contain all of the pages, depending on when the page is compiled
> * it will be empty on the backend, because we evaluate but don't save compiled pages to avoid memory leaks

Seems like `reflex-clerk` 3rd party package was using `app.pages` which was renamed to `app._pages`

We could let the package fix the problem by using _pages, but if they're going to use it, we might as well expose it in a public API ?

masenf:
reflex-clerk should just be using `app.add_page` which is the public API for adding the page.

what it's doing has been broken for sometime, since we started saving unevaluated pages; the current compile step clears out `app._pages` at the beginning and repopulates it from `app._unevaluated_pages`.

the whole reason `app._pages` is private, is because we don't want third party components and apps messing with it.

masenf:
https://github.com/kroo/reflex-clerk/blob/02540611f00094498dc635ed98a4ac68e19bc420/custom_components/reflex_clerk/lib/reflex_clerk.py#L53

https://github.com/kroo/reflex-clerk/blob/02540611f00094498dc635ed98a4ac68e19bc420/custom_components/reflex_clerk/lib/reflex_clerk.py#L101

Both of these should just be using `add_page`.

PR #4842, opened 2025-02-18
upgrade deps as per python 3.10

adhami3310:
there are lots of stuff that we're doing that we don't need to be doing anymore, this is a collection of such stuff


masenf (commented):
good cleanup.

some of the 3.11+ removals make me a bit nervous, is there a lower-bound 3.10.x release that this requires to possibly get some backported features?

adhami3310:
> some of the 3.11+ removals make me a bit nervous

which ones? as far as i checked, everything should work with 3.10

masenf (approved):
double checked on 3.10.0 and i think it's fine

PR #4847, opened 2025-02-19
Allow any user to set `show_built_with_reflex=False` in any mode

masenf:
Based on discussion with users and contributors, the team has decided not to restrict the `show_built_with_reflex` config option to paid subscribers.

By default, apps will show the branding when exported, but setting the config knob to false is sufficient for removing the badge from the exported app without logging in or subscribing.

### DIFF
--- a/reflex/reflex.py
+++ b/reflex/reflex.py
@@ -23,8 +23,6 @@ typer.core.rich = None  # pyright: ignore [reportPrivateImportUsage]
 cli = typer.Typer(add_completion=False, pretty_exceptions_enable=False)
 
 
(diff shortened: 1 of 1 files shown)


Lendemor (approved):

benedikt-bartscher:
This is awesome, thank you @reflex-dev!

m4thfr34k:
The self hosting [documentation](https://reflex.dev/docs/hosting/self-hosting/#remove-reflex-branding-from-your-self-hosted-app) mentions there is a $50/mo fee to do this. Is this still the case or does the documentation need to be updated?

PR #4848, opened 2025-02-19
react 19 time

adhami3310:
as downstream packages are expecting react 19 more and more, we should upgrade as well

### DIFF
--- a/reflex/.templates/web/components/reflex/radix_themes_color_mode_provider.js
+++ b/reflex/.templates/web/components/reflex/radix_themes_color_mode_provider.js
@@ -10,7 +10,9 @@ import {
 export default function RadixThemesColorModeProvider({ children }) {
   const { theme, resolvedTheme, setTheme } = useTheme();
   const [rawColorMode, setRawColorMode] = useState(defaultColorMode);
-  const [resolvedColorMode, setResolvedColorMode] = useState("dark");
+  const [resolvedColorMode, setResolvedColorMode] = useState(
+    defaultColorMode === "dark" ? "dark" : "light"
+  );
 
   useEffect(() => {
     if (isDevMode) {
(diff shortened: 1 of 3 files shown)


masenf:
For future reference, the test_event_chain.py changes were required due to https://github.com/vercel/next.js/issues/66210#issuecomment-2133687655

Specifically `useEffect` is not invoked twice at hydration time, even with strict mode enabled. After the app is loaded, triggering an unmount/mount, for example by navigating between pages, will trigger the `useEffect` twice. This is rather unfortunate, but apps should be designed to not depend on an effect running exactly some number of times -- effects "should" be idempotent, although in the case the reflex, we're actually doing IO during the effect, so it can never be technically pure.

Some lovely sample code to demonstrate the issue

```python
import reflex as rx


class State(rx.State):
    n_mounts: int = 0
    n_unmounts: int = 0

    def on_mount(self):
        self.n_mounts += 1
        return rx.console_log(f"on_mount:{self.n_mounts}")

    def on_unmount(self):
        self.n_unmounts += 1
        return rx.console_log(f"on_unmount:{self.n_unmounts}")


def index() -> rx.Component:
    return rx.container(
        rx.vstack(
            rx.heading("Welcome to (Un)Mount!", size="9"),
            rx.text(
                "Refreshing the page increments n_mounts by 1. "
                "Navigating to Other and then back increments n_mounts by 2. "
                "Both actions increment n_mounts by 2 in react<19"
            ),
            rx.text(
                f"n_mounts:{State.n_mounts}, n_unmounts:{State.n_unmounts}",
                size="5",
                on_mount=State.on_mount,
                on_unmount=State.on_unmount,
            ),
            rx.link("Other", href="/other-page"),
        ),
        rx.logo(),
    )


def other_page():
    return rx.vstack(
        rx.heading("Other Page"),
        rx.link("Index", href="/"),
    )


app = rx.App()
app.add_page(index)
app.add_page(other_page)
```

masenf (approved):
git-r-dun 🚢

PR #4852, opened 2025-02-20
simplify and fix set_color_mode

adhami3310:
still not callable from the backend, because it has no context for "setColorMode"


masenf (approved):
let's take this for 0.7.1 as well

PR #4853, opened 2025-02-20
simplify toast banner logic

adhami3310:


reflex/components/sonner/toast.py
```diff
@@ -277,6 +274,13 @@ def send_toast(
         else:
             toast = toast_command.call(message)
 
+        if fallback_to_alert:
+            toast = ternary_operation(
+                toast_ref.bool(),
+                toast,
+                FunctionVar("window.alert").call(message),
```
masenf, line 281: should we replace the `<br/>` with `\n` when we're doing an alert?

adhami3310: good point, but message doesn't contain `<br/>`, it only contains a str, description seems a bit more weird, i will try it out

adhami3310: window.alert doesn't really have a form of description, so we could ignore it if that's ok :/

masenf (approved):

PR #4858, opened 2025-02-21
fix autoscroll on stateful children

adhami3310:


adhami3310:
tested with
```py
import reflex as rx

app = rx.App()


class State(rx.State):
    field: str = "Hello, World!"

    @rx.event
    def double_field(self) -> None:
        self.field *= 2


@app.add_page
def index():
    return rx.vstack(
        rx.box(
            rx.auto_scroll(
                rx.text(State.field),
                height="100%",
                width="100%",
                overflow="auto",
            ),
            height=200,
            width=200,
        ),
        rx.button("Double", on_click=State.double_field),
    )
```

tgberkeley (approved):

PR #4867, opened 2025-02-24
put default values for computed var in case it errors

### LINKED ISSUE #4864: Exception in user code in an rx.var is suppressed and difficult to debug from the downstream exception results
**Describe the bug**
If a user has an exception in his code for an rx.var, and that var is used in a foreach, then the error that will be reported will be very difficult to relate back to the user's exception. The actual exception displayed will be something like:

[Reflex Frontend Exception]
 TypeError: Cannot read properties of undefined (reading 'map')

**To Reproduce**
Here is a trivial app to reproduce the issue with an obvious divide by zero exception in the user code

```
import reflex as rx


class State(rx.State):
    @rx.var
    def some_vars(self) -> list[str]:
        # test = 3 / 0
        # print(test)
        return ["3", "2", "1"]


def index() -> rx.Component:
    return rx.container(
        rx.text("Should see a list below:"),
        rx.vstack(
            rx.foreach(State.some_vars, rx.text),
            spacing="5",
            justify="center",
        ),
    )


app = rx.App()
app.add_page(index)

```
If you "reflex run" the above code you see a simple page. If you uncomment the two comment lines which will cause an exception in the user's code, you'll get a very difficult to diagnose [Reflex Frontend Exception]

**Expected behavior**
The actual python exception raised in the user's code should be communicated to the user in some fashion.

 - Python Version: 3.13
 - Reflex Version: 0.1.7
 - OS:  Windows 11

adhami3310:
fixes #4864 (makes it easier to debug)

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -986,12 +986,15 @@ class App(MiddlewareMixin, LifespanMixin):
 
     def _setup_sticky_badge(self):
         """Add the sticky badge to the app."""
-        for k, component in self._pages.items():
-            # Would be nice to share single sticky_badge across all pages, but
-            # it bungles the StatefulComponent compile step.
+        from reflex.components.component import memo
+
+        @memo
+        def memoized_badge():
             sticky_badge = sticky()
             sticky_badge._add_style_recursive({})
-            self._pages[k] = Fragment.create(sticky_badge, component)
+            return sticky_badge
+
+        self.app_wraps[(0, "StickyBadge")] = lambda _: memoized_badge()
 
     def _apply_decorated_pages(self):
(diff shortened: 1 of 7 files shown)


benedikt-bartscher:
I do not think this actually improves the debug experience. This just hides the error even more. If a computed var cannot be computed during compile time (f.e. due to a service which is not accessible or empty db) we should suggest users using the `initial_value` attribute of the rx.var

masenf:
> We honestly could raise an actual error here

I think this is the move, and then the traceback will point directly into the user's code without ambiguity.

Lendemor (approved):

PR #4872, opened 2025-02-25
reduce badge perf impact on pages

Lendemor:
The sticky badge showing Built with Reflex is now added as an app_wrap instead of injected in each page, reducing the impact on compilation length.


masenf (approved):
much nice

PR #4878, opened 2025-02-26
remove temporary pin on radix ui themes

adhami3310:
added in #4370, but radix-ui/themes#627 has been since resolves


masenf (approved):
both dialog flicker and scroll area issues are resolved in 3.2.0 based on testing in `ci_template` and `customer_data_app` templates

PR #4890, opened 2025-03-01
json patch

adhami3310:


reflex/state.py
```diff
@@ -1888,9 +2014,12 @@ def _dirty_computed_vars(
             if include_backend or not self.computed_vars[cvar]._backend
         }
 
-    def get_delta(self) -> Delta:
+    def get_delta(self, *, token: str | None = None) -> StateDelta:
         """Get the delta for the state.
 
+        Args:
+            token: The reflex delta
```
masenf, line 2021: 🤔

reflex/state.py
```diff
@@ -108,10 +109,135 @@
...
-Delta = dict[str, Any]
 var = computed_var
 
 
+@dataclasses.dataclass
+class StateDelta:
+    """A dictionary representing the state delta."""
+
+    data: dict[str, Any] = dataclasses.field(default_factory=dict)
+    reflex_delta_token: str | None = dataclasses.field(default=None)
```
masenf, line 120: seems like `client_token` would be more aligned with the existing code. unless i'm misunderstanding what this field is for

reflex/state.py
```diff
@@ -108,10 +109,135 @@
...
+    def items(self):
+        """Get the items of the delta.
+
+        Returns:
+            The items of the delta.
+        """
+        return self.data.items()
+
+
+LAST_DELTA_CACHE: dict[str, StateDelta] = {}
```
masenf, line 211: if we don't save this in redis, then we could get bugs in prod mode if a subsequent request (like upload) comes through a different app instance.

masenf:

Lots of conflict and perhaps not compatible with upcoming changes to use replicache-style syncing 🤷 . Either way, rewrite will be necessary 😂

PR #4898, opened 2025-03-04
treat component prop class as immutable when provided a non literal var

adhami3310:


masenf (commented):
looks good, needs test case

masenf (approved):

PR #4904, opened 2025-03-06
deprecate Component __init__

adhami3310:


masenf (commented):
the pyi changes shouldn't be here. we were explicitly ignoring these before, they should be ignored now

(i tried to debug pyi_generator and determine where these were coming from, but i couldn't tell)

PR #4906, opened 2025-03-06
double down on bun over fnm/npm

adhami3310:
makes initial setup faster (hopefully)


adhami3310:
reflex-web is failing because of oven-sh/bun#17014

masenf (commented):
my system had an old version of node, that got picked up by `bun run`. we should use `bun run --bun` to make sure it uses bun runtime

adhami3310:
the issue with not using node is that reflex web would fail to compile :/

Alek99:
> the issue with not using node is that reflex web would fail to compile :/

Is this just because of spline?

adhami3310:
> Is this just because of spline?

It is because of *bun* failing to compile an app with spline.

masenf (approved):

PR #4914, opened 2025-03-07
fix: every computed var should get it's own static deps

benedikt-bartscher:


masenf (approved):

benedikt-bartscher:
Thanks for the quick merge @masenf

PR #4929, opened 2025-03-08
Prepend hydrate event when connecting

masenf:
Avoid backend on_mount events triggering a `reload` message, which ends up requeuing hydrate event.

```
import reflex as rx


class State(rx.State):
    def fooooooo(self):
        print("Foooo!")

    def on_load(self):
        print("on_load scrode")


def index() -> rx.Component:
    return rx.box("Hola Mundo", on_mount=[State.fooooooo, State.fooooooo, State.fooooooo])


app = rx.App()
```

Before the fix, this code triggers on_load four times 🫨


benedikt-bartscher:
> Before the fix, this code triggers on_load four times 🫨

...and after it only once, awesome!

adhami3310 (approved):

PR #4933, opened 2025-03-10
use noopener for window.open

adhami3310:
resolves such code from printing the same client token after being redirected in a new tab
```py
import reflex as rx


class State(rx.State):
    @rx.event
    def what_is_my_token(self):
        print(self.router.session.client_token)


def index() -> rx.Component:
    return rx.vstack(
        rx.button("What is my token?", on_click=State.what_is_my_token),
        rx.button("Redirect to /", on_click=rx.redirect("/", is_external=True)),
    )


app = rx.App()
app.add_page(index)
```

### DIFF
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -179,7 +179,7 @@ export const applyEvent = async (event, socket) => {
   // Handle special events
   if (event.name == "_redirect") {
     if (event.payload.external) {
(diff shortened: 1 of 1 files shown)


adhami3310:
i swear whenever i touch a JS file we get new formatting changes, we should at some point add prettier to our precommit, but i don't want to

Alek99 (approved):

PR #4941, opened 2025-03-11
prettier

adhami3310:


masenf (commented):
plz add it to pre-commit so we don't regress

masenf (approved):

PR #4944, opened 2025-03-11
update pyproject.toml

adhami3310:


masenf:
this would probably be a good time to do the `poetry-dynamic-versioning` as well, we have an example now on the flexdown and enterprise repos

masenf (approved):
good cleanup

PR #4953, opened 2025-03-12
Disconnect old websockets

masenf:
Disconnect websockets more aggressively
* page beforeunload and unload
* ~when a new event comes in for the same token with a different sid~ (removed in this PR)

### DIFF
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -446,6 +446,13 @@ export const connect = async (
     }
   }
 
+  const disconnectTrigger = (event) => {
+    if (socket.current?.connected) {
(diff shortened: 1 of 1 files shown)


Lendemor:
Related to #4607 I think.

Disconnecting on new sid for same token means the original page will get disconnected?

masenf:
> Disconnecting on new sid for same token means the original page will get disconnected?

yes; this is intended to handle the case where page is refreshed (in an iframe) or hot reloaded and ends up with a new websocket, but never closes out the old one.

that said, i think the frontend logic might be good enough, and i wasn't actually able to hit that log message during actual testing.

adhami3310 (approved):

PR #4982, opened 2025-03-18
add libsass to the dev deps

adhami3310:


masenf (dismissed):

tests/units/compiler/test_compiler.py
```diff
@@ -145,12 +147,12 @@ def test_compile_stylesheets(tmp_path: Path, mocker):
...
 def test_compile_stylesheets_scss_sass(tmp_path: Path, mocker):
-    try:
-        import sass  # noqa: F401
-    except ImportError:
+    if importlib.util.find_spec("sass") is None:
         pytest.skip(
             'The `libsass` package is required to compile sass/scss stylesheet files. Run `pip install "libsass>=0.23.0"`.'
         )
+    if os.name == "nt":
+        pytest.skip("Skipping test on Windows")
```
masenf, line 155: the typical way to do conditional skips is with the `skipif` mark: https://docs.pytest.org/en/stable/how-to/skipping.html#skip-all-test-functions-of-a-class-or-module

```python
@pytest.mark.skipif(sys.platform == "win32", reason="does not run on windows")
```

masenf (approved):

tests/units/compiler/test_compiler.py
adhami3310: sounds good, i will get it in a later cleanup

PR #4923, opened 2025-03-07
Generate sitemap for dynamic deploy url

bhatia2akshit:
### All Submissions:

- [ ] Have you followed the guidelines stated in [CONTRIBUTING.md](https://github.com/reflex-dev/reflex/blob/main/CONTRIBUTING.md) file?
- [ ] Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/reflex-dev/reflex/pulls ) for the desired changed?

### Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### New Feature Submission:

- [ ] Does your submission pass the tests? 
- [ ] Have you linted your code locally prior to submission?

### Changes To Core Features:

- [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your core changes, as applicable?
- [ ] Have you successfully ran tests with your changes locally?

### **After** these steps, you're ready to open a pull request.

    a. Give a descriptive title to your PR.

    b. Describe your changes.

    c. Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such).

PR #4984, opened 2025-03-18
remix over next

adhami3310:
almost, usable, but needs to port 404, figure out titles and such, make sure events aren't broken, and bring back sitemap, among other things

### DIFF
--- a/reflex/.templates/jinja/web/package.json.jinja2
+++ b/reflex/.templates/jinja/web/package.json.jinja2
@@ -1,5 +1,6 @@
 {
   "name": "reflex",
+  "type": "module",
   "scripts": {
     "dev": "{{ scripts.dev }}",
     "export": "{{ scripts.export }}",
--- a/reflex/.templates/jinja/web/pages/_app.js.jinja2
+++ b/reflex/.templates/jinja/web/pages/_app.js.jinja2
@@ -6,8 +6,9 @@ import '$/styles/styles.css'
 {% endblock %}
 
 {% block declaration %}
-import { EventLoopProvider, StateProvider, defaultColorMode } from "$/utils/context.js";
-import { ThemeProvider } from 'next-themes'
+import { EventLoopProvider, StateProvider, defaultColorMode } from "$/utils/context";
+import { ThemeProvider } from '$/utils/react-theme';
+import { Outlet, Scripts, ScrollRestoration } from 'react-router';
 {% for library_alias, library_path in  window_libraries %}
 import * as {{library_alias}} from "{{library_path}}";
 {% endfor %}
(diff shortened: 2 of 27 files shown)


Alek99:
:)

masenf (commented):
still a lot more to go on this, but i like the direction.

- [x] internal nav without breaking websocket
- [ ] support dynamic routes (and hopefully translate existing syntax)
- [ ] pre-render HTML and use hydrateRoot

In the future it would also be awesome to expose an interface for devs to specify additional dynamic routes that should be pre-rendered.

Some grunt work to get the tests passing, but i think we should try to keep this branch up to date with `main` and try to move over as soon as its stable and has feature parity.

It would be nice if things were implemented in a way that would allow us to support both, but that kind of refactoring work is probably not something we can prioritize over the wins from simplifying and shrinking our frontend stack, so probably makes sense to just move fast at this point and hard cut away from next.

adhami3310:
> internal nav without breaking websocket
> support dynamic routes (and hopefully translate existing syntax)

i was a bit stuck on these two, the code for them was very tied to nextjs way of doing things, so help there is appreciated

> pre-render HTML and use hydrateRoot

we do have an SSR option but i don't know how much it does

there are also other things:

- [x] titles/meta tags aren't working quite the same in nextjs
- [x] sitemap needs to be restored (#4923 should handle this)
- [x] dynamic rendering needs to be moved off of nextjs syntax

adhami3310:
Here's the current checklist:

- [x] implement sitemap functionality
- [ ] fix the two integration tests for event actions and event chains
- [x] verify things like dynamic routes are largely working (we currently only support [arg], not the other things like catchall)

masenf (commented):
in dev mode, this file is served publicly as `/reflex.json`, and it probably shares too much information.

```
export const version = "0.8.0.dev1";
export const project_hash = 7.572570981449851e+37;
export const last_version_check_datetime = "2025-06-10 22:01:20.356475";
export const last_reflex_run_datetime = "2025-06-10 22:01:23.733495";
export default {
  version,
  project_hash,
  last_version_check_datetime,
  last_reflex_run_datetime,
};
```

adhami3310:
> and it probably shares too much information

isn't that already happening? state.js imports it so it has to get it (in dev and prod, but in prod it prob gets a tad but more bundled)

masenf:
reflex-chakra no longer working

```
  File "/Users/masenf/code/reflex-dev/reflex/.venv/lib/python3.13/site-packages/reflex_chakra/components/navigation/breadcrumb.py", line 4, in <module>
    from reflex_chakra.components.navigation.link import Link
  File "/Users/masenf/code/reflex-dev/reflex/.venv/lib/python3.13/site-packages/reflex_chakra/components/navigation/link.py", line 5, in <module>
    from reflex.components.next.link import NextLink
ModuleNotFoundError: No module named 'reflex.components.next.link'
```

masenf (approved):
🤩 🔥 🚀

PR #4990, opened 2025-03-18
handle range value types [ENG-5143]

### LINKED ISSUE #4989: `rx.foreach` over range gives Var with incorrect type
**Describe the bug**
```python
import reflex as rx

def wtf(ix):
    print(repr(ix))
    return rx.text(ix)

_ = rx.foreach(range(1), wtf)
```

**Expected behavior**

0.7.1 behavior

```python
Var.__init_subclass__.<locals>.ToNumberOperation(_js_expr='ix', _var_type=<class 'int'>, _var_data=None, _original=Var(_js_expr='ix', _var_type=<class 'int'>, _var_data=None))
```

Currently on `0.7.2`+ the var loses its type

```python
Var(_js_expr='ix', _var_type=typing.Any, _var_data=None)
```

- Python Version: 3.13
- Reflex Version: 0.7.2 (regression from 0.7.1)

adhami3310:
fixes #4989

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -1671,6 +1671,7 @@ class BaseState(Base, ABC, extra=pydantic.Extra.allow):
 
         raise TypeError(
             f"Your handler {handler.fn.__qualname__} must only return/yield: None, Events or other EventHandlers referenced by their class (i.e. using `type(self)` or other class references)."
+            f" Returned events of types {', '.join(map(str, map(type, events)))!s}."
         )
 
     async def _as_state_update(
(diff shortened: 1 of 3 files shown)


masenf (commented):
i'm not sure this is the fix we want.

the way the code was written before, a `range` got translated into a JS expression that evaluated to an Array on the frontend.

with this code, the python `range` will be expanded on the python side into a LiteralArrayVar of a list of all items in the range, which is what you get when using `ArrayVar.range`.

adhami3310:
@masenf what do you think about using rangevar

masenf (approved):

PR #4993, opened 2025-03-19
Allow multiple env_file to be specified, split on os pathsep

masenf:


adhami3310:
we already have some logic for environment variables, we use `:` for all of them, do you want to integrate the logic together?

masenf:
@adhami3310 so i tried to be a good citizen in the codebase and reuse the splitting function, but it actually screws up windows paths 😔 

https://github.com/reflex-dev/reflex/actions/runs/13956577429/job/39069078707

The `tmp_path` has a colon in it

```
'C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-0/test_env_file_file_map2__path_0'
```

Should we fix the other ones to use an os-dependent separator?

```python
    # Additional paths to include in the hot reload. Separated by a colon.
    REFLEX_HOT_RELOAD_INCLUDE_PATHS: EnvVar[list[Path]] = env_var([])

    # Paths to exclude from the hot reload. Takes precedence over include paths. Separated by a colon.
    REFLEX_HOT_RELOAD_EXCLUDE_PATHS: EnvVar[list[Path]] = env_var([])
```

Seems like these would be hard to use with an absolute path on windows

adhami3310:
> Should we fix the other ones to use an os-dependent separator

i think s

masenf:
kept the `os.pathsep` for the env_file. will fix the other ones in a subsequent PR

adhami3310 (approved):

PR #5006, opened 2025-03-21
make pyright happier

adhami3310:
this is not useful per se, but i'm noticing some stuff that can be problematic at times


reflex/vars/base.py
```diff
@@ -1058,7 +1060,9 @@ def bool(self) -> BooleanVar:
 
         return boolify(self)
 
-    def __and__(self, other: Var | Any) -> Var:
+    def __and__(
+        self, other: Var[OTHER_VAR_TYPE] | Any
+    ) -> Var[VAR_TYPE | OTHER_VAR_TYPE]:
```
masenf, line 1065: is this correct? i thought this operation returned a bool-like

masenf (approved):
lots of good cleanup here 🙌

reflex/vars/base.py
adhami3310: yes, `&` is operating like python `and`and javascript `&&`

PR #5013, opened 2025-03-22
disable ipv6

adhami3310:


adhami3310:
we should add an escape hatch for this one

masenf (commented):
my only hesitation with this one, is what happens if the user only has ipv6? (i know very rare and impractical in today's world, but we're possibly introducing tech debt)

masenf (approved):
confirmed working on ipv6 only and ipv6 dual stack network

PR #5019, opened 2025-03-24
add ndigits to round dunder method

adhami3310:


masenf (dismissed):
i recommend adding a few lines to `test_var_operations.py` just to make sure we have some coverage for this

masenf (approved):

PR #5021, opened 2025-03-25
allow arguments to be passed to rx memo event handlers

### LINKED ISSUE #4985: Passing event handlers with arguments with rx.memo
**Describe the bug**
Using event handlers that take arguments does not work in cunjunction with rx.memo. If the event handler is passed to e.g. an on_change of an input, the frontend throws an error. If the event handler is used with partial argument, the compilation already fails.

**To Reproduce**

```python
import reflex as rx
from reflex.event import passthrough_event_spec

class State(rx.State):
    @rx.event
    async def state_event(self, value: str) -> None:
        print(f"Passed {value} to event")

# This works with the folling line commented out
@rx.memo
def my_memoed_component(
    some_value: str,
    event: rx.EventHandler[passthrough_event_spec(str)],
) -> rx.Component:
    return rx.vstack(
        # Fails at compilation here
        # with a
        #  ValueError: Invalid event chain: ((...args) => (event(someValue, ...args))) of type typing.Callable
        rx.button(some_value, on_click=event(some_value)),
        # If the line above is commented out, compilation works, but
        # then the page throws an exception on the frontend:
        #  Unhandled Runtime Error
        #  RangeError: Maximum call stack size exceeded
        #  hasBinary
        #  node_modules/socket.io-parser/build/esm/is-binary.js (24:17)
        #  ...
        # when entering text in the field.
        rx.input("Your string:", on_change=event)
    )

def index() -> rx.Component:
    return rx.hstack(my_memoed_component(some_value="some_value", event=State.state_event))

app = rx.App()
app.add_page(index, "/")
```

**Expected behavior**
I would expect that rx.memo works with passed event handlers from State and ComponentState classes and they can be used as when using a function component.

- Python Version: 3.13
- Reflex Version: 0.7.3
- OS: mac OS 15.3
- Browser (Optional): Arc

adhami3310:
fixes #4985

### DIFF
--- a/reflex/compiler/utils.py
+++ b/reflex/compiler/utils.py
@@ -318,7 +318,7 @@ def compile_custom_component(
     }
 
     # Concatenate the props.
-    props = [prop._js_expr for prop in component.get_prop_vars()]
+    props = list(component.props)
 
     # Compile the component.
     return (
--- a/reflex/components/component.py
+++ b/reflex/components/component.py
@@ -48,12 +48,13 @@ from reflex.constants import (
 from reflex.constants.compiler import SpecialAttributes
 from reflex.constants.state import FRONTEND_EVENT_STATE
 from reflex.event import (
-    EventActionsMixin,
     EventCallback,
     EventChain,
     EventHandler,
     EventSpec,
     no_args_event_spec,
+    parse_args_spec,
+    run_script,
 )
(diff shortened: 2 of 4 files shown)


masenf (approved):
added an integration test from the original bug

masenf (approved):

PR #5032, opened 2025-03-26
allow event handlers to be turned into vars

adhami3310:
```py
import reflex as rx


class SnakeState(rx.State):
    game_over = False

    @rx.event
    def start_game(self):
        self.game_over = False

    @rx.event
    def pause_game(self):
        self.game_over = True


def index():
    return rx.el.button(
        rx.cond(
            SnakeState.game_over,
            "Start Game",
            "Pause Game",
        ),
        on_change=rx.cond(
            SnakeState.game_over, SnakeState.start_game, SnakeState.pause_game
        ),
    )


app = rx.App()

app.add_page(index, route="/")
```

### DIFF
--- a/reflex/event.py
+++ b/reflex/event.py
@@ -1607,7 +1607,7 @@ def get_fn_signature(fn: Callable) -> inspect.Signature:
     return signature.replace(parameters=(new_param, *signature.parameters.values()))
 
 
-class EventVar(ObjectVar, python_types=EventSpec):
+class EventVar(ObjectVar, python_types=(EventSpec, EventHandler)):
     """Base class for event vars."""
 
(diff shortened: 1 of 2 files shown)


masenf (approved):
tentatively approve as it just didn't work at all before. but as discussed, i still think it's a bit of an API discrepancy that the trigger signature doesn't get passed to the event handler when used in a cond.

Workaround i suppose would be to pass an event chain where the first event stores the value where each of the possible event handlers could access it

adhami3310:
> Workaround i suppose would be to pass an event chain where the first event stores the value where each of the possible event handlers could access it

sounds like var data?

PR #5043, opened 2025-03-27
allow none as a child of components

adhami3310:


masenf (commented):
we probably should add an integration test, to make sure that when we switch frameworks or stop using jsx or whatever that pass `null` as a component child still works

masenf (approved):

PR #5064, opened 2025-03-31
reload config on __post_init__

adhami3310:
we were also making a large string for no reason

### DIFF
--- a/tests/integration/test_extra_overlay_function.py
+++ b/tests/integration/test_extra_overlay_function.py
@@ -11,8 +11,6 @@ from reflex.testing import AppHarness, WebDriver
 def ExtraOverlay():
     import reflex as rx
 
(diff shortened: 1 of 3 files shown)


tests/integration/test_extra_overlay_function.py
```diff
@@ -26,6 +24,7 @@ def index():
         )
 
     app = rx.App()
+    rx.config.get_config().extra_overlay_function = "reflex.components.moment.moment"
```
masenf, line 27: i'm assuming it still works if we, for example, added this inside the `index()` page function

masenf (approved):

tests/integration/test_extra_overlay_function.py
adhami3310: yea, we just have to make after app has been inited, we do something similar for the tailwind test

PR #5066, opened 2025-04-01
only include devIndicators if it's False

adhami3310:


masenf (approved):
nice i like this fix!

PR #5074, opened 2025-04-01
remove pyi files

adhami3310:


hatch_build.py
```diff
@@ -0,0 +1,35 @@
...
+        if not (pathlib.Path(self.root) / "scripts").exists():
+            return
+
+        for file in pathlib.Path(self.root).rglob("**/*.pyi"):
+            file.unlink(missing_ok=True)
+
+        import subprocess
+
+        subprocess.run(
+            ["python", "-m", "reflex.utils.pyi_generator"],
```
masenf, line 33:
```suggestion
            import sys
            [sys.executable, "-m", "reflex.utils.pyi_generator"],
```

must use the same python executable running the script when chaining execution.

this fixes the build issue on windows

masenf (commented):
i like this approach, but i feel like it will be harder now to notice unanticipated changes in our pyi files. idk what to do about that. but like for example if we add a field to `rx.Component`, and forget to update `pyi_generator` to ignore that field, now all of a sudden it appears in all the pyi files that we ship, but we'll basically never notice.

adhami3310:
> now all of a sudden it appears in all the pyi files that we ship, but we'll basically never notice

that's a good point and something i was thinking

here's a wild idea, we can store the hash of all pyi files and their paths and we have to change that file manually if the changes in that PR has updated a PYI file

masenf (approved):

PR #5079, opened 2025-04-02
Add value and default_value to rx.el.select

masenf:
correcting an apparent miss from way back

(opening PR without `pyi_hashes.json` change to see if it catches the diff)


adhami3310:
pre-commit works :fireworks: (which means it doesn't work)

adhami3310:
locally, pre-commit fails :thinking:

adhami3310 (approved):

PR #5091, opened 2025-04-04
Added param context to add_page for extended functionality for plugins, flows, etc.

hjpr:
Last commit was failing for unknown reason. Reforked and changed only 4 lines, reran per contributing.md.

Added context to add_page function so that we can grab page-specific information in state or middleware for custom flows like auth or redirects.


hjpr:
so when running pre-commit it changes a bunch of hashes in a .pyi file. when i git add . it stages a .pyi file as well to push

should i only be committing the app.py file and ignoring that .pyi file?

hjpr:
maybe that was it

hjpr:
its automated when i run the precommit, it fails on my end until it rewrites all the hashes and then afterwards it passes. i just assumed that needed to get pushed as well

adhami3310 (approved):

adhami3310:
> its automated when i run the precommit, it fails on my end until it rewrites all the hashes and then afterwards it passes. i just assumed that needed to get pushed as well

it seems the pyi_generator on your end produces a different code, what version of python are you using?

adhami3310:
ok i see the issue, python 3.10-3.12 produces a different hash compared to 3.13 which the CI runs, i didn't notice because i was using 3.13

adhami3310:
image

PR #5098, opened 2025-04-07
allow int and float typing for input elements behind a warning

adhami3310:


tgberkeley:
Can we also add `bool` as I hit the same error as above with checkbox

adhami3310:
> Can we also add bool as I hit the same error as above with checkbox

Checkbox always have the same value, on_change is actually fairly useless there, it always has the value `on` (it's actually more complicated, but checkbox has the same `.value`, but not the same `.checked`)

adhami3310:
blocked on reflex-dev/reflex-web#1268

Alek99 (approved):

masenf (approved):

PR #5112, opened 2025-04-10
'ImmutableMutableProxy' object cannot be converted to 'PyDict' with P…

### LINKED ISSUE #5103: 'ImmutableMutableProxy' object cannot be converted to 'PyDict' with Pydantic v2
**Describe the bug**

Manipulating any Pydantic model type from a state raise a library error `pydantic_core._pydantic_core.PydanticSerializationError: ... 'ImmutableMutableProxy' object cannot be converted to 'PyDict'`.

**To reproduce**

- Define two Pydantic v2 models, `A` and `B`, `B` a child if the first
- Assign `B` to the app state as `b: B = B()`
- Execute `TypeAdapter(A).dump_json(self.b)`
- See the error

**Expected behavior**

Any manipulation of a Pydantic model inside the state work like it should out of the state.

**Context**

- Python version: v3.13
- Reflex version: v0.7.0
- Pydantic version: v2.10.6
- OS: Debian Bookworm
- Browser (Optional): n/a

kshitiz305:
…ydantic v2 

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

    a. 'ImmutableMutableProxy' object cannot be converted to 'PyDict' with Pydantic v2 

    b. modified reflex/state.py to add a __pydantic_serializer__ method to the MutableProxy class. This method ensures that when Pydantic v2 attempts to serialize the proxy object, it receives the underlying Pydantic model instance, resolving the incompatibility.

    c. Closes #5103

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -94,6 +94,12 @@ from reflex.utils.types import (
     true_type_for_pydantic_field,
     value_inside_optional,
 )
+try:
+    # Pydantic v2 SerializationInfo
+    from pydantic import SerializationInfo
+except ImportError:
+    SerializationInfo = Any  # type: ignore
+
 from reflex.vars import VarData
 from reflex.vars.base import (
     ComputedVar,
@@ -4023,6 +4029,10 @@ class MutableProxy(wrapt.ObjectProxy):
             Tuple of (wrapped class, empty args, class __getstate__)
         """
         return self.__wrapped__.__reduce_ex__(protocol_version)
+
+    def __pydantic_serializer__(self, info: SerializationInfo):
+        # Return the underlying object for Pydantic v2 serialization
+        return self.__wrapped__
 
 
 @serializer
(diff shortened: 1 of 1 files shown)


reflex/state.py
```diff
@@ -94,6 +94,12 @@
     true_type_for_pydantic_field,
     value_inside_optional,
 )
+try:
+    # Pydantic v2 SerializationInfo
+    from pydantic import SerializationInfo
```
adhami3310, line 99: it seems this is only relevant for type checking, so you can do
```py
if TYPE_CHECKING:
    from pydantic import SerializationInfo
```
where `TYPE_CHECKING` is imported from `typing`. No need for try catches there.

kshitiz305: I'll update reflex/state.py to reflect this improvement.

Import TYPE_CHECKING from typing.
Replace the try...except block with if TYPE_CHECKING:.

masenf (commented):
why is this patch removing `set_state` and changing the signature of `get_state`?

PR #5113, opened 2025-04-10
dynamic versioning and auto-publish on release

Lendemor:
Note: Need to configure the PYPI token for publishing in secrets before merging.

Converted to Draft: Doesn't seems to work with release branches.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,25 @@
+name: Build and Publish
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+
+permissions:
+  contents: read # Required for checkout
+  id-token: write # For trusted publishing if using OIDC
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - uses: ./.github/actions/setup_build_env
+        with:
+          python-version: "3.10"
```
adhami3310, line 20: the reflex repo requires 3.13 for development now (because of how the ast for pyi files work)

adhami3310:
so how does one make a release with this? do we have to make pre-releases in the release tab as well?

Lendemor:
> so how does one make a release with this? do we have to make pre-releases in the release tab as well?

Yeah, just make a release in the Github interface.
Pre-release are the same, you make a tag `0.X.YaZ` and check the "mark as pre-release"

PR #5120, opened 2025-04-11
Read environment secrets in state at import time [ENG-5426] [ENG-5455]

adhami3310:


masenf (commented):
what sucks about this, is that someone is going to put `env_file` in their `rxconfig.py` and wonder why it's values are not applied when the app modules are imported...

is there anything we can reasonably do about that without calling `get_config` at module time tho?

Lendemor (approved):

PR #5127, opened 2025-04-14
don't use jsx syntax

adhami3310:


itsmeadarsh2008:
What about renaming the files to `*.jsx`?

adhami3310:
> What about renaming the files to *.jsx?

The issue is not that particularly. It's that JSX is rather oriented towards user-readability which is fairly very low on the things we care about on computer generated code.

The strong reason for moving away from JSX is that it breaks JS into two lands, JSX and JS, and we have to have interactions between the two very carefully. If there's no division between the two then code gets a bit easier to maintain.

reflex/.templates/jinja/web/pages/utils.js.jinja2
```diff
@@ -1,116 +1,79 @@
...
 {#     component: component dictionary #}
 {% macro render_tag(component) %}
-<{{component.name}} {{- render_props(component.props) }}>
-{{ component.contents }}
-{% for child in component.children %}
-{{ render(child) }}
-{% endfor %}
-</{{component.name}}>
+jsx(
+{% if component.name|length %}{{ component.name }}{% else %}Fragment{% endif %},
```
masenf, line 30: do we for sure always have `Fragment` in scope? seems like yes, because we wrap all page components in Fragment

reflex/compiler/compiler.py
```diff
@@ -39,8 +39,10 @@ def _compile_document_root(root: Component) -> str:
     Returns:
         The compiled document root.
     """
+    document_root_imports = root._get_all_imports()
+    document_root_imports.setdefault("@emotion/react", []).append(ImportVar("jsx"))
```
masenf, line 43: feels like this should get pulled out into an `apply_common_imports` helper or something to reduce the diff when we may eventually drop this helper and move to `createElement` everywhere

reflex/components/base/bare.py
```diff
@@ -171,9 +172,9 @@ def _get_all_refs(self) -> set[str]:
     def _render(self) -> Tag:
         if isinstance(self.contents, Var):
             if isinstance(self.contents, (BooleanVar, ObjectVar)):
-                return Tagless(contents=f"{{{self.contents.to_string()!s}}}")
-            return Tagless(contents=f"{{{self.contents!s}}}")
-        return Tagless(contents=str(self.contents))
+                return Tagless(contents=f"{self.contents.to_string()!s}")
+            return Tagless(contents=f"{self.contents!s}")
+        return Tagless(contents=f'"{json.dumps(self.contents)}"')
```
masenf, line 177: what circumstance hits this case? and do we have a test for it?

the code looks weird, like what happens if `self.contents` is a str?

```python
>>> s = "hello world"; f'"{json.dumps(s)}"'
'""hello world""'
```

reflex/components/component.py
```diff
@@ -268,6 +268,11 @@ class Component(BaseComponent, ABC):
     # The alias for the tag.
     alias: str | None = pydantic.v1.Field(default_factory=lambda: None)
 
+    # Whether the component is a global scope tag. True for tags like `html`, `head`, `body`.
+    _is_tag_in_global_scope: bool = pydantic.v1.PrivateAttr(
```
masenf, line 272: why not a ClassVar?

reflex/.templates/jinja/web/pages/utils.js.jinja2
adhami3310: i could do the thing where i add Fragment all the time like jsx, would that be good?

reflex/compiler/compiler.py
adhami3310: that makes sense :+1:

reflex/components/base/bare.py
adhami3310: lmao yea that seems wrong, it should be a simple json dumps no need for params, i think there's hardly a way to enter that path but it's there for safety kinda thing

reflex/components/component.py
adhami3310: i wouldn't know (read: pydantic kinda destroyed my sense of what is supposed to be a classvar, but thinking about it, yea that sounds like something that should be a classvar)

reflex/.templates/jinja/web/pages/utils.js.jinja2
masenf: if we're going to have it in the template, then we should make sure it can never blow up due to missing import

masenf (approved):

PR #5148, opened 2025-04-18
fix: pin radix-ui to 1.2.0

benedikt-bartscher:
mitigate https://github.com/radix-ui/primitives/issues/3485


adhami3310:
by the time we get this merge they prob would have fixed it (i hope)

chaance:
Should be fixed in 1.3.1

benedikt-bartscher:
for now, they only have published a release candidate [1.3.1-rc.1744998730501](https://www.npmjs.com/package/radix-ui/v/1.3.1-rc.1744998730501)

Alek99:
> Should be fixed in 1.3.1

Thanks!

PR #5154, opened 2025-04-19
click over typer

adhami3310:
needs release from hosting-cli


reflex/custom_components/custom_components.py
```diff
@@ -10,15 +10,35 @@
...
+    pass
+
+
+def _set_loglevel_from_cli(loglevel: str | None):
+    console.set_log_level(
+        constants.LogLevel.from_string(loglevel) or get_config().loglevel
+    )
+
+
+loglevel_option = click.option(
```
masenf, line 34: we should use `callback` and `is_eager` to automatically handle setting the log level, so that the other commands don't need to do anything special

reflex/testing.py
```diff
@@ -933,8 +934,14 @@ def _start_frontend(self):
...
-            reflex.reflex.export(
-                zipping=False,
+
+            if reflex.utils.prerequisites.needs_reinit(frontend=True):
+                reflex.reflex._init(
+                    name=get_config().app_name, loglevel=reflex.constants.LogLevel.INFO
+                )
+
+            export(
+                zipping=True,
```
masenf, line 944: why is it zipping now?

adhami3310: because i'm being dummy :clown_face:

masenf (approved):

PR #5163, opened 2025-04-22
remove minor version from ci

adhami3310:
this should simplify updating our CI


masenf (approved):
honestly seems faster w/o the cache 😂

PR #5164, opened 2025-04-23
error when var is given something not a str

adhami3310:


Lendemor:
With 
```python
class Foo(rx.Base):
    field: str | rx.Var[str]    
```

the following expression: 
`Foo(field=["1", "bar"]).field`

will still give  
> `Var(_js_expr="1", _var_type="bar", _var_data=None)`

instead of raising an error because a list was passed.

adhami3310:
> instead of raising an error because a list was passed.

that seems an issue on pydantic and not Var

Lendemor:
> > instead of raising an error because a list was passed.
> 
> that seems an issue on pydantic and not Var

yeah, I just left that comment so we keep a trace of this issue

Lendemor (approved):

PR #5169, opened 2025-04-23
move tailwind to its own module

adhami3310:
this helps for the eventual upgrade to v4


reflex/plugins/base.py
```diff
@@ -0,0 +1,100 @@
...
+class PreCompileContext(CommonContext):
+    """Context for pre-compile hooks."""
+
+    add_task: AddTaskProtcol
+
+
+class Plugin:
+    """Base class for all plugins."""
+
+    def get_frontend_development_dependancies(
```
masenf, line 45:
```suggestion
    def get_frontend_development_dependencies(
```

🪩 🕺

masenf (commented):
this is awesome. need to do some more testing, but i like the approach

reflex/.templates/jinja/web/tailwind.config.js.jinja2
itsmeadarsh2008, line 1: Why this file was deleted? We can refer to this template from other files too? It kind of breaks consistency of templates

reflex/plugins/tailwind_v3.py
```diff
@@ -0,0 +1,248 @@
...
+@once
+def tailwind_config_js_template():
+    """Get the Tailwind config template.
+
+    Returns:
+        The Tailwind config template.
+    """
+    from reflex.compiler.templates import from_string
+
+    source = """
```
itsmeadarsh2008, line 40: We can refer to the template here so that it doesn't ruin the consistency.

reflex/.templates/jinja/web/tailwind.config.js.jinja2
adhami3310: the idea in this PR is to move all tailwind functionality to one file, such that it can be removed with ease later on, or replaced with a newer version, it does introduce inconsistencies but they are desirable for the mean time

masenf (approved):

PR #5208, opened 2025-04-29
Update translated docs

adhami3310:


masenf (commented):
changes look reasonable, but i also do not speak all of these languages.

i suggest reaching out on discord and seeing if any community members with native tongues are able to review the changes

masenf (approved):

PR #5227, opened 2025-05-01
decentralized event handlers

adhami3310:
```py
import reflex as rx


class State(rx.State):
    field: int = 23

@rx.event
def update_field(state: State, new_value: int):
    state.field = new_value


def index() -> rx.Component:
    return rx.vstack(
        State.field,
        rx.button(
            "Update Field",
            on_click=update_field(321),
        ),
    )


app = rx.App()
app.add_page(index)
```


masenf (approved):
looks good to me

PR #5229, opened 2025-05-01
what if we did not call basemodel setattr

adhami3310:


adhami3310:
i thought this would improve the performance but it seems barely noticable

PR #5251, opened 2025-05-02
[0.8] don't use reserve words for variable names

adhami3310:


Lendemor (commented):
Seems fine, will approve when it pass CI 👍

reflex/components/core/foreach.py
```diff
@@ -132,11 +132,11 @@ def _render(self) -> IterTag:
 
         if len(params) >= 1:
             # Determine the arg var name based on the params accepted by render_fn.
-            props["arg_var_name"] = params[0].name
+            props["arg_var_name"] = params[0].name + "_rx_foreach_"
```
masenf: should use FIELD_MARKER

masenf (approved):

PR #5259, opened 2025-05-05
pin all frontend npm packages

adhami3310:


adhami3310:
can be found with `    library[: ].*[^ 0-9]"` regex pattern

PR #5275, opened 2025-05-08
update to 508 lucide

adhami3310:
technically no icons were added but they fixed the missing panda (i want the panda)


masenf (approved):
image

PR #5289, opened 2025-05-13
remove pydantic as base class of component

adhami3310:


masenf (approved):
been working well for me over the past few days

Issue #5301, opened 2025-05-16
[Alembic] Creates migrations of tables that's migrations already exist


FoketzoRAI:
**Describe the bug**
After updating from 0.7.5 to 0.7.11, when I launch my reflex app, for some reason it notifies me that I have some changes (I don't have) and I have to make migrations.

I tried to understand what's happening:

1. I see that on reflex 0.7.5 ... 0.7.10 [here](https://github.com/reflex-dev/reflex/blob/main/reflex/model.py#L430) gives me False (which is OK) (if I'm not installing from 0.7.11)

<details>

<summary>Screenshots (OK & Downgrade from 0.7.11 to 0.7.9 and 0.7.8):</summary>

![Image](https://github.com/user-attachments/assets/67ad041a-f021-4291-b844-2586f7eb8e30)

![Image](https://github.com/user-attachments/assets/af157c1b-fec3-4ba2-a189-f8455e0ee3c0)

</details>

3. But in 0.7.11 (and if I downgrade till 0.7.9) it gives me True with a migration with all my models in it.

<details>

<summary>Screenshots:</summary>

![Image](https://github.com/user-attachments/assets/005ce597-d1be-4b14-b922-f1fe74f691a3)

![Image](https://github.com/user-attachments/assets/93b71f51-1ae5-40cb-8d90-11bcc9291d8e)

![Image](https://github.com/user-attachments/assets/2cc83bca-d600-47c8-aedd-d2cfe8322fe9)

</details>

**To Reproduce**

- You should have some migrations on old reflex 0.7.5
- Update to 0.7.11
- Launch app and see message about making migrations.

**Expected behavior**
As I have existing migrations it shouldn't notify me about changes.

**Specifics (please complete the following information):**

- Python Version: 0.13.2
- Reflex Version: 0.7.11 (and if downgrade from it till 0.7.9)
- OS: Ubuntu 24.04
- Browser (Optional): -

**Additional context**

Created migration by checker (kinda change) is my all DB models in one migration.

PR #5307, opened 2025-05-17
Fix: Skip Unnecessary Schema Checks for Compatible Reflex Versions (0.7.5–0.7.11) and Add Version Compatibility Test

mdehsan873:
This PR addresses an issue where Reflex incorrectly prompts for database migrations when upgrading between compatible versions (0.7.5 to 0.7.11), even if no schema changes are required. The following changes are included:
Logic Update:
The check_schema_up_to_date() function now skips schema checks for Reflex versions between 0.7.5 and 0.7.11 (inclusive), preventing unnecessary migration prompts during upgrades within this range.
Test Coverage:
Added a unit test to verify that schema checks are correctly skipped for compatible versions and performed for later versions (e.g., 0.7.12+). The test ensures future changes do not regress this behavior.
Why:
This fix improves the developer experience by avoiding redundant migration notifications during minor upgrades, ensuring migrations are only prompted when truly necessary.
#5301

### DIFF
--- a/reflex/utils/prerequisites.py
+++ b/reflex/utils/prerequisites.py
@@ -1576,6 +1576,14 @@ def check_schema_up_to_date():
     """Check if the sqlmodel metadata matches the current database schema."""
     if get_config().db_url is None or not environment.ALEMBIC_CONFIG.get().exists():
         return
+
+    # Get the current reflex version
+    current_version = version.parse(constants.Reflex.VERSION)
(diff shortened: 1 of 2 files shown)


adhami3310:
what's so special about 0.7.5 and 0.7.11 to warrant hardcoding this? couldn't the schema change while being on the same version and never upgrading?

PR #5312, opened 2025-05-19
pick up env vars prefixed by REFLEX_

Lendemor:


reflex/config.py
```diff
@@ -969,7 +971,15 @@ def update_from_env(self) -> dict[str, Any]:
...
             # The env var name is the key in uppercase.
-            env_var = os.environ.get(key.upper())
+            for prefix in self._prefixes:
+                if env_var := os.environ.get(f"{prefix}{key.upper()}"):
+                    break
+            else:
+                # Default to non-prefixed env var is other are not found.
+                if env_var := os.environ.get(key.upper()):
+                    console.warn(
+                        f"Usage of deprecated {key.upper()} env var detected. Prefer `REFLEX_` prefix when setting env vars."
```
masenf, line 981:
```suggestion
                        f"Usage of deprecated {key.upper()} env var detected. Prefer `{self._prefixes[0]}` prefix when setting env vars."
```

That way the enterprise package can print this message as `REFLEX_ENTERPRISE_` prefix?

Also, maybe we shouldn't have anything special for enterprise and just use `REFLEX_` for any and all config (even use defined props)

masenf (commented):
not seeing how this affects `EnvironmentVariables` though... those seem to be outside the Config

PR #5319, opened 2025-05-20
do not delete the .web dir when init-ing

### LINKED ISSUE #5313: Permissions of the .web directory are overwritten
**Describe the bug**
Without any particular action, permission of the .web directory get updated to rw-------
I have to chmod the directory to get reflex running again. I have no idea where this comes from, and that has happened multiple times.

- Python Version: 3.13.2
- Reflex Version: 0.7.11
- OS: Mac OS 15.4.1
- Browser (Optional): Brave

adhami3310:
fix #5313


masenf (approved):
so with this we're basically deleting the contents of the dest if it exists, but leaving the dest itself

PR #5322, opened 2025-05-21
do not broadcast event to all clients if modify state is misused

adhami3310:


masenf (dismissed):
might also want to check

```python
if sid not in self.sid_to_token:
    console.warn(f"Attempting to send delta to disconnected websocket {sid}")
```

masenf (approved):

PR #5336, opened 2025-05-23
adjust rechart types

Lendemor:


Lendemor:
> Does this fix any of the recharts problems or just remove component

The current PR only allow for passing props dict as value for `label` props, which will be applied to the default label.

Passing Component would require some of the rx.memo we have planned but not done yet.

Alek99 (dismissed):

Alek99 (approved):

PR #5345, opened 2025-05-27
run get_app before checking for schema updates

### LINKED ISSUE #5301: [Alembic] Creates migrations of tables that's migrations already exist

### LINKED ISSUE #5341: Priority order for event handlers is wrong
# Repro snippet

```python
from reflex.components.component import Component

# Custom component hierarchy for testing
class Child(Component):
    """A child component inheriting from Component."""

    tag = "div"

    def get_event_triggers(self):
        """Return custom event triggers for Child."""
        return {
            "on_child_click": rx.event.no_args_event_spec,
            "on_child_hover": rx.event.no_args_event_spec,
        }


class GrandChild(Child):
    """A grandchild component inheriting from Child."""

    tag = "span"

    on_grandchild_special: rx.EventHandler[rx.event.no_args_event_spec]


# Test the component hierarchy
child_instance = Child.create()
grandchild_instance = GrandChild.create()

print("Child event triggers:", list(child_instance.get_event_triggers().keys()))
print(
    "GrandChild event triggers:", list(grandchild_instance.get_event_triggers().keys())
)
```

## Current result

Child event triggers: ['on_child_click', 'on_child_hover']
GrandChild event triggers: ['on_child_click', 'on_child_hover']

## Expected result

Child event triggers: ['on_child_click', 'on_child_hover']
GrandChild event triggers: ['on_child_click', 'on_child_hover', ''on_grandchild_special']

# Reasoning

In this case, child component intentionally omit the `super().get_event_triggers()` because default events are not wanted for that component.

This is the correct way to remove inherited event triggers, but it also remove the event triggers defined in the grandchild component (because 'on_grandchild_special' is usually pushed via `super().get_event_triggers()` which doesn't quite make sense.

tartansandal:
This change ensures `prerequisites` has access to the application models before checking for schema changes. 

Since version 0.7.9, I've been getting spurious "Detected database changes..." warnings when starting up my app, However, when I run `reflex db makemigrations`, nothing is produced.  I tweaked the code to show the phantom migration it was warning me and, to my surprise, it wanted to delete all the objects in my database.   I believe this is because the prerequisites code could not see the app's models when checking for migrations.  This PR resolve that, but it may not be the most sensible approach. :sweat_smile: 

This may be related to and fix #5301

### DIFF
--- a/reflex/reflex.py
+++ b/reflex/reflex.py
@@ -218,6 +218,7 @@ def _run(
         raise click.exceptions.Exit(1)
 
     # Warn if schema is not up to date.
+    prerequisites.get_app()
     prerequisites.check_schema_up_to_date()
 
     # Get the frontend and backend commands, based on the environment.


adhami3310:
i understand now 😅 you are correct in that it's not where we need to do particularly but it does give me a clue on how to do it, thanks for the contribution!

PR #5354, opened 2025-05-27
fix infinite loop in pyi generator

adhami3310:
fixes #5343


adhami3310:
> Just FYI: managed to reproduce this while trying to install reflex on ARM with `uv sync`

on this branch or on main?

not-matthias:
> on this branch or on main?
on main, just wanted to let you know in case something like this happens again. cause i noticed you don't have arm benches that could've caught this in advance.

Lendemor (approved):

PR #5389, opened 2025-05-31
no need to kill nextjs on package change

adhami3310:
also adds helpful messages

### DIFF
--- a/reflex/utils/exec.py
+++ b/reflex/utils/exec.py
@@ -188,8 +188,8 @@ def run_process_and_launch_url(
                         get_different_packages(last_content, new_content)
                     )
                     last_content, last_hash = new_content, new_hash
-                    console.print(
-                        "[bold yellow]Detected changes in package.json.[/bold yellow]\n"
+                    console.info(
+                        "Detected changes in package.json.\n"
                         + format_change("Dependencies", dependencies_change)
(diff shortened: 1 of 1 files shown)


reflex/utils/exec.py
```diff
@@ -27,26 +28,102 @@
...
+    dependencies_change = get_changes(
+        old_package_json_content.get("dependencies", {}),
+        new_package_json_content.get("dependencies", {}),
+    )
+    dev_dependencies_change = get_changes(
+        old_package_json_content.get("devDependencies", {}),
+        new_package_json_content.get("devDependencies", {}),
+    )
+
+    return dependencies_change, dev_dependencies_change
```
Lendemor, line 126:
```suggestion
    return tuple(
        get_changes(
            old_package_json_content.get(deps_key, {}),
            new_package_json_content.get(deps_key, {}),
        )
        for deps_key in PackageJson.__annotations__
    )
```

adhami3310: this makes the typing looser, so i would prefer to keep explicit for now

Lendemor: Make sense for the explicit.

Are the intermediary variable just there for easier reading ? because we could do explicit in the return expression right?

adhami3310: > just there for easier reading

yea it's also to declutter a long return line

PR #5401, opened 2025-06-02
add on_scroll_end event

Lendemor:
From experimentation, on_scroll is kinda spammy during scroll events, so it might be useful to have one that fire once per scroll, when it end.

It's a `no_args_event_spec` for now because I didn't find much useful data inside the event that is fired on that trigger.


adhami3310:
safari doesn't support this, do we care?

Lendemor:
> safari doesn't support this, do we care?

That's a Safari problem isn't it? 
We follow the MDN specifications, if Safari doesn't want to support it, not much we can do, but no reason to keep other browsers from using the feature.

masenf (approved):
tentative approval provided docs are appropriately updated

PR #5410, opened 2025-06-03
add a check if app_name is wrong in rxconfig

Lendemor:


reflex/utils/prerequisites.py
```diff
@@ -334,21 +334,39 @@ def npm_escape_hatch() -> bool:
...
             "Cannot get the app module because `app_name` is not set in rxconfig! "
             "If this error occurs in a reflex test case, ensure that `get_app` is mocked."
         )
         raise RuntimeError(msg)
 
+    # Skip folder structure validation if using custom app_module_import or in testing environment
+    if config.app_module_import is not None or is_testing_env():
+        return
+
+    # Check folder structure matches app_name
```
adhami3310: can't we try with find spec?

PR #5419, opened 2025-06-07
Respect GRANIAN_WORKERS env var in production

tartansandal:
Ensures that `GRANIAN_WORKERS` environment variable is not overridden by `--workers` parameter when starting the production server. 

While testing docker containers on my laptop, I was surprised that I was getting 41 workers when I had set GRANIAN_WORKERS=1. Had to use the deprecated GUNICORN_WORKERS env var to work around this.

I'm not sure if the fix is in the right place, with the config and setup code in flux with the move to granian and 0.8.0, but it does solve the immediate problem.

### DIFF
--- a/reflex/utils/exec.py
+++ b/reflex/utils/exec.py
@@ -698,13 +698,15 @@ def run_granian_backend_prod(host: str, port: int, loglevel: LogLevel):
 
         command = [
             "granian",
-            *("--workers", str(_get_backend_workers())),
             *("--log-level", "critical"),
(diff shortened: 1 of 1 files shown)


Lendemor:
We've started merging the PRs for 0.8, so we will soon make one that remove the deprecated config props like `gunicorn_workers`.

So I don't know if we'll take that PR, but we'll make sure that problem is solved in 0.8.

adhami3310:
superseded by #5422

PR #5430, opened 2025-06-10
[0.9] remove ids refs from form_data

### LINKED ISSUE #5418: Exception: "ReferenceError: refs is not defined"
**Describe the bug**
When submitting a form the exception "ReferenceError: refs is not defined" is raised.

Image

**To Reproduce**

Bring up the app in the included python file below with:
`reflex run --env prod`

Image

Click the Login link and enter any email and password:

Image

Press the "Log in" button

Exception will appear.

Running the same test but launching with:
`reflex run`
Does not produce the exception

For consistent results, I ran
`reflex init`
before each test but not positive this was necessary.

The core element to the test case is having two forms on different pages/routes but each with a field that has the same id string. In this test case it is id="UserEmailAddress".

- Code/Link to Repo:
[reflex_issue_refs_is_not_defined.zip](https://github.com/user-attachments/files/20635986/reflex_issue_refs_is_not_defined.zip)

**Expected behavior**
No exception should occur.

**Screenshots**

- Python Version: 3.13.3
- Reflex Version: 0.7.14
- OS: Ubuntu 24.04.1
- Browser (Optional): Brave

Lendemor:
Passing both name and ids refs can lead to duplicate data, we should only pass form data by name.

Fix #5418 (when extracting hooks in stateful components, the form hook still refer `refs` but the import is removed, leading to the error)

### DIFF
--- a/pyi_hashes.json
+++ b/pyi_hashes.json
@@ -29,7 +29,7 @@
   "reflex/components/el/element.pyi": "3519fdc9a74aab7b02f475568def5f95",
   "reflex/components/el/elements/__init__.pyi": "baeddd04d4d3a82799420b2a6df368f6",
   "reflex/components/el/elements/base.pyi": "2c962706dd6c28821497e9c8044cab4e",
-  "reflex/components/el/elements/forms.pyi": "0a75b05da0e53f581e356e464d3675d2",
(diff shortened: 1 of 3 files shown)


FarhanAliRaza:
already fixed compiler plugin work in PR #6260 and the rx.memo promotion in PR #6517.

PR #5460, opened 2025-06-17
make run dev run react router dev instead of vite dev

adhami3310:
hopefully fixes the cyclical imports issue?

### DIFF
--- a/reflex/utils/processes.py
+++ b/reflex/utils/processes.py
@@ -320,7 +320,8 @@ def stream_logs(
 
     # Windows uvicorn bug
     # https://github.com/reflex-dev/reflex/issues/2335
-    accepted_return_codes = [0, -2, 15] if constants.IS_WINDOWS else [0, -2]
+    # 130 is the exit code that react router returns when it is interrupted by a signal.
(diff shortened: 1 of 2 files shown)


reflex/utils/processes.py
```diff
@@ -320,7 +320,8 @@ def stream_logs(
 
     # Windows uvicorn bug
     # https://github.com/reflex-dev/reflex/issues/2335
-    accepted_return_codes = [0, -2, 15] if constants.IS_WINDOWS else [0, -2]
+    # 130 is the exit code that react router returns when it is interrupted by a signal.
+    accepted_return_codes = [0, -2, 15, 130] if constants.IS_WINDOWS else [0, -2, 130]
```
masenf, line 324:
```suggestion
    accepted_return_codes = [0, -2, 15] if constants.IS_WINDOWS else [0, -2, 130]
```

seems 130 isn't strictly needed on windows, but probably doesn't hurt to have it there

masenf (approved):

reflex/utils/processes.py
adhami3310: it's ok then

PR #5484, opened 2025-06-21
clean up random fields that aren't marked correctly as not javascript fields

### LINKED ISSUE #5427: Some components are missing / not imported in compiled js files
**Describe the bug**
Some js components are missing (because they're not imported) in the compiled nextjs application.

**To Reproduce**
We have a common template (header, footer, menu, etc.) which we use for every page. Therefore we have a function which given the content componnent of the, it generates the whole page component.
For instance:

`return tool_layout(
   rx.vstack(...)
)`

where tool_layout is a function which does something like that:

`def tool_layout(component):
    return rx.vstack(
        header(),
        menu(),
        component,
        footer(),
    )
)`

Everything used to work fine on Reflex 0.5.x (and on other projects running on Reflex 0.6.x the same architecture works well too).

I'm not sure the issue is related to this, but I can't find any other plausible cause.

**Expected behavior**
All the components used in a given page should be imported.

**Screenshots**

Image

Image

**Specifics:**

- Python Version: 3.12.4
- Reflex Version: 0.7.14
- OS: Mac OS
- Browser (Optional): Chrome

adhami3310:
this is a bit sus i will be looking at the tests, fixes #5427


adhami3310:
no longer fixes #5427 hehe

reflex/components/base/meta.py
```diff
@@ -28,31 +29,19 @@ def render(self) -> dict:
...
-    char_set: str | None = None
-
-    # The value of meta.
-    content: str | None = None
-
-    # The name of metadata.
-    name: str | None = None
-
-    # The type of metadata value.
-    property: str | None = None
```
masenf, line 41: why are we removing these?

adhami3310: they are inherited

PR #5500, opened 2025-06-26
Include default metas for char_set and viewport

masenf:
If the user does not override the char_set or viewport metas with their own, then include these as defaults.

Fixes the annoying hydration error that appears when remix sends its dumb "Hey developer  👋" message on the console, which happened when the app didn't specify `utf-8` as its char_set (which would happen if the app used _any_ custom `head_components`)

### DIFF
--- a/reflex/compiler/utils.py
+++ b/reflex/compiler/utils.py
@@ -359,20 +359,33 @@ def create_document_root(
     Returns:
         The document root.
     """
-    head_components = [
-        *(
-            head_components
-            or [
-                # Default meta tags if user does not provide.
-                Meta.create(char_set="utf-8"),
-                Meta.create(
-                    name="viewport", content="width=device-width, initial-scale=1"
-                ),
-            ]
-        ),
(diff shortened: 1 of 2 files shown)


adhami3310 (approved):

reflex/compiler/utils.py
```diff
@@ -359,21 +359,34 @@ def create_document_root(
...
-        ),
-        # Always include the framework meta and link tags.
+    existing_meta_types = set()
+
+    for component in head_components or []:
+        if isinstance(component, Meta):
+            if component.char_set:  # pyright: ignore[reportAttributeAccessIssue]
+                existing_meta_types.add("char_set")
+            if component.name == "viewport":  # pyright: ignore[reportAttributeAccessIssue]
+                existing_meta_types.add("viewport")
```
adhami3310: this guy fields are actually Vars and not python types

adhami3310 (approved):

PR #5510, opened 2025-06-30
reload env file on hot reload in granian

adhami3310:


masenf (approved):

gi0baro:
@adhami3310 it seems to me you can actually get rid of the hook and directly pass `env_files` to `Granian` constructor (with 2.4.0 it loads those files on reload both on the main process and in the workers when they start).

edit: see https://github.com/emmett-framework/granian/blob/v2.4.0/granian/server/common.py#L479

adhami3310:
@gi0baro that seems cleaner, i will look into it, thanks for the comment :D

PR #5521, opened 2025-07-02
fix gunicorn in prod mode

Lendemor:


reflex/utils/exec.py
```diff
@@ -580,7 +580,14 @@ def run_uvicorn_backend_prod(host: str, port: int, loglevel: LogLevel):
     command = (
         ["uvicorn", *("--host", host), *("--port", str(port)), "--factory", app_module]
         if constants.IS_WINDOWS
-        else ["gunicorn", "--preload", *("--bind", f"{host}:{port}"), f"{app_module}()"]
+        else [
+            "gunicorn",
+            "--preload",
+            "--worker-class",
+            "uvicorn.workers.UvicornWorker",
```
masenf: https://github.com/reflex-dev/reflex/blob/84189926c63f7220e8684fcd4d864f904b641650/reflex/config.py#L283

shouldn't we use `uvicorn.workers.UvicornH11Worker`, which we had before?

Also, this being removed may break some people who were relying on overriding `granian_worker_class` config option to work on Azure... maybe they can just work on granian, but we haven't tested granian in an environment that doesn't support per message deflate

masenf (approved):

PR #5524, opened 2025-07-03
load plugins from env variable

adhami3310:
`REFLEX_PLUGINS=reflex.plugins.tailwind_v3.TailwindV3Plugin`


masenf (commented):
Working for me with 2 local plugins separated by colon

A test case would be nice though, I think we already have some test cases for interpreting other env types in `test_config.py` (probably should move those to `test_environment.py`)

masenf:
This should give us some impetus to expand the plugin API to allow for more interesting integrations, like `post_compile` to give a chance for custom modifications in the `.web` directory.

PR #5525, opened 2025-07-03
use deterministic hash for stateful component names

adhami3310:


reflex/components/component.py
```diff
@@ -479,6 +480,57 @@ def _components_from(
...
+    if isinstance(value, str):
+        return int(md5(f'"{value}"'.encode()).hexdigest(), 16)
+    if isinstance(value, (tuple, list)):
+        # Hash tuples by hashing each element.
+        return _deterministic_hash(
+            "[" + ",".join(map(str, map(_deterministic_hash, value))) + "]"
+        )
+    if isinstance(value, enum.Enum):
+        # Hash enums by their name.
+        return _deterministic_hash(value.name)
```
masenf: idk about this one

```python
>>> class E1(enum.Enum):
...     A = 1
...     B = 2
...     C = 3
... 
>>> 
>>> class E2(enum.Enum):
...     A = 1
...     B = 2
...     C = 3
... 
>>> E1.A.name
'A'
>>> E2.A.name
'A'
```

maybe the value would be better? or hashing `str(value)` so it includes the class name?

reflex/components/component.py
```diff
@@ -479,6 +480,57 @@ def _components_from(
...
+    if isinstance(value, dict):
+        # Sort the dictionary to ensure consistent hashing.
+        return _deterministic_hash(
+            tuple(sorted((k, _deterministic_hash(v)) for k, v in value.items()))
+        )
+    if isinstance(value, int):
+        # Hash numbers and booleans directly.
+        return int(value)
+    if isinstance(value, float):
+        return _deterministic_hash(str(value))
```
masenf, line 512: why not hash the float directly, like the `int`?

adhami3310: _deterministic_hash has to return an int, so I'm not sure how to map it

PR #5438, opened 2025-06-12
move pydantic, sqlmodel, alembic to optional dependencies

adhami3310:

PR #5557, opened 2025-07-10
move rx config away from pydantic

adhami3310:
part of #5438

### DIFF
--- a/reflex/config.py
+++ b/reflex/config.py
@@ -6,6 +6,7 @@ import os
 import sys
 import threading
 import urllib.parse
+from collections.abc import Sequence
 from importlib.util import find_spec
 from pathlib import Path
 from types import ModuleType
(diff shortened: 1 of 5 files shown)


reflex/config.py
```diff
@@ -257,21 +233,50 @@ class Config:  # pyright: ignore [reportIncompatibleVariableOverride]
...
+
+    config = rx.Config(
+        app_name="myapp",
+        api_url="http://localhost:8000",
+    )
+    ```
+
+    Every config value can be overridden by an environment variable with the same name in uppercase.
+    For example, `db_url` can be overridden by setting the `DB_URL` environment variable.
+
```
masenf, line 259: probably should update this while we're here
```suggestion
    Every config value can be overridden by an environment variable with the same name in uppercase and a REFLEX_ prefix.
    For example, `db_url` can be overridden by setting the `REFLEX_DB_URL` environment variable.

```

masenf (approved):

reflex/config.py
adhami3310: i will get in a follow up PR

PR #5560, opened 2025-07-11
Add db status command

### LINKED ISSUE #1753: Consult migrations status in db cli
We need to add a `reflex db status` command to consult the current state of the migrations: 
(see applied migrations, and created but not applied migrations)

Lendemor:
fix #1753


masenf (approved):
i think we're trying to get rid of our alembic integration and doc how to use alembic directly... but this is cool

Lendemor:
> i think we're trying to get rid of our alembic integration and doc how to use alembic directly... but this is cool

This can be moved to an external package later if we want I guess?

PR #5564, opened 2025-07-11
check port with bind over connect_ex

adhami3310:


sorousch9:
@adhami3310 In Windows Subsystem for Linux WSL, the recursive change_port call can overflow the call stack and trigger a `RecursionError`.

PR #5574, opened 2025-07-15
add event_action flags to rx.event decorator

Lendemor:
Allow to attach default value for event actions flag on an event_handler decorated by `@rx.event`

### DIFF
--- a/reflex/event.py
+++ b/reflex/event.py
@@ -109,7 +109,7 @@ class EventActionsMixin:
         """
         return dataclasses.replace(
             self,
-            event_actions={"stopPropagation": True, **self.event_actions},
+            event_actions={**self.event_actions, "stopPropagation": True},
         )
 
     @property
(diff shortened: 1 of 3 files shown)


reflex/event.py
```diff
@@ -2238,6 +2256,11 @@ def __new__(
         Args:
             func: The function to wrap.
             background: Whether the event should be run in the background. Defaults to False.
+            stop_propagation: Whether to stop the event from bubbling up the DOM tree.
+            prevent_default: Whether to prevent the default behavior of the event.
+            throttle: Throttle the event handler to limit calls (in milliseconds).
+            debounce: Debounce the event handler to delay calls (in milliseconds).
+            temporal: Whether the event should be temporal.
```
masenf, line 2263:
```suggestion
            temporal: Whether the event should be dropped when the backend is down.
```

masenf (approved):

PR #5608, opened 2025-07-21
fix FOUC on refresh

Lendemor:


reflex/utils/misc.py
```diff
@@ -90,3 +90,38 @@ def with_cwd_in_syspath():
...
+        const systemPreference = window.matchMedia("(prefers-color-scheme: dark)").matches ? "dark" : "light";
+        const resolvedTheme = theme === "system" ? systemPreference : theme;
+
+        // Apply theme immediately - blocks until complete
+        document.documentElement.className = resolvedTheme;
+        document.documentElement.style.colorScheme = resolvedTheme;
+
+    } catch (e) {
+        // Fallback to system preference on any error
+        document.documentElement.className = "system";
```
adhami3310: 1) we are removing other class names by setting ours
2) it seems classname would expect a resolved theme, not the theme preference, but unsure how that makes a difference or how it's being used

PR #5613, opened 2025-07-22
snip source correctly for dep tracker

### LINKED ISSUE #5580: `get_var_value` behaves strangely.
**Describe the bug**

I want to call `get_var_value` on some field of some state class. The field is called `loaded_experiment`. Reflex seems to want to work with a field that is called `loaded_experimen` (notice who the last `t` is missing). See screenshot below.

**To Reproduce**
Steps to reproduce the behavior

This repo contains a minimum example:

- Code/Link to Repo: https://github.com/mohelm/reflex-get-var-value-debug

Steps: clone, `uv init`, `reflex run`.

In words, I init a new reflex app with the default template ("blank" I think), and I try to run a app with this code added: 

```{python}
class State(rx.State):
    load_experiment: str ="hello" 
    """The app state."""

class Metadata(rx.State):

    @rx.var
    async def metadata(self)->str:
        d  =await self.get_var_value(State.load_experiment)
        print(d)
        return d
```

Reflex cannot determine the depedency because it looks for a varaible `load_experimen`. See screenshot below.

**Expected behavior**
It should work with `load_experiment`

**Screenshots**

Image

- Python Version: 3.13
- Reflex Version: 0.8.2
- OS: macos
- Browser (Optional): 

**Additional context**

-

adhami3310:
fixes #5580

### DIFF
--- a/reflex/vars/dep_tracking.py
+++ b/reflex/vars/dep_tracking.py
@@ -254,14 +254,10 @@ class DependencyTracker:
         # Create a python source string snippet.
         if len(source) > 1:
             snipped_source = "".join(
-                [
-                    *source[0][start_column:],
-                    *(source[1:-2] if len(source) > 2 else []),
-                    *source[-1][: end_column - 1],
-                ]
+                [*source[0][start_column:], *source[1:-1], *source[-1][:end_column]]
             )
         else:
-            snipped_source = source[0][start_column : end_column - 1]
+            snipped_source = source[0][start_column:end_column]
         # Evaluate the string in the context of the function's globals and closure.
         return eval(f"({snipped_source})", self._get_globals(), self._get_closure())
 


masenf (commented):
can we get a unit test added for this

adhami3310:
> can we get a unit test added for this

the code is a bit weird, but i will try

tests/units/vars/test_dep_tracking.py
```diff
@@ -0,0 +1,337 @@
...
+    """Test complex cross-state dependency scenarios."""
+
+    class StateA(State):
+        value_a: int = 1
+
+    class StateB(State):
+        value_b: int = 2
+
+        async def get_a_value(self):
+            return (await self.get_state(StateA)).value_a
```
masenf: this construct seems unused in the test

tests/units/vars/test_dep_tracking.py
```diff
@@ -0,0 +1,337 @@
...
+        StateB.get_full_name(): {"value_b"},
+    }
+    assert tracker.dependencies == expected_deps
+
+
+def test_lambda_function_dependencies():
+    """Test tracking dependencies in lambda functions."""
+
+    def lambda_func(self: DependencyTestState):
+        return self.count * 2
```
masenf: this isn't a lambda. this test is equivalent to `test_simple_attribute_access`

PR #5620, opened 2025-07-24
allow leak and block detection on handlers

Lendemor:


reflex/monitoring/__init__.py
adhami3310, line 1: we can just have this in monitoring.py

pyproject.toml
```diff
@@ -29,6 +29,7 @@ dependencies = [
   "platformdirs >=4.3.7,<5.0",
   "psutil >=7.0.0,<8.0; sys_platform == 'win32'",
   "pydantic >=1.10.21,<3.0",
+  "pyleak >=0.1.14,<1.0",
```
adhami3310: can we have this be an optional dep?

Lendemor: Made it optional, and added it to dev dependencies too.

masenf:
```
import asyncio

import reflex as rx


class State(rx.State):
    @rx.event(background=True)
    async def bg_task(self):
        return [
            State.another_bg_task,
            State.quick_thing,
        ]

    @rx.event(background=True)
    async def another_bg_task(self):
        await asyncio.sleep(1)

    @rx.event
    def quick_thing(self):
        print("Quick thinking event triggered!")


def index() -> rx.Component:
    return rx.button("bg task", on_click=State.bg_task)


app = rx.App()
app.add_page(index)
```

masenf:
also, i think we should move the module to `reflex.utils.monitoring`

masenf (commented):
New repro code

```python
import asyncio
import time

import reflex as rx


class State(rx.State):
    @rx.event(background=True)
    async def bg_task(self):
        return [
            State.another_bg_task,
            State.quick_thing,
        ]

    @rx.event(background=True)
    async def another_bg_task(self):
        await asyncio.sleep(1)

    @rx.event
    def quick_thing(self):
        print("Quick thinking event triggered!")
        time.sleep(2)


def index() -> rx.Component:
    return rx.vstack(
        rx.button("bg task", on_click=State.bg_task),
        rx.button("quick thing", on_click=State.quick_thing),
    )


app = rx.App()
app.add_page(index)
```

```python
from pyleak.base import LeakAction
import reflex as rx

config = rx.Config(
    app_name="repro_pyleak",
    plugins=[
        rx.plugins.SitemapPlugin(),
        rx.plugins.TailwindV4Plugin(),
    ],
    enable_pyleak_monitoring=True,
    pyleak_action=LeakAction.LOG,
)
```

When I click the "bg task" button, i end up seeing two event loop block detection logs.

When I click the "quick thing" button, i only see one event loop block detections.

I think we need some protection to avoid re-entering the block detection context if it is already active.

masenf (approved):
working well for me.

from a code organization perspective, implementing the set/reset of `monitoring_active` as a contextmanager itself would have fit nicely in the existing ExitStack/AsyncExitStack pattern without requiring the outer `try`/`finally`. but we can take this and refine it later if we so desire, it's a self-contained module in what amounts to debug code, so not a big deal.

PR #5658, opened 2025-08-01
extract env json into its own file

adhami3310:
that would make all reflex builds output `reflex-env-<hash>.js` inside of `build/client/assets/*`


adhami3310:
i might catch that guy later, bbut it's typically difficult to not have a compiled output when the original is a json file

PR #5659, opened 2025-08-03
make pydantic-core compatible with 3.14.0rc1

xychelsea:
### All Submissions:

### Type of change

### New Feature Submission:

NOTE: Passes tests but using 3.13.5

    b. Describe your changes.


reflex/state.py
```diff
@@ -508,7 +508,7 @@ def __init_subclass__(cls, mixin: bool = False, **kwargs):
 
         new_backend_vars = {
             name: value
-            for name, value in cls.__dict__.items()
+            for name, value in list(cls.__dict__.items())
             if types.is_backend_base_variable(name, cls)
```
adhami3310: as far as i know, no modification is to `cls.__dict__` is happening inside of `types.is_backend_base_variable`, is this aimed to be a performance improvement or a bugfix?

pyi_hashes.json
adhami3310, line 1: you might want to run pre-commit with python 3.13, not very sure if python 3.14 changed their ast output 🤔

pyproject.toml
```diff
@@ -28,7 +28,7 @@ dependencies = [
   "packaging >=24.2,<26",
   "platformdirs >=4.3.7,<5.0",
   "psutil >=7.0.0,<8.0; sys_platform == 'win32'",
-  "pydantic >=1.10.21,<3.0",
+  "pydantic >=2.12.0a1,<3.0",
```
adhami3310: we only specify a minimum dep, so it should be possible for downstream to specify latest pydantic (and that's what `pip install` usually do if pydantic is unspecified), the only thing this does is declare that `pydantic>=1.10.21<2.12.0a1` doesn't work in reflex, which it most likely does, maybe not in python 3.14 though

if we want to be overly correct, we can specify version requirement depending on python version, although that would be best handled in pydantic

Alek99:
Thanks for the pr!

pyproject.toml
xychelsea: this makes sense, my goal was only to flag that this is what i had to do to get it working in 3.14.0rc

reflex/state.py
xychelsea: python 3.14.0rc required it to build reflex

PR #5663, opened 2025-08-04
fix: add missing size property to DynamicIcon class

### LINKED ISSUE #5660: The rx.icon "size" attribute is parsed incorrectly in some statements
**Describe the bug**
The rx.icon "size" attribute is parsed incorrectly in some statements

Test Code: 
```python

class State(rx.State):
    icon: str = "home"
...

rx.icon("home", size=16),
rx.icon(State.icon, size=16),

...

```
Now, you can see two icons of different sizes displayed on the page

The size of the first icon is correct.The parsed HTML code looks like this:
```html
<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" ...></svg>
```

However, the parsed HTML for the second icon looks like this:
```html

<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" class="lucide css-15oa6zw" ... </svg>

```

```css
.css-15oa6zw {
    size: 16px;
}
```
The "size" option in CSS doesn't control the size of an SVG.

As a temporary solution, I  use the following code to achieve the desired effect:

```python
rx.icon(State.icon, class_name="w-4 h-4")
```

Image

--------------
- Python Version: 3.11
- Reflex Version: 0.8.4
- OS: win 11
- Browser (Optional): Chrome

geekalaa:
### All Submissions:

### Type of change

### Changes To Core Features:

## Description

Fixes #5660 where `rx.icon` size attribute was ignored when using dynamic state variables vs static strings.
The DynamicIcon class was missing the size: Var[int] property definition, causing size to be treated as CSS style instead of a direct component prop. This resulted in dynamic icons falling back to Lucide React's default size of 24px instead of using the specified size.

### Problem
- `rx.icon("home", size=16)` worked correctly (16x16 pixels)
- `rx.icon(State.icon, size=16)` was broken (fell back to Lucide's default 24px)

### Root Cause
The `DynamicIcon` class was missing the `size: Var[int]` property definition, causing the size parameter to be treated as a CSS style instead of a direct component       
prop.

### Solution
Added the missing `size: Var[int]` property to the `DynamicIcon` class, ensuring consistent size handling between static and dynamic icons.

### Testing
- Verified both static and dynamic icons now render with proper component props
- All existing icon tests pass

closes #5660

### DIFF
--- a/reflex/compiler/utils.py
+++ b/reflex/compiler/utils.py
@@ -23,8 +23,7 @@ from reflex.constants.state import FIELD_MARKER
 from reflex.istate.storage import Cookie, LocalStorage, SessionStorage
 from reflex.state import BaseState, _resolve_delta
 from reflex.style import Style
-from reflex.utils import console, format, imports, path_ops
-from reflex.utils.exec import is_in_app_harness
+from reflex.utils import format, imports, path_ops
 from reflex.utils.imports import ImportVar, ParsedImportDict
 from reflex.utils.prerequisites import get_web_dir
 from reflex.vars.base import Field, Var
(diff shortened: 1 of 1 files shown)


adhami3310:
that looks correct, running `pre-commit run --all-files` should fix the CI issue, thanks for the PR!

PR #5665, opened 2025-08-04
fix: add missing size property to DynamicIcon class

### LINKED ISSUE #5660: The rx.icon "size" attribute is parsed incorrectly in some statements

geekalaa:
### All Submissions:

### Type of change

### Changes To Core Features:

## Description

Fixes #5660 where `rx.icon` size attribute was ignored when using dynamic state variables vs static strings.
The DynamicIcon class was missing the size: Var[int] property definition, causing size to be treated as CSS style instead of a direct component prop. This resulted in dynamic icons falling back to Lucide React's default size of 24px instead of using the specified size.

### Problem
- `rx.icon("home", size=16)` worked correctly (16x16 pixels)
- `rx.icon(State.icon, size=16)` was broken (fell back to Lucide's default 24px)

### Root Cause
The `DynamicIcon` class was missing the `size: Var[int]` property definition, causing the size parameter to be treated as a CSS style instead of a direct component       
prop.

### Solution
Added the missing `size: Var[int]` property to the `DynamicIcon` class, ensuring consistent size handling between static and dynamic icons.

### Testing
- Verified both static and dynamic icons now render with proper component props
- All existing icon tests pass

closes #5660

### DIFF
--- a/pyi_hashes.json
+++ b/pyi_hashes.json
@@ -40,7 +40,7 @@
   "reflex/components/el/elements/tables.pyi": "e3f299e59bb8ff87aa949c6b606551c9",
   "reflex/components/el/elements/typography.pyi": "b4ec4ffb448f7a9b5f94712404448d9e",
   "reflex/components/gridjs/datatable.pyi": "98a7e1b3f3b60cafcdfcd8879750ee42",
-  "reflex/components/lucide/icon.pyi": "c1090e2c68a5d7653584dd74e8b8f1cf",
+  "reflex/components/lucide/icon.pyi": "91ccfd65214b968f1b18caba58a872ed",
   "reflex/components/markdown/markdown.pyi": "2f84254a548e908020949564fc289339",
   "reflex/components/moment/moment.pyi": "93fa4c6009390fe9400197ab52735b84",
   "reflex/components/plotly/plotly.pyi": "4311a0aae2abcc9226abb6a273f96372",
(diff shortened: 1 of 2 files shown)


pyi_hashes.json
```diff
@@ -40,7 +40,7 @@
   "reflex/components/el/elements/tables.pyi": "e3f299e59bb8ff87aa949c6b606551c9",
   "reflex/components/el/elements/typography.pyi": "b4ec4ffb448f7a9b5f94712404448d9e",
   "reflex/components/gridjs/datatable.pyi": "98a7e1b3f3b60cafcdfcd8879750ee42",
-  "reflex/components/lucide/icon.pyi": "c1090e2c68a5d7653584dd74e8b8f1cf",
+  "reflex/components/lucide/icon.pyi": "91ccfd65214b968f1b18caba58a872ed",
```
adhami3310:
```suggestion
  "reflex/components/lucide/icon.pyi": "b69b861d50bbaecf0e0fde600192ed18",
```

adhami3310:
thanks for the fix :)

PR #5675, opened 2025-08-05
stateful screenshots

adhami3310:


reflex/app.py
```diff
@@ -1503,6 +1508,65 @@ def _write_stateful_pages_marker(self):
...
+                found_new = False
+                for state in all_states:
+                    for substate in state.substates.values():
+                        substate._was_touched = True
+                        if substate not in all_states:
+                            all_states.append(substate)
+                            found_new = True
+
+            await self.state_manager.set_state(
+                new_token + "_" + new_state.get_full_name(), new_state
```
masenf: should use `_substate_key` from `reflex.state` for consistency with the rest of the code base

reflex/reflex.py
```diff
@@ -452,6 +452,191 @@ def export(
...
+    is_flag=True,
+    help="Whether to run the browser in stateful mode (preserving session) or stateless mode (fresh session).",
+)
+@click.option(
+    "--output-dir",
+    type=Path,
+    default=".",
+    help="The directory to save screenshots to.",
+)
+def take_screenshots(
```
masenf, line 511: i think this function should exist outside of the cli so it can also be called as a library

reflex/app.py
```diff
@@ -1503,6 +1508,65 @@ def _write_stateful_pages_marker(self):
...
 
+    def add_active_connections_endpoint(self):
+        """Add an endpoint to the app that returns the active connections."""
+        if not self._api:
+            return
+
+        async def active_connections(_request: Request) -> Response:
+            if not self.event_namespace:
+                return JSONResponse({})
+            return JSONResponse(self.event_namespace.token_to_sid)
```
masenf, line 1527: this is just a tad bit crazy, because the client_token is basically the keys to the kingdom for a user's state and it's assumed to be private.

we need to log a big huge fat warning when this setting is enabled letting the user know that their app privacy is completely compromised.

with these two endpoints it's theoretically possible to steal cookies or other private data from any session.

PR #5684, opened 2025-08-06
Fiex: ensure `cls.__dict__` compat w/ python3.14

xychelsea:
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

Title:
Fix: wrap cls.__dict__.items() in list() for Python 3.14 compatibility

Description:
This change wraps cls.__dict__.items() in list() when building new_backend_vars in BaseState.__init_subclass__.
In Python 3.14 (currently in release candidate), dict_items views have updated semantics that cause iteration issues in this context. Wrapping in list() restores the expected iteration behavior and allows the application to build and run under Python 3.14 without impacting previous Python versions.

Reason for inclusion:
Maintains forward compatibility with upcoming Python releases while preserving behavior on all supported versions. See previous closed Pull Request 5659

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -508,7 +508,7 @@ class BaseState(EvenMoreBasicBaseState):
 
         new_backend_vars = {
             name: value
-            for name, value in cls.__dict__.items()
+            for name, value in list(cls.__dict__.items())
             if types.is_backend_base_variable(name, cls)
         }
         # Add annotated backend vars that may not have a default value.


adhami3310:
i'm not opposed to having this but what changed in python 3.14? is there a changelog about this?

adhami3310:
ok i see, `is_backend_base_variable` mutates `cls`, that's not very nice of it, but i see why it would break

there seems to be other issues in reflex though, so this alone doesn't bring it to 3.14 compat

xychelsea:
I'm building and running `reflex` in Python 3.14 for a future-proofed project, so idk, it works in 3.14 now with this change and with a different `pydantic` version in the dependencies

xychelsea:
From a Perplexity search:

"Annotation Evaluation is Now Deferred at Runtime (PEP 649)
In Python 3.14, PEP 649 introduces deferred evaluation of annotations using descriptors, meaning:

Annotations are no longer stored as strings (as per PEP 563).

Instead, they are stored unevaluated in a descriptor (__annotate__) and only evaluated the first time they're accessed.

This improves support for forward references and ensures evaluation in the correct context and scope.

This shift eliminates the need to rely on from __future__ import annotations, which is now deprecated in Python 3.14."

adhami3310 (approved):
it will be sometime before all of our deps also are 3.14 compatible, but it's a nice first step

PR #5703, opened 2025-08-11
feat: nullable exception handlers

### LINKED ISSUE #3776: Allow to disable frontend exception handler with None
@maximvlah

maximvlah:
Closes https://github.com/reflex-dev/reflex/issues/3776

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -1880,8 +1880,12 @@ class BaseState(EvenMoreBasicBaseState):
         except Exception as ex:
             telemetry.send_error(ex, context="backend")
 
+            app = prerequisites.get_and_validate_app().app
+
             event_specs = (
-                prerequisites.get_and_validate_app().app.backend_exception_handler(ex)
+                app.backend_exception_handler(ex)
+                if app.backend_exception_handler
+                else None
             )
 
             yield await state._as_state_update(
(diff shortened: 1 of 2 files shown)


maximvlah:
@masenf @benedikt-bartscher ready for review

benedikt-bartscher (approved):

reflex/state.py
```diff
@@ -1742,8 +1742,11 @@ async def _as_state_update(
         except Exception as ex:
             state._clean()
 
+            app = prerequisites.get_and_validate_app().app
             event_specs = (
-                prerequisites.get_and_validate_app().app.backend_exception_handler(ex)
+                app.backend_exception_handler(ex)
+                if app.backend_exception_handler
```
adhami3310: it would be a tad better to check with `is not None` just in case a Var might come here, otherwise looks good!

maximvlah: @adhami3310 thx, just fixed it

adhami3310 (dismissed):

adhami3310:
the only issue is that this doesn't disable the frontend `onError` from triggering, it just that when it goes to the backend, it does nothing

benedikt-bartscher (changes requested):
> the only issue is that this doesn't disable the frontend `onError` from triggering, it just that when it goes to the backend, it does nothing

That was the original intention/idea in https://github.com/reflex-dev/reflex/issues/3776. Can you adjust it @maximvlah?

masenf (approved):
seems fine to me, are we wanting to disable the entire frontend exception handling mechanism if the handler is None? because that's definitely more work

adhami3310:
i think to truly resolve the associated ticket, we do want to actually disable onError

adhami3310:
closing this as it doesn't disable onError

PR #5709, opened 2025-08-13
add rx memo marker to memo fields

adhami3310:


reflex/components/component.py
```diff
@@ -2005,18 +2005,20 @@ def get_args_spec(key: str) -> types.ArgsSpec | Sequence[types.ArgsSpec]:
 
         super()._post_init(
             event_triggers={
-                key: EventChain.create(
+                key + MEMO_MARKER: EventChain.create(
                     value=props[key],
                     args_spec=get_args_spec(key),
-                    key=key,
+                    key=key + MEMO_MARKER,
```
masenf: we can probably leave this one as it's only used for the error message, so it would keep the error message using the key that the user provides

masenf (approved):

Issue #5674, opened 2025-08-05
Since v0.8 redirects from `on_load` and `on_mount` handlers are lost when serving from a non-root frontend path.


tartansandal:
For context, I'm serving a number of reflex apps from different frontend paths, using a reverse proxy to deliver the appropriate frontend code and proxy the websocket requests to the appropriate backend. This this works well, provided I set `REFLEX_FRONTEND_PATH` and `REFLEX_API_URL` appropriately during the build and when running the apps, plus doing a little path wrangling in the reverse proxy config. All good.

Since v0.8, however, I've had a problem with `on_load` and `on_mount` handlers that I use to redirect users based on authentication status. Here's a minimal app to demonstrate the problem that just focuses on the `on_load` handlers.

```python
import reflex as rx  # noqa: D100


@rx.page("/")
def index() -> rx.Component:  # noqa: D103
    return rx.container(
        rx.vstack(
            rx.heading("This is the index page!"),
            rx.button("Go to Subpage!", on_click=rx.redirect("/subpage")),
        ),
    )


@rx.page("/subpage")
def subpage() -> rx.Component:  # noqa: D103
    return rx.container(
        rx.vstack(
            rx.heading("This is the sub page!"),
            rx.button("Go to index!", on_click=rx.redirect("/")),
            rx.button("Bounce to index!", on_click=rx.redirect("/bouncer")),
        )
    )


@rx.page("/bouncer", on_load=rx.redirect("/"))
def bouncer() -> rx.Component:  # noqa: D103
    return rx.container(
        rx.vstack(
            rx.heading("This is the bouncer page!"),
            rx.text("You should not be here!"),
            rx.button("Go to index!", on_click=rx.redirect("/")),
        ),
    )


app = rx.App()
```

This all works as expected when run in either development mode or in production with a single port and serving everything from root:

1. Clicking on the "Go to index!" and "Go to subpage!" buttons do just that.
2. Clicking on the "Bounce to index!" button causes the "/bouncer" page's `on_load` handler to redirect back to the index page.

However, if I adjust the environment variables to build and serve from "/prefix"

1. The "Go to index!" and "Go to subpage!" buttons work as expected
2. Clicking on the "Bounce to index!" button displays the "/bouncer" page but does not then redirect to the index page. (I do note that an extra "/" is appended to the path though).
3. Reloading the "/bouncer/" page redirects to the index page.

I get the same behaviour using both Caddy and Nginx as reverse proxies.

[...]

PR #5712, opened 2025-08-14
Add integration tests for frontend path

tartansandal:
Introduces an integration test for #5674.
This xfails on `main` but passes on  #5698

_Edit:_ also fixed an apparent bug in `test_link_hover.py`. There may be a deeper issue here.
_Edit:_ removed xfail after merging main which now includes #5698.
_Edit:_ added a test for 404s.

### DIFF
--- a/reflex/testing.py
+++ b/reflex/testing.py
@@ -1050,8 +1050,13 @@ class AppHarnessProd(AppHarness):
             / reflex.utils.prerequisites.get_web_dir()
             / reflex.constants.Dirs.STATIC
         )
+        # When frontend_path is set, the build output (including 404.html) is
+        # nested under that prefix directory, mirroring what sirv does in real
+        # prod deployments (see PackageJson.Commands.get_prod_command).
+        frontend_path = reflex.config.get_config().frontend_path.strip("/")
+        error_dir = web_root / frontend_path if frontend_path else web_root
         error_page_map = {
-            404: web_root / "404.html",
+            404: error_dir / "404.html",
         }
         with Subdir404TCPServer(
             ("", 0),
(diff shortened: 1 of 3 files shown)


tartansandal:
Note I'm using simple `environment` module tweaks to simulate changing the environment. An alternative would be to modify `AppHarness` to take a `config` object. That would be a much larger change, but might be useful testing some configuration scenarios.

shahshahii (approved):
Shah

masenf:
needs to be updated to remove the new Environment variables (those are exposed by default), and have the test update the `Config` instance directly.

tartansandal:
Thanks for taking another look at this! 🙏

I've rebased onto current main and revised the approach based on your feedback:

- **Removed all `environment.py` changes** — those env vars have become defaults since this PR was originally created, so the additions were redundant.
- **Switched to `os.environ`** for setting `REFLEX_FRONTEND_PATH` and `REFLEX_SHOW_BUILT_WITH_REFLEX` in the test fixtures. I initially tried mutating the `Config` instance directly (e.g. `get_config().frontend_path = "/prefix"`), but `AppHarness._initialize_app` calls `get_config(reload=True)` which creates a fresh Config from disk and wipes out any prior mutations. Setting the values via `os.environ` means they survive the reload through `Config.update_from_env()` — and it mirrors how users actually configure these in practice.
- **Fixed `AppHarnessProd._run_frontend`** — it was hardcoding the 404.html path at the web root, but when `frontend_path` is set the build output (including 404.html) is nested under that prefix. Updated it to match what `sirv` does in real prod deployments.
- **Added prod-mode prefix tests** — the original fixtures were inconsistently using `AppHarnessProd` for baselines but `AppHarness` (dev mode) for prefix cases. Now both dev and prod are covered for the prefix scenarios.

All 7 tests passing, pre-commit checks green ✅

tartansandal:
TIL: `monkeypatch.setenv` :smile:

PR #5725, opened 2025-08-20
improve literal handling in pyi generator

adhami3310:


adhami3310:
it's for reflex enterprise, it occasionally fails :/

PR #5742, opened 2025-08-26
add few missing svg elements

adhami3310:


masenf (approved):
is this just for completeness? or is the LLM generating these kind of structures?

adhami3310:
react flow has a few of those in their examples

PR #5744, opened 2025-08-28
optimize various functions

adhami3310:


reflex/vars/base.py
```diff
@@ -1153,18 +1164,6 @@ def _without_data(self):
...
-
-        Args:
-            instance: The instance to get the var from.
-            owner: The owner of the var.
-
-        Returns:
-            The var.
-        """
-        return self
-
```
masenf, line 1167: 🤔  why was this here originally?

masenf (approved):

adhami3310:
> 🤔 why was this here originally?

i believe we were hoping typing catches on it so we can make Var[str] into StringVar[str] but it got abandoned half way through

PR #5745, opened 2025-08-28
add option to disable hmr

adhami3310:


reflex/compiler/templates.py
```diff
@@ -537,12 +539,25 @@ def vite_config_template(base: str):
...
+  }};
+}}
+
 export default defineConfig((config) => ({{
   plugins: [
     alwaysUseReactDomServerNode(),
     reactRouter(),
     safariCacheBustPlugin(),
-  ],
+  ].concat({"[fullReload()]" if force_full_reload else "[]"}),
```
masenf, line 560: why `concat` and not just `{"fullReload()" if force_full_reload else ""}` in the plugins list?

masenf (approved):

reflex/compiler/templates.py
adhami3310: not a strong reason i just wanted it to be at least in theory extendable but it's not different either way

PR #5771, opened 2025-09-09
 add extra headers to upload endpoint from environment variables

adhami3310:


reflex/event.py
```diff
@@ -890,9 +890,7 @@ def as_event_spec(self, handler: EventHandler) -> EventSpec:
             ),
             (
                 Var(_js_expr="extra_headers"),
-                LiteralVar.create(
-                    dict(environment.REFLEX_UPLOAD_ENDPOINT_EXTRA_HEADERS.get())
-                ),
+                LiteralVar.create(self.extra_headers or {}),
```
masenf: if `self.extra_headers` is a `Var`, this will blow up. better to use `is not None` or just let it be a literal None Var

masenf (approved):

PR #5796, opened 2025-09-16
avoid runtime errors on iterating over non-array

adhami3310:


adhami3310:
that has less to do with iterating over an object and more of calling entries on a null

masenf:
the motivation for this change is to minimize the number of runtime errors that users (AI) accidentally write when dealing with state and foreach; so i think we should extend this to Object.entries at least in the case of `rx.foreach`, it shouldn't blow up if the value is null or undefined.

adhami3310:
i made it such that keys values and entries wouldn't throw a runtime error if any value gets passed to them

PR #5804, opened 2025-09-19
emit_update: take `token` instead of `sid`

masenf:
This allows the app to be more resilient in the face of websocket reconnects. The event is processed against a token, so there's no reason to maintain websocket affinity for event processing. Whenever the update is ready to send, it will be sent to the current websocket/sid associated.


masenf:
merged into #5805

PR #5818, opened 2025-09-24
Upgrade to react-player v3.3.3

masenf:
This is a breaking change as many props and event handlers have changed name and signature. Some care has been taken to maintain a compatible API for Reflex users of `rx.video`.


adhami3310:
greptile caught a type :hand_over_mouth:

adhami3310:
react player no longer requires NoSSRComponent as far as i can see

adhami3310 (approved):

PR #5823, opened 2025-09-25
bring back manual exception printing

adhami3310:


adhami3310:
i would have used `_print_exception_bltin` but i think that's pretty new in python

Issue #5629, opened 2025-07-25
rx.Field for backend vars


benedikt-bartscher:
Currently, there is no obvious way to use rx.Field and rx.field for backend vars (vars prefixed with `_`)
this is especially annoying with some ruff rules which prevent mutable default variables and suggest you to add a ClassVar annotations

benedikt-bartscher:
A backend: bool argument for rx.field would also be nice to allow backend vars regardless of the underscore prefix

adhami3310:
is it not possible to use `rx.Field` with them now? or does that not work?

benedikt-bartscher:
There are multiple issues atm, one of them is this:

```python
import reflex as rx


class State(rx.State):
    _backend: rx.Field[str] = rx.field("")

    @rx.event
    def hi(self) -> None:
        print(self._backend)
        self._backend = "hello"


def index() -> rx.Component:
    return rx.button("hi", on_click=State.hi)


app = rx.App()
app.add_page(index)
```

clicking the button twice emits this log:

```log
Field(default='', is_var=True, annotated_type=typing.Any)
hello
```

benedikt-bartscher:
seems like this is a regression. it used to work in older reflex versions

PR #5833, opened 2025-09-29
correctly handle backend rx.Field default values

### LINKED ISSUE #5829: Backend-only vars no longer mutable
**Describe the bug**
In Reflex 0.7.x, I was able to define a **backend-only** variable such as `_list_1: rx.Field[list[int]] = rx.field([1])` and then call methods such as `self._list_1.clear()` or `self._list_1.append(2)`. This no longer works in 0.8.12.

**To Reproduce**

```python
import reflex as rx

class State(rx.State):
    """The app state."""
    _list_1: rx.Field[list[int]] = rx.field([1])

    def do_something(self):
        self._list_1.append(1)
        print("success")


def index() -> rx.Component:
    return rx.container(
        rx.button("do something", on_click=State.do_something),
        spacing="5",
        justify="center",
        min_height="85vh",
    )


app = rx.App()
app.add_page(index)
```

Shown error: **AttributeError: 'Field' object has no attribute 'append'**

**Expected behavior**
It should just work (like it does for normal State variables that are not prepended with `_`.

- Python Version: 3.12.3
- Reflex Version: 0.8.12
- OS: Linux

adhami3310:
fixes #5829

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -533,7 +533,7 @@ class BaseState(EvenMoreBasicBaseState):
         cls._check_overridden_computed_vars()
 
         new_backend_vars = {
-            name: value
+            name: value if not isinstance(value, Field) else value.default_value()
             for name, value in list(cls.__dict__.items())
             if types.is_backend_base_variable(name, cls)
         }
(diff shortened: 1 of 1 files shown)


benedikt-bartscher:
Closes https://github.com/reflex-dev/reflex/issues/5629 ?

adhami3310:
> Closes #5629 ?

yes!

PR #5841, opened 2025-10-01
replace click exit with system exit

adhami3310:


reflex/custom_components/custom_components.py
```diff
@@ -694,11 +694,11 @@ def _collect_details_for_gallery():
             console.error(
                 f"{package_name} is owned by another user. Unable to update the information for it."
             )
-            raise click.exceptions.Exit(code=1)
+            raise SystemExit(1)
         response.raise_for_status()
     except httpx.HTTPError as he:
         console.error(f"Unable to complete request due to {he}.")
-        raise click.exceptions.Exit(code=1) from he
+        raise SystemExit(1) from None
```
masenf, line 701: wouldn't `raise SystemExit(1)` suffice here if we didn't want to chain the exception?

masenf (approved):

reflex/custom_components/custom_components.py
adhami3310: ruff complains if we don't day from None

PR #5859, opened 2025-10-08
python3.14 compat

masenf:
Compatibility for python3.14 (released October 7, 2025)


adhami3310 (approved):

masenf:
turns out there actually is a nice api for this, will update

PR #5891, opened 2025-10-15
0816dev

adhami3310:


tests/units/components/test_tag.py
```diff
@@ -30,7 +32,7 @@ def test_format_props(props: dict[str, Var], test_props: list):
     ("prop", "valid"),
     [
         (1, True),
-        (3.14, True),
+        (math.pi, True),
```
masenf:
```suggestion
        (3.16, True),
```

tests/units/test_base.py
```diff
@@ -35,7 +37,7 @@ def test_json(child):
     Args:
         child: A child class.
     """
-    assert child.json().replace(" ", "") == '{"num":3.14,"key":"pi"}'
+    assert child.json().replace(" ", "") == '{"num":3.141592653589793,"key":"pi"}'
```
masenf: i think this is a bit insane. can we just use a different constant?

PR #5918, opened 2025-10-21
add context for lock expiry

adhami3310:


reflex/istate/manager/__init__.py
```diff
@@ -71,21 +72,27 @@ async def get_state(self, token: str) -> BaseState:
         """
 
     @abstractmethod
-    async def set_state(self, token: str, state: BaseState):
+    async def set_state(
+        self, token: str, state: BaseState, *, context: Event | None = None
```
masenf: making the `context` an `Event` in the public API seems a bit limiting... what if we need to pass something else through later? then we need to break the public API or add another named kwarg.

i would suggest either making this `event_context` or some more generic type that we can attach the event to (similar to the Plugin contexts)

adhami3310: My only concern was that we would be making a type for each event, bug i guess that would be not an issue considering what else we do

masenf (approved):

PR #5922, opened 2025-10-23
ENG-8113: handle_frontend_exception triggers auto reload

masenf:
For configurable types of errors refresh once per configurable period in an attempt to automatically clear spurious issues when reloading.

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -2487,8 +2487,6 @@ class FrontendEventExceptionState(State):
         re.compile(re.escape("TypeError: null is not an object")),  # Safari
         re.compile(r"TypeError: can't access property \".*\" of null"),  # Firefox
     ]
-    # Reload the page only once per cooldown period to avoid reload loops.
-    auto_reload_cooldown_time_ms: ClassVar[int] = 10000  # 10 seconds
 
     @event
     def handle_frontend_exception(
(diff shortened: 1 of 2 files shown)


reflex/state.py
```diff
@@ -2473,8 +2474,17 @@ def wrapper() -> Component:
 class FrontendEventExceptionState(State):
     """Substate for handling frontend exceptions."""
 
+    # If the frontend error message contains any of these strings, automatically reload the page.
+    auto_reload_on_errors: ClassVar[list[str]] = [
+        "TypeError: Cannot read properties of null",
```
adhami3310: this is a browser specific error, check against firefox/safari/chrome for more solid coverage

reflex/state.py
```diff
@@ -2471,15 +2472,23 @@ def wrapper() -> Component:
...
+    auto_reload_on_errors: ClassVar[list[re.Pattern]] = [
+        re.compile(  # Chrome/Edge
+            re.escape("TypeError: Cannot read properties of null")
+        ),
+        re.compile(re.escape("TypeError: null is not an object")),  # Safari
+        re.compile(r"TypeError: can't access property \".*\" of null"),  # Firefox
     ]
     # Reload the page only once per cooldown period to avoid reload loops.
-    auto_reload_cooldown_time_ms: ClassVar[int] = 10000
+    auto_reload_cooldown_time_ms: ClassVar[int] = 10000  # 10 seconds
```
adhami3310: using `10_000` might be better than adding a comment

reflex/state.py
```diff
@@ -2471,15 +2472,23 @@ def wrapper() -> Component:
...
 
     # If the frontend error message contains any of these strings, automatically reload the page.
-    auto_reload_on_errors: ClassVar[list[str]] = [
-        "TypeError: Cannot read properties of null",
+    auto_reload_on_errors: ClassVar[list[re.Pattern]] = [
+        re.compile(  # Chrome/Edge
+            re.escape("TypeError: Cannot read properties of null")
+        ),
+        re.compile(re.escape("TypeError: null is not an object")),  # Safari
+        re.compile(r"TypeError: can't access property \".*\" of null"),  # Firefox
```
adhami3310, line 2488: do we want those to be configurable via env var?

reflex/state.py
adhami3310: might need to come from an env var as well

reflex/state.py
masenf: these i don't think there's a nice way with the escaping and stuff.

if users want to customize the `auto_reload_on_errors`, they can import this class in rxconfig.py and modify the list

reflex/state.py
masenf: added this as an env var

adhami3310 (approved):

PR #5925, opened 2025-10-23
more auto_reload_on_error patterns

masenf:
Might be more that we want to add based on testing.


reflex/state.py
```diff
@@ -2486,6 +2486,14 @@ class FrontendEventExceptionState(State):
         ),
         re.compile(re.escape("TypeError: null is not an object")),  # Safari
         re.compile(r"TypeError: can't access property \".*\" of null"),  # Firefox
+        # Firefox: property access is on a function that returns null.
```
adhami3310, line 2489: technically it's not a function, it just outputs whatever expression it was if it was not literally null

adhami3310 (approved):

PR #5938, opened 2025-10-29
allow state mutation with contextmanager from normal (non-background) events

benedikt-bartscher:
this makes it much easier to write reusable code which can be used in background and "normal" events and avoids a lot of dirty workarounds to make it possible

we can also choose to make this the default instead of adding just another config option

### DIFF
--- a/reflex/environment.py
+++ b/reflex/environment.py
@@ -727,11 +727,6 @@ class EnvironmentVariables:
     # How long to wait between automatic reload on frontend error to avoid reload loops.
     REFLEX_AUTO_RELOAD_COOLDOWN_TIME_MS: EnvVar[int] = env_var(10_000)
 
-    # Whether to allow entering a state mutation with `async with self` in normal (non background) event handlers
-    REFLEX_STATE_ALLOW_CONTEXTMANAGER_WITHOUT_BACKGROUND_TASK: EnvVar[bool] = env_var(
-        False
-    )
-
 
 environment = EnvironmentVariables()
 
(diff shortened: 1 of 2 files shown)


masenf (commented):
yeah lets bring it without the config option. this is fine

masenf:
one other thing to keep in mind is that `async with self` does an implicit `yield` of interim update to the frontend. if we just allow `async with self` to return the real state in non-background tasks, then the behavior would be slightly different.

benedikt-bartscher:
> yeah lets bring it without the config option. this is fine

I hoped for that answer

> one other thing to keep in mind is that `async with self` does an implicit `yield` of interim update to the frontend. if we just allow `async with self` to return the real state in non-background tasks, then the behavior would be slightly different.

That's true, but I found that in many use cases that's not really an issue

benedikt-bartscher:
> one other thing to keep in mind is that `async with self` does an implicit `yield` of interim update to the frontend. if we just allow `async with self` to return the real state in non-background tasks, then the behavior would be slightly different.

While reading this again: I was aware of that, but never really thought deeply about this. Is there a simple way to avoid the implicit `yield`? For some use-cases it could be nice to allow only locking the state for backend var modification without sending the implicit update to the frontend. But that's another topic and should not be part of this PR

masenf:
the implicit yield is an implementation detail of using `async with app.modify_state` inside the background task `StateProxy`; because it happens out-of-band or normal event processing, we always send a `StateUpdate` when using `app.modify_state`. There's not really an easy way to do that with the current mechanism for processing events.

But i'm inclined to paper over that, with maybe a TODO comment for now, since i'm planning to refactor the event processing pipeline this weekend and will make it handle this case, and possibly even make the implicit yield an option that can be disabled.

PR #5944, opened 2025-11-04
0819dev

adhami3310:


reflex/components/datadisplay/code.py
```diff
@@ -382,7 +382,7 @@ class Theme:
 class CodeBlock(Component, MarkdownComponentMap):
     """A code block."""
 
-    library = "react-syntax-highlighter@15.6.6"
+    library = "react-syntax-highlighter@16.1.0"
```
masenf: i think there was a reason we had this one pinned on 15, something about how the `inline` prop gets passed from the markdown component. iirc it was making both inline code spans and code blocks look like code blocks

masenf (approved):

PR #5954, opened 2025-11-06
Making `rx.download()` more robust

riebecj:
Ran into issues with formatting when downloading files with PEM and DER formats due to the data passing through `urllib.parse.quote(data)`. So, I followed a similar approach to  `getBackendURL(env.UPLOAD)` and add a `DOWNLOAD_AS_BLOB:` prepend to the data which is checked for and processed. If `as_blob` is set to True, it is the first thing processed, otherwise the function logic continues as previously designed.

Adds two new kwargs to `rx.download()`:
     - `as_blob`: tells the JS event to use `const blob = new Blob(...)` and `URL.createObjectURL(blob)` to download the object.
     - `mime_type`: Optional mime-type specification when downloading `as_blob`. The default is `"application/octet-stream"` as JS blobs are typically byte data, but the user can get more specific (e.g. `application/x-pem-file`).

Also, there are no tests for `rx.download()` so no new tests were added.

> NOTE: There's still an issue when taking something like a PEM or DER formatted string and passing it as `bytes` to `rx.download()` even if using `as_blob`. The issue is the data get's downloaded as a raw bytearray rather than the true data format prior to bytearray conversion. Still not sure how to fix it but the functionality I need is added with this PR.

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

### DIFF
--- a/reflex/.templates/web/utils/state.js
+++ b/reflex/.templates/web/utils/state.js
@@ -274,7 +274,8 @@ export const applyEvent = async (event, socket, navigate, params) => {
         [event.payload.url.replace("DOWNLOAD_AS_BLOB:", "")],
         { type: "application/octet-stream" },
       );
-      a.href = URL.createObjectURL(blob);
+      const url = URL.createObjectURL(blob);
+      a.href = url;
       window.setTimeout(function () {
         window.URL.revokeObjectURL(url);
       }, 60000); // Wait 60 seconds


adhami3310:
closing in favor of https://github.com/reflex-dev/reflex/pull/5957, thanks for the help!

masenf:
@riebecj we ended up taking a more generic fix, but I'm curious if you had a repro case to share where you were hitting issues.

I tried to reproduce this and wasn't able to.

```python
"""Test files generated with:

ssh-keygen -t rsa -m PEM -f ./foo.pem
openssl rsa -outform der -in ./foo.pem -out foo.der
"""
from pathlib import Path

import reflex as rx


class State(rx.State):
    @rx.event
    def download_pem(self):
        return rx.download(
            data=Path("foo.pem").read_text(),
            filename="certificate.pem",
        )

    @rx.event
    def download_der(self):
        return rx.download(
            data=Path("foo.der").read_bytes(),
            filename="certificate.der",
        )



def index() -> rx.Component:
    return rx.fragment(
        rx.color_mode.button(),
        rx.vstack(
            rx.link("Download PEM file", on_click=State.download_pem),
            rx.link("Download DER file", on_click=State.download_der),
        ),
    )


app = rx.App()
app.add_page(index)
```

riebecj:
> @riebecj we ended up taking a more generic fix, but I'm curious if you had a repro case to share where you were hitting issues.
> 
> I tried to reproduce this and wasn't able to.
> 
> ```python
> """Test files generated with:
> 
> ssh-keygen -t rsa -m PEM -f ./foo.pem
> openssl rsa -outform der -in ./foo.pem -out foo.der
> """
> from pathlib import Path
> 
> import reflex as rx
> 
> 
> class State(rx.State):
>     @rx.event
>     def download_pem(self):
>         return rx.download(
>             data=Path("foo.pem").read_text(),
>             filename="certificate.pem",
>         )
> 
>     @rx.event
>     def download_der(self):
>         return rx.download(
>             data=Path("foo.der").read_bytes(),
>             filename="certificate.der",
>         )
> 
> 
> 
> def index() -> rx.Component:
>     return rx.fragment(
>         rx.color_mode.button(),
>         rx.vstack(
>             rx.link("Download PEM file", on_click=State.download_pem),
>             rx.link("Download DER file", on_click=State.download_der),
>         ),
>     )
> 
> 
> app = rx.App()
> app.add_page(index)
> ```

Instead of reading a file directly, pass is as an `rx.Var` like this:
```python
rx.download(
    data=rx.Var.create(Path("foo.pem").read_text()),
    filename="certificate.pem",
)
```
You'll see what I mean.

riebecj:
@masenf @adhami3310 See my [comment](https://github.com/reflex-dev/reflex/pull/5957#issuecomment-3502836288)

PR #5957, opened 2025-11-06
add mime_type to download

adhami3310:
based on #5954

### DIFF
--- a/reflex/event.py
+++ b/reflex/event.py
@@ -4,7 +4,6 @@ import dataclasses
 import inspect
 import sys
 import types
-import urllib.parse
 from base64 import b64encode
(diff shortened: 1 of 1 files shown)


riebecj:
This still doesn't fix my problem when the `data` is an instance of `rx.Var` because you're double string-quoting. The result is a downloaded file as a `"My multiline\nstring"` The fix for this is changing [this line](https://github.com/reflex-dev/reflex/pull/5957/files#diff-bcdf9d6a8341b46fa93d8f132dd65b3781bc6fd6930e4d941a6bef6849579493R1288) to `f"data:{mime_type}," + data.to_string(use_json=False),`

PR #5961, opened 2025-11-07
Improve handling `Var`'s when passing them as `data` to `rx.download(...)`

riebecj:
Related to #5954 and #5957.

When `data` is provided as a `Var`, strings could sometimes have formatting issues and bytes were `ArrayVar`'s that were really not handled at all, as there is no such thing as byte-strings in JavaScript. They were correctly converted to byte arrays, but the resultant downloaded file was just a string representation of the byte array. 

- `Var[bytes]`: The byte array is coerced into a `Uint8Array` and converted to a `Blob` using the specified `mime_type`. The object URL is then created from the Blob. 
- `Var[str]`: The string data is left intact to preserve formatting. 
- `Var[Any]`: Any other supplied state or non-state variable is stringify-ed. 
- **Base64 Encoded**: To ensure correct data translation, the var data (except for bytes) is base64 encoded.

All of this is done within the variable ecosystem, as there seems to be some aversion to changing any of the Reflex embedded JS (not a gripe, just an observation).

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

### DIFF
--- a/reflex/vars/blob.py
+++ b/reflex/vars/blob.py
@@ -19,6 +19,9 @@ if TYPE_CHECKING:
 class Blob:
     """Represents a JavaScript Blob object."""
 
+    data: str | bytes = ""
+    mime_type: str = ""
+
 
 BLOB_T = TypeVar("BLOB_T", bound=bytes | str, covariant=True)
 
@@ -27,10 +30,34 @@ class BlobVar(Var[BLOB_T], python_types=Blob):
     """A variable representing a JavaScript Blob object."""
 
     @classmethod
+    def _determine_mime_type(cls, value: str | bytes | Blob | Var) -> str:
+        mime_type = ""
+        if isinstance(value, str | bytes | Blob):
+            match value:
+                case str():
+                    mime_type = "text/plain"
+                case bytes():
+                    mime_type = "application/octet-stream"
+                case Blob():
+                    mime_type = value.mime_type
+
+        elif isinstance(value, Var):
+            if isinstance(value._var_type, str):
+                mime_type = "text/plain"
+            if isinstance(value._var_type, bytes):
+                mime_type = "application/octet-stream"
+
+        if not mime_type:
+            msg = "Unable to determine mime type for blob creation."
+            raise ValueError(msg)
+
+        return mime_type
+
+    @classmethod
     def create(
         cls,
-        value: str | bytes | Var,
-        mime_type: str | Var,
+        value: str | bytes | Blob | Var,
+        mime_type: str | Var | None = None,
         _var_data: VarData | None = None,
     ):
         """Create a BlobVar from the given value and MIME type.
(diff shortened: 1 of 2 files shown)


riebecj:
I don't think the failed test has anything to do with my code changes, but I can update if necessary.

reflex/event.py
```diff
@@ -1285,7 +1288,16 @@ def download(
             url = cond(
                 is_data_url,
                 data.to(str),
-                f"data:{mime_type}," + data.to_string(),
+                (
+                    CREATE_OBJECT_URL.call(data.to_blob(mime_type=mime_type))
+                    if isinstance(data, ArrayVar)
```
masenf: not generally safe to convert any `ArrayVar` to a blob, for example an array of strings or array of objects gets converted into something weird now.

i like the new "if Var is a string, then encode it and send it as is", but i don't like the automatic `to_blob`, i would contend that's something the caller should be doing before calling `rx.download`.

we can update the `is_data_url` method to look for the `blob:` prefix in addition to `data:`.

riebecj:
@masenf, I can work on that. 

What about handling of Python `bytes`? Currently, nothing in the Var system is designed to ensure handling of Python byte strings to JS byte arrays. Since `bytes` are a Python primitive, I would suggest that it gets first class treatment before Blobs, urls, and base64 encoding. If the idea is to NOT handle bytes and require conversion to `str`, it should not be an accepted type anywhere.

riebecj:
Converting to draft to use as a work space for `BytesVar` and `BlobVar`.

This will be abandoned and a new one created when the architecture and implementation of the referenced vars is agreed upon.

riebecj:
@masenf , Take a look at `BytesVar` and `BlobVar`. Let me know where things should be changed/improved/etc.

I ran it through multiple example tests using direct State vars, computed vars, and even creating them directly. All seem to function correctly, download with proper formatting, and supports Unicode.

riebecj:
All the type hinting seems to be there too. 

I was a little wary about using `encode` and `decode` because of the underlying `_decode` for Vars, but it felt the most Pythonic given its implementation type.

reflex/vars/blob.py
```diff
@@ -0,0 +1,118 @@
...
+
+        Returns:
+            A BlobVar instance representing the JavaScript Blob object.
+        """
+        if not isinstance(value, Var):
+            value = LiteralVar.create(value)
+        if not isinstance(mime_type, Var):
+            mime_type = LiteralVar.create(mime_type)
+        elif type(value).__qualname__.endswith("BytesCastedVar"):
+            value = f"new Uint8Array({value})"
```
masenf: if you know it's a Var, then you should be able to get `._var_type` as `bytes` instead of string scanning the class name

reflex/vars/blob.py
```diff
@@ -0,0 +1,118 @@
...
+    var_operation_return,
+)
+
+if TYPE_CHECKING:
+    from reflex.vars import Var
+
+
+@dataclasses.dataclass
+class Blob:
+    """Represents a JavaScript Blob object."""
```
masenf, line 20: it seems like this structure should include some sequence `value` and a `mime_type` so you could actually do something like

`data = Var.create(Blob(file_contents, mime_type="application/octet-stream"))`

Otherwise it doesn't really make sense to have a LiteralBlobVar, if there is no actual functional python type backing it.

masenf (commented):
overall, this looks quite good. i left a couple of comments.

hopefully this work makes it easier to deal with blobby things, like media streams, in reflex

riebecj:
@masenf Made the changes you suggested.

riebecj:
@masenf Any more suggestions?

PR #5966, opened 2025-11-10
Fix/Socket reconnection rehydrates

### LINKED ISSUE #5963: When socket reconnects, and state is expired, it should rehydrate
Users are reporting issues where after the state expires on the backend, if their socket reconnects, the old state lingers in the client until they send some other event which triggers a `reload`, which triggers a rehydrate.

The new behavior should be to always `reload` when reconnecting if the target state has no `router_data`. ie: on socket connect, if we have seen the token before, but the state is new, emit a reload event and let the frontend take care of rehydrating.

ParagGhatage:
### All Submissions:

- [X] Have you followed the guidelines stated in [CONTRIBUTING.md](https://github.com/reflex-dev/reflex/blob/main/CONTRIBUTING.md) file?
- [X] Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/reflex-dev/reflex/pulls) for the desired changed?

### Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Changes To Core Features:

# Fix: Reload frontend on socket reconnect when backend state has expired

## Description

This PR fixes the issue where the frontend retains stale state when the backend state has expired, and the socket reconnects silently.

### Problem
Previously, the backend only emitted a `reload` event when an incoming frontend event (e.g., a click) triggered state access.  
When the socket reconnected automatically - without any user interaction - the stale frontend state remained until the next user action caused rehydration.

### Fix
- Extended `on_connect()` in `app.py` to check the backend state immediately after reconnect.  
- If the state corresponding to the client token no longer exists or has an empty `router_data`, a `"reload"` event is emitted instantly to the client.  
- This ensures the frontend rehydrates as soon as it reconnects, keeping UI and backend state in sync.

### Behavior Summary
✅ Detects expired backend state right after reconnect.  
✅ Emits `"reload"` to prompt frontend rehydration.  
✅ Prevents users from seeing stale UI after backend restarts or token expiry.

## Linked Issue
Closes #5963

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -2092,27 +2092,32 @@ class EventNamespace(AsyncNamespace):
             return
 
         try:
-            #  Check if backend state already expired BEFORE linking the token
-            substate_key = _substate_key(token, self.app.state_manager.state)
-            state = await self.app.state_manager.get_state(substate_key)
+            # Detect reconnect (token seen before)
+            is_reconnect = token in self._token_manager.token_to_sid
 
-            if not getattr(state, "router_data", None):
-                console.debug(
-                    f"[Reflex] socket reconnect detected with expired token {token}, emitting reload"
-                )
-                await self.emit("reload", data={"reason": "state expired"}, to=sid)
-                return  # no need to proceed further since frontend will reload
-
-            # Only link after confirming state is still valid
+            # Always link first to ensure mappings are valid
             await self.link_token_to_sid(sid, token)
 
+            # If this is a reconnect and backend has lost state, trigger reload
+            if is_reconnect:
+                substate_key = _substate_key(token, self.app.state_manager.state)
+                state = await self.app.state_manager.get_state(substate_key)
+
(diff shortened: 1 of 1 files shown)


ParagGhatage:
@masenf ,
I have rectified the logical error mentioned by greptile in the second commit.
Please take a look.

ParagGhatage:
> this doesn't work for me at all. it tries to send `"reload"` on the initial connection with an invalid payload.
> 
> did you test this? did it work for you?

@masenf,
Sorry about the earlier code breaking - I realize now that the "reload" was being triggered even on the initial connection before the socket was fully established.
I’ve updated the logic in my latest commit to ensure the reload only fires when a reconnection is detected and the backend state has actually expired.

Also saw your PR #5969 - makes total sense to move this logic to the frontend and handle rehydration directly through the backend instead of a full reload. 
That approach looks much cleaner.

PR #5985, opened 2025-11-18
Adding `BlobVar` and `BytesVar`.

riebecj:
Provides better support for converting Var data to JS Blobs. Adds `BytesVar` to back Python `bytes` with JS `Uint8Array`. To keep things consistent and pythonic, `.encode(...)` and `.decode(...)` methods have been added to `StringVar`. Also, both `StringVar` and `BytesVar` have a `to_blob(...)` method that assists in creating a `new Blob(...)` from the Python data. The `BlobVar` has a `.create_object_url(...)` method that returns the URL string for downloading with `rx.download(...)`. It also seems to handle Unicode quite nicely.

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

### DIFF
--- a/docs/app/assets/tailwind-theme.css
+++ b/docs/app/assets/tailwind-theme.css
@@ -1699,6 +1699,14 @@
         /* padding: 0rem 0.125rem 0rem 0.125rem; */
     }
 
+    /* Override Radix Code (rt-Code) accent coloring so inline code matches slate theme */
+    code.rt-Code,
+    code.rt-Code.rt-variant-soft,
+    .rt-Code.rt-variant-soft {
+        color: var(--c-slate-11) !important;
+        background-color: var(--c-slate-3) !important;
+    }
+
     .code-error-style {
         font-family: var(--font-jetbrains);
         font-size: 0.835rem;
@@ -1803,6 +1811,44 @@
         }
     }
 
+    .code-block button,
+    .code-block > button {
+        border: none !important;
+        border-width: 0 !important;
+        background: transparent !important;
+        background-color: transparent !important;
+        opacity: 0 !important;
+        transition: opacity 0.15s ease-out !important;
+        pointer-events: none;
+        display: inline-flex !important;
+        align-items: center !important;
+        gap: 0.375rem !important;
+        font-size: 0.8125rem !important;
+        font-weight: 500 !important;
+        color: var(--c-slate-11) !important;
+        top: 8px !important;
+        right: 12px !important;
+        padding: 6px 8px !important;
+    }
+
+    .code-block button::after,
+    .code-block > button::after {
+        content: "Copy";
+    }
+
+    .code-block:hover button,
+    .code-block:hover > button,
+    .code-block button:focus-visible {
+        opacity: 1 !important;
+        pointer-events: auto;
+    }
+
+    .code-block button:hover,
+    .code-block > button:hover {
+        background: var(--c-slate-3) !important;
+        background-color: var(--c-slate-3) !important;
+    }
+
 
     .tab-style {
         color: var(--c-slate-9);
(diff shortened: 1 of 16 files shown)


reflex/vars/sequence.py
```diff
@@ -1287,6 +1327,120 @@ def json(self) -> str:
...
+
+    Args:
+        value: The BytesVar to decode.
+        encoding: The character encoding to use for decoding.
+
+    Returns:
+        A StringVar containing the decoded string.
+    """
+    return var_operation_return(
+        f"(new TextDecoder({encoding})).decode(new Uint8Array({value}))",
```
masenf: > `rx.Var.create(b"foobar").decode()`

Produces code that duplicates the Uint8Array (it still works, but it seems suboptimal)

```
(new TextDecoder("utf-8")).decode(new Uint8Array(new Uint8Array([102, 111, 111, 98, 97, 114])))
```

reflex/vars/sequence.py
```diff
@@ -1287,6 +1327,120 @@ def json(self) -> str:
...
+        Args:
+            value: The string or bytes value to create the var from.
+            _var_data: Additional hooks and imports associated with the Var.
+
+        Returns:
+            A LiteralBytesVar representing the bytes value.
+        """
+        if isinstance(value, str):
+            value = value.encode()
+        return LiteralVar.create(value, _var_data=_var_data)
```
masenf: The non-literal BytesVar should not define `create` like this -- that's what the `LiteralBytesVar` is for.

The non-literal Var just holds the valid operations for that type, and generally speaking the `_js_expr` will be some variable name (imagine a State variable of type `bytes`, the `_js_expr` would be the state+var name).

reflex/vars/blob.py
```diff
@@ -0,0 +1,163 @@
...
+        """Create a BlobVar from the given value and MIME type.
+
+        Args:
+            value: The data to create the Blob from (string, bytes, or Var).
+            mime_type: The MIME type of the Blob (string or Var).
+            _var_data: Optional variable data.
+
+        Returns:
+            A BlobVar instance representing the JavaScript Blob object.
+        """
```
masenf: Combine this with LiteralBlobVar.create and remove the definition in the non-literal BlobVar

reflex/vars/blob.py
```diff
@@ -0,0 +1,163 @@
...
+class LiteralBlobVar(LiteralVar, BlobVar):
+    """A literal version of a Blob variable."""
+
+    _var_value: Blob = dataclasses.field(default_factory=Blob)
+
+    @classmethod
+    def create(
+        cls,
+        value: bytes | str | Blob,
+        mime_type: str | None = None,
```
masenf: i think this method should just take a `Blob` object to simplify the API.

it also needs to assign that `Blob` object to `_var_value` so it can be pulled out on the python side later.

reflex/vars/blob.py
```diff
@@ -0,0 +1,163 @@
...
+            if isinstance(value._var_type, str):
+                mime_type = "text/plain"
+            if isinstance(value._var_type, bytes):
+                mime_type = "application/octet-stream"
+
+        if not mime_type:
+            msg = "Unable to determine mime type for blob creation."
+            raise ValueError(msg)
+
+        return mime_type
```
masenf: suggest moving this logic into a `create` function for the `Blob` dataclass

reflex/vars/blob.py
```diff
@@ -0,0 +1,163 @@
...
+            mime_type = LiteralVar.create(mime_type)
+
+        if isinstance(value, str | bytes):
+            value = LiteralVar.create(value)
+
+        elif isinstance(value, Blob):
+            value = LiteralVar.create(value.data)
+
+        if isinstance(value._var_type, bytes):
+            value = f"new Uint8Array({value})"
```
masenf: if the value is a BytesVar, then you don't need this additional conversion, because the BytesVar already includes the array cast in its `_js_expr`, so when you format it into the expression, it will already appear like this

`str(rx.Var(f"{rx.Var.create(b"foo")}"))` -> `'new Uint8Array([102, 111, 111])'`

reflex/vars/sequence.py
```diff
@@ -1287,6 +1327,120 @@ def json(self) -> str:
...
+        mime_type: The MIME type of the blob.
+
+    Returns:
+        A Blob object containing the data.
+    """
+    if isinstance(value, BytesVar):
+        return var_operation_return(
+            js_expression=f"new Blob([new Uint8Array({value})], {{ type: {mime_type} }})",
+            var_type=Blob,
+        )
```
masenf: this case is unnecessary; if the `value` is a `BytesVar`, then simply formatting it into the js_expression below will use it's JS-form which already includes the Uint8Array cast

masenf (commented):
try to reduce the redundant paths in the new code. there should never be two separate paths to format what is essentially the same value; the Var system is cumulative, so if you have already correctly dealt with some subtype in a different Var, you can just include that directly in other expressions.

pay attention to the diference between the base type Var which corresponds to some JS-native type (StringVar, BytesVar, ObjectVar, etc) and the LiteralVar which corresponds to some Python-native type. The LiteralVar is responsible for representing the python value in the JS world, and the base type Var is responsible for defining what operations on that type can be performed in JS on the frontend.

it would be helpful to add some test case, at the very least a few more examples in `tests/integration/test_var_operations.py` where these various conversions are used end-to-end in the browser to validate their functionality, but more importantly to notice when some future change in reflex causes their behavior to regress.

PR #5990, opened 2025-11-20
Allow computed vars to yield events and update state reliably

### LINKED ISSUE #5956: Should be able to `yield` chained events from a computed var
**Describe the Features**
When a computed var is being computed, you cannot reliably modify the state nor chain to any other event (especially background events).

It would be really nice if you could just yield chained events from a computed var, then return the final value for the var.

It would also be really nice if the `get_delta` computation in state, would re-run the computed var calculation until some interation limit or until no further dirty vars are generated. That way computed vars could directly modify state state vars and have them reliably persisted and sent to the frontend.

- What is the purpose of the feature?
Make computed vars more useful/less edge cases.

- Show an example / use cases for the new feature.
```python
class MyState(rx.State):
    _user_token: str
    logged_in: bool = False

    @rx.var
    def user_is_valid(self) -> bool:
        if not self._user_token:
            return False
        if not validate(self._user_token):
            yield rx.toast("Token was found to be invalid")
            self.logged_in = False
            self._user_token = ""
            return False
        return True
```

In this case, the `user_is_valid` setting `self._user_token` to blank would trigger a re-calculation of `user_is_valid` immediately, which would take the early exit path. The delta will include `logged_in = False`. A toast will be displayed to the user from the event which triggered this recalculation of `user_is_valid`.

Since computed vars are always recalculated during delta computation and the lock is held during this process, it's always safe to manipulate the state, the problem was that the final delta was not updated based on changes that occured to the state during computed var resolution.

aibysid:
```markdown
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

---

### Description

This PR enables **Computed Vars** to reliably trigger side effects, such as yielding events or modifying other state variables, during their computation.

Previously, computed vars were expected to be pure functions. Attempting to `yield` events (like `rx.toast`) resulted in the property returning a generator object instead of the calculated value. Additionally, if a computed var modified another state variable, that change was often missed in the current state update cycle because `get_delta` did not re-evaluate dependencies after the initial computation.

**Key Changes:**
1.  **Support for Generator Computed Vars**: Updated `ComputedVar.__get__` (in `reflex/vars/base.py`) to detect generator return values. It now iterates through the generator, collects yielded events, and returns the final value (captured from `StopIteration`).
2.  **Event Collection**: Added `_computed_var_events` to `BaseState` to store these yielded events and updated `_as_state_update` to include them in the response sent to the frontend.
3.  **Iterative State Updates**: Modified `BaseState.get_delta` (in `reflex/state.py`) to iterate (up to a limit) when calculating the delta. This ensures that if a computed var modifies a base var (marking it dirty), the loop continues to capture the ripple effects of that change in the same update cycle.

### Example Usage

```python
class MyState(rx.State):
    count: int = 0
    status: str = "Normal"

    @rx.var
    def check_status(self) -> str:
        if self.count > 10:
            # This side effect (event) is now captured
            yield rx.toast("Count is too high!")
            
            # This side effect (state modification) is now reliably captured in the same update
            self.status = "High" 
            return "Alert"
        return "OK"
```

### Tests
Added `tests/units/test_computed_var_side_effects.py` which verifies:
- A computed var correctly yields an event (e.g., `rx.window_alert`).
- A computed var correctly modifies another state variable, and the change is present in the state delta.

Closes #5956
```

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -366,6 +366,9 @@ class BaseState(EvenMoreBasicBaseState):
     # The parent state.
     parent_state: BaseState | None = field(default=None, is_var=False)
 
+    # Events triggered by computed vars.
+    _computed_var_events: list[EventSpec] = field(default_factory=list, is_var=False)
+
     # The substates of the state.
     substates: builtins.dict[str, BaseState] = field(
         default_factory=builtins.dict, is_var=False
@@ -440,6 +443,9 @@ class BaseState(EvenMoreBasicBaseState):
 
         # Create a fresh copy of the backend variables for this instance
         self._backend_vars = copy.deepcopy(self.backend_vars)
+
+        # Initialize computed var events list
+        self._computed_var_events = []
 
     def __repr__(self) -> str:
         """Get the string representation of the state.
@@ -1795,6 +1801,14 @@ class BaseState(EvenMoreBasicBaseState):
         try:
             # Get the delta after processing the event.
             delta = await state._get_resolved_delta()
+
+            # Collect events from computed vars
+            computed_var_events = state._collect_computed_var_events()
+            if computed_var_events:
+                fixed_events.extend(
+                    fix_events(self._check_valid(handler, computed_var_events), token)
+                )
+
             state._clean()
 
             return StateUpdate(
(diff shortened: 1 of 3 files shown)


aibysid:
One more test done by actual dummy app :
Screenshot 2025-11-20 at 16 58 12
Screenshot 2025-11-20 at 16 58 05

aibysid:
> this feels a bit hacky, but i'm willing to entertain it for now since my larger event processing refactor will take some more time to be ready.
> 
> greptile makes valid comments that should be addressed here, particularly the subdelta one

Yeah I agree that got overlooked will try to make a new commit with those getting addressed

masenf:

PR #6044, opened 2025-12-15
turn on inconsistentCjsInterop again

adhami3310:
it caused failures in `react-simple-code-editor`


masenf (approved):
working with recharts and moment

PR #6053, opened 2025-12-16
0824dev

adhami3310:


masenf (approved):
lol that greptile sequence diagram is comically useless

but hey lucide added some great new icons, like `stone` and `circle_pile` 🤠

PR #6056, opened 2025-12-17
hide __getattribute__ from type checking

adhami3310:
this improves pyright


reflex/istate/shared.py
```diff
@@ -190,11 +190,11 @@ async def _link_to(self, token: str) -> Self:
         if not token:
             msg = "Cannot link shared state to empty token."
             raise ReflexRuntimeError(msg)
-        if self._linked_to == token:
+        if self._linked_to == token:  # pyright: ignore[reportAttributeAccessIssue]
```
masenf: for the non-test code, can we add an `if not isinstance(self, SharedState): raise ReflexRuntimeError("Can only call `_link_to` on SharedState instances")`

PR #6062, opened 2025-12-21
Fix src file lookup when subfolder is provided in asset()

martiniifun:
When using `rx.asset(path="logo.jpg", shared=True, subfolder="images")`, the current implementation looks for `.../logo.jpg` instead of `.../images/logo.jpg`. This PR updates the logic to include the `subfolder` in the source file path check.


masenf:

Issue #6019, opened 2025-12-04
Need a better error path when backend sends unprocessable delta


masenf:
**Describe the Features**
Currently if the backend sends a delta that the frontend cannot understand, it (hopefully) throws `dispatch[substate] is not a function` in the browser, which is not useful for end users and if the user is not checking the browser console, it makes the app appear broken after the hydrate event. Ideally the backend should be made aware of this error as python devs often check their terminal for error logs before the browser console

Instead, reflex should internally catch this error in the `socket.on("event")` handler and raise an error with an actionable message that includes the failed state name and potential remediation steps (rebuild frontend, check api_url, etc). The error _should_ be fatal, because we never want to run an app with a mismatched frontend and backend as that indicates some deployment error that we want to catch during testing.

- What is the purpose of the feature?
More actionable error when frontend/backend have mismatched states.

bertbarabas:
Rather than implementing "The error should be fatal" as described above,  consider having the client and server recognize that the backend has changed since the last time the client was connected and the client should completely refresh itself.

PR #6100, opened 2026-01-24
Explicit state and event id minification

benedikt-bartscher:
## Summary

Add state and event name minification with CLI management tools.
- Introduces `minify.json` config file to map state/event names to short IDs (e.g., `myapp.state.AppState` → `"a"`)
- Adds `reflex minify` CLI command group for managing minification
- Reduces frontend bundle size and WebSocket payload size by replacing long state/event paths with compact identifiers

## CLI Commands

```bash
# Initialize minify.json with IDs for all states and events
reflex minify init

# Sync minify.json after code changes
reflex minify sync
reflex minify sync --prune --reassign-deleted

# Validate config against current codebase
reflex minify validate

# List state tree with minified names
reflex minify list
reflex minify list --json

# Lookup a minified path
reflex minify lookup "a.bU"
```

based on #6098


reflex/reflex.py
```diff
@@ -842,6 +843,229 @@ def rename(new_name: str):
...
+        # Calculate new prefix for children
+        child_prefix = prefix + ("    " if is_last else "|   ")
+
+        # Print event handlers
+        handlers = state_data["event_handlers"]
+        substates = state_data["substates"]
+        has_substates = len(substates) > 0
+
+        if handlers:
+            console.log(f"{child_prefix}|-- Event Handlers:")
```
abulvenz: Very tiny optical nit, cause sometimes EventHandlers is the last sub item.
It might work like this, didn't test it.
```suggestion
            console.log(f"{child_prefix}{'|' if has_substates else '`'}-- Event Handlers:")
```

benedikt-bartscher:
TODO: best-effort-mode
- walk trough the state id registry after the app module get's imported
- filter out all states with explicit state id
- sort the remaining states without state id by full name
- assign state ids - start with the highest user defined state id - users might have fragmentation by choice

a similar thing can be implemented for event handler id's

edit: not that easy, as state names are baked into to many things (events, vars etc). we would need to defer init_subclass and possibly parts of @rx.var and @rx.event

edit2: i have found a better approach with a minify.json which makes this obsolete

benedikt-bartscher:
Open for discussion: Should reflex reserve the first 5/10 state id's for internal states? Current reflex uses 0-3, but that might change in the future.

edit: with sibling uniqueness we would just need to reserve the state id's on the first rx.State subclass level.

edit2: i have found a better approach with a minify.json which makes this obsolete

benedikt-bartscher:
TODO: think about sibling uniqueness for states. currently using global uniqueness

edit: adjusted in [6b4c5fa](https://github.com/reflex-dev/reflex/pull/6100/commits/6b4c5fa88de5e41170491f58b5eb5a39e44576da)

benedikt-bartscher:
Idea: store ids in a minify.json file

- allows overwriting state ids for upstream dependencies and reflex core
- one centralized place for the ids
- keep track of reserved numbers
- better automatic instrumentation - state and event ids can all be managed with cli - without manual code changes
- can be loaded once on startup and is frozen - probably faster
- no best effort mode needed

reflex/components/recharts/cartesian.py
```diff
@@ -242,10 +242,8 @@ class Brush(Recharts):
     end_index: Var[int]
 
     # The fill color of brush
-    fill: Var[str | Color]
 
     # The stroke color of brush
-    stroke: Var[str | Color]
```
benedikt-bartscher: ?

reflex/utils/format.py
```diff
@@ -439,15 +440,25 @@ def format_props(*single_props, **key_value_props) -> list[str]:
...
-        The state and function name.
+        The state and function name (possibly minified based on minify.json).
     """
+    from reflex.event import EventHandler
+    from reflex.minify import is_event_minify_enabled
+    from reflex.state import State
+
+    assert isinstance(handler, EventHandler), (
+        f"Expected EventHandler, got {type(handler)}"
+    )
```
masenf: assertions in non-test code might be optimized out if one runs with `python -O`

since this runs as part of the compiler, it's better to check the condition and (in this case) `raise TypeError`

reflex/reflex.py
```diff
@@ -842,6 +842,411 @@ def rename(new_name: str):
...
+    path = _get_minify_json_path()
+    if path.exists():
+        console.error(
+            f"{MINIFY_JSON} already exists. Use 'reflex minify sync' to update "
+            "or delete the file to reinitialize."
+        )
+        raise SystemExit(1)
+
+    # Load the user's app to register all state classes
+    prerequisites.get_app()
```
masenf: just getting the app might not be sufficient to slurp all of the states. some states, like ComponentState or dynamically created states might only be created while evaluating pages.

maybe it's okay to leave those out of the minification?

masenf (commented):
just started reviewing this, will need to test on it a bit more.

the explicit management of the minified names via cli solves our earlier problems of mismatching the minified names between compile and build time, but it seems like a lot of extra api surface area for a mostly under the covers feature.

would it make sense to auto-minify and store the minify.json as a compilation artifact in the .web/backend directory? we already do something similar to save which pages need to be re-evaluated in backend-only mode in order to have a consistent state tree.

reflex/utils/format.py
benedikt-bartscher: this is mainly used to satisfy the type checker. A pattern which i personally prefer over `cast`. At runtime this should always be `EventHandler`. However, no strong preference, i just changed it to raise instead

reflex/reflex.py
benedikt-bartscher: yeah, i was aware of this limitation, maybe i can figure it out later. I'd love to support ComponentState as well

benedikt-bartscher:
Thanks you @masenf!

> the explicit management of the minified names via cli solves our earlier problems of mismatching the minified names between compile and build time, but it seems like a lot of extra api surface area for a mostly under the covers feature.

Yeah, it is. Advanced users and big deployments might prefer managing this manually. I think about it like db migrations. Manual management makes it possible to rename states and event handlers without any issues in production.

> would it make sense to auto-minify and store the minify.json as a compilation artifact in the .web/backend directory? we already do something similar to save which pages need to be re-evaluated in backend-only mode in order to have a consistent state tree.

The minify.json should be persisted/commited to have stable id's. However if we decide that minify becomes default, we can certainly auto-update the minify.json during `reflex run` and/or `reflex compile`

reflex/reflex.py
benedikt-bartscher: dynamic states are now supported

packages/reflex-base/src/reflex_base/utils/format.py
```diff
@@ -444,43 +444,55 @@ def format_props(*single_props, **key_value_props) -> list[str]:
...
+    cached = handler.__dict__.get(_FORMATTED_NAME_CACHE_ATTR)
+    if cached is not None:
+        return cached
     state, name = get_event_handler_parts(handler)
-    if state == "":
-        return name
-    return f"{state}.{name}"
+    full = name if state == "" else f"{state}.{name}"
+    object.__setattr__(handler, _FORMATTED_NAME_CACHE_ATTR, full)
+    return full
```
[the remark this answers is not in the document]
benedikt-bartscher: Documented design constraint rather than a defect of this cache. The per-class `get_name` / `get_full_name` lru_caches have exactly the same property (they predate this PR and became name-resolver-sensitive with it), so keying only the handler cache by context would fix nothing: the state half of the formatted name still comes from `get_full_name`. Dropping the cache puts a registry + resolver dispatch back on the per-event hot path. The only non-test fork is `AppHarness`, where the handlers shared between parent and fork are framework handlers whose names no resolver rewrites. The docstring states the one-resolver-per-process rule; per-context caching is a follow-up if multi-resolver processes ever become a requirement.

packages/reflex-base/src/reflex_base/registry.py
```diff
@@ -263,3 +351,102 @@ def get_substates(
...
+        slot.
+
+        Args:
+            resolver: The resolver to install. Pass :class:`DefaultNameResolver`
+                to revert to built-in names.
+        """
+        from reflex_base.utils.format import _FORMATTED_NAME_CACHE_ATTR
+
+        object.__setattr__(self, "name_resolver", resolver)
+        for cls in self.base_states.values():
```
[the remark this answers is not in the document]
benedikt-bartscher: No production path swaps the resolver after registration. It is installed when `reflex.state` is imported and re-checked in `prerequisites.get_app` immediately before the user module imports, so it is always in place before any user state class exists. The post-registration swaps are the minify CLI (which exits without compiling a frontend afterwards) and unit tests (which swap deliberately). Rebuilding VarData and dependency metadata to support a swap nothing performs would be dead complexity; the requirement is documented on `set_name_resolver` and `install_minify_resolver`.

reflex/compiler/compiler.py
```diff
@@ -203,6 +210,28 @@ def _resolve_default_color_mode(theme: Component | None) -> str:
...
+
+    Returns:
+        The names under the active name resolver.
+    """
+    # ``@event(...)`` types the attribute as an EventCallback; the registry
+    # holds the EventHandler the class rewrote it into.
+    return templates.InternalEventNames(
+        main_state_name=State.get_name(),
+        on_load_internal=format_event_handler(
+            OnLoadInternalState.event_handlers["on_load_internal"]
```
masenf, line 224: why does this one get the value through `.event_handlers` but the others are directly accessing the event handler by attribute

reflex/minify.py
```diff
@@ -0,0 +1,779 @@
...
+
+#: Modules whose ``BaseState`` subclasses can never be minified — their
+#: :class:`Var` hooks are baked at framework-import time before any user
+#: resolver can run. ``reflex.istate.dynamic`` is *not* listed: that's where
+#: ``ComponentState.create()`` puts user-owned dynamic classes.
+_FRAMEWORK_STATE_MODULES: frozenset[str] = frozenset({
+    "reflex.state",
+    "reflex.istate.shared",
+    "reflex.custom_components.custom_components",
+})
```
masenf, line 293: it would be nice to get around this limitation, but i don't think that needs to block us for this initial implementation.

masenf (commented):
the part that gives me hesitation is that the frontend can build with one minification scheme and then run against a backend that's loaded with a potentially different minification scheme. similar to the version mismatch check, it seems that we need to hash the minification scheme and compare with the backend to assert that both the frontend and backend agree on what the names on the wire actually mean.

i think we can drive this one in soon (finally)

----------------

an aside: while there are downsides in implementation complexity and network traffic, i wonder if a more robust implementation _compiles_ the minification map into the frontend, and then the frontend tells the backend what names it's going to be mapping to what real state/event handler names. obvious downsides here are now we're doing per-connection mapping even when in most cases all connections will have the same mapping, and it's still long names on the wire (albeit once, not for every event).

reflex/minify.py
benedikt-bartscher: done

tests/units/test_minification.py
```diff
@@ -0,0 +1,1810 @@
...
+        # config falls back to None — every lookup returns None.
+        assert resolver.config is None
+        assert resolver.resolve_state_name(State) is None
+        assert resolver.resolve_handler_name(State, "any") is None
+
+
+class TestFrameworkStateMinification:
+    """Framework states are ordinary states to the resolver."""
+
+    def test_root_state_name_follows_config(self, temp_minify_json, monkeypatch):
```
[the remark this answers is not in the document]
benedikt-bartscher: The fixture already does this, one level up.

These tests take `temp_minify_json`, which runs the body inside `RegistrationContext.get().fork()` and on teardown does `monkeypatch.undo()` (restoring cwd/env) followed by `clear_config_cache()`. That re-installs the resolver from the restored cwd and, through `set_name_resolver`, clears the per-class `get_name` / `get_full_name` / `get_class_substate` caches and re-keys the registry — i.e. exactly the `finally` block you're describing.

Verified rather than assumed, by appending a probe test immediately after `TestFrameworkStateMinification`:

```python
def test_names_are_back_to_default():
    assert State.get_name() == "reflex___state____state"
```

It passes, and the full unit suite passes under `pytest-randomly` across repeated runs.

packages/reflex-base/src/reflex_base/compiler/templates.py
```diff
@@ -357,30 +393,34 @@ def context_template(
...
-    internal_events.push(ReflexEvent('{state_name}.{constants.CompileVars.ON_LOAD_INTERNAL}'));
+    internal_events.push(ReflexEvent('{internal_events.on_load_internal}'));
 
     return internal_events;
 }}
 
 // The following events are sent when the websocket connects or reconnects.
 export const initialEvents = () => [
-    ReflexEvent('{state_name}.{constants.CompileVars.HYDRATE}'),
+    ReflexEvent('{internal_events.hydrate}'),
```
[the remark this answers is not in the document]
benedikt-bartscher: Valid concern, but it is neither specific to this line nor introduced by it.

**Not new here.** Any id that changes between a deployed bundle and the backend breaks in-flight clients — user state and event ids were already minified before this change. `minify.json` is checked in precisely so the ids stay stable across deploys.

**Enforcement largely exists already.** `state.js` has the #6019 path: when a delta arrives for a substate the frontend has no dispatch function for, it sets `backend_state_mismatch`, logs an actionable message, reports it to the backend over `client_error`, and stops processing further updates.

**What genuinely changed is the failure mode.** `hydrate` used to have a constant name, so a skewed client would hydrate successfully and then hit the mismatch path above on its first delta. Now hydrate itself can miss, and the unknown-handler branch in `EventProcessor._process_queue` only does `logger.exception` — nothing reaches the client — so the page sits unhydrated instead of surfacing the existing error.

That last part is worth addressing, but a minification-scheme handshake is a protocol decision for the feature as a whole rather than part of making the framework states minifiable. The cheap version is routing the unknown-handler branch through `_handle_backend_exception` so skew surfaces client-side. Happy to do either here if @masenf wants it in this PR; otherwise I would rather track it separately.

reflex/compiler/compiler.py
benedikt-bartscher: `@event(supersedes=True)` types that attribute as an `EventCallback` rather than the `EventHandler` the metaclass installs (the other two use `@typing_event`, which types it correctly), so it needed the class-dict lookup — but rather than document that, f6b8211f5a routes all three through a shared `_event_name()` helper so they read identically.

packages/reflex-base/src/reflex_base/environment.py
```diff
@@ -552,6 +552,13 @@ class PathExistsFlag:
 ExistingPath = Annotated[Path, PathExistsFlag]
 
 
+class MinifyMode(enum.Enum):
+    """Mode for minification of state/event IDs."""
+
+    ENABLED = "enabled"
+    DISABLED = "disabled"
```
masenf: is there expected to be other minify modes? seems like REFLEX_MINIFY_* could be bool, which would make writing `REFLEX_MINIFY_STATES=true` or `REFLEX_MINIFY_STATES=1` or `REFLEX_MINIFY_EVENTS=0` work as users expect other bool env vars to work

masenf (commented):
performing production readiness review via Fable 5.1

packages/reflex-base/src/reflex_base/environment.py
benedikt-bartscher: i initially wanted to add more modes, f.e. one where users can explicitly set id's for states and events in python - however i decided against that because the minify cli provides much better DX imo

will migrate to bool

reflex/reflex.py
```diff
@@ -1050,6 +1052,444 @@ def rename(new_name: str):
...
+    path_to_id: dict[str, str | None] = {}
+    for s in collect_all_states(State):
+        path = get_state_full_path(s)
+        entry = config["states"].get(path)
+        path_to_id[path] = entry["id"] if entry is not None else None
+
+    parts = minified_path.split(".")
+    # Accept a path copied verbatim from the frontend, which leads with the
+    # root state's own name; the lookup below walks from the root's children.
+    if parts[0] == State.get_name():
```
masenf: **Bug: `lookup` result depends on the caller's shell env vars.**

`State.get_name()` only returns the root's minified id when `REFLEX_MINIFY_STATES=enabled` is set in the shell running the CLI. Without it the root is `reflex___state____state`, so a path copied from the browser is never stripped and resolves against the wrong level.

Reproduced on a sample app where `reflex.state.State` is `a` and `AppState` is `a`:

```
$ reflex minify lookup a.a.l
No state or event handler found for minified segment 'l' in path 'a.a.l'
$ REFLEX_MINIFY_STATES=enabled reflex minify lookup a.a.l
minidemo.minidemo.AppState
minidemo.minidemo.AppState.setvar
```

Suggest stripping the first segment when it matches either the root's configured id (`config["states"].get(get_state_full_path(State))`) or `RegistrationContext.default_state_name(State)`, independent of the env vars. The same env-independence would be good for `list` and `validate` (see the other comment on `_open_minify_session`).

---

reflex/minify.py
```diff
@@ -0,0 +1,810 @@
...
+
+    Args:
+        root_state: Optional subtree root.
+
+    Returns:
+        A complete :class:`MinifyConfig`.
+    """
+    states: dict[str, StateEntry] = {}
+    events: dict[str, dict[str, str]] = {}
+    sibling_counter: dict[type[BaseState] | None, int] = {}
```
masenf, line 639: **Root and first-level ids collide, which makes `lookup` inherently ambiguous.**

Ids restart at `a` per sibling group, and the root `State` is alone in its group, so `reflex.state.State` gets `a` and the first user state under it also gets `a`. With `init` on my sample app the wire name for `DeepDeep.go` was `a.a.a.a.a.a`, and `lookup a.a` cannot tell "root → AppState" from "AppState → SubState" (the CLI currently picks the former only when the env var happens to be set).

Options, roughly in order of preference:
1. Do not minify the root `State` at all: it is a single fixed segment, and the handshake digest already protects against mismatch.
2. Reserve the root's id from the first-level pool so the two never coincide.
3. Always require the root segment in `lookup` and drop the "root is optional" behaviour from the docstring.

---

reflex/reflex.py
```diff
@@ -1050,6 +1052,444 @@ def rename(new_name: str):
...
+        )
+        raise SystemExit(1)
+
+    # Loading the app runs arbitrary module-level code, which may write to
+    # stdout directly; reserving only covers Reflex's own logging, so send
+    # everything the import prints to stderr rather than into the document.
+    with (
+        contextlib.redirect_stdout(sys.stderr) if for_json else contextlib.nullcontext()
+    ):
+        _load_app_for_minify()
```
masenf, line 1146: **`validate` crashes on the exact misconfiguration it is meant to catch, and the runtime error never mentions `minify.json`.**

With two siblings both assigned `"a"` and `REFLEX_MINIFY_STATES=enabled`, loading the app here installs the resolver first, so `__init_subclass__` raises before `validate_minify_config` ever runs:

```
reflex_base.utils.exceptions.StateValueError: The substate class 'a' has been
defined multiple times. Shadowing substate classes is not allowed.
```

The same traceback is what a user sees from `reflex run`, with nothing pointing at the config file. Without the env var, `validate` correctly reports `Duplicate state_id='a' under '...AppState'`.

Two suggestions:
- The inspection commands (`validate`, `list`, `lookup`, `sync`) should load the app with the resolver forced to `DefaultNameResolver`, since they read the config rather than apply it. That also removes the env-dependence noted on `lookup`.
- When a non-default resolver is active and the sibling-name check in `BaseState.__init_subclass__` fires, the message should say the name came from `minify.json` and point at `reflex minify validate`.

---

packages/reflex-base/src/reflex_base/.templates/web/utils/state.js
```diff
@@ -721,6 +743,17 @@ export const connect = async (
...
   };
 
+  // The backend resolves wire names with its own copy of the minification
+  // scheme. If it disagrees with the one this bundle was built against, every
+  // name we send is meaningless to it, so stop before the first event.
+  socket.current.on(SCHEME_MISMATCH_EVENT, (detail) => {
+    backend_state_mismatch = true;
+    event_queue.length = 0;
+    console.error(
+      `Cannot talk to the backend: it resolves state and event names with a different minification scheme (frontend "${detail?.frontend ?? ""}", backend "${detail?.backend ?? ""}"). Try refreshing the page or clearing your browser cache. If you are the developer of this app, rebuild the frontend against the same minify.json the backend is running.`,
```
masenf: **A scheme mismatch is invisible to end users.**

I served a minified prod bundle against a backend started without the env vars. The socket connects, `scheme_mismatch` arrives, and the page then does nothing: no hydration, no toast, no connection-error UI. The only signal is this `console.error`, and its "try refreshing the page" advice is aimed at someone who will never open the console.

The most likely real-world trigger is a cached bundle after a deploy that changed ids. Suggest either surfacing this through the existing connection-error UI (`connectErrors` / `rx.connection_toaster`), or performing a single guarded `location.reload()` since a fresh bundle usually fixes it. Silent-dead-page is worse than the connection error the user would get if the socket had failed outright.

---

reflex/compiler/compiler.py
```diff
@@ -776,6 +845,8 @@ def compile_contexts(
     Returns:
         The path and code of the compiled context.
     """
+    _assert_state_names_are_bound()
```
masenf, line 851: **Suggestion: warn at compile time when minification is enabled but the config is stale.**

I added a new substate while `reflex run` was up with both modes enabled. It compiled and dispatched fine with the mixed name `a.a.minidemo___minidemo____new_state.ping`, but nothing told me the config was behind until I ran `validate` by hand. Since the registry is walked here anyway, a one-line warning listing states/handlers that have no entry (the same check `validate_minify_config` does for `missing`) would make the workflow self-correcting.

---

reflex/app.py
```diff
@@ -2083,9 +2087,32 @@ async def on_connect(self, sid: str, environ: dict):
...
+            await self.emit(
+                str(constants.SocketEvent.SCHEME_MISMATCH),
+                {"frontend": client_scheme, "backend": server_scheme},
+                to=sid,
+            )
+            # The client queues its initial events as soon as it sees CONNECT,
+            # and only learns of the mismatch a tick later, so drop whatever it
+            # sends meanwhile: a name from the other scheme could resolve to a
+            # real -- but wrong -- handler here.
+            self._scheme_mismatch_sids.add(sid)
```
masenf, line 2116: **The upload endpoint bypasses this guard.**

`on_event` drops events from a mismatched sid, but `_upload` is a plain HTTP route that resolves its handler by name from the form data. A stale bundle whose scheme differs can still post an upload whose handler name resolves to a real but wrong handler on this backend, which is exactly the hazard described in the comment above. Consider carrying the scheme digest on the upload request too (header or form field) and rejecting it when it differs from `scheme_digest()`.

---

reflex/minify.py
```diff
@@ -0,0 +1,810 @@
...
+    from reflex_base.registry import RegistrationContext
+
+    ctx = RegistrationContext.try_get()
+    resolver = ctx.name_resolver if ctx is not None else None
+    if not isinstance(resolver, MinifyNameResolver):
+        return ""
+    return resolver.digest()
+
+
+def clear_config_cache() -> None:
```
masenf, line 485: **Docstring over-promises.** It says to call this after editing `minify.json` or changing the env vars at runtime, but once any state class exists its `VarData` keeps the old names, and the next `compile_contexts` raises via `_assert_state_names_are_bound`. Either document that this is only safe before user states are imported (tests, fresh processes), or make it rebuild what the assertion checks.

---

reflex/state.py
```diff
@@ -2718,20 +2754,43 @@ def get_component(cls, *children, **props) -> Component:
...
+            # identifier string is rejected here rather than further in.
+            if not isinstance(_state_key, str) or not _state_key.isidentifier():
+                msg = (
+                    "_state_key must be a string that is a valid Python identifier, "
+                    f"got {_state_key!r}."
+                )
+                raise ValueError(msg)
+            # Doubled separator: an unkeyed name is always ``_n`` followed by
+            # digits, so no key can produce one.
+            state_cls_name = f"{cls.__name__}__{_state_key}"
```
masenf, line 2828: **`_state_key` semantics and error message.**

Two things I hit while trying this:

1. Reusing a key raises the generic sibling-shadowing error with the long default name: `The substate class 'reflex___istate___dynamic____counter__cart' has been defined multiple times.` A check here that says "`_state_key='cart'` is already used by another `Counter` instance" would be much clearer.
2. Worth deciding explicitly whether a repeated key should error or return the existing state class. A "cart" counter rendered in a shared layout on several pages is a natural use, and users will likely expect one key to mean one shared state. If erroring is intended, the news fragment and docs should say so.

Also: this is new public API on `ComponentState.create`, and there is no page in `docs/` for it yet.

---

packages/reflex-base/src/reflex_base/event/__init__.py
```diff
@@ -3166,6 +3219,7 @@ def BaseState(self) -> "type[BaseState]":  # noqa: N802
 
 event = EventNamespace
 event.event = event  # pyright: ignore[reportAttributeAccessIssue]
+event.typing_event = staticmethod(typing_event)  # pyright: ignore[reportAttributeAccessIssue]
```
masenf: This exposes `rx.event.typing_event` as public API. It is a pyright-only no-op used twice in `reflex/state.py`. Suggest keeping it private (`_typing_event`) and not attaching it to the `event` namespace, so it does not become something to document and support long-term.

---

packages/reflex-base/src/reflex_base/compiler/templates.py
```diff
@@ -345,7 +380,11 @@ def context_template(
         rf"""
 export const state_name = "{state_name}"
 
-export const exception_state_name = "{constants.CompileVars.FRONTEND_EXCEPTION_STATE_FULL}"
+export const main_state_name = "{internal_events.main_state_name}"
```
masenf, line 372: Nit: `state_name` and `main_state_name` are always the same value here (`_compile_contexts` passes `state.get_name()` for the app's root `State`, and `InternalEventNames.main_state_name` is `State.get_name()`). If `state_name` is kept for compatibility, a comment saying so would help; otherwise one export is enough.

---

tests/units/test_minification.py
```diff
@@ -0,0 +1,1967 @@
...
+    Returns:
+        A ``CliRunner`` ready to invoke ``reflex.reflex.cli`` commands.
+    """
+    from reflex.utils import prerequisites
+
+    monkeypatch.setattr(prerequisites, "get_compiled_app", lambda *a, **kw: mock.Mock())
+    return CliRunner()
+
+
+class TestIntToMinifiedName:
```
masenf: The repo guidelines (CLAUDE.md, Testing) ask for module-level test functions rather than test classes; every test in this file is wrapped in a class. The naming rule there also says unit test files mirror the source module, so this should be `tests/units/test_minify.py` for `reflex/minify.py`.

---

news/+minify-json-stdout.bugfix.md
```diff
@@ -0,0 +1 @@
+`reflex minify list --json` and `reflex minify lookup --json` write only the JSON document to stdout; warnings and anything the app prints while it loads go to stderr, so the output parses.
```
masenf: This fragment, `+minify-scheme-handshake.bugfix.md`, and `packages/reflex-components-core/news/6100.bugfix.md` describe behaviour of a feature that has never shipped, so they read as bugfixes for something users never had. Suggest folding them into `6100.feature.md`. `packages/reflex-base/news/+event-handler-surfaced.bugfix.md` is a genuine behaviour change (unhandled events now reach `backend_exception_handler`) but is unrelated to minification, so it is worth calling out in the PR description or splitting.

---

.gitignore
```diff
@@ -27,6 +27,7 @@ node_modules
 package-lock.json
 *.pyi
 .pre-commit-config.yaml
+uploaded_files/*
```
masenf: Unrelated to this PR; please drop it or land it separately.

---

masenf (commented):
I built a sample app (4-level state tree, sibling substates, a handler literally named `a`, computed vars, `setvar`, `rx.LocalStorage`, form submit, background handler, `call_script` callback, `get_state` across parent/sibling, backend and frontend exceptions, dynamic route with `on_load`, two unkeyed plus one `_state_key` `ComponentState`), ran `reflex minify init`, and drove it with Playwright in dev mode and in a production build with both env vars enabled.

**Works:** every interaction dispatched correctly through fully minified names (deepest handler was `a.a.a.a.a.a`, exercising the `_skip_self` fix); client-storage keys, `hydrate`, `on_load_internal`, `update_vars_internal` and `handle_frontend_exception` all resolved; the handshake accepts a matching digest and emits `scheme_mismatch` and drops events for a wrong or empty one; a minified prod bundle against an unminified backend stops before sending any event; hot-adding an unsynced state produced a mixed name that still dispatched, and `sync` then assigned it the next free id. Unit tests pass, ruff is clean.

**Inline comments cover:** two `lookup` correctness problems (env-dependent root stripping; root/first-level id collision), `validate` crashing on duplicate ids with a message that never mentions `minify.json`, the silent dead page on scheme mismatch, the upload endpoint bypassing the mismatch guard, a stale-config compile warning, `clear_config_cache` docs, `_state_key` semantics, `typing_event` public exposure, a duplicated context export, test layout vs. repo guidelines, news fragments, and an unrelated `.gitignore` line.

**Not anchored to a line:** there is no documentation under `docs/` for the `reflex minify` CLI, the two `REFLEX_MINIFY_*` env vars, the `minify.json` format, the requirement that a deployed frontend and backend share the same file and modes, or `_state_key`.

Not verified: the Redis state manager path (no Redis here). The granian worker crash with `state_auto_setters=True` and `AppState.setvar(...)` that I hit reproduces on `main` too and is not this PR's.

---

reflex/reflex.py
benedikt-bartscher: Fixed. The root-prefix strip now compares against the root's *configured* id (from `minify.json`) as well as `RegistrationContext.default_state_name(State)`, instead of `State.get_name()` — so both spellings are accepted and the result no longer depends on the CLI's environment. Verified `lookup a.b.a.a` returns identical output with and without `REFLEX_MINIFY_*` set.

It was also bidirectional: `lookup a.a.a` silently returned *different* answers with and without the env var, rather than just failing.

reflex/minify.py
benedikt-bartscher: Took option 2, generalised. The collision isn't only root↔first-level: ids restart per sibling group, so **every** state shared an id with its first child.

That turned out to be a silent runtime bug, not just a `lookup` ambiguity — it breaks the `_skip_self` heuristic, so `_always_dirty_substates` (which stores relative single-segment names) resolves to the parent, and a `@rx.var(cache=False)` on the alphabetically-first substate stops being recomputed:

```
minified:  State._get_potentially_dirty_states() = ['State']      # wrong
baseline:  State._get_potentially_dirty_states() = ['AppState']   # correct
```

So ids now skip the parent's own id at every depth. Costs zero bytes (one integer burned per group; ids only lengthen past 54 siblings), and it makes the "root is optional" docstring true rather than aspirational. `validate` gained an error for a child reusing its parent's id.

FWIW option 1 measured cheaper than expected but wouldn't have fixed the above: the root name occurs 115× in the compiled output of a small app (~2.5 kB raw), but only ~100 B gzipped — LZ77 eats the repeated literal.

Deliberately **not** bumping `SCHEMA_VERSION`, so anyone who already ran `init` on this branch should delete `minify.json` and re-run it.

reflex/reflex.py
benedikt-bartscher: Both done.

One wrinkle on the first suggestion: forcing the default resolver inside `_open_minify_session` is too late — `list` and `lookup` do `from reflex.state import …` in their function bodies *before* calling it, and by then the framework states have baked their names into `VarData`, so switching resolvers afterwards trips `_assert_state_names_are_bound`. It now happens in the `minify` group callback, scoped to the invocation via `ctx.with_resource()`.

Second suggestion done too: when a non-default resolver is active the message now names `minify.json` and points at `reflex minify validate`.

packages/reflex-base/src/reflex_base/.templates/web/utils/state.js
benedikt-bartscher: Surfaced through `connectErrors`, and shared with the pre-existing `missing_substates` path (`state.js:772`) — that one was silently dead in exactly the same way, so both go through one `fatalMismatch` helper now.

Went with the connection UI rather than `location.reload()`: `test_unprocessable_delta_is_fatal_without_reload` already pins "the frontend must not reload on its own" for the sibling path.

Two things that made the naive version not actually work: the toaster gates on `connectErrors.length >= 2`, and a fatal error never retries so there's never a second one — it needs a debounce bypass; and the toast's Reload action has to be raw JS, since a `ToastAction.on_click` would be routed through `processEvent` and dropped by the very loop that's dead.

reflex/compiler/compiler.py
benedikt-bartscher: Added at that call site.

Worth flagging that the handshake doesn't cover this case: `_compute_digest` hashes the *config*, not the state tree, so a stale config produces matching digests on both sides and nothing notices.

Measured ~344 µs at 305 states — cheaper than `_assert_state_names_are_bound` already on the line above, and ~0.1% of a compile. It reads the config off the installed resolver (no disk re-read) and is gated per mode, so the `reflex minify` commands' own dry-run compiles never trip it. `logger.warning`, not an error: a missing entry falls back to the full name, and `reflex minify validate` already exits 1 for CI.

reflex/app.py
benedikt-bartscher: Confirmed and fixed. It reproduces: with two upload handlers whose ids swap between builds, a scheme-A bundle's POST resolves to `BetaState.upload_beta` on a scheme-B backend and gets dispatched with the files — no error, no log line.

Uploads now carry the digest in a `Reflex-Scheme` header, checked in `_require_upload_headers` before the body is read (409 on mismatch). Header absent → `""`, which matches a non-minified backend, so nothing else changes.

One layering note: `reflex-components-core` doesn't depend on `reflex`, so `digest()` moved onto the `NameResolver` protocol and `scheme_digest()` moved to `reflex_base.registry` — both guard sites now call the same helper, and it let `scheme_digest()` drop its `isinstance`.

reflex/minify.py
benedikt-bartscher: Documented the constraint. It's a bit worse than "once any state class exists": `reflex.state.State` itself ends up unbound, so with `REFLEX_MINIFY_STATES` it's broken in *any* process that imported `reflex.state`. The two halves are asymmetric — handler names do re-propagate, state names don't — so the docstring says that.

Didn't make it rebuild: Vars are frozen with no back-reference, `ComputedVar` `VarData` can merge several states, and Vars already embedded in component trees aren't reachable from the registry at all — so it'd be either silently incomplete or a full component-graph walk. It also has no production callers (test-only), which argues for documenting rather than enforcing.

reflex/state.py
benedikt-bartscher: Message fixed — it now names the key and the module-qualified owner.

It was worse than what you saw, though: via the compiler path `_modify_exception` scans to the first `.` (the sentence-ending period) and title-cases the rest, so in a real app the message was `The substate class '.Cart' Has Been Defined Multiple Times.` — neither `_state_key`, nor the component, nor the key survived. That `_modify_exception` fragility is separate and not fixed here.

On point 2, kept it erroring. You're right that two pages mean two `create()` calls, but the one-line hoist works and is already the idiom the docs teach. Sharing has a demonstrated silent failure: `get_component` is documented to mutate the class from props (the docs' own `EditableText` does `cls.__fields__["text"].default = …`), so two keyed call sites with different `initial_value`s would give both pages the last one, decided by page-registration order. Error→share is a backwards-compatible loosening later; the reverse isn't. Documented under *Naming the State* on the ComponentState page.

Also: a repeated key isn't strictly per-component — two `ComponentState` subclasses sharing a `__name__` collide even unkeyed — so the message reports the owner's module.

packages/reflex-base/src/reflex_base/event/__init__.py
benedikt-bartscher: Agreed — but un-attaching it alone breaks the runtime import: `sys.modules[__name__] = event` replaces the module with the `EventNamespace` class, so a module-level name is only importable if it's explicitly attached (documented in `tests/units/test_event.py:357`). Dropping line 3222 would have given `ImportError` at `reflex/state.py:40`, invisible to pyright.

So it moved into `reflex/state.py` as `_typing_event`, next to its only two call sites. It's gone from `rx.event` entirely; added a test asserting that.

packages/reflex-base/src/reflex_base/compiler/templates.py
benedikt-bartscher: Removed `state_name`. It's a bit more than a nit: it's either exactly `main_state_name`, or — when a substate is passed as the app state — `get_name()`, a bare segment that matches no `initialState`/delta key at all. So it has no valid distinct meaning.

Its one consumer (`state.js:792`, the `is_hydrated` gate) wanted the framework root anyway, since that's where `is_hydrated` lives, so it now reads `main_state_name`. The parameter and the now-unreachable `ValueError` went with it.

tests/units/test_minification.py
benedikt-bartscher: Done — and it needed a split rather than a rename, since the file covered four source modules (there's no `reflex/minification.py` for it to mirror):

| destination | tests |
|---|---|
| `tests/units/test_minify.py` | 55 |
| `tests/units/test_state.py` | 13 |
| `tests/units/test_reflex.py` | 13 |
| `tests/units/reflex_base/test_registry.py` | 9 |
| `tests/units/components/base/test_bare.py` | 1 |

All module-level now; the classes carried no shared state, so un-nesting was mechanical. Shared helpers live in `tests/units/minify_helpers.py` and `temp_minify_json` moved to `tests/units/conftest.py`. Same test count before and after.

news/+minify-json-stdout.bugfix.md
benedikt-bartscher: Done. `+minify-json-stdout` dropped and `+minify-scheme-handshake` folded into `6100.feature.md` — both described behaviour of commands/handshake added in this same PR.

`packages/reflex-components-core/news/6100.bugfix.md` stayed a bugfix but got reworded: the change also narrows a *shipped* false positive (`value.startswith("reflex___state")` → `FIELD_MARKER`-anchored), so it isn't purely never-shipped.

`+event-handler-surfaced` kept in this PR and renamed. `git log -S` puts it in the scheme-handshake commit, but the code and its two regression tests have no minification coupling — called out in the PR description; happy to split it if you'd rather.

Separately: renamed every `+<slug>` fragment to `6100.*`. towncrier treats `+` fragments as orphans and blanks the issue, so they'd have shipped with no PR link.

.gitignore
benedikt-bartscher: Dropped. Worth landing on its own, though — this PR's first commit accidentally committed `uploaded_files/test.txt` (removed again in `1c5114d5ee`), and nothing on main ignores it, so `rx.get_upload_dir()` litters the repo root whenever an app is run from the checkout.

packages/reflex-base/src/reflex_base/environment.py
benedikt-bartscher: done

packages/reflex-base/src/reflex_base/registry.py
```diff
@@ -1,23 +1,115 @@
...
+
+    A frontend bundle and the backend it talks to must agree on what the names
+    on the wire mean. Memoized on the resolver instance, so installing another
+    resolver produces a fresh digest without coordination.
+
+    Returns:
+        A short hex digest, or ``""`` when no name is rewritten.
+    """
+    ctx = RegistrationContext.try_get()
+    return ctx.name_resolver.digest() if ctx is not None else ""
```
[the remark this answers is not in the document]
benedikt-bartscher: `NameResolver` is introduced by this PR — it doesn't exist anywhere on `upstream/main` (no `set_name_resolver`, no `name_resolver` field, no `reflex/minify.py`), and the latest release is `v0.9.11a2`. There are no existing custom resolvers to stay compatible with.

More importantly, the suggested fallback would defeat the thing the digest exists for. A custom resolver that *does* rewrite names would report `""`; two *different* such resolvers would both report `""`, match, and let events resolve to real-but-wrong handlers — exactly the mismatch this handshake was added to catch. A loud failure is safer than a silent empty digest.

It's also loud and early: `scheme_digest()` is called from `_compile_contexts`, so a resolver missing `digest()` fails at compile time in development, not at runtime on the upload path. Implementing the contract is one line (`DefaultNameResolver.digest()` returns `""`), pyright flags the incomplete Protocol statically, and `isinstance(x, NameResolver)` returns `False` since it's `@runtime_checkable`.

Leaving as-is.

reflex/minify.py
```diff
@@ -0,0 +1,921 @@
...
+
+    path_to_parent_key = {get_state_full_path(s): get_parent_key(s) for s in all_states}
+
+    # Reserve every configured id within its sibling group — orphans under
+    # their recorded parent, so their ids are never reused. Live entries get
+    # their stored parent healed to the actual value.
+    parent_key_to_existing_ids: dict[str | None, set[int]] = {}
+    for state_path, entry in new_states.items():
+        parent_key = path_to_parent_key.get(state_path, entry["parent"])
+        entry["parent"] = parent_key
```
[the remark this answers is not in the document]
benedikt-bartscher: Confirmed — `sync` was writing a config its own `validate` rejects:

```
   a   MoveParent             parent=None
   a   MoveParent.MovedChild  parent=MoveParent   <- healed parent, preserved id
```

There's a second direction with the same root cause: a parent *assigned in this run* can land on a preserved child's id. Insert a new intermediate state above an existing one and the new parent draws from its own sibling pool while the child keeps its id, so re-checking only after the parent-healing step would close one direction and leave the other.

Fixed with a single pass after assignment (`_rehome_ids_colliding_with_parent`) that walks the tree depth-first, so a parent is final before its children are checked — no loop needed. The child always moves, never the parent, so the result doesn't depend on which collided first. Moving a preserved id is normally what `sync` avoids, but in both cases the state's wire path already changed (re-parented, or a state inserted above it), so the id couldn't have meant the same thing to an existing client anyway.

Regression test per direction, plus a fuzz over 1500 random configs (id churn, re-parenting to random/stale/None parents, dropped entries, both flag combinations) asserting `validate` is clean after every `sync`.

packages/reflex-base/src/reflex_base/.templates/web/utils/state.js
```diff
@@ -721,6 +748,28 @@ export const connect = async (
     });
   };
 
+  // A mismatch is terminal: no event will ever be answered again. Push it onto
+  // connectErrors as well, or the page sits dead with the reason only in the
+  // console, where no end user will look.
+  const fatalMismatch = (developerMessage) => {
```
masenf, line 760: **Retested at ff94c48: the fatal toast never appears for a scheme mismatch in a real app.**

Setup: `reflex export --frontend-only` with both `REFLEX_MINIFY_*=1`, served the static build, ran `reflex run --env prod --backend-only` without the env vars. Six page loads, headless Chromium, waited up to 15 s each: the `console.error` from this helper is printed every time, `connectErrors`-driven UI never shows. Same bundle does show "Cannot connect to server: websocket error" when the backend is down, so the toaster itself works.

Cause is ordering. The server emits `scheme_mismatch` from inside `on_connect`, so the packet reaches the client **before** the namespace CONNECT ack (I captured the frames: `42/_event,["scheme_mismatch",…]` then `40/_event,{"sid":…}`). socket.io-client buffers events that arrive before CONNECT and flushes them in `onconnect()` via `emitBuffered()` *before* `emitReserved("connect")`. So this handler pushes the fatal error, and then the `connect` handler at line 699 runs `setConnectErrors([])` and wipes it.

Suggested fix: make the `connect` handler keep a fatal entry (e.g. `setConnectErrors((errs) => errs.at(-1)?.fatal ? errs : [])`), or defer the fatal push to after connect. Patching the built bundle that way made the toast with the Reload action appear. `test_unprocessable_delta_is_fatal_without_reload` covers the delta path only; a Playwright test for the scheme-mismatch toast would have caught this.

---

reflex/reflex.py
```diff
@@ -1050,6 +1052,464 @@ def rename(new_name: str):
...
+        the file is absent.
+    """
+    from reflex.minify import (
+        MINIFY_JSON,
+        _get_minify_json_path,
+        _load_minify_config_uncached,
+    )
+
+    if for_json:
+        log.reserve_stdout()
```
masenf, line 1127: **Test pollution: the stdout reservation leaks out of the `--json` CLI tests.**

`reserve_stdout()` sets a module global that nothing in this code path clears (a real CLI process exits, so that is fine at runtime). Under `CliRunner` it survives the test, so any later test that captures the log sink's stdout sees empty output:

```
uv run pytest tests/units/test_reflex.py::test_lookup_accepts_full_wire_path \
  tests/units/test_event.py::test_arg_mismatch_warning_renders_brackets_verbatim
# 1 failed: assert 'expects (dict[str, typing.Any]) ...' in ''
```

Either pair the reservation with a reset when the command finishes (`ctx.call_on_close(log._reset)` style, or a `try/finally` in the `--json` commands), or add an autouse fixture in `tests/units/test_reflex.py` that restores `log._stdout_reserved`. Only one test there (line ~743) does so today.

---

masenf (commented):
Retested at ff94c48 against the same sample app (4-level state tree, sibling substates, a handler named `a`, computed vars, `setvar`, `rx.LocalStorage`, form submit, background handler, `call_script` callback, `get_state` across parent and sibling, backend and frontend exceptions, dynamic route with `on_load`, two unkeyed plus one keyed `ComponentState`), in dev mode and in an exported production build.

**Verified fixed (10 of 11 threads):**

| Finding | Evidence |
|---|---|
| `lookup` env dependence | `a.b.a.b.a.a`, `a.b.l` and `b.a.a` return byte-identical output with and without `REFLEX_MINIFY_*` set |
| Parent/child id collision | `init` now assigns `AppState`=b under root=a, `SubState`=a under b, `Deep`=b under a; no state shares its parent's id at any depth |
| `validate` crashing on duplicate ids | With both env vars set it now reports `Duplicate state_id='a' under 'minidemo.minidemo.State.AppState'` and exits 1 |
| Duplicate-id runtime message | Names the state path, the resolved id, the sibling scope, `minify.json` and `reflex minify validate` |
| Stale-config warning | Hot-reloading in a new substate logged the missing `state:` and `event:` entries plus the `sync` hint |
| Upload guard | `Reflex-Scheme: wrong` gets 409 with both digests named; matching header dispatches normally |
| `clear_config_cache` docs | Asymmetry between handler and state names is documented |
| `_state_key` duplicate | `_state_key='cart' is already used by a minidemo.counter.Counter instance…` |
| `typing_event` | Gone from `rx.event`, now `_typing_event` in `reflex/state.py` |
| `state_name` / news / `.gitignore` / test layout | All as described; `test_minify.py` split is module-level, 55 tests |

The parent-id skip is a better fix than what I suggested: my comment only described the root/first-level case, and it does apply at every depth.

Functional behaviour is unchanged. All 24 browser-driven checks pass in dev with minification on, including the deepest handler at `a.b.a.b.a.a`, background events, `call_script` callbacks, client storage rehydration and cross-state `get_state`. Ruff and format clean.

**Two new issues, both inline:**

1. The fatal mismatch toast does not appear in a real app. The `scheme_mismatch` packet arrives before the socket.io CONNECT ack, so it is buffered and flushed by `emitBuffered()` ahead of `emitReserved("connect")`; the `connect` handler's `setConnectErrors([])` then wipes the fatal entry. Six loads, toast never rendered, only the console error. Patching the built bundle to preserve a fatal entry makes the toast and its Reload action appear, which confirms the cause.
2. `reserve_stdout()` leaks across tests. `tests/units/test_reflex.py` + `tests/units/test_event.py` fails `test_arg_mismatch_warning_renders_brackets_verbatim`; the same pair passes on `main`, and the file's own line 743 already works around it for one test.

Everything else I could check is clean. Not retested: Redis state manager (none available here).

---

packages/reflex-base/src/reflex_base/.templates/web/utils/state.js
benedikt-bartscher: Confirmed and fixed in 9d31345. You're right, and my earlier "ordering, verified safe" note on this thread was wrong — I'd asserted the opposite ordering without checking it.

Verified your mechanism against the vendored client rather than by observation:

```js
onconnect(t,s){ …; this.connected=!0, this.emitBuffered(), this.ft(!0), this.emitReserved("connect") }
```

`emitBuffered()` really does run before `emitReserved("connect")`, so the buffered `scheme_mismatch` is replayed first and the `connect` handler then wipes the entry. The `missing_substates` path was unaffected only because it fires long after connect, which is why the existing test stayed green.

Took the first option: the `connect` handler now keeps a fatal entry rather than clearing unconditionally. That also reads correctly beyond this bug — reconnecting cannot change what the names mean, so a fatal mismatch should outlive a reconnect.

Added the Playwright test you asked for, extending the existing `test_client_error.py` app instead of adding one. It patches `reflex.app.scheme_digest` after the harness compiles, so the bundle keeps the real digest and only the backend disagrees. Confirmed it is a real regression test: with the fix it passes, with the fix reverted it fails on a 30 s timeout waiting for the toast.

reflex/reflex.py
benedikt-bartscher: Reproduced your exact command (1 failed, `assert '…' in ''`) and fixed in 9d31345.

Went with the first option rather than a test fixture: `_open_minify_session` now pairs the reservation with `ctx.call_on_close(lambda: log.reserve_stdout(False))`, matching the `ctx.with_resource(force_default_names())` already in the `minify` group callback. That also makes it correct for any programmatic CLI invocation, not just under `CliRunner` — an autouse fixture would have hidden the leak rather than removed it.

Added a parametrized regression test asserting `log.is_stdout_reserved()` is `False` after `minify list --json` and `minify lookup --json`.

reflex/state.py
```diff
@@ -1177,19 +1228,31 @@ def get_substates(cls) -> set[type[BaseState]]:
         """
         return RegistrationContext.get().get_substates(cls)
 
+    # Unbounded on purpose: the resolved name depends on the active resolver,
+    # which lives in a ContextVar, so an evicted entry can be recomputed from a
+    # scope that has no context and pin the default name process-wide. One
+    # entry per state class is already bounded by the registry.
     @classmethod
-    @functools.lru_cache
+    @functools.cache
```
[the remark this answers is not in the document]
benedikt-bartscher: Investigated; the retention is real but not caused by this cache, so I'm not taking the suggested change.

`reload_state_module` drops the class from `all_base_state_classes` and from the substates set, but never removes it from `ctx.base_states`, which keeps a strong reference. Measured on a reloaded class:

```
in base_states?                 : True
in base_state_substates values? : False
alive after gc                  : True
```

A/B against the pre-PR bounded cache, five same-name reload cycles each:

```
bounded (lru_cache, maxsize=128) -> classes still alive: 5 / 5
unbounded (functools.cache)      -> classes still alive: 5 / 5
```

Identical, so the cache is not what retains them and the growth predates this PR. A weak-key cache would free nothing while `base_states` holds the reference, and it would reintroduce the eviction hazard this cache is unbounded to avoid: `get_name` resolves through a ContextVar-scoped resolver, so an evicted entry recomputed from a scope with no context pins the default name process-wide, in an app with more than 128 state classes. That is a data-correctness bug, reproduced, and covered by `test_resolved_name_survives_cache_pressure`.

The genuine issue here is that `reload_state_module` leaks into `base_states`; it is pre-existing and worth its own fix rather than one in this PR.

The test-side half of this review (the filler classes leaking into `all_base_state_classes` and the caches) was valid and is fixed in 282d430 — they are dropped in a `finally`, verified with a session-scoped probe: 0 leaked names, cache back to the session's real states.

reflex/minify.py
```diff
@@ -0,0 +1,965 @@
...
+    for parent_key, children in parent_key_to_new_children.items():
+        existing_ids = parent_key_to_existing_ids.get(parent_key, set())
+        parent_entry = new_states.get(parent_key) if parent_key is not None else None
+        if parent_entry is not None:
+            existing_ids = existing_ids | {minified_name_to_int(parent_entry["id"])}
+        assigned = _assign_next_ids(children, existing_ids, reassign_deleted)
+        for state_path, minified_name in assigned.items():
+            new_states[state_path] = StateEntry(id=minified_name, parent=parent_key)
+
+    _rehome_ids_colliding_with_parent(new_states, all_states, reassign_deleted)
```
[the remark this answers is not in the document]
benedikt-bartscher: Fixed in e1ba2bb, and the live case turned out to be sharper than orphans: `sync` can create the duplicate itself, with no orphan involved.

```
config: Moved(id=c, parent=Root)    # code says its parent is Q
        Sibling(id=c, parent=Q)

sync heals Moved.parent -> Q and keeps id "c"
  c  Moved    parent=Q
  c  Sibling  parent=Q
validate: Duplicate state_id='c' under '...Root.Q'
```

Same mechanism as the parent-id collision — heal the parent, keep the id — just colliding sideways, which the old pass walked straight past.

`_rehome_ids_colliding_with_parent` is now `_rehome_conflicting_ids` and compares against the whole sibling scope, retained orphans included. On who gives way: an entry this sync re-parented yields, since its scope changed and its id could not have meant the same thing to a served client; a child sharing its parent's id yields, because moving the parent would cascade; an entry sync did not touch keeps its id. A duplicate that was already in the file is left for `validate` to report — silently renumbering what the user wrote would break the id stability that is the point of `sync`.

In the case above: `Moved` -> `d`, `Sibling` keeps `c`.

Regression test per direction. Worth noting the fuzz I had been leaning on does not generate this pattern — it passed against the reverted fix — so the deterministic tests are what actually cover it.

reflex/minify.py
```diff
@@ -0,0 +1,959 @@
...
+    ambiguous in a relative path. The child moves rather than the parent, so
+    the result does not depend on which collided first. Walks ancestors first,
+    so a parent is already final when its children are checked.
+
+    Args:
+        states: The state entries to fix up, modified in place.
+        all_states: State classes in depth-first order.
+        reassign_deleted: Whether a replacement id may fill a gap.
+    """
+    for state_cls in all_states:
```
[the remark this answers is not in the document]
benedikt-bartscher: Belated answer on this one — it was acted on but never replied to.

Reproduced: with default flags `sync` retained an orphan holding `id: "a"` under a live parent also holding `"a"`, and `validate` then rejected the file `sync` had just written.

I did not take the suggested repair, though, because moving the orphan would defeat what the reservation is for. An orphan's id is reserved *at that exact value* so a state added later cannot take it while a frontend built against the old file is still using it — that is why `StateEntry.parent` is kept for dead entries at all. Rehoming the orphan frees the original id for reuse, which is precisely the hazard the reservation exists to prevent.

The invariant that a state must not share its parent's id exists to keep a relative path unambiguous at runtime. An orphan has no class, so it resolves no name and cannot create that ambiguity. The over-strict side was `validate`, not `sync`, so the parent-id check now applies only to live entries and the orphan keeps `"a"`. If the deleted class ever comes back the entry becomes live and the repair pass moves it then, so it is self-correcting.

Covered by `test_validate_exempts_an_orphan_holding_its_parent_id`.

This also exposed a flaw in one of my own tests: `test_validate_detects_a_child_reusing_its_parent_id` built its config from a path with no class behind it — an orphan — while claiming to check a live child. It uses a real subclass now.

The live-sibling variant you raised separately is a different mechanism and is fixed in e1ba2bb.

benedikt-bartscher:
consider exposing "minify.json is out of date, so these names compile unminified" as const - also consider disabling this warning if pytest env is set?

Issue #3564, opened 2024-06-25
[REF-3190] Add testing to verify props propagation of all reflex components


Lendemor:
Creating this issue to track an idea suggested in #3541.

<sub>[REF-3190](https://linear.app/reflex-dev/issue/REF-3190/add-testing-to-verify-props-propagation-of-all-reflex-components)</sub>

benedikt-bartscher:
@abulvenz in https://github.com/reflex-dev/reflex/issues/3541 you mentioned "In the first draft I tried to do it with automatic component instanciation, but as I wrote there are many exceptions from the rule.".

Maybe we can introduce a `create_fake`/`create_demo` classmethod in `rx.Component` which raises `NotImplementedError` by default. Every component which has a `create_fake`/`create_demo` method could then be included in a dynamic pytest fixture. In the future, we could also use this classmethod to generate docs/examples.

abulvenz:
> @abulvenz do you want me to assign this to you? (since you have a draft working on it)

Thanks for asking @Lendemor . Right now this has low prio to me. If you assign it to me, it might get done in some bright future when I find the time.

ShubhamAXS19:
can i work on this issue ??

picklelo:
@ShubhamAXS19 yes, just assigned you! Let us know if you have issues / questions

picklelo:
@ss-mathew just assigned you!

masenf:
complete prop testing will be added for the new component library, we won't be backfilling coverage for radix or older components. their behavior is currently pinned by a handful of integration tests.

PR #6105, opened 2026-01-29
fix: apply ID to debounceinput even when it is a var.

abulvenz:
### All Submissions:

When using `value` and `on_change` and an `rx.Var` as `input.id` prop, it will not be correctly set on the input, because debounce input only copies IDs that are not vars.

We had some problems with IDs before. Maybe it's about time for https://github.com/reflex-dev/reflex/issues/3564

Minimal example with working and not working input. Interestingly the var in evaluated in both cases.

```python
import reflex as rx
class State(rx.State):
    val: str = ""

    @rx.event
    def on_change(val):
        return val

def index() -> rx.Component:
    return rx.container(
        rx.input(
            value=State.val, on_change=State.on_change,
            id=rx.Var.create("broken"),
        ),
        rx.input(
            value=State.val, on_change=State.on_change,
            id=f"{rx.Var.create('works')}",
        ),
    )

app = rx.App()
app.add_page(index)

### DIFF
--- a/reflex/components/core/debounce.py
+++ b/reflex/components/core/debounce.py
@@ -111,6 +111,7 @@ class DebounceInput(Component):
         child_ref = child.get_ref()
         if props.get("input_ref") is None and child_ref:
             props["input_ref"] = Var(_js_expr=child_ref, _var_type=str)
+        if child.id is not None:
             props["id"] = child.id
 
         # Set the child element to wrap, including any imports/hooks from the child.


benedikt-bartscher:
ref https://github.com/reflex-dev/reflex/issues/3564

masenf (approved):

abulvenz:
> ref #3564

Ah, thanks! You are right. I linked the wrong issue in the description. Corrected now.

PR #6111, opened 2026-02-01
fix(data_table): resolve blank cells for non-camelCase column keys

### LINKED ISSUE #6108: rx.data_table shows blank cells for some column keys (possibly case-related?)
### What I’m seeing

I might be missing something obvious, but I’m running into a weird behavior with `rx.data_table`: after loading data, the table headers render, but only one column (“from”) shows values. The other two (“Route” and “To”) stay blank even though the underlying data seems present.

### Minimal repro

```python
import reflex as rx
import pandas as pd


class State(rx.State):
    data: list = []

    def load_data(self):
        df = pd.DataFrame(
            {
                "Route": ["1", "2", "3"],
                "To": ["Central", "Kowloon", "Airport"],
                "from": ["Tsuen Wan", "Mong Kok", "Tung Chung"],
            }
        )
        self.data = df.to_dict("records")


def index():
    return rx.vstack(
        rx.button("Load Data", on_click=State.load_data),
        rx.data_table(
            data=State.data,
            columns=["Route", "To", "from"],
            sort=True,
        ),
    )


app = rx.App()
app.add_page(index)
```

### Steps to reproduce

1. Run the app.
2. Click **Load Data**.
3. Observe the table.

### Expected

All three columns should display values.

### Actual

* Column **from** displays correctly (`Tsuen Wan`, `Mong Kok`, `Tung Chung`)
* Columns **Route** and **To** appear but their cells are empty.

### Screenshot

(Attached) — only the `from` column shows values.

Image

### Guess / question

Not sure if this is intended, but I wonder if the table is normalizing column keys internally (e.g., lowercasing `"Route"` / `"To"`), which might cause it not to match the dict keys coming from `df.to_dict("records")`. Totally possible I’m using `columns` wrong though.

### Workaround I found

If I rename the DataFrame columns to lowercase (e.g., `route`, `to`, `from`) it seems to work.

### Environment

* Reflex: `0.8.26`
* Python: `3.12`
* OS: `Mac`
* Browser: `Google Chrome`

zain2983:
fix(data_table): resolve blank cells for non-camelCase column keys (#6108)

---

### All Submissions:
- [x] Have you followed the guidelines stated in [[CONTRIBUTING.md](https://github.com/reflex-dev/reflex/blob/main/CONTRIBUTING.md)](https://github.com/reflex-dev/reflex/blob/main/CONTRIBUTING.md) file?
- [x] Have you checked to ensure there aren't any other open [[Pull Requests](https://github.com/reflex-dev/reflex/pulls)](https://github.com/reflex-dev/reflex/pulls) for the desired changed?

### Type of change

### Changes To Core Features:

---

### Description

When `rx.data_table` receives `data` as a runtime `Var` and `columns` as a static list of strings, any column key that isn't already fully lowercase renders with blank cells.

This happens because Grid.js infers each column's lookup `id` by camelCasing the column name string. So a key like `"Route"` gets inferred as `"route"`, which no longer matches the actual key in the data. Since `data` is a runtime `Var`, the keys aren't available at build time, so normalising columns alone can't fix it.

The fix converts each plain column string into the `{ id, name }` object format that Grid.js also supports. When `id` is set explicitly, Grid.js skips the camelCase inference entirely and uses it directly as the lookup key. This ensures the key always matches regardless of casing. Columns already in object format are left untouched, so this is fully backward-compatible.

Closes #6108


reflex/components/gridjs/datatable.py
```diff
@@ -14,58 +14,52 @@
...
-
+    # -----------------------------
+    # Component Props
+    # -----------------------------
+    data: Any  # The data to display (list of lists or DataFrame)
+    columns: Var[Sequence]  # Columns to display (optional if using DataFrame)
+    search: Var[bool]  # Enable search
+    sort: Var[bool]  # Enable sorting
+    resizable: Var[bool]  # Enable column resizing
+    pagination: Var[bool | dict]  # Enable pagination
```
masenf, line 43: these should not be reformatted; our doc generator specifically looks for comments preceeding props, so if you write it like this, our doc generator breaks

reflex/components/gridjs/datatable.py
```diff
@@ -14,58 +14,52 @@
...
+
+    # -----------------------------
+    # Component creation
+    # -----------------------------
```
masenf, line 47: reflex code style avoids these kind of fancy block comments

reflex/components/gridjs/datatable.py
```diff
@@ -14,58 +14,52 @@
...
-        Raises:
-            ValueError: If a pandas dataframe is passed in and columns are also provided.
+            DataTable: The created DataTable component.
         """
         data = props.get("data")
         columns = props.get("columns")
 
-        # The annotation should be provided if data is a computed var. We need this to know how to
-        # render pandas dataframes.
+        # 1️⃣ Ensure computed Vars have type annotations
```
masenf, line 62: Remove the emoji from the comments

reflex/components/gridjs/datatable.py
```diff
@@ -118,14 +124,53 @@ def _render(self) -> Tag:
...
+            elif isinstance(self.columns, Var):
+                self.columns = self.columns._replace(
+                    _js_expr=(
+                        f"{self.columns._js_expr}.map("
+                        "(col) => typeof col === 'string' "
+                        "? ({ id: col, name: col }) "
+                        ": col)"
+                    ),
+                    _var_type=list[Any],
+                )
```
masenf, line 171: prefer to use Var Operations to construct the javascript expression

```python
self.columns = self.columns.foreach(
    lambda col: rx.cond(col.js_type() == 'string', {"id": col, "name": col}, col)
)
```

The code it generates is not as "pretty", but it's more likely to remain functional as the framework evolves and the code generation changes.

pyi_hashes.json
```diff
@@ -1,122 +1,3 @@
...
-  "reflex/components/react_player/video.pyi": "998671c06103d797c554d9278eb3b2a0",
-  "reflex/components/react_router/dom.pyi": "3042fa630b7e26a7378fe045d7fbf4af",
-  "reflex/components/recharts/__init__.pyi": "6ee7f1ca2c0912f389ba6f3251a74d99",
-  "reflex/components/recharts/cartesian.pyi": "cfca4f880239ffaecdf9fb4c7c8caed5",
-  "reflex/components/recharts/charts.pyi": "013036b9c00ad85a570efdb813c1bc40",
-  "reflex/components/recharts/general.pyi": "d87ff9b85b2a204be01753690df4fb11",
-  "reflex/components/recharts/polar.pyi": "ad24bd37c6acc0bc9bd4ac01af3ffe49",
-  "reflex/components/recharts/recharts.pyi": "c41d19ab67972246c574098929bea7ea",
-  "reflex/components/sonner/toast.pyi": "3c27bad1aaeb5183eaa6a41e77e8d7f0"
+  "reflex/components/gridjs/datatable.pyi": "e1f34ade3873a931770da4a35586f298"
```
masenf, line 2: the other hashes in this file are important too, only the changed files should be updated here

masenf (changes requested):

masenf:
please reopen when comments are fixed

PR #6118, opened 2026-02-07
cache get_type_hints for heavy init subclass stuff - improve import performance

benedikt-bartscher:
For one of my apps this reduces the app import time over 50%


adhami3310:
in a followup, we should also add
```patch
index 89698cfa..b87f2e4a 100644
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -569,14 +569,14 @@ class BaseState(EvenMoreBasicBaseState):
 
         new_backend_vars = {
             name: value if not isinstance(value, Field) else value.default_value()
-            for mixin_cls in [*cls._mixins(), cls]
+            for mixin_cls in (*cls._mixins(), cls)
             for name, value in list(mixin_cls.__dict__.items())
             if types.is_backend_base_variable(name, mixin_cls)
         }
         # Add annotated backend vars that may not have a default value.
         new_backend_vars.update({
             name: cls._get_var_default(name, annotation_value)
-            for mixin_cls in [*cls._mixins(), cls]
+            for mixin_cls in (*cls._mixins(), cls)
             for name, annotation_value in mixin_cls._get_type_hints().items()
             if name not in new_backend_vars
             and types.is_backend_base_variable(name, mixin_cls)
@@ -764,7 +764,7 @@ class BaseState(EvenMoreBasicBaseState):
         return getattr(cls, unique_var_name)
 
     @classmethod
-    def _mixins(cls) -> list[type]:
+    def _mixins(cls) -> list[type[BaseState]]:
         """Get the mixin classes of the state.
 
         Returns:
```
to improve type safety around those functions that call this

masenf (commented):
do you get the same performance improvements by pulling the `get_type_hints` out of the for loop without adding the functools cache?

benedikt-bartscher:
> do you get the same performance improvements by pulling the `get_type_hints` out of the for loop without adding the functools cache?

no, 4,103s total vs 7,244s total

benedikt-bartscher:
@adhami3310 thanks for your input, i just migrated it to tuple and applied stricter typing

PR #6119, opened 2026-02-08
fix: enforce server-side upload size and file count limits

### LINKED ISSUE #6115: Denial of Service (DoS) via Memory Exhaustion in File Upload Handling
### Summary
A potential Denial of Service (DoS) vulnerability exists in Reflex's backend file upload handling. The current implementation reads the entire content of uploaded files into memory (`io.BytesIO`) without enforcement of size limits, which can lead to rapid memory exhaustion and server crashes.

### Technical Analysis
In `reflex/app.py` (L1767), the code iterates through incoming files and performs a full read into a `BytesIO` object:

```python
# reflex/app.py
file_copies = []
for file in files:
    content_copy = io.BytesIO()
    content_copy.write(await file.read())  # <--- Critical Point
    content_copy.seek(0)
    # ...
```

### Impact
An attacker can exploit this by sending multiple concurrent requests with large files. Since `await file.read()` loads the full payload into RAM, it will quickly consume all available system memory, potentially triggering the OOM killer or rendering the server unresponsive.

### Proof of Concept (Isolated Test)
The following script replicates the vulnerable logic and demonstrates the linear memory spike:

```python
import asyncio
import io
import os
import psutil

class MockUploadFile:
    def __init__(self, size):
        self.size = size
        
    async def read(self):
        return b"A" * self.size

async def simulate_reflex_upload():
    process = psutil.Process(os.getpid())
    print(f"[*] Initial Memory Usage: {process.memory_info().rss / 1024 / 1024:.2f} MB")
    
    FILE_SIZE = 500 * 1024 * 1024 # 500MB
    file = MockUploadFile(FILE_SIZE)
    
    print(f"[*] Executing Reflex logic: io.BytesIO().write(await file.read())...")
    content_copy = io.BytesIO()
    content_copy.write(await file.read())
    
    print(f"[*] Post-Execution Memory Usage: {process.memory_info().rss / 1024 / 1024:.2f} MB")

if __name__ == "__main__":
    asyncio.run(simulate_reflex_upload())
```

### Recommended Mitigation
1. **Configurable Size Limits:** Introduce a `MAX_UPLOAD_SIZE` setting to reject large files before they are read.
2. **Streaming/Spilling to Disk:** Leverage FastAPI's `UploadFile` ability to spill to disk or use `tempfile.TemporaryFile` instead of `io.BytesIO` for large payloads.

tsaipraveen99:
## Problem

The upload handler in `reflex/app.py` reads entire uploaded files into memory
(`io.BytesIO`) with no size limit. The frontend `Upload` component has
`max_size`/`max_files` props, but these are client-side only and trivially
bypassed via direct HTTP requests. An attacker can send arbitrarily large or
numerous files to exhaust server memory and crash the process.

Closes #6115

## Solution

Add server-side enforcement with three layers of defense, controlled by two new
config fields:

| Config field | Env var | Default | Description |
|---|---|---|---|
| `upload_max_size` | `REFLEX_UPLOAD_MAX_SIZE` | `10485760` (10 MB) | Per-file size limit in bytes. `0` disables. |
| `upload_max_files` | `REFLEX_UPLOAD_MAX_FILES` | `10` | Max files per request. `0` disables. |

**Layer 1 — Content-Length pre-check.** Before Starlette parses the multipart
body, the request's `Content-Length` header is compared against
`upload_max_size × upload_max_files`. Oversized requests are rejected
immediately with HTTP 413. Malformed `Content-Length` values return HTTP 400.

**Layer 2 — Per-file `file.size` header check.** After form parsing, each
file's reported size is checked against `upload_max_size`. Files that declare a
size over the limit are rejected without reading any bytes.

**Layer 3 — Chunked read with byte counting.** File content is streamed in 1 MB
chunks during the copy loop. If accumulated bytes exceed `upload_max_size`, the
read is aborted. This catches files that omitted or spoofed their size header.

## Files changed

- **`reflex/config.py`** — Add `upload_max_size` and `upload_max_files` to
  `BaseConfig`. Add validation in `Config._post_init` (after `update_from_env`
  applies overrides, so invalid env values like `REFLEX_UPLOAD_MAX_SIZE=-1` are
  caught).
- **`reflex/app.py`** — Add all three enforcement layers plus file count check
  to `upload_file()`. A `_cleanup()` helper ensures all file handles and
  already-copied `BytesIO` buffers are closed on every rejection path. Error
  messages do not include the attacker-controlled filename.
- **`tests/units/test_upload_size_limit.py`** *(new)* — 14 tests covering
  config defaults, negative value rejection, Content-Length handling
  (over/under/malformed), per-file rejection (via size header and chunked read),
  file count enforcement, disabled-limit behavior, and filename sanitization.
  All tests invoke the actual `upload()` handler with mocked requests.

## Migration note

Existing apps that accept uploads larger than 10 MB or more than 10 files per
request will need to adjust their config:

```python
# rxconfig.py
config = rx.Config(
    app_name="myapp",
    upload_max_size=50 * 1024 * 1024,  # 50 MB
    upload_max_files=25,
)
```

Or via environment variables:

```bash
REFLEX_UPLOAD_MAX_SIZE=0 reflex run  # disable size limit entirely
```

## Design decisions

- **HTTP 413** (Payload Too Large) for size violations, **HTTP 400** (Bad
  Request) for file count violations — correct HTTP semantics for each case.
- **`0` disables the limit** rather than a separate boolean — follows the
  convention used by nginx (`client_max_body_size 0`), Apache, and others.
- **Filename omitted from error messages** — `file.filename` is
  attacker-controlled input; echoing it in the response risks reflected XSS if
  the JSON is ever rendered as HTML downstream.
- **`_cleanup()` takes explicit parameters** rather than closing over the
  `file_list` variable — avoids a fragile closure dependency and also closes
  `BytesIO` buffers from already-copied files when a later file triggers
  rejection.

## Test plan

- [x] New tests: `pytest tests/units/test_upload_size_limit.py` — 14 passed
- [x] Existing upload handler tests: `pytest tests/units/test_app.py -k upload` — 9 passed
- [x] Existing upload component tests: `pytest tests/units/components/core/test_upload.py` — 6 passed
- [x] Config tests: `pytest tests/units/test_config.py` — 31 passed

### DIFF
--- a/reflex/app.py
+++ b/reflex/app.py
@@ -1911,15 +1911,46 @@ def upload(app: App):
         """
         from reflex.utils.exceptions import UploadTypeError, UploadValueError
 
+        config = get_config()
+        upload_max_size = config.upload_max_size
+        upload_max_files = config.upload_max_files
+
+        # Reject based on Content-Length before Starlette buffers the body.
+        # Content-Length covers the entire multipart request, so use
+        # (per-file limit * max files) as the upper bound.
+        if upload_max_size > 0 and upload_max_files > 0:
+            content_length_str = request.headers.get("content-length")
+            if content_length_str is not None:
+                try:
+                    content_length = int(content_length_str)
+                except ValueError:
+                    return JSONResponse(
+                        status_code=400,
+                        content={"detail": "Invalid Content-Length header."},
+                    )
+                max_request_size = upload_max_size * upload_max_files
+                if content_length > max_request_size:
+                    return JSONResponse(
+                        status_code=413,
+                        content={"detail": f"Upload exceeds the maximum allowed size of {max_request_size} bytes."},
+                    )
+
         # Get the files from the request.
         try:
             files = await request.form()
         except ClientDisconnect:
             return Response()  # user cancelled
-        files = files.getlist("files")
-        if not files:
+        file_list = files.getlist("files")
+        if not file_list:
             msg = "No files were uploaded."
             raise UploadValueError(msg)
+
+        # Enforce max file count.
+        if upload_max_files > 0 and len(file_list) > upload_max_files:
+            return JSONResponse(
+                status_code=400,
+                content={"detail": f"Too many files uploaded ({len(file_list)}). Maximum allowed is {upload_max_files}."},
+            )
 
         token = request.headers.get("reflex-client-token")
         handler = request.headers.get("reflex-event-handler")
@@ -1966,19 +1997,54 @@ def upload(app: App):
             )
             raise UploadValueError(msg)
 
+        async def _cleanup(upload_files, copied_files):
+            """Close all uploaded files and any already-copied BytesIO buffers.
+
+            Args:
+                upload_files: The raw uploaded file list from the request.
+                copied_files: The list of UploadFile copies made so far.
+            """
+            for f in upload_files:
+                if isinstance(f, StarletteUploadFile):
+                    await f.close()
+            for f in copied_files:
+                f.file.close()
+
         # Make a copy of the files as they are closed after the request.
         # This behaviour changed from fastapi 0.103.0 to 0.103.1 as the
         # AsyncExitStack was removed from the request scope and is now
         # part of the routing function which closes this before the
         # event is handled.
         file_copies = []
-        for file in files:
+        for file in file_list:
             if not isinstance(file, StarletteUploadFile):
+                await _cleanup(file_list, file_copies)
                 raise UploadValueError(
                     "Uploaded file is not an UploadFile." + str(file)
                 )
+            # Enforce upload size limit: early rejection via file.size header
+            if upload_max_size > 0 and file.size is not None and file.size > upload_max_size:
+                await _cleanup(file_list, file_copies)
+                return JSONResponse(
+                    status_code=413,
+                    content={"detail": f"File exceeds the maximum upload size of {upload_max_size} bytes."},
+                )
             content_copy = io.BytesIO()
-            content_copy.write(await file.read())
+            # Read in chunks to enforce limit even when file.size is not reported
+            bytes_read = 0
+            while True:
+                chunk = await file.read(1024 * 1024)  # 1 MB chunks
+                if not chunk:
+                    break
+                bytes_read += len(chunk)
+                if upload_max_size > 0 and bytes_read > upload_max_size:
+                    content_copy.close()
+                    await _cleanup(file_list, file_copies)
+                    return JSONResponse(
+                        status_code=413,
+                        content={"detail": f"File exceeds the maximum upload size of {upload_max_size} bytes."},
+                    )
+                content_copy.write(chunk)
             content_copy.seek(0)
             file_copies.append(
                 UploadFile(
@@ -1989,12 +2055,10 @@ def upload(app: App):
                 )
             )
 
-        for file in files:
-            if not isinstance(file, StarletteUploadFile):
-                raise UploadValueError(
-                    "Uploaded file is not an UploadFile." + str(file)
-                )
-            await file.close()
+        # Close the raw uploaded files (copies are kept for event processing).
+        for file in file_list:
+            if isinstance(file, StarletteUploadFile):
+                await file.close()
 
         event = Event(
             token=token,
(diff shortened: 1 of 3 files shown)


tests/units/test_upload_size_limit.py
```diff
@@ -0,0 +1,284 @@
...
+def _make_request_mock(files, content_length=None):
+    """Create a mock Starlette Request with the given files."""
+    request_mock = unittest.mock.Mock()
+    headers = {
+        "reflex-client-token": "test-token",
+        "reflex-event-handler": "fake_state.handle_upload",
+    }
+    if content_length is not None:
+        headers["content-length"] = str(content_length)
+    request_mock.headers = headers
```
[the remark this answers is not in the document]
tsaipraveen99: Good catch — fixed in the latest push.

`_make_request_mock` now uses a sentinel default and only calls `str()` on `int` inputs. Raw string values (including `""`, `"   "`, etc.) are passed through directly to the header dict, so the mock accurately represents what Starlette would surface.

The malformed Content-Length test is now parametrized across five edge cases:

| Input | `int()` behavior | Expected |
|---|---|---|
| `"not-a-number"` | `ValueError` | 400 |
| `""` | `ValueError` | 400 |
| `"   "` | `ValueError` | 400 |
| `"12.5"` | `ValueError` | 400 |
| `"1e3"` | `ValueError` | 400 |

All 18 tests pass (up from 14 — the parametrized test adds 4 new cases).

reflex/app.py
```diff
@@ -1966,19 +1997,54 @@ async def upload_file(request: Request):
...
+                return JSONResponse(
+                    status_code=413,
+                    content={"detail": f"File exceeds the maximum upload size of {upload_max_size} bytes."},
+                )
             content_copy = io.BytesIO()
-            content_copy.write(await file.read())
+            # Read in chunks to enforce limit even when file.size is not reported
+            bytes_read = 0
+            while True:
+                chunk = await file.read(1024 * 1024)  # 1 MB chunks
```
adhami3310, line 2036: wouldn't it be simpler to read the maximum size? and then see if there's anything left

tsaipraveen99: Good question! I considered that approach, but went with chunked reads intentionally for memory reasons.

If `upload_max_size` is set to something large (e.g. 200 MB), a single `await file.read(upload_max_size + 1)` allocates up to ~200 MB in one shot — even for a 1 KB file that would finish quickly in the loop. The chunked approach caps the in-flight read buffer to 1 MB regardless of the configured limit, and aborts early if the file exceeds the limit partway through.

The full content does end up in a `BytesIO` for valid files, so peak memory for *accepted* files is similar either way. But for *rejected* files (the attack scenario this is defending against), the chunked approach avoids buffering up to `upload_max_size` bytes before deciding to reject — it stops as soon as bytes read crosses the threshold.

adhami3310: > a single await file.read(upload_max_size + 1) allocates up to ~200 MB in one shot

are you sure? i don't see much of that on the internet

reflex/config.py
```diff
@@ -260,6 +260,13 @@ class BaseConfig:
     # The transport method for client-server communication.
     transport: Literal["websocket", "polling"] = "websocket"
 
+    # Maximum file upload size in bytes. Files larger than this will be rejected with HTTP 413.
+    # Defaults to 10 MB. Set to 0 to disable the limit.
+    upload_max_size: int = 10 * 1024 * 1024  # 10 MB
+
+    # Maximum number of files per upload request. Set to 0 to disable the limit.
+    upload_max_files: int = 10
+
```
masenf, line 269: if we are going to take this, these should default to unset. If the values are unset during the upload processing event, then a deprecation warning should be emitted that tells users to set these values as they will be enforced in the next minor release (either 0.10 or 1.0).

masenf (commented):
Now that reflex also supports `upload_file_chunk`, if we're going to impose limits, then they need to apply to both upload methods.

masenf:
closing this PR for now as it has not had any updates in 3 months. If you address the concerns and refresh the branch from `main`, reopen the PR and we can keep the conversation going.

PR #6129, opened 2026-02-12
fix: Handle periods in dynamic route segments for production mode

### LINKED ISSUE #5803: Can't have periods '.' in dynamic routes in prod because they break sirv
**Describe the bug**
We have periods (like... /data/v1.2.3/foo) that work in dev mode, but break in production mode. It looks like this is due to sirv treating them like not found assets and having a different code path.

**To Reproduce**
1. Make a dynamic route ('/data/[version]/[prop]')
2. run with --env prod
3. navigate to that route with a value with a period in it ('curl http://localhost:3000/data/v1.2.3/foo')
4. the 404.html fallback page is not loaded. just an empty 404 response directly from sirv.

- Code/Link to Repo:

**Expected behavior**

**Screenshots**

- Python Version: 3.12
- Reflex Version: 0.8.11
- OS: Linux
- Browser (Optional):

**Additional context**
https://github.com/lukeed/sirv/issues/132 describes the upstream bug that impacts reflex.

and sirv's documentation reads:

>Any asset requests (URLs that end with an extension) ignore --single behavior and will send a 404 response instead of the "index.html" fallback. To ignore additional paths, pass URL patterns to the --ignores argument.

# Don't include "/blog*" or "/portfolio*" pages into SPA
$ sirv public --single --ignores "^/blog" --ignores "^/portfolio"
You may pass a string to customize which file should be sent as fallback.
In other words, --single shell.html will send the directory's shell.html file instead of its index.html file.

so it seems quite clear that this does not work in sirv.

timon0305:
## Summary
Fixes dynamic routes with periods (e.g., `/data/v1.2.3/info`) breaking in production mode. sirv treats URLs with extensions as asset requests, returning 404 instead of falling back to the SPA handler.

## Changes
- Added `get_dynamic_route_prefixes()` in [reflex/route.py](reflex/route.py) to extract route patterns with dynamic segments
- Updated `_compile_package_json()` in [reflex/utils/frontend_skeleton.py](reflex/utils/frontend_skeleton.py) to accept dynamic route prefixes
- Wired route collection in [reflex/app.py](reflex/app.py) during compilation to generate sirv `--ignores` flags

## Example
For routes `/data/[version]/info` and `/posts/[slug]`, the generated package.json now includes:
```json
"prod": "sirv ./build/client --single 404.html --host --ignores '^/data/' --ignores '^/posts/'"
```

## Test plan
- [x] All 27 route unit tests pass
- [x] 2589 unit tests pass
- [x] Manual testing with test app confirmed `/data/v1.2.3/info` works in production mode

Fixes #5803

### DIFF
--- a/reflex/route.py
+++ b/reflex/route.py
@@ -132,6 +132,54 @@ def replace_brackets_with_keywords(input_string: str) -> str:
     )
 
 
+def get_dynamic_route_prefixes(routes: list[str]) -> list[str]:
+    """Get route prefixes for dynamic routes to use with sirv --ignores.
+
+    sirv treats URLs with extensions (periods) as asset requests, not SPA routes.
+    This function extracts route patterns that have dynamic segments, so they can
+    be excluded from asset detection using sirv's --ignores flag.
+
+    Example:
+        >>> get_dynamic_route_prefixes(["/posts/[slug]", "/data/[version]/info"])
+        ['^/posts/', '^/data/']
+
+    Args:
+        routes: List of route strings.
+
+    Returns:
+        List of regex patterns for sirv --ignores flag.
+    """
+    prefixes = []
+    for route in routes:
+        # Check if route has dynamic segments
+        if get_route_args(route):
+            # Extract the prefix up to the first dynamic segment
+            parts = route.split("/")
+            prefix_parts = []
+            for part in parts:
+                # Stop at first dynamic segment
+                if (
+                    constants.RouteRegex.ARG.match(part)
+                    or constants.RouteRegex.OPTIONAL_ARG.match(part)
+                    or part == constants.RouteRegex.SPLAT_CATCHALL
+                ):
+                    break
+                prefix_parts.append(part)
+
+            # Build prefix pattern (e.g., "/posts" -> "^/posts/")
+            prefix = "/".join(prefix_parts)
+            # Ensure prefix starts with / for absolute path matching
+            if not prefix.startswith("/"):
+                prefix = "/" + prefix
+            if prefix and prefix != "/":
+                # Add trailing slash and anchor to start for sirv pattern
+                pattern = f"^{prefix}/"
+                if pattern not in prefixes:
+                    prefixes.append(pattern)
+
+    return prefixes
+
+
 def route_specificity(keyworded_route: str) -> tuple[int, int, int]:
     """Get the specificity of a route with keywords.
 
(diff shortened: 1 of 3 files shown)


reflex/utils/frontend_skeleton.py
```diff
@@ -167,23 +167,39 @@ def _update_react_router_config(config: Config, prerender_routes: bool = False):
...
+    """Compile package.json with optional dynamic route handling.
+
+    Args:
+        dynamic_route_prefixes: List of route prefix patterns for sirv --ignores.
+    """
+    # Build prod command with --ignores flags for dynamic routes
+    prod_command = constants.PackageJson.Commands.PROD
+    if dynamic_route_prefixes:
+        # Add --ignores flag for each dynamic route prefix
+        ignores = " ".join(f"--ignores '{prefix}'" for prefix in dynamic_route_prefixes)
```
masenf, line 180: use `shlex.join` instead so these values get properly shell quoted

reflex/route.py
```diff
@@ -132,6 +132,56 @@ def replace_brackets_with_keywords(input_string: str) -> str:
...
+                prefix_parts.append(part)
+
+            # Build prefix pattern (e.g., "/posts" -> "^/posts/")
+            prefix = "/".join(prefix_parts)
+            # Ensure prefix starts with / for absolute path matching
+            if not prefix.startswith("/"):
+                prefix = "/" + prefix
+            if prefix and prefix != "/":
+                # Add trailing slash and anchor to start for sirv pattern
+                pattern = f"^{prefix}/"
```
masenf, line 178: should use `re.escape(prefix)` here to ensure the prefix will never contain regex injection

masenf (commented):
cool fix, but it doesn't work for me

masenf:
is this even a bug anymore? on current `main`:

```python
import reflex as rx


def index():
    return rx.heading(rx.State.id, rx.State.version, rx.State.slug)


app = rx.App()
app.add_page(index, route="/foo/[id]/[version]/[slug]")
app.add_page(index, route="/data/[version]/info")
app.add_page(index, route="/posts/[slug]")
app.add_page(index, route="/[id]")
app.add_page(index)
```

all of `/data/v1.2.3/info` and various combinations of periods are rendering correctly. the `/[id]` one does not work with periods, but that's probably okay.

maybe the bug got fixed upstream? 

either way, when i run with the `--ignores` arguments from this PR, the routes below them don't seem to render at all

PR #6136, opened 2026-02-17
0828dev

adhami3310:


reflex/components/recharts/cartesian.py
```diff
@@ -241,12 +241,6 @@ class Brush(Recharts):
     # The default end index of brush. If the option is not set, the end index will be calculated by the length of data.
     end_index: Var[int]
 
-    # The fill color of brush
-    fill: Var[str | Color]
-
-    # The stroke color of brush
-    stroke: Var[str | Color]
-
```
masenf, line 249: were these removed upstream? i wonder if we need some kind of warning to any users that were passing these props?

masenf (dismissed):

reflex/components/recharts/cartesian.py
adhami3310: they are defined above, ruff was flagging it

PR #6137, opened 2026-02-18
feat: Add function to skip serialization.

abulvenz:
This is needed for permissions and roles on states.

### DIFF
--- a/tests/units/test_state.py
+++ b/tests/units/test_state.py
@@ -4475,13 +4475,13 @@ async def test_rebind_mutable_proxy(mock_app: rx.App, token: str) -> None:
 class NormalState(rx.State):
     """This state should be serialized."""
 
-    value: rx.Field[int] = rx.field(42)
+    value: int = 42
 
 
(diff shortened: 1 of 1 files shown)


abulvenz:
Locally the test passes.

reflex/state.py
```diff
@@ -2059,6 +2059,15 @@ def _dirty_computed_vars(
             if include_backend or not self.computed_vars[cvar]._backend
         }
 
+    @property
```
abulvenz, line 2062: In order to address greptile's name mangling apprehension and to circumvent `EventHandlerShadowsBuiltInStateMethodError` this is a RO-property now.

adhami3310:
is there value from having this as an instance method as opposed to a classmethod? just more flexibility?

benedikt-bartscher:
> is there value from having this as an instance method as opposed to a classmethod? just more flexibility?

when used for permissions, you most likely want to check values of the state instance

PR #6148, opened 2026-03-01
fix: unwrap MutableProxy values in _mark_dirty before writing

benedikt-bartscher:


masenf:
@benedikt-bartscher what is the use case / bug fixed by this patch? Is there a problem if the data structures get proxy objects added to them? When serializing to redis, we don't serialize the proxy objects, so they get refreshed each time the state is reloaded for processing another event.

Also noticed that if the args is a mutable proxy of a list that contains other mutable proxy as items, then these items will not be unwrapped, not sure if that was intentional or not.

benedikt-bartscher:
@masenf MutableProxy caused issues with sqlalchemy realtions. If i have spare time for it, i may add a unit test with sqlalchemy

benedikt-bartscher:
related https://github.com/reflex-dev/reflex/pull/6168

benedikt-bartscher:
i guess #6168 solves my issues - if not i am going to add more tests and re-open this

thanks @masenf

PR #6155, opened 2026-03-05
fix: make disable_plugins accept Plugin types instead of strings

### LINKED ISSUE #6150: Inconsistent types between `plugins` and `disable_plugins` in `rx.Config` (str vs Plugin)
## Summary
In `rx.Config`, the `disable_plugins` parameter only accepts a list of strings, while the `plugins` parameter correctly expects a list of `Plugin` instances. This creates an inconsistency and produces warnings when trying to disable built‑in plugins such as `SitemapPlugin`, which is defined as a `Plugin` object rather than a string.

## Expected Behavior
Both `plugins` and `disable_plugins` should follow the same type pattern. Since `plugins` accepts a list of `Plugin` instances, it would be more consistent for `disable_plugins` to also accept a list of `Plugin` objects (or at least support both types).

## Actual Behavior
- `disable_plugins` only accepts `list[str]`.
- Adding `"SitemapPlugin"` to `disable_plugins` triggers warnings because the sitemap is defined as a `Plugin` instance.
- This makes disabling built‑in plugins unintuitive and error‑prone.

## Example

```python
import reflex as rx

config = rx.Config(
    plugins=[rx.plugins.SitemapPlugin()],
    disable_plugins=[rx.plugins.SitemapPlugin()],  # Causes warnings
)

### PR DESCRIPTION
**Pull request by masenf**
Change `disable_plugins` from `list[str]` to `list[type[Plugin]]` so both `plugins` and `disable_plugins` use consistent Plugin-based types (#6150).

Strings and instances are still accepted at runtime with a deprecation warning, and normalized to the Plugin class. Environment variable support via REFLEX_DISABLE_PLUGINS continues to work through a new `interpret_plugin_class_env` interpreter, which `interpret_plugin_env` now reuses for class resolution.

Fix #6150

### DIFF
--- a/reflex/config.py
+++ b/reflex/config.py
@@ -383,12 +383,6 @@ class Config(BaseConfig):
             if isinstance(entry, type) and issubclass(entry, Plugin):
                 normalized.append(entry)
             elif isinstance(entry, Plugin):
-                console.deprecate(
-                    feature_name="Passing Plugin instances to disable_plugins",
-                    reason="pass Plugin classes directly instead, e.g. disable_plugins=[SitemapPlugin]",
-                    deprecation_version="0.8.28",
-                    removal_version="0.9.0",
-                )
                 normalized.append(type(entry))
             elif isinstance(entry, str):
                 console.deprecate(

### CODE REVIEW

File: reflex/config.py [hunk 1]
```diff
@@ -369,16 +372,58 @@ def _post_init(self, **kwargs):
...
+        for entry in self.disable_plugins:
+            if isinstance(entry, type) and issubclass(entry, Plugin):
+                normalized.append(entry)
+            elif isinstance(entry, Plugin):
+                console.deprecate(
+                    feature_name="Passing Plugin instances to disable_plugins",
+                    reason="pass Plugin classes directly instead, e.g. disable_plugins=[SitemapPlugin]",
+                    deprecation_version="0.8.28",
+                    removal_version="0.9.0",
+                )
```
masenf:
```suggestion
```

PR #6174, opened 2026-03-13
fix: avoid eagerly copying uploaded files into memory

### LINKED ISSUE #3517: File upload appears to hold everything in memory until the end
**Describe the bug**
1. For large files, memory upload consumes significant memory and it grows even more when writing out the file iobuffer.
2. Upload is very slow and consumes significant CPU when everything is on the same host (both reflex and the browser). Uploading files shouldn't really consume much CPU at all...

**To Reproduce**
 - Just use the standard upload example and upload a multi-gigabyte file.

**Expected behavior**
Very little memory should be used because the data should be incrementally written to a file.
To avoid re-writing the file after upload, I'd expect to be able to specify the target or if you choose to write to a temporary file I'd expect to be able to rename the temporary file to it's final home.

 - Python Version: 3.12.4
 - Reflex Version: 0.5.4
 - OS: WSL Ubuntu 22.04.4 LTS  on Windows 11 pro 
 - Browser (Optional): Brave 1.67.116 ( Chromium 126.0.6478.71)

FarhanAliRaza:
Previously the upload handler read every file into a BytesIO buffer before passing it to the event handler, doubling peak memory for large uploads. Now the original Starlette temp files are passed through directly and form_data.close() is deferred until the streaming response completes.

- Pass file.file directly instead of copying into BytesIO
- Defer form_data.close() to finally block in streaming generator
- Close form_data on error path to prevent resource leaks
- Replace mock form data in tests with real Starlette FormData
- Add test verifying files are not read before handler runs

After these fixes, 4 GB will peak at about 4 GB of RAM when the user actually reads the file. Before there were two spikes, once when copying and once when the user read the file in its view. 

### All Submissions:

### Type of change

closes #3517

### DIFF
--- a/tests/units/test_app.py
+++ b/tests/units/test_app.py
@@ -1056,13 +1056,22 @@ async def test_upload_file_keeps_form_open_until_stream_completes(
     assert not bio1.closed
     assert not bio2.closed
 
-    # Each yield from the handler produces an intermediate update;
-    # assert both file handles stay open throughout streaming.
-    async for _ in streaming_response.body_iterator:
-        assert not bio1.closed
-        assert not bio2.closed
+    # Drive the response through the full ASGI lifecycle so that
+    # _UploadStreamingResponse.__call__ invokes the on_finish callback.
+    scope = {"type": "http"}
+    done = asyncio.Event()
 
-    # After the stream completes, form_data.close() should have been called,
+    async def receive():
+        await done.wait()
+        return {"type": "http.disconnect"}
+
+    async def send(message):  # noqa: RUF029
+        if message.get("type") == "http.response.body" and not message.get("body"):
+            done.set()
+
+    await streaming_response(scope, receive, send)
+
+    # After the ASGI call completes, form_data.close() should have been called,
     # closing both underlying file handles.
     assert form_close.await_count == 1
     assert bio1.closed


FarhanAliRaza:
There is also another method to solve this issue. But that will require a more involved refactor. We kind of use Pheonix live view method. We upload the files in chunks from the frontend and the backend writes those chunks to files. That will be the nuclear option that can reduce Ram a lot. I have to fully research it . It sounds possible.

tests/units/test_app.py
```diff
@@ -1008,6 +1000,69 @@ def getlist(key: str):
...
+    upload_fn = upload(app)
+    streaming_response = await upload_fn(request_mock)
+
+    assert isinstance(streaming_response, StreamingResponse)
+    assert file1.read.await_count == 0
+    assert form_close.await_count == 0
+    assert not bio.closed
+
+    async for _ in streaming_response.body_iterator:
+        pass
```
masenf: we should `assert not bio.closed` for each iteration of this loop

tests/units/test_app.py
```diff
@@ -974,15 +974,7 @@ async def test_upload_file(tmp_path, state, delta, token: str, mocker: MockerFix
...
-        files_mock = unittest.mock.Mock()
-
-        def getlist(key: str):
-            assert key == "files"
-            return [file1, file2]
-
-        files_mock.getlist = getlist
-
-        return files_mock
+        return FormData([("files", file1), ("files", file2)])
```
masenf, line 975: can you extend this test:

1. use actual different BinaryIO instances for each file
2. set a `_tmp_path` in the file upload state using `app.modify_state` (instead of the current `get_state`). This path is where the handler will actually write the data that it reads from the "files"
3. update `multi_handle_upload` to `yield` at the end of processing each UploadFile
4. each iteration of the streaming loop should assert that both BinaryIO handles remain unclosed
5. at the end of processing, also check that the files got written to the tmp dir correctly and that both BinaryIO handles got closed

FarhanAliRaza:
I have fixed the reviews. `_UploadStreamingResponse` finally can be reverted, it will not affect 99% of request it is just extra safety.

reflex/app.py
```diff
@@ -1920,107 +1936,122 @@ async def upload_file(request: Request):
...
+                # Process the event.
+                async with app.state_manager.modify_state_with_links(
+                    event.substate_token
+                ) as state:
+                    async for update in state._process(event):
+                        # Postprocess the event.
+                        update = await app._postprocess(state, event, update)
+                        yield update.json() + "\n"
+            finally:
+                await _close_form_data()
```
masenf: yeah i think cleaning up this path since we handle it with `_UploadStreamingResponse` makes the code more maintainable

```suggestion
            # Process the event.
            async with app.state_manager.modify_state_with_links(
                event.substate_token
            ) as state:
                async for update in state._process(event):
                    # Postprocess the event.
                    update = await app._postprocess(state, event, update)
                    yield update.json() + "\n"
```

masenf (commented):
suggestion based on your last comment.

i like the new approach of baking the close into the streaming response.

reflex/app.py
FarhanAliRaza: done.

FarhanAliRaza:
Integration tests pass locally . The test is maybe flaky.

PR #6180, opened 2026-03-16
Avoid spurious re-renders in context providers

masenf:
Components that read the color mode or event loop context currently re-render when their provider updates for unrelated reasons. This change keeps context values stable so unrelated `ThemeProvider` updates and router navigation do not invalidate those consumers.

`ThemeProvider` memoizes its context values and color-mode callbacks, and attaches the system-preference listener once on mount. It returns its provider elements directly. `EventLoopProvider` memoizes its provider element while `[addEventsLocal, connectErrors, children]` stay unchanged, preserving its context value's identity. The module-level dispatcher assignments still run on every render.

Validation: 286 compiler/template tests, the appearance browser test covering light/dark selection and system-preference changes, and all pre-commit hooks pass. A package changelog fragment describes the effect on context consumers.


tests/units/compiler/test_context_js_template.py
```diff
@@ -0,0 +1,58 @@
...
+    Returns:
+        The source of the generated ``EventLoopProvider`` component.
+    """
+    rendered = context_template(
+        is_dev_mode=False,
+        default_color_mode="light",
+        initial_state={"state": {}},
+        state_name="state",
+    )
+    start = rendered.index("export function EventLoopProvider")
```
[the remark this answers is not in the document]
masenf: Keeping the named delimiter for this scoped change. This helper renders our own fixed template and locates both exports explicitly; if either boundary changes it fails loudly. A brace counter is not a JavaScript parser (strings/template literals can contain braces), while introducing parser infrastructure just for this assertion would broaden the PR. Collected as follow-up: standardize structural extraction across the compiler-template tests, with a shared parser if we adopt one.

tests/units/compiler/test_react_theme_js_template.py
```diff
@@ -0,0 +1,74 @@
...
+def test_theme_provider_color_mode_setters_are_stable() -> None:
+    """``setColorMode``/``toggleColorMode`` must not be rebuilt on every render.
+
+    They are handed to consumers through ``ColorModeContext``, so an unstable
+    identity defeats the memoization of the context value itself.
+    """
+    body = _theme_provider_body()
+
+    assert "const setColorMode = useCallback(" in body
+    assert "const toggleColorMode = useCallback(" in body
```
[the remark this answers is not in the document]
masenf: The implementation already has the correct dependencies: `setColorMode` captures only the stable React setter and uses `[]`; `toggleColorMode` captures `setColorMode` and `resolvedTheme` and lists both. I am keeping this unit test as a structural smoke test rather than duplicating the implementation's exact dependency-array text. Such a snapshot would reject equivalent dependency choices without proving callback identity. The existing browser test covers live color changes, and I am extending it to exercise system-preference changes through the memoized providers. Collected follow-up: measure consumer render counts/callback identity in a reusable context-provider browser test.

tests/units/compiler/test_react_theme_js_template.py
```diff
@@ -0,0 +1,74 @@
...
+        body.index('window.matchMedia("(prefers-color-scheme: dark)")') :
+    ]
+    cleanup_end = listener_effect.index("mediaQuery.removeEventListener")
+    assert (
+        listener_effect[cleanup_end:]
+        .lstrip()
+        .startswith(
+            'mediaQuery.removeEventListener("change", handleChange);\n    };\n  }, []);'
+        )
+    ), "the media-query effect should declare an empty dependency array."
```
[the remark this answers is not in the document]
masenf: Added browser coverage in the existing `test_appearance_color_toggle`: after selecting system mode, it emulates dark and then light preferences and verifies the resolved mode, conditional content, and rendered theme all update. I temporarily removed the listener registration and confirmed the new dark-mode assertion fails (it remains light); after restoring the implementation the test passes. This guards the actual listener behavior without relying on its source spelling.

tests/units/compiler/test_react_theme_js_template.py
```diff
@@ -0,0 +1,74 @@
...
+
+    assert "const themeContextValue = useMemo(" in body, (
+        "ThemeContext value should be memoized."
+    )
+    assert "const colorModeContextValue = useMemo(" in body, (
+        "ColorModeContext value should be memoized."
+    )
+    assert "[themeContextValue, colorModeContextValue, children]" in body, (
+        "ThemeProvider should return a memoized element keyed on its context values."
+    )
```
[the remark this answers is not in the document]
masenf: Extended the existing browser test to exercise system-preference changes while raw mode stays `system`: dark and light updates must reach both the resolved-color consumers and the rendered theme through the memoized providers. The original dark/light selection assertions already cover changing the raw mode. The strengthened test passes, and removing the listener was confirmed to make it fail. Keeping the unit assertions as structural smoke coverage instead of adding a snapshot of the complete memo implementation; consumer render-count/identity coverage is recorded as follow-up alongside the callback discussion.

packages/reflex-base/src/reflex_base/.templates/web/utils/react-theme.js
```diff
@@ -106,21 +107,33 @@ export function ThemeProvider({ children, defaultTheme = "system" }) {
...
+        ThemeContext.Provider,
+        { value: themeContextValue },
+        createElement(
+          ColorModeContext.Provider,
+          { value: colorModeContextValue },
+          children,
+        ),
+      ),
+    [themeContextValue, colorModeContextValue, children],
   );
```
FarhanAliRaza: This outer `useMemo` has no effect. Please remove it and return the `createElement(...)` call directly.

`themeContextValue` and `colorModeContextValue` are already memoized. A provider that gets the same `value` and the same `children` reference already bails out in React.

I measured this with render counters in an example app (a plain component, a `ColorModeContext` consumer, and an `EventLoopContext` consumer, each in the page and in an app wrap). I removed this `useMemo` from the compiled `react-theme.js` and repeated the full run: color mode set, toggle, system mode, OS scheme changes, and navigation. Every render count was identical to the run with the `useMemo`.

tests/units/compiler/test_react_theme_js_template.py
FarhanAliRaza: Please remove this assertion together with the outer element `useMemo` in `react-theme.js`. It pins a memo that does not change any render count.

packages/reflex-base/news/6180.performance.md
```diff
@@ -0,0 +1 @@
+Context providers no longer re-render their subtrees spuriously: `ThemeProvider` memoizes the `ThemeContext`/`ColorModeContext` values (and the `setColorMode`/`toggleColorMode` callbacks it hands out), and `EventLoopProvider` memoizes its provider element so router hook changes no longer cascade through the app. The `ThemeProvider` media-query listener is also attached once on mount instead of being torn down and re-added on every render.
```
FarhanAliRaza: Please shorten this and correct the scope. On `main`, a component that does not read these contexts does not re-render on navigation. Only the context consumers re-render, so there is no subtree cascade to remove.

Suggested text:

> Components that read the color mode or event loop context no longer re-render on unrelated `ThemeProvider` updates or on router navigation. The `ThemeProvider` system-preference listener is now attached once on mount.

packages/reflex-base/src/reflex_base/compiler/templates.py
```diff
@@ -507,14 +507,19 @@ def context_template(
     initialEvents,
     clientStorage,
   )
-  // Publish the dispatchers so JSX literals constructed outside the
-  // React-tree path (e.g. ``ErrorBoundary.onError``) can call ``addEvents``.
+  // Populate the module-level dispatchers so JSX literals constructed
+  // outside the React-tree path (e.g. ``ErrorBoundary.onError``) can call
+  // ``addEvents`` without needing the events hook hoisted in their scope.
```
FarhanAliRaza: This comment rewrite is not related to the memoization. Please restore the original two-line comment to keep the diff minimal.

packages/reflex-base/src/reflex_base/.templates/web/utils/react-theme.js
masenf: Removed the outer element `useMemo` in 47f6c78fa; `ThemeProvider` now returns `createElement(...)` directly. The context values and callbacks remain memoized. All 286 compiler tests, the appearance browser test (including system-preference changes), and all pre-commit hooks pass.

tests/units/compiler/test_react_theme_js_template.py
masenf: Removed that assertion in 47f6c78fa along with the redundant outer memo. Also corrected the nearby test descriptions to describe context-consumer invalidation rather than a subtree cascade.

packages/reflex-base/news/6180.performance.md
masenf: Replaced the fragment with your suggested two sentences in 47f6c78fa. The PR description and affected test docstrings now describe the benefit to context consumers without claiming a subtree cascade.

packages/reflex-base/src/reflex_base/compiler/templates.py
masenf: Restored the original two-line comment verbatim in 47f6c78fa.

FarhanAliRaza (approved):

PR #6181, opened 2026-03-16
Create separate context providing components for each state

masenf:
The generated `StateProvider` currently owns every substate reducer, so each delta re-renders that combined provider and recreates the provider elements for all substates. Give each substate its own reducer and provider so a delta only re-renders the providers it touches. Consumer updates continue to depend on their context values.

Dispatchers register into a stable shared registry during the layout phase, before the event loop connects the WebSocket. `DispatchProvider`'s memo depends only on `children`, since the registry ref has a stable identity.

Validation: 282 compiler tests, all eight router browser cases across development and production modes, and all pre-commit hooks pass. The new browser case extends the existing router app: its initial on-load event updates two sibling substates, and the test verifies both rendered values plus a single received delta containing both updates. It replaces the source-string test that did not establish runtime ordering.


masenf:
The final CodSpeed check flagged `test_var_access[mutable_dict]` (3.22% lower efficiency) with a different-runtime-environment warning. This PR changes the generated state-provider template; the measured Python mutable-proxy code is unchanged. A matched local comparison gave 1.472 ms on main and 1.492 ms here (about 1.36%, below the reported change). These short wall-clock samples cannot establish a causal regression. I have rerun the performance job and am keeping unrelated proxy work out of this PR.

packages/reflex-base/news/6181.performance.md
```diff
@@ -0,0 +1 @@
+Give each substate its own React context provider in the generated `utils/context.js`. A state delta now re-renders only the provider for the substate it touched, instead of recreating every `StateContexts` element under a single combined provider. Dispatchers register into a stable shared registry during the layout phase, so they are in place before the event loop connects the websocket.
```
FarhanAliRaza: Please keep only the first two sentences. The last sentence (dispatcher registry, layout phase) is an implementation detail. The changelog guidance asks for what changed for a downstream user; the PR holds the rest.

tests/units/compiler/test_compiler.py
```diff
@@ -1669,6 +1669,68 @@ def test_context_template_owner_stack_pin(disable_owner_stacks: bool):
...
+        initial_state={"reflex___state____state": {}},
+        state_name="reflex___state____state",
+    )
+
+    assert "useIsomorphicLayoutEffect(() => {" in rendered
+    assert "dispatchers[substateName] = dispatchSubstate;" in rendered
+    assert "delete dispatchers[substateName];" in rendered
+    # ``useLayoutEffect`` warns during SSR, where no effect runs at all.
+    assert (
+        'const useIsomorphicLayoutEffect =\n  typeof document !== "undefined" '
```
FarhanAliRaza: This assert depends on the exact whitespace of the generated source, so a reformat of the template breaks it. The test also does not prove the ordering that its docstring describes: it only checks that the strings exist. A Playwright case in an existing integration app, with a delta that names two or more substates right after connect, would cover the runtime behavior. I checked that behavior by hand and it works.

packages/reflex-base/src/reflex_base/compiler/templates.py
```diff
@@ -518,19 +508,58 @@ def context_template(
...
+// Holds the mutable substate -> dispatch registry that ``SubstateProvider``
+// writes into and ``EventLoopProvider`` reads. The registry object identity is
+// stable for the lifetime of the tree, so neither adding a dispatcher nor
+// updating a substate re-renders the consumers of ``DispatchContext``.
+const DispatchProvider = ({{ children }}) => {{
+  const dispatchers = useRef({{}});
+  return useMemo(
+    () =>
+      createElement(DispatchContext, {{ value: dispatchers.current }}, children),
+    [children, dispatchers],
```
FarhanAliRaza: `dispatchers` is a ref object, so its identity never changes. `[children]` is enough here.

FarhanAliRaza (commented):
I tested this PR in a real app against a `main` baseline. The two builds differ only in `templates.py`.

The app has a root state, sibling substates, a nested substate, a computed var, an event that changes two substates through `get_state`, a generator event, and two pages with `on_load`. I added 300 generated substates to make the provider cost visible. A React commit hook counted the components that rendered in each commit and recorded the commit duration.

On `main`, each delta renders `StateProvider` and the one consumer. With this PR, each delta renders one `SubstateProvider` and the one consumer. Consumer renders are the same on both builds. With 310 substates, the median commit duration goes from 1.1 ms to 0.2 ms over 200 clicks.

I also exercised a two-substate delta, a 50-click burst, client navigation between the pages, a hard load on the second page, and a dev hot reload that adds a state var. All deltas dispatched correctly and the browser console stayed clean. A frontend export succeeds.

The requested changes are in the inline comments.

packages/reflex-base/news/6181.performance.md
masenf: Kept only the first two sentences in ca3d12153. The dispatcher-registry and layout-phase explanation now lives in the PR description.

tests/units/compiler/test_compiler.py
masenf: Replaced the source-string ordering test in ca3d12153 with a Playwright case in the existing RouterQueryApp. Its first on-load handler updates RouterQueryState and a sibling ConnectedState in one event. The test waits for both rendered values and decodes the received Socket.IO packets to verify that a single initial delta contains both substate updates. All eight router cases pass across dev and production, along with 282 compiler tests and every pre-commit hook.

packages/reflex-base/src/reflex_base/compiler/templates.py
masenf: Removed the stable ref from the dependency array in ca3d12153; the memo now depends only on `[children]`.

packages/reflex-base/news/6181.performance.md
```diff
@@ -0,0 +1 @@
+Give each substate its own React context provider in the generated `utils/context.js`. A state delta now re-renders only the provider for the substate it touched, instead of recreating every `StateContexts` element under a single combined provider.
```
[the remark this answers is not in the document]
masenf: Rewritten in 8b23c9add as: “Reduce browser rendering overhead for state updates, especially in apps with many substates.” This describes the downstream benefit without generated paths or provider internals, and avoids claiming fewer consumer re-renders. Changelog hooks pass.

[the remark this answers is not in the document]
masenf: Addressed in 8b23c9add by removing the generated-file reference entirely. The fragment now describes the user-facing performance benefit, which also resolves the incorrect extension without retaining implementation details in the changelog.

PR #6188, opened 2026-03-17
Support FunctionVar handlers in EventChain rendering

### LINKED ISSUE #6185: Fix EventChain to separately invoke addEvents for mixed event types
## Background: Event Types in Reflex

There are three kinds of events that can be bound to a component's event trigger (e.g. `on_click`):

1. **EventHandler** — a reference to an `@rx.event` decorated function defined in an `rx.State` subclass. When bound to a trigger, this generates JS code that adds the event (mapping any trigger arguments) to the frontend queue to be sent to the server over the websocket and processed on the backend.

2. **Special frontend events** — built-in events like `_call_script`, `_console_log`, etc. These also generate JS code that adds the event to the frontend event queue, but they are handled in a special frontend path without going to the backend.

3. **EventChainVar** — an escape hatch where a `Var` (representing some expression in the compiled JS) is passed directly as an event trigger. It gets executed inline on the frontend, bypassing all Reflex event loop machinery.

Types 1 and 2 can be "chained" by passing them together in a list to an event trigger prop. Type 3 is special and currently **can only be passed by itself**.

## Problem

Currently, it's not possible to mix normal state event handlers with custom `FunctionVar` event handlers in an event chain:

```python
import reflex as rx

class State(rx.State):
    @rx.event
    def do_a_thing(self):
        print("Doing a thing!")

log_after_timeout = rx.vars.FunctionStringVar.create(
    "(...args) => { setTimeout(() => console.log('Timeout reached!', args), 1000); }"
)

def index():
    return rx.button(
        "Do both",
        on_click=[
            State.do_a_thing,       # EventHandler — needs addEvents()
            log_after_timeout,       # FunctionVar — should be called directly
        ]
    )
```

This fails because `LiteralEventChainVar.create()` tries to pass everything through a single `addEvents([ ... ], args, actions)` call. `FunctionVar` items don't serialize as `ReflexEvent(...)` objects, so they can't go through `addEvents`.

The current workaround is `rx.call_function(log_after_timeout)`, but this is a cludge — it still generates code that routes through `addEvents`.

## Goal

Make EventChain generate a JS function body that handles each item appropriately: `addEvents()` for `EventHandler`/`EventSpec` items, and direct function calls for `FunctionVar` items. The generated code should be a single function that executes each step in sequence.

Currently there are type checks that require a Var to be an `EventChainVar`, but this should be relaxed to allow any `FunctionVar` to be used directly in the chain.

## Example Expected Output

For the example above, the generated JS should look something like:

```javascript
(_event) => {
    addEvents([ReflexEvent("state.do_a_thing", {})], _event, {});
    ((...args) => { setTimeout(() => console.log('Timeout reached!', args), 1000); })(_event);
}
```

Rather than trying to pack both into a single `addEvents([...])` array.

## Acceptance Criteria

- [ ] `LiteralEventChainVar.create()` groups consecutive `EventHandler`/`EventSpec` events into `addEvents()` (`EventChain.invocation`) calls and invokes `FunctionVar`/`EventVar` events as direct function calls
- [ ] The generated JS function body is a sequence of statements (not a single expression), using `explicit_return` or a block body
- [ ] Mixed chains work: `on_click=[State.handler, some_function_var, State.other_handler]` produces 3 separate invocations (addEvents, direct call, addEvents)
- [ ] Pure EventHandler/EventSpec chains (the common case) still produce a single `addEvents()` call — no regression
- [ ] Pure FunctionVar chains produce direct calls without `addEvents`
- [ ] Event actions (preventDefault, stopPropagation) are applied correctly regardless of mix
- [ ] Unit tests covering: pure EventSpec chain, pure FunctionVar chain, mixed chain, mixed chain with event actions

## Key Files

| File | Purpose |
|------|---------|
| `reflex/event.py:2041-2113` | `LiteralEventChainVar.create()` — this is the main code to modify. Currently builds a single `invocation.call(events_array, args, actions)` |
| `reflex/event.py:449-555` | `EventChain.create()` — constructs the EventChain from mixed input types |
| `reflex/vars/function.py:348-403` | `ArgsFunctionOperation` — used to build the JS arrow function. May need `explicit_return=True` or a block-body variant |
| `reflex/vars/function.py:410-459` | `ArgsFunctionOperationBuilder` — builder pattern variant, base of `LiteralEventChainVar` |
| `reflex/constants/compiler.py` | `CompileVars.ADD_EVENTS`, `Hooks.EVENTS`, `Imports.EVENTS` |
| `reflex/utils/format.py` | `format_prop()` — where EventChain is converted to a Var for rendering |

## Notes

- `EventChain.events` is typed as `Sequence[EventSpec | EventVar | EventCallback]` — you'll need to distinguish which items are "addEvents-compatible" (EventSpec/EventHandler/EventVar) vs "direct call" (FunctionVar)
- The `ArgsFunctionOperation` supports `explicit_return=True` which generates `{ return expr }` blocks — you may need a multi-statement block body instead
- Look at how `rx.call_function` currently wraps FunctionVar to understand the existing workaround

FarhanAliRaza:
Allow EventChain to accept frontend FunctionVar handlers alongside EventSpec, EventVar, and EventCallback values.

When a chain contains FunctionVars, keep backend events grouped through addEvents(...) and invoke frontend functions inline with the trigger arguments so mixed chains preserve execution order and DOM event actions like preventDefault and stopPropagation.

Wrap inline arrow functions before emitting VarOperationCall JS so direct invocation renders valid JavaScript, add unit coverage for pure/mixed event-chain formatting and creation, and move upload exception docs to the helper that actually raises them to satisfy darglint.

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

closes #6185

### DIFF
--- a/reflex/event.py
+++ b/reflex/event.py
@@ -1,5 +1,7 @@
 """Define event classes to connect the frontend and backend."""
 
+from __future__ import annotations
+
 import dataclasses
 import inspect
 import sys
@@ -237,7 +239,7 @@ class EventHandler(EventActionsMixin):
         """
         return getattr(self.fn, BACKGROUND_TASK_MARKER, False)
 
-    def __call__(self, *args: Any, **kwargs: Any) -> "EventSpec":
+    def __call__(self, *args: Any, **kwargs: Any) -> EventSpec:
         """Pass arguments to the handler to get an event spec.
 
         This method configures event handlers that take in arguments.
@@ -345,7 +347,7 @@ class EventSpec(EventActionsMixin):
         object.__setattr__(self, "client_handler_name", client_handler_name)
         object.__setattr__(self, "args", args or ())
 
-    def with_args(self, args: tuple[tuple[Var, Var], ...]) -> "EventSpec":
+    def with_args(self, args: tuple[tuple[Var, Var], ...]) -> EventSpec:
         """Copy the event spec, with updated args.
 
         Args:
@@ -361,7 +363,7 @@ class EventSpec(EventActionsMixin):
             event_actions=self.event_actions.copy(),
         )
 
-    def add_args(self, *args: Var) -> "EventSpec":
+    def add_args(self, *args: Var) -> EventSpec:
         """Add arguments to the event spec.
 
         Args:
@@ -452,7 +454,7 @@ class CallableEventSpec(EventSpec):
 class EventChain(EventActionsMixin):
     """Container for a chain of events that will be executed in order."""
 
-    events: Sequence["EventSpec | EventVar | FunctionVar | EventCallback"] = (
+    events: Sequence[EventSpec | EventVar | FunctionVar | EventCallback] = (
         dataclasses.field(default_factory=list)
     )
 
@@ -463,11 +465,11 @@ class EventChain(EventActionsMixin):
     @classmethod
     def create(
         cls,
-        value: "EventType",
+        value: EventType,
         args_spec: ArgsSpec | Sequence[ArgsSpec],
         key: str | None = None,
         **event_chain_kwargs,
-    ) -> "EventChain | Var":
+    ) -> EventChain | Var:
         """Create an event chain from a variety of input types.
 
         Args:
@@ -1312,7 +1314,7 @@ def download(
 
 def call_script(
     javascript_code: str | Var[str],
-    callback: "EventType[Any] | None" = None,
+    callback: EventType[Any] | None = None,
 ) -> EventSpec:
     """Create an event handler that executes arbitrary javascript code.
 
@@ -1353,7 +1355,7 @@ def call_script(
 
 def call_function(
     javascript_code: str | Var,
-    callback: "EventType[Any] | None" = None,
+    callback: EventType[Any] | None = None,
 ) -> EventSpec:
     """Create an event handler that executes arbitrary javascript code.
 
@@ -1389,7 +1391,7 @@ def call_function(
 
 def run_script(
     javascript_code: str | Var,
-    callback: "EventType[Any] | None" = None,
+    callback: EventType[Any] | None = None,
 ) -> EventSpec:
     """Create an event handler that executes arbitrary javascript code.
 
@@ -1407,7 +1409,7 @@ def run_script(
     return call_function(ArgsFunctionOperation.create((), javascript_code), callback)
 
 
-def get_event(state: "BaseState", event: str):
+def get_event(state: BaseState, event: str):
     """Get the event from the given state.
 
     Args:
@@ -1420,7 +1422,7 @@ def get_event(state: "BaseState", event: str):
     return f"{state.get_name()}.{event}"
 
 
-def get_hydrate_event(state: "BaseState") -> str:
+def get_hydrate_event(state: BaseState) -> str:
     """Get the name of the hydrate event for the state.
 
     Args:
@@ -1773,11 +1775,11 @@ def call_event_fn(
     fn: Callable,
     arg_spec: ArgsSpec | Sequence[ArgsSpec],
     key: str | None = None,
-) -> list[EventSpec | FunctionVar]:
+) -> list[EventSpec | FunctionVar | EventVar]:
     """Call a function to a list of event specs.
 
-    The function should return a single EventSpec, a list of EventSpecs, or a
-    single Var.
+    The function should return a single event-like value or a heterogeneous
+    sequence of event-like values.
 
     Args:
         fn: The function to call.
@@ -1785,7 +1787,7 @@ def call_event_fn(
         key: The key to pass to the event handler.
 
     Returns:
-        The event specs from calling the function or a Var.
+        The event-like values from calling the function.
 
     Raises:
         EventHandlerValueError: If the lambda returns an unusable value.
@@ -1806,9 +1808,9 @@ def call_event_fn(
     # Call the function with the parsed args.
     out = fn(*[*parsed_args][:number_of_fn_args])
 
-    # Convert the output to a list.
-    if not isinstance(out, list):
-        out = [out]
+    # Normalize common heterogeneous event collections into individual events
+    # while keeping other scalar values for validation below.
+    out = list(out) if isinstance(out, (list, tuple)) else [out]
 
     # Convert any event specs to event specs.
     events = []
@@ -1828,7 +1830,8 @@ def call_event_fn(
                 hint = " Hint: use `fn.partial(...)` instead of calling the FunctionVar directly."
             msg = (
                 f"Invalid event chain for {key}: {fn} -> {e}: A lambda inside an EventChain "
-                "list must return `EventSpec | EventHandler | FunctionVar` or heterogeneous list of these types. "
+                "list must return `EventSpec | EventHandler | EventChain | EventVar | FunctionVar` "
+                "or a heterogeneous sequence of these types. "
                 f"Got: {type(e)}.{hint}"
             )
             raise EventHandlerValueError(msg)
@@ -1979,7 +1982,7 @@ class LiteralEventVar(VarOperationCall, LiteralVar, EventVar):
         cls,
         value: EventSpec | EventHandler,
         _var_data: VarData | None = None,
-    ) -> "LiteralEventVar":
+    ) -> LiteralEventVar:
         """Create a new LiteralEventVar instance.
 
         Args:
@@ -2066,7 +2069,7 @@ class LiteralEventChainVar(ArgsFunctionOperationBuilder, LiteralVar, EventChainV
         cls,
         value: EventChain,
         _var_data: VarData | None = None,
-    ) -> "LiteralEventChainVar":
+    ) -> LiteralEventChainVar:
         """Create a new LiteralEventChainVar instance.
 
         Args:
@@ -2189,39 +2192,39 @@ class EventCallback(Generic[Unpack[P]], EventActionsMixin):
 
     @overload
     def __call__(
-        self: "EventCallback[Unpack[Q]]",
-    ) -> "EventCallback[Unpack[Q]]": ...
+        self: EventCallback[Unpack[Q]],
+    ) -> EventCallback[Unpack[Q]]: ...
 
     @overload
     def __call__(
-        self: "EventCallback[V, Unpack[Q]]", value: V | Var[V]
-    ) -> "EventCallback[Unpack[Q]]": ...
+        self: EventCallback[V, Unpack[Q]], value: V | Var[V]
+    ) -> EventCallback[Unpack[Q]]: ...
 
     @overload
     def __call__(
-        self: "EventCallback[V, V2, Unpack[Q]]",
+        self: EventCallback[V, V2, Unpack[Q]],
         value: V | Var[V],
         value2: V2 | Var[V2],
-    ) -> "EventCallback[Unpack[Q]]": ...
+    ) -> EventCallback[Unpack[Q]]: ...
 
     @overload
     def __call__(
-        self: "EventCallback[V, V2, V3, Unpack[Q]]",
+        self: EventCallback[V, V2, V3, Unpack[Q]],
         value: V | Var[V],
         value2: V2 | Var[V2],
         value3: V3 | Var[V3],
-    ) -> "EventCallback[Unpack[Q]]": ...
+    ) -> EventCallback[Unpack[Q]]: ...
 
     @overload
     def __call__(
-        self: "EventCallback[V, V2, V3, V4, Unpack[Q]]",
+        self: EventCallback[V, V2, V3, V4, Unpack[Q]],
         value: V | Var[V],
         value2: V2 | Var[V2],
         value3: V3 | Var[V3],
         value4: V4 | Var[V4],
-    ) -> "EventCallback[Unpack[Q]]": ...
+    ) -> EventCallback[Unpack[Q]]: ...
 
-    def __call__(self, *values) -> "EventCallback":  # pyright: ignore [reportInconsistentOverload]
+    def __call__(self, *values) -> EventCallback:  # pyright: ignore [reportInconsistentOverload]
         """Call the function with the values.
 
         Args:
@@ -2234,11 +2237,11 @@ class EventCallback(Generic[Unpack[P]], EventActionsMixin):
 
     @overload
     def __get__(
-        self: "EventCallback[Unpack[P]]", instance: None, owner: Any
-    ) -> "EventCallback[Unpack[P]]": ...
+        self: EventCallback[Unpack[P]], instance: None, owner: Any
+    ) -> EventCallback[Unpack[P]]: ...
 
     @overload
-    def __get__(self, instance: Any, owner: Any) -> "Callable[[Unpack[P]]]": ...
+    def __get__(self, instance: Any, owner: Any) -> Callable[[Unpack[P]]]: ...
 
     def __get__(self, instance: Any, owner: Any) -> Callable:
         """Get the function with the instance bound to it.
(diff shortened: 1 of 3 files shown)


FarhanAliRaza:
Does it require documentation?

reflex/vars/function.py
```diff
@@ -239,7 +317,13 @@ def _cached_var_name(self) -> str:
         Returns:
             The name of the var.
         """
-        return f"({self._func!s}({', '.join([str(LiteralVar.create(arg)) for arg in self._args])}))"
+        func_expr = str(self._func)
+        if _starts_with_arrow_function(func_expr) and not format.is_wrapped(
```
masenf, line 321: do we really need the complex logic to detect an arrow function or can we just unconditionally wrap the expression in parentheses if it's not already wrapped

reflex/event.py
```diff
@@ -2098,17 +2120,80 @@ def create(
...
+                if value.event_actions.get("preventDefault"):
+                    statement_js.append(
+                        "if (_reflex_dom_event?.preventDefault) "
+                        "{_reflex_dom_event.preventDefault();}"
+                    )
+                if value.event_actions.get("stopPropagation"):
+                    statement_js.append(
+                        "if (_reflex_dom_event?.stopPropagation) "
+                        "{_reflex_dom_event.stopPropagation();}"
+                    )
```
masenf: code for handling the event actions (with the addition of throttle and debounce) already exists in `state.js`, but it's inside the `addEvents` functions. i would suggest pulling this logic out as an exported function in `state.js` that takes the event arg and a function which could either be an `addEvents` partial or some user-passed function.

instead of collecting all the event actions and combining them into the EventChain, it probably is better structure to leave the event actions on the individual EventSpec, and apply any further event actions to the event chain itself. Then use the same function to process these actions at each level.

For example, you might have `applyEventActions(target, event_actions, ...args)`; in this case the `target` could be the entire EventChain function if the chain has its own actions, or the actions could be passed to each `addEvents` call (which internally would call `applyEventActions`).

reflex/event.py
```diff
@@ -482,6 +484,20 @@ def create(
...
+                return value
+            if isinstance(value, FunctionVar):
+                if event_chain_kwargs:
+                    warnings.warn(
+                        f"event_chain_kwargs {event_chain_kwargs!r} are ignored for "
+                        "FunctionVar values.",
+                        stacklevel=2,
+                    )
                 return value
             if isinstance(value, EventVar):
```
masenf: include FunctionVar here and remove the special case for it above. We should now be treating EventVar and FunctionVar in the same way: creating a new event chain that includes these values in the `events` list so they can be composed and formatted as an EventChainVar later.

reflex/event.py
```diff
@@ -2098,17 +2120,80 @@ def create(
         if invocation is not None and not isinstance(invocation, FunctionVar):
             msg = f"EventChain invocation must be a FunctionVar, got {invocation!s} of type {invocation._var_type!s}."
             raise ValueError(msg)
+        assert invocation is not None
+
+        has_function_var = any(isinstance(e, FunctionVar) for e in value.events)
+
+        if not has_function_var:
```
masenf: we should strive to reduce special cases in this function. i think it's absolutely fine if we process both mixed and homogenous cases with the same logic.

it's actually better if we have a separate `addEvents` call for each EventSpec because then each of those can have different event actions attached to them and the system is more composable.

reflex/event.py
```diff
@@ -2098,17 +2120,80 @@ def create(
...
+                queueable_group.clear()
+
+            for event in value.events:
+                if isinstance(event, FunctionVar):
+                    flush_queueable_group()
+                    function_call = event.call(*call_args)
+                    statement_js.append(f"{function_call!s};")
+                    statement_var_data.append(function_call._get_all_var_data())
+                else:
+                    queueable_group.append(event)
```
masenf: just to reiterate, i don't think we need or want to batch up these groups; each non-FunctionVar should get its own `addEvents` invocation.

tests/units/test_app.py
```diff
@@ -1719,23 +1719,40 @@ def test_app_wrap_compile_theme(
             "export function Layout"
         )
     ].strip()
+
+    def _wrap_event_handler_calls(handler_js: str) -> str:
```
masenf: seems like it might be better to just make these replacements in the asserted string

reflex/event.py
```diff
@@ -483,24 +482,24 @@ def create(
         """
         # If it's an event chain var, return it.
         if isinstance(value, Var):
-            if isinstance(value, EventChainVar):
+            if isinstance(value, LiteralEventChainVar):
```
masenf, line 492: is there a reason this didn't stay as `EventChainVar`? seems a bit redundant with the extra case added below

reflex/vars/function.py
FarhanAliRaza: It breaks the integration test without this.

FarhanAliRaza: It emitts invalid code like `((typeof)(value))`

FarhanAliRaza:
The failing test is flaky
It can be fixed like this but i dont know if it is right decision.

```
diff --git a/reflex/istate/manager/redis.py b/reflex/istate/manager/redis.py
index 63fc9058..e59d57a8 100644
--- a/reflex/istate/manager/redis.py
+++ b/reflex/istate/manager/redis.py
@@ -153,6 +153,10 @@ class StateManagerRedis(StateManager):
         default_factory=dict,
         init=False,
     )
+    _lock_updates_subscribed: asyncio.Event = dataclasses.field(
+        default_factory=asyncio.Event,
+        init=False,
+    )
     _lock_task: asyncio.Task | None = dataclasses.field(default=None, init=False)
 
     # Whether debug prints are enabled.
@@ -797,8 +801,12 @@ class StateManagerRedis(StateManager):
         }
         async with self.redis.pubsub() as pubsub:
             await pubsub.psubscribe(**handlers)  # pyright: ignore[reportArgumentType]
-            async for _ in pubsub.listen():
-                pass
+            self._lock_updates_subscribed.set()
+            try:
+                async for _ in pubsub.listen():
+                    pass
+            finally:
+                self._lock_updates_subscribed.clear()
 
     def _ensure_lock_task(self) -> None:
         """Ensure the lock updates subscriber task is running."""
@@ -954,6 +962,9 @@ class StateManagerRedis(StateManager):
             lock_id: The ID of the lock.
         """
         token = lock_key.decode().rsplit("_lock", 1)[0]
+        if self._oplock_enabled:
+            self._ensure_lock_task()
+            await self._lock_updates_subscribed.wait()
         if (
             # If there's not a line, try to get the lock immediately.
             not self._n_lock_waiters(lock_key)
@@ -964,8 +975,6 @@ class StateManagerRedis(StateManager):
                     f"{SMR} [{time.monotonic() - start:.3f}] {lock_key.decode()} instaque by {lock_id.decode()}"
                 )
             return
-        # Make sure lock waiter task is running.
-        self._ensure_lock_task()
         async with (
             self._lock_waiter(lock_key) as lock_released_event,
             self._request_lock_release(lock_key, lock_id),

```

tests/units/test_app.py
masenf: seems this function can be removed?

reflex/event.py
```diff
@@ -2098,17 +2118,71 @@ def create(
...
+                _args=FunctionArgs(arg_def),
+                _return_expr=invocation.call(
+                    LiteralVar.create([
+                        LiteralVar.create(event) for event in value.events
+                    ]),
+                    arg_def_expr,
+                    value.event_actions,
+                ),
+                _var_value=value,
+            )
```
masenf: this special case creates a behavior difference when two EventSpec/EventHandler have different event actions:

```python
import reflex as rx


class State(rx.State):
    @rx.event
    def do_a_thing(self, value: str):
        print(f"Doing a thing! {value}")


log_after_timeout = rx.vars.FunctionStringVar.create("(...args) => { setTimeout(() => console.log('Timeout reached!', args), 1000); }")

def index():
    return rx.vstack(
        rx.button(
            "Do single debounce",
            on_click=[
                State.do_a_thing("first x 1000").debounce(1000),
            ]
        ),
        rx.button(
            "Do two debounce",
            on_click=[
                State.do_a_thing("first x 1000").debounce(1000),
                State.do_a_thing("second x 200").debounce(200),
            ]
        ),
        rx.button(
            "Do two debounce with log",
            on_click=[
                State.do_a_thing("first x 1000").debounce(1000),
                State.do_a_thing("second x 200").debounce(200),
                log_after_timeout.partial("arbitrary args", "can be passed"),
                log_after_timeout,
            ]
        ),
    )

app = rx.App()
app.add_page(index)
```

in the first case, it works as expected, since there is only one debounce action.

in the second case, the debounce actions get rolled up together in the `addEvents` function, which results in a single debounce being called for 200ms. this is a long standing issue we've had in the framework with the composability of event actions.

in the third case, the debounce actions are each applied separately because separate `addEvents` calls are used.

i think if we just get rid of this special case wherein we combine non-FunctionVar events, then we finally fix this bug and allow each event spec to have its own event actions.

i do like how the EventChain actions are applied over the entire function if the chain has its own actions. Currently this isn't really exposed in the framework because the typing does not allow you to directly pass an EventChain to an event trigger prop, but likely in the future we will allow this.

masenf:
> The failing test is flaky It can be fixed like this but i dont know if it is right decision.

made the fix in https://github.com/reflex-dev/reflex/pull/6196

you can merge from `main` into your branch to bring it in

reflex/event.py
FarhanAliRaza:
```python

def make_timeout_logger():
    return rx.vars.FunctionStringVar.create(
        "(...args) => { setTimeout(() => console.log('Timeout reached!', args), 1000); }"
    ).to(EventChain)

def test_event_chain_create_wraps_event_chain_typed_function_var_kwargs():
    """FunctionVars cast to EventChain should still compose with chain-level kwargs."""
    frontend_handler = make_timeout_logger()

    chain = EventChain.create(
        frontend_handler,
        args_spec=lambda: (),
        event_actions={"preventDefault": True},
    )

    assert isinstance(chain, EventChain)
    assert chain.events == [frontend_handler]
    assert chain.event_actions == {"preventDefault": True}
```

This case fails without this. `EventChainVar` if is redundant, so I removed that if here https://github.com/reflex-dev/reflex/pull/6188/commits/1d7e57e420c73264aa77a3fd560eec44bd586fcc

masenf: okay sounds good. appreciate the code comment for future readers

tests/units/test_event.py
```diff
@@ -745,23 +760,52 @@ def do_a_thing(self):
...
         args_spec=lambda: (),
-        event_actions={"preventDefault": True},
     )
     rendered = str(LiteralVar.create(chain))
 
     assert "applyEventActions(" not in rendered
-    assert rendered.count("addEvents(") == 1
+    assert rendered.count("addEvents(") == 2
+    assert '["debounce"] : 1000' in rendered
+    assert '["debounce"] : 200' in rendered
```
masenf: probably also assert that these each appear once in the rendered output

tests/units/test_event.py
```diff
@@ -745,23 +760,52 @@ def do_a_thing(self):
...
+        args_spec=lambda: (),
+        event_actions={"preventDefault": True},
+    )
+    rendered = str(LiteralVar.create(chain))
+
+    assert "applyEventActions(" in rendered
+    assert rendered.count("addEvents(") == 2
+    assert '["debounce"] : 1000' in rendered
+    assert '["debounce"] : 200' in rendered
     assert '["preventDefault"] : true' in rendered
```
masenf: similarly here, lets assert that each of these action strings appear exactly once in the code.

masenf:
@FarhanAliRaza I added some commits to catch some additional cases that were missing while i was testing manually.

Specifically code like this:

```python
        rx.form(
            rx.input(
                name="input_value",
                on_blur=lambda v: rx.EventChain.create(
                    value=log_after_timeout.partial(v.to(str).capitalize()),
                    args_spec=lambda: (),
                ),
            ),
            rx.button(
                "Submit",
            ),
            on_submit=[
                log_after_timeout,
                lambda fd: [
                    log_after_timeout.partial(fd.to(dict)["input_value"]),
                    rx.console_log(fd.to(dict)["input_value"]),
                ],
                State.on_submit,
            ],
        ),
```

2 behavior changes:
- When a FunctionVar is passed, the arguments dictated by the args_spec for the event trigger must be applied as a `.partial` so the handler gets the transformed args 
- When a lambda is passed in the chain, it should be able to return the same types of heterogeneous values that are accepted by the chain itself.

Additionally, lambda return value validation was consolidated into `call_event_fn`.

Can you explore more edge cases around this and add some additional test cases to the unit tests suite.

masenf:
test failure was from another interesting case that we don't have a unit test for yet

```python
            on_key_down=lambda key: rx.cond(
                rx.Var.create(["a", "s", "d", "w"]).contains(key),
                GlobalHotkeyState.update_key(key),
                rx.console_log(key)
            )
```

here the event handler is a lambda that returns some conditional code that runs on the frontend before deciding whether to trigger a backend event or do something else.

Stuff like this still doesn't work though

```python
            rx.input(
                name="input_value",
                on_blur=lambda v: rx.cond(
                    v == "foo",
                    log_after_timeout.partial("Input was foo!"),
                    State.do_a_thing(v),
                ),
            ),
```

When we have an `rx.cond` that returns heterogenous values (either FunctionVar or EventVar), then that gets rejected by the new lambda return type checking logic. Idk if we want to allow that to return any arbitrary var, or if requiring the user to say `rx.cond(...).to(rx.event.EventVar)` is good enough for now. I'm probably more inclined to the latter.

FarhanAliRaza:
I fixed it locally by adding this.

Is it correct?


```python

 elif isinstance(e, Var) and safe_issubclass(
            e._var_type, (EventSpec, EventChain)
        ):
            # Helpers like rx.cond() can return event-typed Vars that should be
            # treated the same as literal EventSpec/EventChain values here.
            e = e.guess_type()
            
```


Update:
But now that integeration test passes so not pushing this.

masenf:
> When we have an rx.cond that returns heterogenous values (either FunctionVar or EventVar), then that gets rejected by the new lambda return type checking logic. Idk if we want to allow that to return any arbitrary var, or if requiring the user to say rx.cond(...).to(rx.event.EventVar) is good enough for now. I'm probably more inclined to the latter.

I think we have some special logic somewhere that is treating the `rx.cond` case specially. Even with the explicit cast to EventVar, i'm not able to get it to work because it wants to treat the whole thing as an EventVar (and thus wrap it in `addEvents`), which just doesn't work if one of the branches of the `rx.cond` is not a real EventSpec.

I think we can just defer on this problem for now; it wasn't working before and still doesn't work now. We could improve it in the future.

reflex/event.py
```diff
@@ -2098,17 +2108,55 @@ def create(
...
+                    {},
+                )
+            )
+            for event in value.events
+        ]
+
+        if not statements:
+            statements.append(invocation.call(LiteralVar.create([]), arg_def_expr, {}))
+
+        statement_var_data = [statement._get_all_var_data() for statement in statements]
```
masenf: i think this might be the line that is causing the CodSpeed regression.

usually if you just combine the vars, the parser will pick up any nested VarData automatically:

```python
            statement_block = Var(
                _js_expr=f"{{{''.join(f'{statement};' for statement in statements)}}}",
            )
```

i would expect the above to be equivalent to how it was written without the explicit VarData merge

FarhanAliRaza: You were right, it is fixed now.

masenf:
This PR looks good and works well. I would just like to see a couple more test cases:

* Pass FunctionVar with a non-empty args spec should apply the spec arguments partially
  * Lambda function that returns FunctionVar should NOT get arg spec applied
* Lambda function should be able to return sequence of mixed type events
* Lambda function should be able to return EventChain explicitly -- this should go through unmodified

PR #6190, opened 2026-03-18
feat: add chunked file upload support Streaming Upload API (rx.upload_files_chunk)

### LINKED ISSUE #6184: Streaming Upload API (rx.upload_files_chunk)
## Problem

Reflex's current upload API (`rx.upload_files`) buffers the entire file into memory on the server before the event handler runs. For large files (video, datasets, etc.), this causes high memory usage and prevents the developer from processing data as it arrives (e.g., streaming to S3, computing checksums incrementally, showing server-side progress).

## Goal

Add a new streaming upload API that allows event handlers to process file data chunk-by-chunk as it arrives, without buffering the entire file.

## Proposed API

`rx.upload_files_chunk(...)` is used in the same way as `rx.upload_files(...)` — provided as an argument to an event handler passed to the `rx.upload` `on_drop` trigger or a button's `on_click`. The difference is that instead of the handler receiving a `list[rx.UploadFile]`, it receives an `rx.UploadChunkIterator`.

```python
import reflex as rx
from pathlib import Path
from typing import BinaryIO

class UploadState(rx.State):
    files_received: list[str] = []

    @rx.event(background=True)
    async def handle_upload_chunk(self, chunk_iter: rx.UploadChunkIterator):
        """Process uploaded file data as it streams in.

        A single upload request may contain multiple files.
        The iterator yields chunks across all files in the request.
        """
        upload_dir = Path(rx.get_upload_dir())
        file_handles: dict[str, BinaryIO | None] = {}
        local_paths: dict[str, Path] = {}
        try:
            async for chunk in chunk_iter:
                # chunk.filename: str — original file name
                # chunk.offset: int — byte offset of this chunk within the file
                # chunk.content_type: str — MIME type
                # chunk.data: bytes — the chunk payload

                # Resolve local path once per filename.
                if chunk.filename not in local_paths:
                    p = upload_dir / chunk.filename
                    p.parent.mkdir(parents=True, exist_ok=True)
                    local_paths[chunk.filename] = p

                fh = file_handles.get(chunk.filename)
                if chunk.filename not in file_handles:
                    # New file — open a fresh handle.
                    fh = open(local_paths[chunk.filename], "wb")
                    file_handles[chunk.filename] = fh
                elif fh is None:
                    # Previously seen file whose handle was closed — reopen for append.
                    fh = open(local_paths[chunk.filename], "ab")
                    file_handles[chunk.filename] = fh
                fh.seek(chunk.offset)
                fh.write(chunk.data)
        finally:
            for fh in file_handles.values():
                if fh is not None:
                    fh.close()

        async with self:
            self.files_received = list(file_handles.keys())


def index():
    return rx.upload.root(
        id="my_upload",
        on_drop=UploadState.handle_upload_chunk(
            rx.upload_files_chunk(
                upload_id="my_upload",
                on_upload_progress=lambda u: rx.console_log(u),
            )
        ),
    )
```

**Key design points:**

- `rx.upload_files_chunk(...)` works like `rx.upload_files(...)` — it is passed to the upload trigger. It supports the same `upload_id` and `on_upload_progress` arguments
- The handler **must** be a `@rx.event(background=True)` task (enforce this with a validation error at compile time)
- The handler **must** accept an `rx.UploadChunkIterator` argument (async iterator yielding `UploadChunk` objects)
- `UploadChunk` is a dataclass with fields: `filename: str`, `offset: int`, `content_type: str`, `data: bytes`
- A single upload request may contain multiple files — the iterator yields chunks across all files interleaved

**Notes**

- Background events handle their own updates to the frontend via `async with self` mechanism, so the upload handler is not responsible for yielding ndjson updates as it does for the current API.

## Acceptance Criteria

- [ ] New `UploadChunk` dataclass defined and exported from `rx` namespace
- [ ] New `UploadChunkIterator` type defined (async iterator of `UploadChunk`)
- [ ] Backend `/_upload` endpoint (or a new streaming variant) supports chunked/streaming multipart parsing — investigate `streaming_form_data` library or similar
- [ ] Handler validation: raise a clear error if the given event handler is not a background task
- [ ] Handler validation: raise a clear error if no `UploadChunkIterator` parameter is found in the handler signature
- [ ] The frontend upload JS (`reflex/.templates/web/utils/helpers/upload.js`) sends file data in a streaming-compatible way (this may mean chunked uploads via multiple requests, or a streaming multipart body)
- [ ] Integration test demonstrating a large file upload processed chunk-by-chunk
- [ ] Existing `rx.upload_files` API continues to work unchanged (backward compatible)

## Key Files

| File | Purpose |
|------|---------|
| `reflex/app.py:1894-2026` | Current `upload_file()` endpoint — reads entire file via `await request.form()` then `await file.read()`. New streaming path goes here or in a parallel endpoint |
| `reflex/app.py:248-285` | `UploadFile` class wrapping Starlette's UploadFile |
| `reflex/.templates/web/utils/helpers/upload.js` | Frontend upload handler — uses XMLHttpRequest + FormData. May need chunking logic |
| `reflex/components/core/upload.py` | `Upload` component, `upload_file()` / `selected_files()` helpers, `FileUpload` event |
| `reflex/event.py:845-937` | `FileUpload` / `upload_files` event class — `as_event_spec()` creates the client-side upload trigger |
| `reflex/constants/event.py` | `Endpoint.UPLOAD` constant |

## Prior Art / Libraries

- [`streaming_form_data`](https://github.com/siddhantgoel/streaming-form-data) — streaming multipart parser for Python; avoids buffering files into memory
  - See ext-upload.tar.gz for an idea of how this was used with Reflex previously, but don't get hung up on this implementation, as the new proposed API will handle chunking with background events and async iterators instead of calling an `on_chunk` event repeatedly, which would take the costly state lock.

FarhanAliRaza:
Implement chunked/streaming file uploads to handle large files without loading them entirely into memory. Moves upload handling logic from app.py to event.py, adds chunked upload JS helpers, and updates the upload component to support the new upload_files_chunk API. Includes unit and integration tests for chunked upload, cancel, and streaming.

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

closes #6184

### DIFF
--- a/reflex/event.py
+++ b/reflex/event.py
@@ -101,7 +101,7 @@ EVENT_ACTIONS_MARKER = "_rx_event_actions"
 UPLOAD_FILES_CLIENT_HANDLER = "uploadFiles"
 
 
-def _handler_name(handler: "EventHandler") -> str:
+def _handler_name(handler: EventHandler) -> str:
     """Get a stable fully qualified handler name for errors.
 
     Args:
@@ -115,7 +115,7 @@ def _handler_name(handler: "EventHandler") -> str:
     return handler.fn.__qualname__
 
 
-def resolve_upload_handler_param(handler: "EventHandler") -> tuple[str, Any]:
+def resolve_upload_handler_param(handler: EventHandler) -> tuple[str, Any]:
     """Validate and resolve the UploadFile list parameter for a handler.
 
     Args:
@@ -154,7 +154,7 @@ def resolve_upload_handler_param(handler: "EventHandler") -> tuple[str, Any]:
     raise UploadValueError(msg)
 
 
-def resolve_upload_chunk_handler_param(handler: "EventHandler") -> tuple[str, type]:
+def resolve_upload_chunk_handler_param(handler: EventHandler) -> tuple[str, type]:
     """Validate and resolve the UploadChunkIterator parameter for a handler.
 
     Args:


tests/units/components/core/test_upload.py
```diff
@@ -44,10 +72,54 @@ def test_get_upload_url():
...
+def test_upload_files_chunk_export():
+    chunk = rx.UploadChunk(
+        filename="foo.txt",
+        offset=0,
+        content_type="text/plain",
+        data=b"hello",
+    )
+
+    assert chunk.filename == "foo.txt"
+    assert isinstance(rx.UploadChunkIterator(), rx.UploadChunkIterator)
```
masenf: 🤔

tests/units/components/core/test_upload.py
```diff
@@ -83,6 +155,51 @@ def test_upload_create():
     assert isinstance(up_comp_4, Upload)
     assert up_comp_4.is_used
 
+    # reset is_used
+    Upload.is_used = False
+
+    up_comp_5 = Upload.create(
+        id="foo_id",
+        on_drop=StreamingUploadStateTest.chunk_drop_handler(
+            cast(Any, rx.upload_files_chunk(upload_id="foo_id"))
```
masenf: use `# pyright: ignore[...]` comments so if we eventually fix the typing down the road, we can detect "useless type ignore" and remove them

reflex/_upload.py
```diff
@@ -0,0 +1,537 @@
+"""Backend upload helpers and routes for Reflex apps."""
```
masenf, line 1: to match the other non plural names throughout reflex, can this be renamed as `reflex/upload.py`

FarhanAliRaza: Renaming it to `_upload` naming it upload contradict with upload component.

masenf: sounds good. i do want to move more toward framework internals being marked with the leading underscore and only exposing a well vetted public API for our upcoming 1.0 release

FarhanAliRaza:
Don't know why pre-commit is failing.

masenf:
> Don't know why pre-commit is failing.

The pyi_generator script created files that different from the last known hash. i.e. "something" in the pyi output changed and it looks like it was something that most components are inheriting, so probably in the base component class 

EDIT: actually i see that you added UploadFile as default import, so every pyi file got its hash changed. do we need UploadFile as a default import?

FarhanAliRaza:
> EDIT: actually i see that you added UploadFile as default import, so every pyi file got its hash changed. do we need UploadFile as a default import?

No we dont need it in default imports.
But it has another issue it was generating verbose pyi like this 
image

even when we have a top-level import UploadFile in `upload.pyi` it still tries to do an absolute import. but can be made better. 
I fixed this in pyi_generator.

reflex/components/core/upload.py
```diff
@@ -212,7 +218,7 @@ class GhostUpload(Fragment):
     """A ghost upload component."""
 
     # Fired when files are dropped.
-    on_drop: EventHandler[_on_drop_spec]
+    on_drop: EventHandler[_on_drop_args_spec]
 
     # Fired when dropped files do not meet the specified criteria.
     on_drop_rejected: EventHandler[_on_drop_spec]
```
masenf: this is a grey line comment, so i'll probably make it into a separate issue, but i don't think this is the right signature for the reject trigger, it shouldn't have anything to do with the `files` arg and i don't think the js object resembles a `list[rx.UploadFile]`

reflex/_upload.py
```diff
@@ -0,0 +1,572 @@
...
+        self._charset = ""
+        self._current_partial_header_name = b""
+        self._current_partial_header_value = b""
+        self._current_part = _UploadChunkPart()
+        self._chunks_to_emit: deque[UploadChunk] = deque()
+        self._seen_upload_chunk = False
+        self._part_count = 0
+        self._emitted_chunk_count = 0
+        self._emitted_bytes = 0
+        self._stream_chunk_count = 0
```
masenf: this should be a dataclass

reflex/_upload.py
```diff
@@ -0,0 +1,572 @@
...
+
+        Returns:
+            The name of the uploaded file.
+        """
+        if self.path:
+            return self.path.name
+        return None
+
+
+@dataclasses.dataclass
```
masenf:
```suggestion
@dataclasses.dataclass(kw_only=True, slots=True)
```

These flags should be considered for all dataclasses:
1. `kw_only` increases modularity by allowing subclasses to have non-default arguments and not worry about argument ordering if the class is reorganized later
2. `slots` increases performance by allocating a specific place for each field, particularly useful for classes that get instantiated or accessed frequently

i think we could probably add these to the `UploadFile` class as well

reflex/event.py
```diff
@@ -90,6 +92,239 @@ def substate_token(self) -> str:
...
 BACKGROUND_TASK_MARKER = "_reflex_background_task"
 EVENT_ACTIONS_MARKER = "_rx_event_actions"
+UPLOAD_FILES_CLIENT_HANDLER = "uploadFiles"
+
+
+@dataclasses.dataclass(
+    init=True,
+    frozen=True,
+)
+class UploadChunk:
```
masenf:
```suggestion
@dataclasses.dataclass(
    init=True,
    frozen=True,
    kw_only=True,
    slots=True,
)
class UploadChunk:
```

can this class also move to the _upload module?

reflex/event.py
```diff
@@ -90,6 +92,239 @@ def substate_token(self) -> str:
...
+    """A chunk of uploaded file data."""
+
+    filename: str
+    offset: int
+    content_type: str
+    data: bytes
+
+
+class UploadChunkIterator(AsyncIterator[UploadChunk]):
```
masenf: can this also move to `_upload` module, in the process of refactoring `event.py` and don't want to add additional implementations here (changes to existing stuff is fine)

reflex/state.py
```diff
@@ -1895,6 +1942,8 @@ async def _process_event(
         except Exception:
             type_hints = {}
 
+        await _normalize_upload_payload(handler, payload)
```
masenf: i dont really like this added in the hot path that all event processing goes through. where in the frontend are we hardcoding "files"? if that's easy enough to resolve on the frontend or maybe in the upload handler itself if possible that would be better.

reflex/event.py
```diff
@@ -1246,6 +1101,11 @@ def as_event_spec(self, handler: EventHandler) -> EventSpec:
 upload_files_chunk = UploadFilesChunk
 
 
+_upload_module = import_module("reflex._upload")
+UploadChunk = _upload_module.UploadChunk
+UploadChunkIterator = _upload_module.UploadChunkIterator
+
+
```
masenf: i think these can be removed from this module as well.

to expose them at the top level `reflex` namespace, add them to the lazy import spec in `reflex/__init__.py`

FarhanAliRaza: Oh, it remained by mistake.

reflex/components/core/upload.py
masenf: #6205

masenf:
@greptile-apps please do a final re-review

PR #6209, opened 2026-03-24
stateful components

benedikt-bartscher:
introduced in https://github.com/reflex-dev/reflex/pull/6199


masenf (commented):
in #6199 we just stopped calling `_compile_stateful_components` in dev mode entirely.

do you have a workflow where this causes an issue?

masenf:
in the next release (0.9.2), we rewrote the compiler and auto-memoization mechanism. it now outputs each stateful component in its own file under `$/utils/components/` in both dev and prod modes

PR #6260, opened 2026-03-31
Add compiler plugin hooks and plugins and move compilation pipeline out of App

### LINKED ISSUE #6210: Single-pass compiler: Introduce CompilerPlugin protocol, CompilerHooks, and context objects
## Summary

Introduce the foundational types for the new single-pass compiler: `CompilerPlugin` (protocol), `CompilerHooks` (dispatcher), `BaseContext`, `PageContext`, and `CompileContext`. These are the building blocks that all subsequent compiler work depends on.

This is part of the **2a. Single Pass Compiler** roadmap item. The goal is to replace the current multi-pass compilation approach (where the component tree is walked multiple times to extract imports, hooks, custom code, dynamic imports, refs, and app wrap components separately) with a single tree walk using a plugin architecture.

## Background & Motivation

Today, compiling a single page involves **at least 6 separate recursive tree walks** over the same component tree:

1. `_get_all_imports()` — collects imports from every component
2. `_get_all_hooks()` — collects React hooks
3. `_get_all_custom_code()` — collects custom JS code snippets
4. `_get_all_dynamic_imports()` — collects dynamic import statements
5. `_get_all_refs()` — collects refs
6. `_get_all_app_wrap_components()` — collects app wrapper components

Each of these methods in `reflex/components/component.py` follows the same pattern: call the non-recursive version on self, then recursively call on all children, merging results. This is wasteful — a single walk could collect all information in one pass.

Additionally, compilation logic is currently spread across `reflex/app.py` (the `_compile()` method, ~400 lines), `reflex/compiler/compiler.py`, and `reflex/compiler/utils.py`, making it hard to extend or adapt for different targets.

## Design

A working demo of the proposed architecture is available as a starting point (see the `CompilerPlugin` protocol, `CompilerHooks`, `BaseContext`, `PageContext`, `CompileContext` classes in the demo code shared in the project brief). The key ideas:

### `CompilerPlugin` Protocol
```python
class CompilerPlugin(Protocol):
    async def eval_page(self, page_fn, /, **kwargs) -> PageContext | None: ...
    async def compile_page(self, page_ctx, /, **kwargs) -> None: ...
    async def compile_component(self, comp, /, **kwargs) -> AsyncGenerator[...]: ...
```

- `eval_page`: Evaluate a page function into a `PageContext`. Returns `None` if the plugin doesn't handle this page.
- `compile_page`: Perform a transformation on the `PageContext` during page compilation.
- `compile_component`: Two-phase async generator for component transformation. **Pre-yield** (descending): sees component before children are visited. **Post-yield** (ascending): runs after children are visited, receives `(component, children)` tuple.

### `CompilerHooks` Dispatcher
Holds a `tuple[CompilerPlugin, ...]` and dispatches hook calls to all registered plugins in order. Plugin ordering controls priority.

### Context Objects
- `BaseContext`: Async context manager that sets itself as a `ContextVar`, allowing any code in the call stack to access the current context via `ContextClass.get()`.
- `PageContext`: Holds per-page compilation state — `name`, `root_component`, `imports`, `module_code`, etc.
- `CompileContext`: Holds the full compilation state — list of pages, hooks, compiled results.

## Acceptance Criteria

- [ ] `CompilerPlugin` protocol is defined in a new module (e.g. `reflex/compiler/plugins.py` or `reflex/compiler/plugin.py`)
- [ ] `CompilerHooks` dataclass is implemented with `_dispatch`, `eval_page`, `compile_page`, and `compile_component` methods
- [ ] `BaseContext` is implemented with `ContextVar`-based async context management, `get()` classmethod, and `ensure_context_attached()` guard
- [ ] `PageContext` is implemented with fields for: `name`, `route`, `root_component`, `imports` (list of `ParsedImportDict`), `module_code` (set of str), `hooks` (dict), `dynamic_imports` (set), `refs` (dict), `app_wrap_components` (dict)
- [ ] `CompileContext` is implemented with fields for: `pages`, `hooks` (CompilerHooks), `compiled_pages` dict, and a `compile()` method that orchestrates page compilation
- [ ] All new types have docstrings and type annotations
- [ ] Unit tests cover: plugin dispatch ordering, context var lifecycle (enter/exit/get), PageContext field accumulation

## Key Files to Understand

- `reflex/components/component.py` — `BaseComponent._get_all_*` methods (lines ~1590-2002) show the current tree-walking pattern
- `reflex/compiler/compiler.py` — `_compile_page()`, `compile_stateful_components()`, `_get_shared_components_recursive()`
- `reflex/app.py` — `App._compile()` method (~lines 1151-1549) is the main compilation orchestrator

## Notes

- The context objects use `ContextVar` so that plugins running during compilation can access `PageContext.get()` or `CompileContext.get()` from anywhere in the call stack without explicit parameter passing.
- Plugin ordering matters: e.g., a `FooPlugin` that transforms components must run before `ConsolidateImportsPlugin` so the transformed components' imports are captured.
- This issue only introduces the types and basic tests. Wiring them into the actual compilation pipeline is a separate issue.

### LINKED ISSUE #6211: Single-pass compiler: Implement core compilation plugins (imports, hooks, custom code, dynamic imports, refs)
## Summary

Implement the default set of `CompilerPlugin`s that replicate the current multi-pass tree walking behavior as single-pass plugins. Each plugin collects one category of information during the single tree walk and stashes it in the `PageContext`.

Depends on #6210 (compiler plugin protocol and context objects).

## Background

Today, `reflex/compiler/compiler.py::_compile_page()` calls these tree-walking methods sequentially:

```python
def _compile_page(component: BaseComponent) -> str:
    imports = component._get_all_imports()          # walk 1
    _apply_common_imports(imports)
    imports = utils.compile_imports(imports)
    return templates.page_template(
        imports=imports,
        dynamic_imports=sorted(component._get_all_dynamic_imports()),  # walk 2
        custom_codes=component._get_all_custom_code(),                 # walk 3
        hooks=component._get_all_hooks(),                              # walk 4
        render=component.render(),                                     # walk 5
    )
```

And `App._compile()` does additional walks for:
- `component._get_all_imports()` (again, for global import collection)
- `component._get_all_app_wrap_components()` (walk 6)

Each walk traverses the entire component tree. With the plugin architecture, all of these become `compile_component` hooks that run during a single recursive walk.

### 1. `DefaultPagePlugin`
- `eval_page`: Evaluates the page function into a `PageContext` (equivalent to `compile_unevaluated_page()` in `reflex/compiler/compiler.py`)
- `compile_page`: Orchestrates the single recursive tree walk by calling `_compile_component_recursive()`
- `_compile_component_recursive()`: The core single-walk method that dispatches `compile_component` hooks for each component

**Reference**: See `DefaultPagePlugin` in the demo code. The key insight is the two-phase generator pattern — pre-yield sees the component going down, post-yield (via `gen.asend((comp, children))`) sees the result going back up.

### 2. `ConsolidateImportsPlugin`
Replaces `Component._get_all_imports()` and `_apply_common_imports()`.

```python
async def compile_component(self, comp):
    comp, _ = yield  # post-yield: get the (potentially transformed) component
    if isinstance(comp, Component) and (imports := comp._get_imports()):
        PageContext.get().imports.append(imports)

async def compile_page(self, page_ctx):
    # After tree walk, collapse and merge all collected imports
    page_ctx.imports = [collapse_imports(merge_imports(*page_ctx.imports))]
```

**Current code**: `Component._get_all_imports()` in `component.py` (~line 1739)

### 3. `ConsolidateHooksPlugin`
Replaces `Component._get_all_hooks()` and `_get_all_hooks_internal()`.

```python
async def compile_component(self, comp):
    comp, _ = yield
    if isinstance(comp, Component):
        hooks = {}
        hooks.update(comp._get_hooks_internal())
        if (h := comp._get_hooks()) is not None:
            hooks[h] = None
        hooks.update(comp._get_added_hooks())
        PageContext.get().hooks.update(hooks)
```

**Current code**: `Component._get_all_hooks()` in `component.py` (~line 1899)

### 4. `ConsolidateCustomCodePlugin`
Replaces `Component._get_all_custom_code()`.

**Current code**: `Component._get_all_custom_code()` in `component.py` (~line 1590)

### 5. `ConsolidateDynamicImportsPlugin`
Replaces `Component._get_all_dynamic_imports()`.

**Current code**: `Component._get_all_dynamic_imports()` in `component.py` (~line 1627)

### 6. `ConsolidateRefsPlugin`
Replaces `Component._get_all_refs()`.

**Current code**: `Component._get_all_refs()` in `component.py` (~line 1933)

### 7. `ConsolidateAppWrapPlugin`
Replaces `Component._get_all_app_wrap_components()`.

**Current code**: `Component._get_all_app_wrap_components()` in `component.py` (~line 1972)

### 8. `ApplyStylePlugin`
Replaces `Component._add_style_recursive()`. Applies styles during the descending phase (pre-yield) of the tree walk.

**Current code**: `Component._add_style_recursive()` in `component.py` (~line 1251)

## Acceptance Criteria

- [ ] Each plugin listed above is implemented as a class conforming to `CompilerPlugin`
- [ ] Each plugin has unit tests that verify it produces identical output to the current `_get_all_*` method for the same component tree
- [ ] A composed pipeline of all plugins produces the same `PageContext` data as the current `_compile_page()` function
- [ ] Integration test: compile a sample page with the new plugin pipeline and compare the resulting JS output to the current compiler output (should be byte-identical)
- [ ] Plugins are collected in a sensible default ordering (style first, then consolidation plugins, with imports last since it needs to see all transformations)

## Key Files

- `reflex/components/component.py` — All `_get_all_*` methods and `_add_style_recursive`
- `reflex/compiler/compiler.py` — `_compile_page()`, `_apply_common_imports()`
- `reflex/compiler/utils.py` — `merge_imports()`, `compile_imports()`, `collapse_imports()`

## Notes

- The individual `_get_imports()`, `_get_hooks()`, `_get_custom_code()`, etc. (non-recursive, single-component versions) remain on the `Component` class. Only the `_get_all_*` (recursive tree-walking) versions are replaced by plugins.
- The `_get_components_in_props()` call in some `_get_all_*` methods means those plugins need to also walk components embedded in props, not just `children`.
- Plugin ordering is critical. For example, `ApplyStylePlugin` must run before any consolidation plugin. The `ConsolidateImportsPlugin.compile_page` phase must run after all `compile_component` hooks have finished collecting imports.

### LINKED ISSUE #6212: Single-pass compiler: Wire plugin pipeline into App._compile() and remove multi-pass tree walks
## Summary

Replace the current compilation orchestration in `App._compile()` with the new `CompileContext`-based plugin pipeline. This is the integration point where the new compiler architecture actually takes over from the old one. Also moves compilation logic out of `reflex/app.py` into the compiler module.

Depends on #6210 and #6211.

## Background

`App._compile()` in `reflex/app.py` (~lines 1151-1549) is currently a ~400-line method that orchestrates the entire compilation process. It:

1. Evaluates unevaluated pages via `compile_unevaluated_page()`
2. Applies styles via `_add_style_recursive()`
3. Collects imports via `_get_all_imports()` on each page
4. Collects app wrap components via `_get_all_app_wrap_components()`
5. Runs `StatefulComponent.compile_from()` on each page (auto-memoization)
6. Compiles custom components via `compile_memo_components()`
7. Compiles shared stateful components via `_compile_stateful_components()`
8. Submits page compilation tasks to a thread/process pool executor
9. Compiles document root, app root, contexts, theme, stylesheets
10. Writes output files

The goal is to replace steps 1-7 with the `CompileContext.compile()` method and the registered plugins, and to move the remaining orchestration out of `App._compile()` into a standalone `compile_app()` function in the compiler module.

### Move compilation out of `App`
The `App` object should be part of the `CompileContext` (it holds config, style, pages, etc.), but the compilation logic itself should live in `reflex/compiler/`. This keeps `App` focused on being the runtime application object.

```python
# New entry point in reflex/compiler/compiler.py (or a new orchestrator module)
async def compile_app(app: App, *, prerender_routes: bool = False, dry_run: bool = False):
    """Compile the entire application using the plugin pipeline."""
    compile_ctx = CompileContext(
        app=app,
        pages=[...],
        hooks=CompilerHooks(plugins=(...)),
    )
    async with compile_ctx:
        await compile_ctx.compile()
    # ... write output files
```

### Replace tree walks with plugin results
Instead of:
```python
for component in self._pages.values():
    all_imports.update(component._get_all_imports())       # walk
    app_wrappers.update(component._get_all_app_wrap_components())  # walk
```

The plugins have already collected this data into `PageContext` during the single walk:
```python
for page_ctx in compile_ctx.compiled_pages.values():
    all_imports.update(merge_imports(*page_ctx.imports))
    app_wrappers.update(page_ctx.app_wrap_components)
```

### Remove `ExecutorSafeFunctions` and executor-based parallelism
The current code uses `concurrent.futures` with thread/process pool executors. As noted in the project brief, the compiler is CPU-bound and doesn't benefit from Python threads (GIL) or multiprocessing (overhead). Remove:
- `ExecutorSafeFunctions` class in `reflex/compiler/compiler.py`
- `ExecutorType.get_executor_from_environment()` usage in `App._compile()`
- The `REFLEX_COMPILE_EXECUTOR` environment variable / config

Replace with straightforward sequential (or later, async) compilation.

### Page template rendering
The final step of rendering each page to JS via `templates.page_template()` should also become a plugin or a step in `CompileContext.compile()` that uses the accumulated `PageContext` data:

```python
code = templates.page_template(
    imports=utils.compile_imports(page_ctx.imports[0]),
    dynamic_imports=sorted(page_ctx.dynamic_imports),
    custom_codes=page_ctx.module_code,
    hooks=page_ctx.hooks,
    render=page_ctx.root_component.render(),
)
```

## Acceptance Criteria

- [ ] `App._compile()` is replaced by a new top-level compilation function in the compiler module (or significantly simplified to just call it)
- [ ] The new function uses `CompileContext` and the default plugin set to compile pages
- [ ] `ExecutorSafeFunctions` class is removed
- [ ] Thread/process pool executor usage for page compilation is removed
- [ ] All existing tests pass (the output should be functionally identical)
- [ ] Benchmark: compile a non-trivial app (e.g., the reflex-web docs site) with both old and new paths to verify no performance regression
- [ ] `App._compile()` no longer directly calls any `_get_all_*` methods on components

## Key Files

- `reflex/app.py` — `App._compile()` (the main target)
- `reflex/compiler/compiler.py` — `ExecutorSafeFunctions`, `compile_page()`, `compile_stateful_components()`
- `reflex/compiler/utils.py` — `write_file()`, output path helpers
- `reflex/environment.py` — `ExecutorType` enum (to be removed)

## Migration Strategy

To reduce risk, consider a phased approach:
1. First, add the new compilation path behind a feature flag / environment variable
2. Run both old and new paths in CI, comparing outputs
3. Once validated, remove the old path

## Notes

- The `_compile()` method also handles non-compilation concerns (asset copying, frontend package installation, plugin `pre_compile` hooks, react-router config). These should remain but be clearly separated from the component compilation pipeline.
- The current progress bar tracking (`rich.progress`) should be preserved — the new pipeline should report progress as pages are compiled.
- The longer-term vision is for `CompileContext` to be adaptable to different targets (React Native, Vue, etc.), so the page template rendering step should be pluggable too.

### LINKED ISSUE #6213: Single-pass compiler: Replace StatefulComponent auto-memoization with rx._x.memo-based approach
## Summary

Replace the current `StatefulComponent` auto-memoization mechanism with one that uses `rx._x.memo` (the experimental memo decorator) to create stateful components that reference `{children}` instead of recursively recreating the component tree. Implement this as a compiler plugin in the new single-pass architecture.

Depends on [ENG-9142](https://linear.app/reflex-dev/issue/ENG-9142/single-pass-compiler-introduce-compilerplugin-protocol-compilerhooks) and [ENG-9143](https://linear.app/reflex-dev/issue/ENG-9143/single-pass-compiler-implement-core-compilation-plugins-imports-hooks).

## Background: How StatefulComponent Works Today

`StatefulComponent` in `reflex/components/component.py` (\~line 2375) is a `BaseComponent` subclass that wraps components which depend on state. The current flow:

1. `StatefulComponent.compile_from(component)` walks the tree recursively
2. For each `Component`, it calls `StatefulComponent.create(component)` which:
   * Checks if the component has `_memoization_mode` enabled
   * Renders the component to a string and hashes it to create a unique tag name (`_get_tag_name`)
   * If a component with that tag already exists in `tag_to_stateful_component`, bumps its `references` counter
   * Otherwise creates a new `StatefulComponent` wrapper
3. `StatefulComponent._render_stateful_code()` renders the wrapper as a memoized React component using `stateful_component_template()`
4. In prod mode, shared components (referenced > 1 time) are extracted to a separate `stateful_components.js` file via `_compile_stateful_components()` / `_get_shared_components_recursive()`

### Problems with this approach:

* **Re-renders the entire component subtree** inside the memoized wrapper — the children are baked into the rendered code, not passed as `{children}`
* **Hash-based identity** is fragile — rendering to string and hashing is expensive and can produce false duplicates or miss actual duplicates
* **Complex shared component tracking** — the `references` counter and `rendered_as_shared` flag add complexity
* **Separate tree walk** — `StatefulComponent.compile_from()` is yet another full traversal of the component tree, done before the main compilation walks

### New `MemoizeStatefulPlugin` compiler plugin

Instead of a separate `StatefulComponent.compile_from()` pass, implement auto-memoization as a `CompilerPlugin` that runs during the single tree walk:

```python
class MemoizeStatefulPlugin(CompilerPlugin):
    async def compile_component(self, comp, /, **kwargs):
        """Wrap stateful components with rx._x.memo."""
        # Pre-yield: check if this component depends on state
        comp, children = yield
        
        if not isinstance(comp, Component):
            yield
            return
            
        if not comp._memoization_mode.disposition:
            yield
            return
            
        # Check if component actually references state vars
        if not self._has_state_dependency(comp):
            yield
            return
        
        # Wrap with memo, using {children} placeholder instead of
        # re-rendering the entire subtree
        memo_comp = self._create_memo_wrapper(comp)
        yield memo_comp
```

### Use `{children}` instead of baking the subtree

The key architectural change: instead of rendering the entire subtree into the memoized component's body, the memo wrapper should accept `{children}` as a prop. This means:

* The memoized component renders as `<MemoComp_{hash}>{children}</MemoComp_{hash}>`
* Children are passed through, not duplicated
* React's own reconciliation handles re-rendering children efficiently
* No need for the `references` / `rendered_as_shared` tracking

### Use `rx._x.memo` decorator

The existing `memo` (alias for `custom_component`) in `component.py` creates `CustomComponent` wrappers. The experimental `rx._x.memo` should be used instead, which provides proper React.memo semantics. The stateful component wrapper should use this decorator to generate the memoized JS output.

## Acceptance Criteria

- [ ] New `MemoizeStatefulPlugin` implements `compile_component` hook
- [ ] Memoized components use `{children}` prop instead of baking the subtree
- [ ] The old `StatefulComponent` class is removed or deprecated
- [ ] `_compile_stateful_components()` and `_get_shared_components_recursive()` in `compiler.py` are removed
- [ ] The `stateful_components.js` shared file is no longer needed (or simplified)
- [ ] `StatefulComponent.compile_from()` tree walk is eliminated
- [ ] Existing apps produce functionally equivalent output (same visual behavior, may differ in JS structure)
- [ ] Unit tests verify that stateful components are properly memoized
- [ ] Integration test: an app with shared stateful components across pages works correctly

## Key Files

* `reflex/components/component.py`:
  * `StatefulComponent` class (\~line 2375) — the class to replace
  * `StatefulComponent.compile_from()` (\~line 2788) — the tree walk to eliminate
  * `StatefulComponent.create()` (\~line 2411) — the memoization logic
  * `StatefulComponent._get_tag_name()` (\~line 2517) — hash-based naming
  * `MemoizationMode` — controls which components get memoized
* `reflex/compiler/compiler.py`:
  * `compile_stateful_components()` — orchestrates the memoization pass
  * `_compile_stateful_components()` — extracts shared components
  * `_get_shared_components_recursive()` — shared component tracking
* `reflex/compiler/templates.py` — `stateful_component_template()`

## Notes

* The `MemoizationMode` class on components should still be respected — components can opt in/out of memoization.
* This change may affect how `StatefulComponent`'s special `_get_all_*` methods work (they currently short-circuit when `rendered_as_shared` is True). With the new approach, this complexity goes away.
* Consider whether the new memo wrapper needs to handle event trigger hooks (`memo_trigger_hooks` on the current `StatefulComponent`).

### LINKED ISSUE #6325: Implement radix-themes as a plugin instead of tightly integrated in the core

For apps that do not use radix-themes, there is no way to opt-out of the radix themes js/css bundle that increases overall app size. Radix themes should only be included if the app is making use of radix themes components.

**What needs to happen**

* Create a `Plugin` that encapsulates any radix-themes functionality that is currently in `App` (for example, expose the `theme=` kwarg on the plugin's constructor)
* The RadixThemesPlugin should not be enabled by default
* If `theme=` is supplied to `App`, the plugin should be enabled, with a deprecation warning.
* If any radix themes components are used in any page, the plugin should be enabled, with a deprecation warning.

FarhanAliRaza:
## Summary

- Introduce a **CompilerPlugin protocol** with `enter_component`/`leave_component`/`eval_page`/`compile_page` hooks, enabling third-party plugins to participate in the frontend compilation pipeline
- Move the frontend compilation pipeline from `App._compile()` into `compiler.compile_app()`, decoupling compilation logic from the App class
- Remove `ExecutorType`/`ExecutorSafeFunctions` abstractions from `environment.py` in favor of a sequential plugin-driven compilation model
- Add built-in plugins: `DefaultPagePlugin`, `ApplyStylePlugin`, `DefaultCollectorPlugin` that replicate the existing compilation behavior
- Fix memo component compilation ordering so app-wrap components are included
- Fix `DefaultPagePlugin` to preserve `Var`-backed page titles instead of replacing them with the default string

## Key Changes

- **`packages/reflex-core/src/reflex_core/plugins/compiler.py`** (new): Core plugin infrastructure — `CompilerPlugin` protocol, `CompileContext`, `PageContext`, `CompilerHooks`, and component tree traversal logic
- **`reflex/compiler/plugins/builtin.py`** (new): Built-in plugins (`DefaultPagePlugin`, `ApplyStylePlugin`, `DefaultCollectorPlugin`) that replicate existing compilation behavior
- **`reflex/compiler/compiler.py`**: Refactored `compile_app()` to use plugin-driven compilation via `CompilerHooks`
- **`reflex/app.py`**: Removed ~430 lines of compilation logic, `App._compile()` now delegates to `compiler.compile_app()`
- **`packages/reflex-core/src/reflex_core/environment.py`**: Removed `ExecutorType` and `ExecutorSafeFunctions` (no longer needed)

## Test Plan

- [ ] New unit tests in `tests/units/compiler/test_plugins.py` (855 lines) covering:
  - Plugin hook dispatch (enter/leave component, eval/compile page)
  - `CompileContext` lifecycle and state management
  - `DefaultPagePlugin` behavior including Var-backed titles
  - `ApplyStylePlugin` style application
  - `DefaultCollectorPlugin` import/custom code collection
  - Route deduplication and error handling
  - End-to-end compilation matching legacy output
- [ ] New test in `tests/units/test_app.py` verifying memo components are written to shared components module
- [ ] Existing test suite passes without modification

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

closes #6210
closes #6211
closes #6212
Closes #6213
Closes #6325


reflex/compiler/plugins.py
```diff
@@ -0,0 +1,431 @@
...
+                else tuple(compiled_component.children)
+            )
+            compiled_children = await self._compile_children(
+                structural_children,
+                **kwargs,
+            )
+
+            if isinstance(compiled_component, Component):
+                for prop_component in compiled_component._get_components_in_props():
+                    await self.compile_component(prop_component, **kwargs)
```
masenf: at first glance it seems like we might need to do something with the potential replacement value returned by this function, like applying it as the prop value on the compiled_component if it differs

masenf (commented):
we need a test case for the child replacement logic when traversing the tree; we might not strictly need it now, but it will be important when moving the StatefulComponent compilation into the plugin system.

masenf:
Khaleel and I were discussing this a bit further, and I think it would be better to add these new plugin hooks and hook dispatching system to the existing Plugin base in `packages/reflex-core`

FarhanAliRaza:
@masenf i dont know why the tests are failing.

packages/reflex-base/src/reflex_base/plugins/base.py
```diff
@@ -117,6 +119,93 @@ def post_compile(self, **context: Unpack[PostCompileContext]) -> None:
...
+        """Evaluate a page-like object into a page context.
+
+        Args:
+            page_fn: The page-like object to evaluate.
+            kwargs: Additional compiler-specific context.
+
+        Returns:
+            A page context when the plugin can evaluate the page, otherwise ``None``.
+        """
+        del page_fn, kwargs
```
masenf: why does it `del` the local vars?

packages/reflex-base/src/reflex_base/plugins/compiler.py
```diff
@@ -0,0 +1,1136 @@
...
+    ["PageContext", "CompileContext"],
+    CompiledEnterHook,
+]
+LeaveHookBinder: TypeAlias = Callable[
+    ["PageContext", "CompileContext"],
+    CompiledLeaveHook,
+]
+
+
+class CompilerPlugin(Protocol):
```
masenf: is this protocol needed since we have the base plugin definition?

packages/reflex-base/src/reflex_base/plugins/compiler.py
```diff
@@ -0,0 +1,1136 @@
...
+        """
+        if replacement is None:
+            return comp, children
+        if isinstance(replacement, tuple):
+            return replacement
+        return replacement, children
+
+
+@dataclasses.dataclass(kw_only=True)
+class BaseContext:
```
masenf, line 585: this class also gets defined in my backend event loop PR as `reflex_base.context.base` depending on whose merges first, we can refactor

packages/reflex-base/src/reflex_base/plugins/compiler.py
```diff
@@ -0,0 +1,1136 @@
...
+                    raise TypeError(msg)
+                page_ctx.root_component = self.app._add_overlay_to_component(
+                    page_ctx.root_component,
+                    overlay_component,
+                )
+
+            if isinstance(page_ctx.root_component, StatefulComponent):
+                self.all_imports = merge_imports(
+                    self.all_imports,
+                    page_ctx.root_component._get_all_imports(),
```
masenf: is this still making the recursive call? isn't this information cached in the PageContext as `imports`?

packages/reflex-base/src/reflex_base/plugins/base.py
FarhanAliRaza: They are gonna be overloaded by the actual plugin. So Claude is i think, just trying to use it. Deleting it.

reflex/experimental/memo.py
```diff
@@ -950,6 +1004,65 @@ def _create_component_wrapper(
...
+    # Snapshot-boundary components (see ``is_snapshot_boundary``) own their
+    # subtree — the ``.children`` slot is internal machinery from the
+    # subclass's ``.create`` (e.g. the dropzone Div built inside
+    # ``Upload.create``), not a user content hole. The memoize plugin wraps
+    # the boundary with no structural children on the page side, so the memo
+    # body renders the full snapshot rather than a ``{children}``-holed
+    # template.
+    snapshot_only = is_snapshot_boundary(component)
+
+    captured_hole_child: list[Component] = []
```
masenf, line 1041: why is this a `list[Component]` and not `Component | None`? The latter seems more semantically accurate to how it's used.

masenf (commented):
i noticed another problem here with `Foreach` and `Cond`, and I think the fix is similar to what i did for `Bare`: basically treat these components as MemoizationLeaf and extract them to their own memoized component without the `{children}` hole.

When we're compiling out a Foreach, Cond, or Match, these need to be extracted as their own auto-memo component if they are stateful. And to ensure that nested Foreach is working, we need to treat the first one we encounter in the tree as a MemoizationLeaf so the nested Foreach can access the outer Foreach loop variables.

Ideally we have some way of separately memoizing these and passing the loop props down if needed, but that's a lot of extra complexity, so we should just get the base case working in the same way the current compiler does it.

The main difference between the approach here and the current compiler is that the current compiler would treat the parent component as stateful if any of its children are Foreach/Cond/Match, but that's not really appropriate any more with the `{children}` passthrough being the default, so these really should be memoized as Fragment containing their contents.

masenf:
With the latest commits, we are very close. I'm noticing a few remaining problems that shouldn't be too hard to fix:

MemoizationLeaf components should still be memoized if anything in their subtree is stateful. i.e. AccordionTrigger is not getting memoized, even though the trigger contains a cond that depends on a state variable.

I think the title and meta components (and potentially other html components) should be marked as MemoizationLeaf so we don't end up memoizing a stateful Bare that is passed to them. These elements cannot have children, so they don't render properly when we do this for example:

```javascript
jsx("title", {}, jsx(Bare_comp_cbb10fd3740c4ca79b576489b0269aed, {})),
```

We should get two test cases and fixes for these.

I'm signing off for now, but I might push commits in a few hours if I don't see anything here.

tests/units/compiler/test_memoize_plugin.py
```diff
@@ -1125,3 +1129,841 @@ def page() -> Component:
...
+    """A snapshot boundary memoizes when its subtree contains state-derived hooks.
+
+    ``LeafComponent`` mirrors the Radix-primitive shape: ``recursive=False``
+    set directly without inheriting from ``MemoizationLeaf``.
+    """
+    boundary = LeafComponent.create(Plain.create(STATE_VAR))
+    assert _should_memoize(boundary)
+
+
+def test_snapshot_boundary_wraps_subtree_once_when_descendant_is_stateful() -> None:
```
masenf, line 1169: this test seems basically the same as `test_memoization_leaf_suppresses_descendant_wrapping` now. i think we can remove `test_memoization_leaf_suppresses_descendant_wrapping` in favor of this test case

tests/units/compiler/test_memoize_plugin.py
```diff
@@ -1125,3 +1129,841 @@ def page() -> Component:
...
+    from reflex_components_core.el.elements.scripts import Script
+
+    return [
+        pytest.param(StyleEl, id="style"),
+        pytest.param(Textarea, id="textarea"),
+        pytest.param(Script, id="script"),
+    ]
+
+
+@pytest.mark.parametrize("cls", _text_only_classes())
```
masenf: `_text_only_classes` should not be a separate function. just make the imports at module level and define the parametrizations inline

tests/integration/tests_playwright/test_memoize_edge_cases.py
```diff
@@ -0,0 +1,219 @@
...
+        def set_css_dark(self):
+            self.css_marker = "memo-css-dark"
+
+        @rx.event
+        def bump(self):
+            self.counter = self.counter + 1
+
+    def index():
+        return rx.box(
+            rx.el.title(MemoState.title_marker),
```
masenf: instead of adding a semantically invalid title, which no real app would actually do, pass the dynamic title to `app.add_page(..., title=MemoState.title_marker)` and then remove the weird inline JS to find the title element and instead just assert against the document.title.

update the description to match the new logic

reflex/compiler/plugins/memoize.py
```diff
@@ -71,12 +71,61 @@ def _compute_memo_tag(component: Component) -> str | None:
...
+
+    Returns:
+        True if the subtree carries event triggers, explicit hooks, or any
+        Var whose merged var_data has ``state`` or ``hooks``.
+    """
+    if _seen is None:
+        _seen = set()
+    if id(component) in _seen:
+        return False
+    _seen.add(id(component))
```
masenf: just because we've seen a component doesn't mean it automatically is not reactive.

maybe split this function up so the `_seen` structure can be a dict that stores whether the given subtree is reactive, then the short circuit can return the correct value that we cached previously.

packages/reflex-components-core/src/reflex_components_core/base/link.py
```diff
@@ -11,6 +12,8 @@ class RawLink(BaseHTML):
 
     tag = "link"
 
+    _memoization_mode = MemoizationMode(recursive=False)
```
masenf: instead of adding MemoizationMode to each element that needs it, you should create new base classes for the RawTextBaseHTML and VoidBaseHTML and then use those to assign the _memoization_mode.

in the future, we might want additional validation on these elements, for example, they should never have children. we don't need this extra validation right now (to avoid breaking existing code), but we should set up the structure for representing these semantics in a more readable way.

tests/units/compiler/test_memoize_plugin.py
```diff
@@ -1125,3 +1129,841 @@ def page() -> Component:
...
+def test_snapshot_boundary_with_event_trigger_descendant_is_wrapped() -> None:
+    """A snapshot boundary with a stateful event-trigger descendant must wrap."""
+    from reflex_base.event import EventChain
+
+    event_var = Var(_js_expr="test_event")._replace(
+        _var_type=EventChain,
+        merge_var_data=VarData(state="TestState", hooks={"useTestState": None}),
+    )
+    inner = Plain.create()
+    inner.event_triggers["on_click"] = event_var
```
masenf: why not just use `inner = Plain.create(on_click=event_var)`?

tests/units/compiler/test_memoize_plugin.py
```diff
@@ -1125,3 +1129,841 @@ def page() -> Component:
...
+    wrapper_tag = next(iter(ctx.memoize_wrappers))
+    assert wrapper_tag.lower().startswith("title_"), (
+        f"Wrapper should be derived from Title, got: {wrapper_tag!r}"
+    )
+    output = page_ctx.output_code or ""
+    assert "Bare_comp" not in output, (
+        "Bare must not be independently memoized as a child of <title>. "
+        "It needs to remain a text interpolation inside the title's snapshot.\n"
+        f"Page output snippet: {output[:2000]}"
+    )
```
masenf: defaulting to empty string and then asserting the absence of a value is not a good assertion.

you need to at least check that the page output contains the memo wrapper call and that the state var name / hook does not appear in the page output. `"Bare_comp"` is not a good thing to check the output for because if we change the generic generated names then it stops working. Instead, all positive and negative assertions should be derived from values that absolutely cannot change or would be inconsistent with the system (like the name of the state var).

this issue and similar patterns should be addressed throughout this file.

tests/units/compiler/test_memoize_plugin.py
```diff
@@ -1125,3 +1129,841 @@ def page() -> Component:
...
+
+    output = page_ctx.output_code or ""
+    assert "useContext(StateContexts" not in output, (
+        "State read leaked into the page module — the trigger's stateful cond "
+        "should be captured inside the snapshot wrapper instead.\n"
+        f"Page output snippet: {output[:2000]}"
+    )
+
+
+def _restricted_content_components() -> list:
```
masenf: again, why is this a function? you're in test code, so move the imports to module level and define the parametrization inline

tests/units/compiler/test_memoize_plugin.py
```diff
@@ -1125,3 +1129,841 @@ def page() -> Component:
...
+    component AND defeat the boundary's "I own my subtree" contract.
+    """
+    inner = LeafComponent.create(Plain.create(STATE_VAR))
+    outer = LeafComponent.create(inner)
+    ctx, _page_ctx = _compile_single_page(lambda: outer)
+
+    assert len(ctx.memoize_wrappers) == 1, (
+        f"Nested snapshot boundaries must collapse to one outer wrapper; got "
+        f"{list(ctx.memoize_wrappers)}"
+    )
```
masenf, line 1555: add an additional assertion that the memo code contains the nested LeafComponent and its nested Plain/Bare components

tests/units/compiler/test_memoize_plugin.py
```diff
@@ -1125,3 +1129,841 @@ def page() -> Component:
...
+    boundary = LeafComponent.create(
+        Plain.create(STATE_VAR),
+        Plain.create(STATE_VAR),
+        WithProp.create(label=STATE_VAR),
+    )
+    ctx, _page_ctx = _compile_single_page(lambda: boundary)
+    assert len(ctx.memoize_wrappers) == 1, (
+        f"Multiple stateful descendants must share the boundary's wrapper; got "
+        f"{list(ctx.memoize_wrappers)}"
+    )
```
masenf, line 1611: again, extend the assertions to check that the generated memo component actually contains its expected children

tests/units/compiler/test_memoize_plugin.py
```diff
@@ -1125,3 +1129,841 @@ def page() -> Component:
...
+    assert len(ctx.memoize_wrappers) == 1, (
+        f"Identical boundary subtrees should share one wrapper; got "
+        f"{list(ctx.memoize_wrappers)}"
+    )
+    wrapper_tag = next(iter(ctx.memoize_wrappers))
+    assert (page_ctx.output_code or "").count(f"jsx({wrapper_tag},") == 2
+
+
+def test_passthrough_wrapper_inside_snapshot_boundary_is_suppressed() -> None:
```
masenf: this seems the same as `test_snapshot_boundary_wraps_subtree_once_when_descendant_is_stateful`, if there are any extra assertions present in this test case, roll them into `test_snapshot_boundary_wraps_subtree_once_when_descendant_is_stateful` and remove this one.

tests/units/compiler/test_memoize_plugin.py
```diff
@@ -1125,3 +1129,841 @@ def page() -> Component:
...
+
+    return [
+        pytest.param(lambda: Title.create("hello", id="t"), id="title_with_id"),
+        pytest.param(lambda: Img.create(src="/x.png", id="logo"), id="img_with_id"),
+        pytest.param(lambda: Br.create(id="br"), id="br_with_id"),
+        pytest.param(lambda: BaseInput.create(id="i"), id="input_with_id"),
+        pytest.param(
+            lambda: Meta.create(name="description", id="m"), id="meta_with_id"
+        ),
+    ]
```
masenf: again, move imports to top of module and define parametrization cases inline

tests/units/compiler/test_memoize_plugin.py
```diff
@@ -1125,3 +1129,841 @@ def page() -> Component:
...
+    A ``client_state`` Var contributes its ``useState``/``useId`` hooks via
+    ``var_data.hooks`` without setting ``var_data.state``. The reactive-Var
+    walk must catch the hooks-only case so client-state-driven content
+    inside a snapshot boundary lands in the memo body.
+    """
+    from reflex_components_core.el.elements.metadata import Title
+
+    from reflex.experimental.client_state import ClientStateVar
+
+    cs_var = ClientStateVar.create("titletest", default="hi", global_ref=False)
```
masenf: parametrize this test on global_ref True/False; expectation should be the same

tests/units/compiler/test_memoize_plugin.py
```diff
@@ -1125,3 +1129,841 @@ def page() -> Component:
...
+        if Path(path).parent.name == "components"
+    }
+    expected = {f"{tag}.jsx" for tag in ctx.memoize_wrappers}
+    assert component_module_names == expected, (
+        f"Per-wrapper file invariant broken. wrappers={sorted(ctx.memoize_wrappers)} "
+        f"files={sorted(component_module_names)}"
+    )
+    assert len(ctx.memoize_wrappers) >= 2, (
+        "Test should exercise multi-wrapper case to be meaningful."
+    )
```
masenf: the test should know the exact amount of expected memoize_wrappers

masenf (commented):
in test code, imports should be at the top of the module. the only exception to this is integration tests which need imports in the app function because that ends up getting written out to its own module by the AppHarness

masenf:
@greptile-apps re-review

masenf:
ugh just realized that dynamic components are broken with this change

```python
import reflex as rx


class State(rx.State):
    count: int = 0

    @rx.event
    def set_count(self, count: int):
        self.count = count

    @rx.var
    def counter_ui(self) -> rx.Component:
        return rx.hstack(
            rx.button("-", on_click=State.set_count(self.count - 1)),
            rx.text(self.count, size="9"),
            rx.button("+", on_click=State.set_count(self.count + 1)),
            spacing="5",
            justify="center",
        )


def index() -> rx.Component:
    return rx.container(
        rx.vstack(
            State.counter_ui,
        ),
    )


app = rx.App()
app.add_page(index)
```

This counter actually works on 0.9.1, but doesn't render the event handlers with this PR.

Generated code on this PR:

```javascript
//__reflex_evaluate
const React = window.__reflex.react;
const { jsx } = window.__reflex["@emotion/react"];
const { Fragment, useEffect } = window.__reflex["react"];
const {
  Button: RadixThemesButton,
  Flex: RadixThemesFlex,
  Text: RadixThemesText,
} = window.__reflex["@radix-ui/themes"];

export default function MySSRComponent() {
  return jsx(
    RadixThemesFlex,
    {
      align: "start",
      className: "rx-Stack",
      direction: "row",
      justify: "center",
      gap: "5",
    },
    jsx(RadixThemesButton, {}, "-"),
    jsx(RadixThemesText, { as: "p", size: "9" }, -1),
    jsx(RadixThemesButton, {}, "+"),
  );
}
```

------

Generated code on 0.9.1

```javascript
//__reflex_evaluate
const React = window.__reflex.react;
const { jsx } = window.__reflex["@emotion/react"];
const { Fragment, useContext, useEffect } = window.__reflex["react"];
const {
  Button: RadixThemesButton,
  Flex: RadixThemesFlex,
  Text: RadixThemesText,
} = window.__reflex["@radix-ui/themes"];
const { EventLoopContext } = window["__reflex"]["$/utils/context"];
const { ReflexEvent, applyEventActions } = window["__reflex"]["$/utils/state"];

export default function MySSRComponent() {
  const [addEvents, connectErrors] = useContext(EventLoopContext);

  return jsx(
    RadixThemesFlex,
    {
      align: "start",
      className: "rx-Stack",
      direction: "row",
      justify: "center",
      gap: "5",
    },
    jsx(
      RadixThemesButton,
      {
        onClick: (_e) =>
          addEvents(
            [
              ReflexEvent(
                "reflex___state____state.repro_dynamic_component___repro_dynamic_component____state.set_count",
                { ["count"]: 0 },
                {},
              ),
            ],
            [_e],
            {},
          ),
      },
      "-",
    ),
    jsx(RadixThemesText, { as: "p", size: "9" }, 1),
    jsx(
      RadixThemesButton,
      {
        onClick: (_e) =>
          addEvents(
            [
              ReflexEvent(
                "reflex___state____state.repro_dynamic_component___repro_dynamic_component____state.set_count",
                { ["count"]: 2 },
                {},
              ),
            ],
            [_e],
            {},
          ),
      },
      "+",
    ),
  );
}
```

masenf (approved):

PR #6267, opened 2026-04-01
ENG-9028: Move event queue to backend

masenf:
This patch moves the primary responsibility of chaining/queueing events to the backend. Previously the frontend was responsible for sending events to the backend. This created additional round-trip latency when chaining events using `yield`.

### Contexts are King

Instead of relying on ClassVar, module globals, and other hacks to track reflex objects, this PR introduces contextvars to the code base and leverages them to manage global state safely from anywhere in the app.

### New Modules (in `reflex_base`)
- **`registry.py`** — ContextVar-based registry for `BaseState` classes and `EventHandler`s, replacing implicit global lookups. Currently used for states, events and stateful components. Eventually will be used for DECORATED_PAGES and other weird globals.
- **`event/context.py`** — `EventContext` (inherits `BaseContext`) holding per-event metadata: token, state manager ref, enqueue/emit/delta callbacks
- **`event/processor/`** — New `BaseStateEventProcessor` and `EventProcessor` classes that own the full event lifecycle (previously spread across `App` and `State` methods)
- **`context/base.py`** — Generic `BaseContext` ContextVar wrapper

### Breaking Changes (0.9.0)
- **`Event.token` field removed** — `Event` no longer carries a `token`; the token lives in `EventContext`
- **`Event.substate_token` replaced** by `Event.state_cls` property (resolved via the registry)
- **`Delta` type refined** from `dict[str, Any]` → `dict[str, dict[str, Any]]` (nested by substate name)
- **`StateManager.create()` no longer takes a `state=` arg** — state classes are discovered from the registry
- **`StateToken` introduced** (`reflex/istate/manager/token.py`) — typed generic token replacing raw `"client_token_substate"` strings in all state managers (disk, memory, redis)
- **`EventHandlerSetVar.state_cls` → `.state`** field rename to align with EventHandler change.
- **`fix_events` removed** from sortof public event API, replaced by `Event.from_event_type()`
- **`get_hydrate_event` removed** — replaced by internal rehydration; simulated pre-hydrated states removed
- **`App._background_tasks` replaced** by `App.event_processor._tasks` (processor manages background task lifecycle in the same way as normal event tasks)
- **`AppHarness` (`reflex/testing.py`) simplified** — `state_manager` property and related methods removed

### State Manager Changes
- All three managers (disk, memory, redis) refactored to use `StateToken[BaseState]` instead of raw string tokens
- State managers now `close()` old locks properly
- `OPLOCK_ENABLED` support in state manager close/tests

### Frontend (`state.js`)
- Params passed around as a ref instead of inline values
- StateUpdate only includes non-empty fields, reducing bytes over the wire.

### Test Infrastructure
- Large test suite overhaul: new fixtures in `conftest.py` for `EventProcessor`, registry, and context
- `get_app` / `mock_app` dependencies removed from tests in favor of the new processor-based approach
- New unit tests for registry, event context, base state processor, and event processor
- Integration tests updated to replace all backend state assertions with in-app equivalents.

### DIFF
--- a/reflex/istate/manager/__init__.py
+++ b/reflex/istate/manager/__init__.py
@@ -15,6 +15,9 @@ from typing_extensions import ReadOnly, Unpack, deprecated
 from reflex.istate.manager.token import TOKEN_TYPE, StateToken
 from reflex.utils import console, prerequisites
 
+if TYPE_CHECKING:
+    from reflex.state import BaseState
+
 
 class StateModificationContext(TypedDict, total=False):
     """The context for modifying state."""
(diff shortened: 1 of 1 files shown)


packages/reflex-base/src/reflex_base/event/__init__.py
```diff
@@ -65,29 +67,84 @@ class Event:
...
+            if isinstance(e, Event):
+                # If the event is already an event, append it to the list.
+                if router_data is not None and e.router_data != router_data:
+                    out.append(
+                        dataclasses.replace(e, router_data=e.router_data | router_data)
+                    )
+                else:
+                    out.append(e)
+                continue
+            # Otherwise, create an event from the event spec.
```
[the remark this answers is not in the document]
masenf: #6270

masenf:
@greptile-apps re-review this PR now that first round of feedback has been applied and all tests are passing.

pay specific attention to things like potentially breaking changes that have not already been called out. pay attention to regressions. watch out for potential performance pitfalls.

FarhanAliRaza (commented):
## Code Review: ENG-9028 — Move event queue to backend
Claude's review of this PR.

Below are findings organized by severity.

---

### High Severity

**1. `disk.py` `set_state` silently drops state updates during debounce**

When a token is already in the write queue, the newer state is discarded:
```python
```
If the same token's state is modified twice within the debounce window, only the **first** modification is persisted. The queue entry should be updated with the latest state.

**2. `compat.py` — `asyncio.ensure_future(loop=)` on Python 3.12**

```python
loop = asyncio.get_event_loop()
todo = {asyncio.ensure_future(aw, loop=loop) for aw in set(aws)}
```
The `loop` parameter was deprecated in 3.10 and removed in later 3.x releases. Since this compat path runs for `sys.version_info < (3, 13)` (i.e. 3.11 and 3.12), verify this doesn't raise `TypeError` on 3.12.

---

### Medium Severity

**3. `BaseContext.__exit__` raises `KeyError` without matching `__enter__`**

```python
def __exit__(self, *exc_info):
    if (token := self._attached_context_token.pop(self)) is not None:
        self._context_var.reset(token)
```
`.pop(self)` raises `KeyError` if `__exit__` is called without a prior `__enter__`. Should be `.pop(self, None)`.

**4. `EventFuture.__post_init__` uses deprecated `loop=` on `asyncio.Future`**

```python
def __post_init__(self) -> None:
    super(EventFuture, self).__init__(loop=self.loop)
```
`asyncio.Future(loop=...)` is deprecated since 3.10 and scheduled for removal. Since `EventFuture` is always created inside a running loop, the `loop` parameter can likely be dropped.

**5. Background task state access after lock release**

In `base_state_processor.py`, after background tasks release the state lock, `root_state._get_resolved_delta()` may be called without holding the lock. This could produce inconsistent deltas if another event modifies the state concurrently. Worth verifying the locking discipline in `chain_updates`.

**6. Shared class-level `asyncio.Lock()` defaults in state managers**

In `memory.py`, `disk.py`, and `redis.py`:
```python
_state_manager_lock: asyncio.Lock = dataclasses.field(default=asyncio.Lock())
```
This shares a single Lock instance across all StateManager instances. Should use `default_factory=asyncio.Lock`. Currently safe (single instance in practice) but would be a bug in tests creating multiple managers.

---

### Low Severity

**7. `state.js` — recursive `processEvent` without depth guard**

```javascript
if (event_queue.length > 0) {
    await processEvent(socket, navigate, params);
}
```
The old `event_processing` guard is removed. Since JS is single-threaded this is safe, but a very large queue could cause deep recursion. Consider an iterative `while` loop.

**8. `Event.state_cls` raises `KeyError` on frontend-only events**

```python
@property
def state_cls(self) -> "type[BaseState]":
    substate_name = self.name.rpartition(".")[0]
    return RegistrationContext.get().base_states[substate_name]
```
Will raise `KeyError` for events like `_redirect` that have no substate. Callers need to be aware, or a guard should be added.

**9. `wrappers.py` `get_state` broken for non-legacy tokens**

When `state_cls is None`, the code uses `_split_substate_key(token)` on what should be a plain client token (UUID). Splitting a UUID on `_` produces wrong results for non-legacy token formats.

**10. `disk.py` `load_state` — blocking sync I/O + silent exception swallowing**

`load_state` calls `token_path.exists()` and `token_path.open()` synchronously in an `async def` method (the write side correctly uses `run_in_thread`). Also, `except Exception: pass` silently swallows deserialization errors — should at least log a warning.

---

### Test Coverage Gaps

- **`chain_updates()` / `process_event()`** in `base_state_processor.py` — only tested indirectly through the full pipeline
- **`_handle_backend_exception` with event return** — the override that calls `chain_updates` with events from the exception handler is untested
- **`compat.py` `as_completed` backport** — zero tests for the Python <3.13 backport
- **Multi-token concurrent ordering** — no test verifies events for different tokens are processed independently and in-order
- **`enqueue_stream_delta` error path** — untested when handler raises during streaming

---

reflex/istate/manager/__init__.py
```diff
@@ -105,8 +105,64 @@ def _coerce_token(token: StateToken[TOKEN_TYPE] | str) -> StateToken[TOKEN_TYPE]
             return BaseStateToken.from_legacy_token(token, root_state=State)  # type: ignore[return-value]
         return token
 
+    if TYPE_CHECKING:
```
adhami3310: why inside of TYPE CHECKING?

masenf: no real reason, maybe import times are reduced if these overloads aren't included? idk i got rid of it tho

adhami3310 (approved):

PR #6289, opened 2026-04-06
Respect root bun.lock during frontend installs

### LINKED ISSUE #4314: renovate/dependabot support
Reflex allows to specify npm deps via `library` and `lib_dependencies` string attributes of a Component.

Renovate and Dependabot will not parse such versions without additional plugins or configurations. It would be awesome if we could implement/document such config/plugins for reflex users and maybe even reflex-internal usage.

My team only uses renovate, I can be a bit more specific about that:
- Renovate is highly configurable, I could write a (complicated and error-prone) regex manager to parse the versions from python files (https://docs.renovatebot.com/modules/manager/regex/). This one will probably be easy, and I will raise a PR documenting those regexes if @reflex-dev is interested.
- For best support, we could later implement our own reflex manager for renovate (in TypeScript)

We could also implement a fancy approach where reflex (alternatively) allows setting `library` and `lib_dependencies` to a Path with a `package.json`. One big advantage is that this approach will probably add support for all dependency management tools. One disadvantage is that we have to leave python to adjust the versions.

### LINKED ISSUE #6268: Support persisting and respecting bun.lock
## Summary

Currently, the `bun.lock` file inside `.web/` is regenerated on each build to pull the most up-to-date versions of all dependencies. This means there is no way for users to lock direct **and** transitive JS dependencies, which is a significant supply chain security concern — especially in enterprise environments.

## Proposed Solution

- Move (or copy) the `bun.lock` file to the **project root directory** alongside the app code, so it can be committed to version control without needing to track files inside `.web/`.
- When `bun.lock` exists in the root, Reflex should copy it into `.web/` **before** invoking the package manager, so that transitive dependencies are not silently bumped.
- Reflex should still be able to bump direct dependencies using bun interfaces when the user requests it (e.g. `reflex update` or similar).
- Provide good UX for persisting and respecting the package manager lock — ideally this should "just work" without requiring manual workarounds.

## Current Workarounds

- Manually committing `.web/bun.lock` to git (fragile, since `.web` is typically gitignored).
- Writing a plugin that overrides `pre_compile` to copy a lock file into `.web/` before the package manager is invoked.

## Additional Notes

Running `reflex --loglevel debug` shows the bun commands being executed, which can help verify whether an existing lock file is being respected. As far as we can tell, nothing in the current code explicitly *ignores* the lock file — the issue is that it gets overwritten.

GautamBytes:
### All Submissions:

### Type of change

### Changes To Core Features:

### Description

This PR fixes Bun lockfile persistence for Reflex apps by treating the project root `bun.lock` as the canonical lockfile and mirroring it into `.web` before frontend package manager commands run.

Without this change, `.web/` reinitialization and frontend package installs could overwrite or ignore the effective Bun lockfile state, which meant users could not reliably commit and reuse Bun’s resolved dependency graph. That made it difficult to pin direct and transitive frontend dependencies.

This change:
- adds helpers to mirror `bun.lock` between the app root and `.web`
- restores `.web/bun.lock` from the root lock during `.web` initialization
- persists the final `.web/bun.lock` back to the root after successful Bun package operations
- updates frontend package install caching so lockfile changes and current extra package state are respected
- removes stale custom frontend packages from `.web/package.json` when they are no longer desired
- documents that `bun.lock` should live at the project root and be committed to version control

closes #6268
closes #4314


GautamBytes:
@masenf can you review this whenever feasible!

masenf (commented):
do we really need all this extra code? couldn't we get away with just syncing the bun lock file between the root and .web and if the bun lock file contents is different on the sync, then delete the delete the `reflex.install_frontend_packages.cached` so it reinstalls?

i'm reading through the new code and it seems like the js package manager should be responsible for making sure the installed files match the lock file.

masenf:
Your latest commit should be covered by #6307

reflex/utils/js_runtimes.py
```diff
@@ -353,24 +353,59 @@ def remove_existing_bun_installation():
...
+
+def _sync_root_bun_lock_for_frontend_install():
+    """Sync the canonical bun.lock into .web and invalidate the install cache when needed."""
+    root_bun_lock_path = frontend_skeleton.get_root_bun_lock_path()
+    web_bun_lock_path = frontend_skeleton.get_web_bun_lock_path()
+    cache_file = _frontend_packages_cache_path()
+
+    if not root_bun_lock_path.exists():
+        if web_bun_lock_path.exists():
+            frontend_skeleton.sync_root_bun_lock_to_web()
```
masenf, line 373: why are we syncing the root bun.lock over the .web bun.lock when root_bun_lock_path does not exist?

masenf (commented):
consider encapsulating common test setup into another fixture to reduce per-test boilerplate that makes it harder to tell what is being tested.

thanks for the contribution, we're working on getting through all of the PR backlog and appreciate your patience

masenf (approved):

PR #6292, opened 2026-04-07
Remove 090 deprecations

adhami3310:


masenf (commented):
update state_auto_setters in BaseConfig

PR #6296, opened 2026-04-07
Improve Lighthouse Scrore

### LINKED ISSUE #6295: Improve lighthouse performance scores (desktop and mobile)
Lighthouse is a webpage performance benchmark built into Chrome browsers. It evaluates pages across four categories: **Performance**, **Accessibility**, **Best Practices**, and **SEO**.

Reflex apps in prod mode should consistently score in the 90s for all categories in the default configuration.

## Current State

Out-of-the-box Reflex apps can score lower than expected on Lighthouse audits, particularly on mobile. Common issues include:

- **Large JS bundle size** — the compiled frontend ships more JavaScript than necessary, increasing Total Blocking Time (TBT) and Time to Interactive (TTI).
- **Render-blocking resources** — CSS and JS that block first paint, hurting First Contentful Paint (FCP) and Largest Contentful Paint (LCP).
- **Missing or suboptimal meta tags** — default templates may lack proper `<meta>` viewport, description, or Open Graph tags, lowering the SEO score.
- **Accessibility gaps** — default component markup may be missing ARIA attributes, proper heading hierarchy, color contrast ratios, or focus management.
- **Cumulative Layout Shift (CLS)** — layout shifts during hydration or lazy-loaded content can degrade the performance score.
- **Image handling** — no built-in support for modern formats (WebP/AVIF), responsive `srcset`, or lazy loading defaults.

### Performance (target: 90+)
- Reduce JS bundle size via tree-shaking, code splitting, and dead code elimination
- Defer non-critical JS and CSS; inline critical CSS
- Optimize hydration to minimize TBT
- Minimize CLS during initial load and state hydration
- Serve assets with proper caching headers and compression (gzip/brotli)

### Accessibility (target: 90+)
- Ensure all default components emit semantic HTML with proper ARIA attributes
- Enforce sufficient color contrast in the default theme
- Ensure proper heading hierarchy and landmark regions in default layouts
- Keyboard navigability and focus management for interactive components

### Best Practices (target: 90+)
- Serve all assets over HTTPS by default in prod
- Avoid deprecated APIs in generated frontend code
- Ensure proper image aspect ratios to avoid layout shifts
- Use passive event listeners where applicable

### SEO (target: 90+)
- Generate proper `<title>`, `<meta description>`, and canonical tags by default
- Ensure pages are crawlable — proper use of `robots.txt` and sitemap generation
- Structured data / JSON-LD support for common page types
- Ensure all links have descriptive text and images have alt attributes

## Approach

- Establish a Lighthouse CI benchmark in the test suite that runs against a reference Reflex app in prod mode
- Track scores per category across releases to catch regressions
- Address low-hanging fruit first (meta tags, compression, defer attributes) then tackle deeper issues (bundle splitting, hydration optimization)
- Document best practices for users to further optimize their own apps

FarhanAliRaza:
### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

closes #6295


benedikt-bartscher:
Shouldn't the compression be left to the reverse-proxy/ingress/deployment? If not, consider adding brotli alongside gzip?

masenf:
> Shouldn't the compression be left to the reverse-proxy/ingress/deployment?

Yes, generally. However, better performance can be achieved by configuring the reverse proxy to serve the pre-compressed files so they don't need to be compressed on the fly when requested.

@FarhanAliRaza a good point is raised here though, I think the vite compress plugin should be able to handle output to gzip, brotli, or zstd for completeness. We should add some documentation to the self-hosting page about how to enable and use pre-compressed responses in caddy and nginx.

The other thing is that `sirv` is going away as the production frontend server, and expected to be replaced by Starlette StaticFiles... however neither this nor the AppHarnessProd support pre-compressed files out of the box.

I found this old, possibly unmaintained starlette plugin: https://github.com/magnuswatn/starlette-precompressed-static

It pretty much does what we want with the exception that it doesn't support `html=True` mode, which we do use. I think we could probably get away with creating our own StaticFiles wrapper that provides this functionality and maintaining it in reflex for `--env prod` behavior.

Because compression does touch on the intersection between application and deployment, we need to make sure these options are configurable by the end user. I don't _think_ anyone is directly reverse proxying to a `reflex run --env prod --single-port` instance, but they definitely could be. And if this ends up serving a compressed response, but the reverse proxy is expecting to compress the response itself, then we could end up with encoding mismatches.

packages/reflex-base/src/reflex_base/.templates/web/utils/helpers/upload.js
```diff
@@ -45,7 +44,7 @@ export const uploadFiles = async (
     // So only process _new_ chunks beyond resp_idx.
     chunks.slice(resp_idx).map((chunk_json) => {
       try {
-        const chunk = JSON5.parse(chunk_json);
+        const chunk = JSON.parse(chunk_json);
```
masenf, line 47: can this reuse the parsing method that the main websocket dispatch loop is using? these chunks could be _anything_ from the backend, including weird floats

packages/reflex-base/src/reflex_base/.templates/apps/blank/code/blank.py
```diff
@@ -11,26 +11,37 @@ class State(rx.State):
...
+                ),
+                rx.button(
+                    rx.link(
+                        "Check out our docs!",
+                        href="https://reflex.dev/docs/getting-started/introduction/",
+                        is_external=True,
+                        underline="none",
+                    ),
+                    as_child=True,
+                    high_contrast=True,
```
masenf, line 32: do these changes actually affect the lighthouse score?

packages/reflex-base/src/reflex_base/components/dynamic.py
```diff
@@ -42,6 +42,17 @@ def reset_bundled_libraries() -> None:
     bundled_libraries.extend(DEFAULT_BUNDLED_LIBRARIES)
 
 
+# Captured during Component serialization so ``collect_window_library_imports``
+# can expose tags reachable only through eval'd code on ``window.__reflex``;
+# without this, ``evalReactComponent`` would fail to resolve them.
+dynamic_component_imports: dict[str, set[imports.ImportVar]] = {}
```
masenf, line 48: we should probably move this to RegistrationContext (I already have #6382 moving some other globals over). The idea with RegistrationContext is that all of the "globals" associated with an App can live there, allowing two different apps to exist with their own config, etc in the same process.

masenf, line 48: I'm also slightly concerned that this might miss dynamic components that are constructed at runtime in event handlers -- the compiler only sees these as `initial_state`, which may not follow conditional paths causing some other imports to be needed.

Ideally, if the user specifically calls `bundle_library`, that library should always get bundled into the frontend regardless of if it's used anywhere

reflex/compiler/compiler.py
```diff
@@ -280,17 +325,139 @@ def _validate_stylesheet(stylesheet_full_path: Path, assets_app_path: Path) -> N
...
+        is True if any Theme has a non-literal (state-driven) color, in which
+        case the caller should fall back to the monolithic stylesheet.
+    """
+    accents: set[str] = set()
+    grays: set[str] = set()
+    has_dynamic = False
+    for root in roots:
+        if root is None:
+            continue
+        for node in _walk_components(root):
```
masenf, line 408: i think we need to update this to plug into the new compiler tree walk now that it's merged in

reflex/compiler/compiler.py
```diff
@@ -280,17 +325,139 @@ def _validate_stylesheet(stylesheet_full_path: Path, assets_app_path: Path) -> N
...
+            if getattr(node, "tag", None) != "Theme":
```
masenf, line 409: other components besides `Theme` can reference theme colors; most radix components have mapped the reflex `color_scheme` prop to the underlying radix themes `color` prop which accepts a color name. We also have the `rx.color` helper which takes radix themes color names and translates them into CSS for use with any component. Not clear if those cases are also handled here.

reflex/testing.py
```diff
@@ -801,115 +801,43 @@ def expect(
...
-            },
-        ) as self.frontend_server:
-            frontend_path = config.frontend_path.strip("/")
-            self.frontend_url = "http://localhost:{1}".format(
-                *self.frontend_server.socket.getsockname()
-            ) + (f"/{frontend_path}/" if frontend_path else "/")
-            self.frontend_server.serve_forever()
+        config = get_config()
+        frontend_app = Starlette()
+        frontend_app.mount(
```
masenf, line 823: for prod mode, i don't think we even need to do this anymore. if we just run the prod backend without `backend_only` mode, then we'll get this kind of behavior out of the backend server and it will more closely resemble how the app actually runs in prod mode.

if we can avoid drift between how the framework really runs and how the test harness runs, we'll be more likely to catch issues during testing.

masenf (commented):
quite a bit of feedback here.

my suggestion would be to break up this PR into smaller chunks so we can bring it in incrementally.

i'd probably start with the test infrastructure for running lighthouse in CI, then bring each component in a separate PR so we can see the effect each change has on the lighthouse score in the test.

for particularly hairy changes or ones that might have edge cases, we can make an informed determination whether the added complexity is worth the perf/accessibility/seo improvement on the test rubric.

PR #6337, opened 2026-04-16
Add AGENTS.md

masenf:
Get some sanity when developing with Claude, Cursor, and friends.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,115 @@
+# Coding Agent Guidelines
+
+Reflex: Python web **framework** compiling to React. Monorepo using uv workspace — main package in `reflex/`, sub-packages in `packages/`, docs site in `docs/`.
+
+## Workflow
+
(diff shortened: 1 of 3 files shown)


FarhanAliRaza:
I think some things can be added here; I don't know if they are just my personal preference or can be added globally. But these things are what i noticed AI does badly generally.

About performance 

* Do not add more work-intensive functions like (isinstance checks) to bypass naive problems (like type checking). Fix code at the fundamental level. 
* Do not repeat validation and be over-defensive. (by adding if conditions everywhere to revalidate already validated data upstream)
* Think about performance in terms of machine-level CPU cycles, like data copy is always slow.

About Testing

* We do not want bogus tests that pass even the thing it is testing is failing. 
* We want to use RED and GREEN test-driven development..
* Tests should fail without bug fixes that we did.

masenf:
> We want to use RED and GREEN test-driven development..
> Tests should fail without bug fixes that we did.

I think these two are already represented in the current doc. I'll add the others

adhami3310 (approved):

PR #6341, opened 2026-04-18
Use full_raw_path for heading anchor hrefs

opadron:
## Summary

`HeadingLink` built its anchor `href` from `router.page.full_path`, the normalized (registered-route) path. When `frontend_path` is set, that path is stripped of the mount prefix, so the generated `href` drops the prefix too — clicking jumps to a route that doesn't exist on the ingress, and anchor-only navigation on the same page lands on a 404. Switch to `full_raw_path`, which is the URL the client actually requested and preserves the `frontend_path` prefix.

When `frontend_path` is unset, `full_path == full_raw_path`, so this is a no-op for the default case.

The paired `rx.set_clipboard(href)` in `on_click` reads the same local, so the "copy link" action and the rendered href stay consistent.

## Regression test

Added `tests/units/reflex_ui_shared/components/blocks/test_headings.py` which walks the rendered component tree and asserts the href references `router.page.full_raw_path` rather than `full_path`.

### DIFF
--- a/packages/reflex-ui-shared/src/reflex_ui_shared/components/blocks/headings.py
+++ b/packages/reflex-ui-shared/src/reflex_ui_shared/components/blocks/headings.py
@@ -112,7 +112,7 @@ class HeadingLink(rx.link.__self__):
             The component.
         """
         id_ = cls.slugify(text)
-        href = rx.State.router.page.full_path + "#" + id_
+        href = rx.State.router.page.full_raw_path + "#" + id_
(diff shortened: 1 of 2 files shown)


carlosabadia (approved):

opadron:
@carlosabadia let's hold off on this for the time being.  @masenf had suggested a better way to fix the issue in question.

masenf:
fixed in main #6397

PR #6359, opened 2026-04-22
Updates to support docs frontend path

carlosabadia:


docs/app/rxconfig.py
```diff
@@ -1,17 +1,20 @@
+import os
+
 import reflex as rx
 
 from agent_files import AgentFilesPlugin
 
 config = rx.Config(
     app_name="reflex_docs",
     deploy_url="https://reflex.dev",
+    frontend_path=os.environ.get("REFLEX_FRONTEND_PATH", "/docs"),
```
masenf: REFLEX_FRONTEND_PATH will take over anything configured here automatically, no need to explicitly read the value here.

masenf (approved):

PR #6370, opened 2026-04-23
ENG-9029: EventHandlerAPIPlugin (enterprise) docs

masenf:


picklelo (dismissed):
working well for me

picklelo:
Tested end-to-end — feature works great. One concrete issue worth fixing before merge:

The **"Driving the app from an LLM"** section uses `http://localhost:3000` in all the example prompts, but in standard dev mode `/.well-known/api-catalog` and `/_reflex/*` routes only exist on the backend port (`:8000`). Hitting `:3000` returns a 404 — confirmed this locally.

The port warning is in the Configuration section but it's far from the LLM examples where someone copy-pasting will actually run into this.

Suggest either:
- Change the LLM example prompts to use `localhost:8000` explicitly, or
- Add a short note in the "Driving the app from an LLM" section like: *"In dev mode, point agents at `localhost:8000` (backend port) not `localhost:3000`"*

Everything else is accurate — seeded tickets, CRUD via curl, `retrieve_state`, streaming NDJSON all work exactly as documented.

picklelo (approved):

PR #6387, opened 2026-04-25
docs: redesign API Reference (interactive demo + props table polish)

Alek99:
## Summary

Redesigns the per-component API Reference block in the docs site:

- **Live demo + code panel side-by-side**: the rendered component sits next to its Python source. The code is parsed from the same preview source the demo uses (so children like `rx.icon`, `placeholder=...` show up correctly), and is bound to the same state vars as the demo — selecting `variant=surface` updates both the rendered button and `variant=\"surface\"` in the code. Manual syntax highlighting (function call / string / bool tokens) is applied via colored spans.
- **Pill-button controls** scoped to styling props (`variant`, `size`, `color_scheme`, `radius`, `high_contrast`, `loading`, `disabled`). Components with no styling props skip the demo + controls and go straight to the props table.
- **Props table rework**: columns are now `prop | type | default | description` with descriptions inline (no more info-icon hover). Literal types render as individual chips that wrap to multiple lines. Long rows clamp to ~4 lines with a fade-out mask across all four columns and a centered `Show more / Show less` toggle row beneath; the data row's bottom border is suppressed so the toggle reads as part of the same logical row.
- **Dark-mode polish**: sidebar bg matches the code-block bg, active sidebar highlight swapped to `slate-3`, active text uses `slate-12` (was violet), the "Build with AI" violet badge + gradient toned down for dark mode.
- **Right-hand sidebar**: "Copy to markdown" and "Ask AI about this page" removed (the AI dropdown above the page already covers these).

## Test plan

- [ ] Visit `/docs/library/forms/button/`: verify the demo + code render side-by-side, pill controls update both, prop table descriptions are inline, type chips wrap.
- [ ] Visit `/docs/library/forms/input/` and scroll to `rx.input.slot`: code panel reflects the actual preview structure (icon child + placeholder), not just `(...)`.
- [ ] Visit a long-prop component (e.g. `Button`) and click `Show more` on the `role` / `color_scheme` rows.
- [ ] Toggle dark mode: confirm the left sidebar bg matches the code-block bg, AI dropdown popup is opaque in both modes, the "Build with AI" badge is no longer overly bright.
- [ ] Visit a component without styling props (e.g. `rx.upload`): confirm no demo + controls panel appears.


docs/app/reflex_docs/pages/docs/component.py
```diff
@@ -79,6 +80,40 @@ class PropDocsState(rx.State):
 ]
 
 
+_PILL_BTN_CLASS = (
+    "font-small cursor-pointer rounded-md px-2.5 py-1 text-slate-11 "
```
carlosabadia: prefer to use `text-sm font-medium` instead of `font-small` (old style)

also, prefer to use the updated palette with the secondary theme variants: `text-secondary-*` and `bg-secondary-*`

docs/app/reflex_docs/pages/docs/component.py
```diff
@@ -507,12 +591,191 @@ def generate_props(
...
+                    rx.code(
+                        prop.name,
+                        class_name="code-style text-nowrap leading-normal text-slate-11",
+                    ),
+                    control,
+                    class_name="grid grid-cols-[8rem_1fr] gap-4 items-start",
+                )
+                for prop, control in interactive_controls
+            ],
+            class_name="flex flex-col gap-3 border border-slate-4 rounded-md p-4 bg-slate-2 mb-4 w-full",
```
carlosabadia: same as before, we can use the `border-secondary-*` and `bg-secondary-*` classes

docs/app/reflex_docs/pages/docs/component.py
```diff
@@ -524,32 +787,30 @@ def generate_props(
                 rx.table.header(
                     rx.table.row(
                         rx.table.column_header_cell(
-                            "Prop",
-                            class_name=table_header_class_name,
+                            "prop",
+                            class_name=table_header_class_name + " w-[9rem]",
```
carlosabadia: we can use the `ui.cn(...)` tailwind merge utility for the class names

docs/app/reflex_docs/templates/docpage/docpage.py
```diff
@@ -773,7 +792,7 @@ def wrapper(*args, **kwargs) -> rx.Component:
                         sidebar,
                         class_name=(
                             "w-[19.5rem] shrink-0 hidden lg:block z-10 border-r border-m-slate-4 dark:border-m-slate-10 sticky left-0 "
-                            "before:content-[''] before:absolute before:top-0 before:bottom-0 before:right-0 before:w-[100vw] before:bg-white-1 dark:before:bg-m-slate-11 before:-z-10 "
+                            "before:content-[''] before:absolute before:top-0 before:bottom-0 before:right-0 before:w-[100vw] before:bg-white-1 dark:before:bg-[#1a1b1d] before:-z-10 "
```
carlosabadia: I'd avoid harcoded colors like this and use the system palette

carlosabadia:
The color grid isn't even 
image

carlosabadia:
i would make the table header uppercase 
image

carlosabadia:
should we add a copy button on the api reference component code snippet 
image

Alek99:
> should we add a copy button on the api reference component code snippet image

This was harder than expected will have to be a follow up

tgberkeley (dismissed):

carlosabadia (dismissed):

Kastier1:
image
rx spinner should have size be configurable. 

image
according orientation doesn't seem to do anything 

image
the progress bar looks bad in the configuration menu

image
can we make the badge a bit bigger? its hard to see the radius changes with it that small

image
we should prob remove this configuration one...

image
missing api reference

image
missing api reference

image
seems like all dynamic render components are missing api refs

image
input has test in it for somereason

image
color never seems to change the switch

image
this one seems kinda usless, either we add more config options or we should remove it 

image
do we expect to have a non hyperlinked link here?

image
lets add more configuration options

image
clicking on html and html embbedding feels weird....

image
missing api reference

image
remove this config example

image
lets remove this config example as well

image
this looks kinda bad ....

image
this config examples does not work
image

image
the align here doesnt seem to do anything

image
align doesnt seem to do anything here as well

image
why do we always add high contrast but almost never add size or a number of the other config options... this one could use some love... honestly same goes for all typographly

Kastier1:
image
these cards here dont really make sense and sometimes they are the same color which makes them look even worse

image
this link doesnt work in `docs/getting-started/installation/` page bottom card

image
these links do not work either `docs/getting-started/introduction/` at the bottom

Okay It looks like all link here a broken that link to `/docs/docs`

Missing API Reference for all dynamic rendering. Example:
`docs/library/dynamic-rendering/auto-scroll/`

only seems to update card 5 here... not sure if intended
`docs/library/layout/grid/#api-reference`

should remove this interactive api ref, it does nothing:
`docs/library/layout/section/#api-reference`

This page is TOOO BIG it takes why to long for me to load this page on local. We need to split up this page somehow
`docs/library/other/html/`

This is missing api ref section
`docs/library/other/memo/`

remove this interactive api ref, it does nothing
`docs/library/overlay/tooltip/#api-reference`

carlosabadia (approved):

PR #6395, opened 2026-04-27
ENG-9414: Remove hardcoded docs urls

carlosabadia:


adhami3310:
i would prefer if we didn't, it makes less digestable to AI

carlosabadia:
> i would prefer if we didn't, it makes less digestable to AI

that's true indeed

see this @Alek99

adhami3310:
this should instead be a CI/check step that all internal doc links are valid

masenf (commented):
one thing we need to solve with this is getting the actual url to be rendered when the user accesses the `.md` link. right now, it's just showing the placeholder, which makes the link unusable.

there's another issue i think we should address #6436, which is indicating the markdown files served via `.md` are utf-8. on my mac, both safari and chrome are defaulting to latin-1 or something and it's garbling some characters.

carlosabadia:
Made a check/ci as suggested https://github.com/reflex-dev/reflex/pull/6448

PR #6411, opened 2026-04-28
ENG-9602: feat: add frontend inspector for mapping DOM nodes to Python source

FarhanAliRaza:
- Add `frontend_inspector`, `frontend_inspector_shortcut`, and `frontend_inspector_editor` config options; "dev" mode fails the prod build (validated at Config init and re-checked at compile time).
- New `reflex_base.inspector` package: state flag, capture registry that records source location per component, and an integration module owning compile-time wiring.
- `Component.create` stamps emitted tags with a `data-rx-source` id when capture is enabled (Fragments skipped).
- Vite config and package.json now react to inspector state and are re-rendered each compile; `launch-editor` pulled in as a dev dep on demand.
- Extract framework-frame walking from `console.py` into shared `reflex_base.utils.frames`, reused by the inspector.
- Exclude `_inspector_id` from generated pyi stubs.
- Docs: new `advanced_onboarding/frontend_inspector.md` page, sidebar entry, and pointer from the configuration guide.
- Tests for config validation, runtime gating, capture lifecycle, frame walking, browser contract, and the component data-attribute.

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

Alt-click actually opens the file in the editor. 

image


packages/reflex-base/src/reflex_base/config.py
```diff
@@ -255,6 +260,12 @@ class BaseConfig:
 
     transport: Literal["websocket", "polling"] = "websocket"
 
+    frontend_inspector: FrontendInspectorMode = "off"
+
+    frontend_inspector_shortcut: str = "alt+x"
+
+    frontend_inspector_editor: str = ""
+
```
adhami3310: i would rather this be implemented as a plugin, and the config options belong in that plugin, similar to sitemap

masenf (commented):
agree with full implementing this as a plugin.

not sure if it's possible, but the first thing i wished this had was the ability to see the _stack_ of where the component was instantiated. maybe i want to jump to the component instantiation? or maybe i want to jump into the page function or sidebar function that called the helper.

lot of potential here though; hopefully we can integrate this into claude code's preview browser to automatically select context from UI 🤤

FarhanAliRaza:
we currently send this data for each component and dedupe it. 

```
    "565": {
        "file": "/home/farhan/code/reflex/docs/app/reflex_docs/docgen_pipeline.py",
        "line": 625,
        "column": 1,
        "component": "AccordionContent"
    },
 ```
 
 
 
 We can store whatever data it is our will, but if we do alot of work it will slow down compilation and increase size of the source map that we send to the frontend. Now for the docs app we send about 500+ kb. That is for development so size is, I think, not an issue.

File: packages/reflex-base/src/reflex_base/plugins/base.py [hunk 2]
```diff
@@ -105,6 +105,19 @@ def get_frontend_dependencies(
         """
         return []
 
+    def start_compile(self, app: "App") -> None:
```
adhami3310: this isn't using the context pattern

FarhanAliRaza: fixed

FarhanAliRaza:
Need to test it with memoized components.

PR #6420, opened 2026-04-29
Fix broken docs links and updates for agents

carlosabadia:


docs/app/agent_files/_plugin.py
Alek99, line 1: The dynamic API docs pages are routed with module.__name__.lower() in apiref.py, so pages like /api-reference/eventhandler/ would map to /api-reference/eventhandler.md when users follow the new “append .md” guidance. This code emits api-reference/event-handler.md instead, so several multi-word API pages get markdown URLs that do not match their page URLs. Either generate markdown with the existing route slug or update the page routes and redirects together.

Alek99 (approved):

PR #6422, opened 2026-04-29
fix(pyi): incremental .pyi regen now follows the import graph

FarhanAliRaza:
Fixed

1. Fixed partial generation, before it dropped the unchanged in files.
2. fixed the inherited changes that were not happening before.

### All Submissions:

### Type of change

### New Feature Submission:

### DIFF
--- a/packages/reflex-base/src/reflex_base/utils/pyi_generator.py
+++ b/packages/reflex-base/src/reflex_base/utils/pyi_generator.py
@@ -1717,10 +1717,14 @@ class PyiGenerator:
 
         if use_json and (file_paths or file_targets):
             file_paths = list(map(Path, file_paths))
-            anchor = (
-                file_paths[0].parent if file_paths else file_targets[0].resolve().parent
-            )
-            pyi_hashes_parent = anchor
+            # Anchor the pyi_hashes.json search at cwd, not at the first written
+            # file. The hash file is workspace-scoped — there is one for the whole
+            # repo — so walking up from cwd is what the caller (make_pyi.py, run
+            # from the repo root) actually means. Anchoring at file_paths[0] would
+            # make the result depend on which file happened to land first in
+            # `written_files`, and could find a package-local hash file in a
+            # leaf-package walk-up before reaching the workspace root.
+            pyi_hashes_parent = Path.cwd().resolve()
             while (
                 pyi_hashes_parent != pyi_hashes_parent.parent
                 and not (pyi_hashes_parent / PYI_HASHES).exists()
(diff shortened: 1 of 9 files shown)


packages/reflex-base/src/reflex_base/utils/pyi_generator.py
```diff
@@ -1607,7 +1607,10 @@ class PyiGenerator:
     modules: list = []
     root: str = ""
     current_module: Any = {}
-    written_files: list[tuple[str, str]] = []
+
+    def __init__(self) -> None:
+        """Initialize per-instance scan state."""
+        self.written_files: list[tuple[str, str]] = []
```
adhami3310: shouldn't the same be done to modules and current module?

packages/reflex-base/src/reflex_base/utils/pyi_generator.py
```diff
@@ -1712,69 +1715,51 @@ def scan_all(
...
-                        formatted_path = file_path.relative_to(
-                            pyi_hashes_parent
-                        ).as_posix()
-                        pyi_hashes[formatted_path] = hashed_content
-
-                    pyi_hashes_file.write_text(
-                        json.dumps(pyi_hashes, indent=2, sort_keys=True) + "\n"
-                    )
+        if use_json and (file_paths or file_targets):
+            file_paths = list(map(Path, file_paths))
```
adhami3310: this might be a good time to take PYI hash generation into its own function

packages/reflex-base/src/reflex_base/utils/pyi_generator.py
FarhanAliRaza: They need to be removed. They are not used.

packages/reflex-base/src/reflex_base/utils/pyi_generator.py
FarhanAliRaza: fixed

tests/units/reflex_base/utils/pyi_generator/test_hash_merge.py
```diff
@@ -0,0 +1,229 @@
...
+from pathlib import Path
+
+from reflex_base.utils.pyi_generator import PyiGenerator
+
+
+def _write_hashes(path: Path, mapping: dict[str, str]) -> None:
+    path.write_text(json.dumps(mapping, indent=2, sort_keys=True) + "\n")
+
+
+def _make_workspace(root: Path) -> Path:
```
masenf: this should be a fixture that returns the workspace path based on the tmp_path fixture

masenf (commented):
i might be reading it wrong, but are we regenerating modules that are imported by modules that changed? the other fixes seem legit.

FarhanAliRaza:
For example if react_player is changed, in the main branch now it only generates react_player.pyi because it only sees this in git diff. But it should have also changed audio.pyi and video.pyi because they import react player amd inherit from it . Before partial generation missed it. Only full regneration catches that. 

So we are regenerating that imports the changed modules.

PR #6424, opened 2026-04-29
Fix flakiness in integration/test_upload

masenf:
Just a huge claude+caffeine fueled overhaul of the upload integration suite.

* Clean up uploads dir before and after tests
  * Make sure `test_cancel_upload` actually gets a chance to retry
* Poll for upload to be cancelled (progress stops updating)
  * Remove unnecessary long static sleeps while keeping the test robust
* Use helper functions for getting progress dicts and getting the upload box, two operations that had fiddly xpath + brittle indexing
* Extract slow network condition fixture
* Remove `secondary` parametrization in favor of `upload_root_id`

### DIFF
--- a/tests/integration/test_upload.py
+++ b/tests/integration/test_upload.py
@@ -398,11 +398,17 @@ def driver(upload_file: AppHarness):
 
 
 @pytest.fixture
-def simulate_slow_network(driver: WebDriver) -> None:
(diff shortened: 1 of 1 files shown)


masenf:
@greptile-apps re-review

adhami3310 (approved):

PR #6425, opened 2026-04-29
delete custom components

adhami3310:
question: should we do this?


Alek99:
Lets hold off for now

masenf (commented):
if we do it, i think we should have a new template repo for reflex component packages. it would be great if this template had all the latest best practices: uv, pyi-on-build, pre-commit, trusted publishing, test suite, AGENTS.md, etc.

i think the community would get better use out of a template repo with everything already set up.

PR #6426, opened 2026-04-30
update mcp docs

adhami3310:
check https://github.com/reflex-dev/agent-skills/pull/10 as well


Alek99:
I don't think this documentation sufficiently covers the landscape of code gen tooling see how they do it https://resend.com/mcp

We should update the MCP docs to cover the following.

Screenshot 2026-04-29 at 6 24 18 PM

Alek99 (approved):

PR #6446, opened 2026-05-03
fix(exec): correct frontend URL when frontend_path lacks a leading slash (#6360)

### LINKED ISSUE #6360: frontend_path without a leading slash results in incorrect message printed
**Describe the bug**
specify `REFLEX_FRONTEND_PATH=noslash reflex run`

In startup the wrong url is printed:

```
[INFO] Started worker-1
c
App running at: http://localhost:3001/noslash/noslash/
Backend running at: http://0.0.0.0:8000
```

**To Reproduce**
Specify `frontend_path` without a leading slash -- it gets formatted incorrectly during to the use of `urljoin` having a different behavior for "relative" vs "absolute" path segments

- Python Version: any
- Reflex Version: 0.9.0
- OS: macOS / any
- Browser (Optional): n/a

ChinmayShringi:
### All Submissions:

### Type of change

### Changes To Core Features:

---

closes #6360

### Description

Setting `REFLEX_FRONTEND_PATH=noslash` (no leading slash) caused startup to print a URL with a duplicated segment:

```
App running at: http://localhost:3001/noslash/noslash/
```

#### Root cause

`urllib.parse.urljoin` treats a relative path differently from an absolute one. Vite already prints its listening URL with the configured base appended (e.g. `http://localhost:3001/noslash/`), so:

- `urljoin("http://localhost:3001/noslash/", "/noslash")` returns `http://localhost:3001/noslash` (absolute path replaces).
- `urljoin("http://localhost:3001/noslash/", "noslash")` returns `http://localhost:3001/noslash/noslash` (relative path appends).

#### Fix

Extract `_apply_frontend_path(base_url, frontend_path)` in `reflex/utils/exec.py` that normalizes the configured path to `/path/` form before joining. The result is now independent of how the user wrote `frontend_path`:

| `frontend_path` | Result |
|---|---|
| `"/app"` | `http://localhost:3001/app/` |
| `"app"` | `http://localhost:3001/app/` |
| `"/app/"` | `http://localhost:3001/app/` |
| `"app/v1"` | `http://localhost:3001/app/v1/` |

The helper is pure, so it is unit-tested directly without mocking the dev-server subprocess.

### Test plan

- `uv run pytest tests/units/utils/test_exec.py` (9 parametrized cases pass)
- `uv run ruff check reflex/utils/exec.py tests/units/utils/test_exec.py`
- `uv run pyright reflex/utils/exec.py tests/units/utils/test_exec.py` (0 errors)
- Manual: `REFLEX_FRONTEND_PATH=noslash reflex run` now prints `http://localhost:3001/noslash/` (one segment).

The new test file covers: empty path, leading-slash, no-leading-slash, trailing-slash, both-slashes, multi-segment path, and the bug scenario where vite has already baked the path into its listening URL.

### DIFF
--- a/reflex/utils/exec.py
+++ b/reflex/utils/exec.py
@@ -159,6 +159,29 @@ def notify_frontend(url: str, backend_present: bool):
     )
 
 
+def _apply_frontend_path(base_url: str, frontend_path: str) -> str:
+    """Apply ``frontend_path`` to a base URL emitted by the frontend dev server.
+
+    ``urljoin`` treats a relative path differently from an absolute one, so
+    a ``frontend_path`` configured without a leading slash (e.g. ``"app"``)
+    used to produce duplicated path segments like
+    ``http://localhost:3001/app/app/`` (issue #6360). Normalizing the
+    configured path to ``/path/`` form before joining makes the behavior
+    consistent regardless of how the user wrote it.
+
+    Args:
+        base_url: The base URL printed by the frontend dev server.
+        frontend_path: The configured ``Config.frontend_path``.
+
+    Returns:
+        The URL with the frontend path correctly applied.
+    """
+    if not frontend_path:
+        return base_url
+    normalized = "/" + frontend_path.strip("/") + "/"
+    return urljoin(base_url, normalized)
+
+
 def notify_backend(host: str | None = None):
     """Output a string notifying where the backend is running.
 
@@ -226,9 +249,9 @@ def run_process_and_launch_url(
                 match = re.search(constants.ReactRouter.FRONTEND_LISTENING_REGEX, line)
                 if match:
                     if first_run:
-                        url = match.group(1)
-                        if get_config().frontend_path != "":
-                            url = urljoin(url, get_config().frontend_path)
+                        url = _apply_frontend_path(
+                            match.group(1), get_config().frontend_path
+                        )
 
                         notify_frontend(url, backend_present)
                         if backend_present:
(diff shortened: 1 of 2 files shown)


masenf:
was there something wrong with #6364

ChinmayShringi:
@masenf #6364 (by @qubydev) takes a different approach to the same issue (#6360): it removes the `urljoin` call entirely on the assumption that Vite already bakes `frontend_path` into the listening URL. That works for the dev server but breaks the prod/preview path where the listening URL doesn't include the base, so the path silently goes missing.

This PR keeps the join but normalizes `frontend_path` to `/path/` form first, so both cases work regardless of whether the user wrote `app`, `/app`, `/app/`, or it's already in the URL. Tests cover all those forms plus the Vite-baked-URL scenario.

Also just pushed a fix for the greptile P2: `frontend_path='/'` was producing `'//'` which urljoin treats as protocol-relative. Now guarded.

masenf:
> That works for the dev server but breaks the prod/preview path where the listening URL doesn't include the base, so the path silently goes missing.

The prod path does not go through `sirv` anymore. When the user does `reflex run --env prod`, under the hood reflex does an export and then serves that export via Starlette StaticFiles

ChinmayShringi:
You're right, my bad, I didn't know the prod path had moved off `sirv` to Starlette `StaticFiles`. Just confirmed: `run_process_and_launch_url` is dev-only now, so the urljoin call really is redundant and #6364's approach is the correct fix. #6364 also normalizes `frontend_path`/`backend_path` at config load time, which addresses the root cause better than my surface-level helper.

Closing this in favor of #6364. Sorry for the noise @qubydev, your PR was the right call.

qubydev:
@ChinmayShringi can you please help me identify the reason of the failure of some test cases although it passes on my local device. (Please checkout the PR)

PR #6447, opened 2026-05-04
feat: add app_wraps to VarData for Var-driven provider injection

FarhanAliRaza:
Lets Vars declare app-level wrapper components in their VarData so the compiler can mount providers (state context, event loop, upload, etc.) based on what's actually used, instead of relying on hardcoded chains or special-case logic. State/event-loop providers now ride along on VarData.from_state and the events-hook helper, and UploadFilesProvider is mounted when selected_files/upload_file is referenced — even without an Upload component on the page. Layout renders the assembled chain so AppWrap reduces to hooks + children.

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

fixes ENG-9152


reflex/compiler/plugins/memoize.py
```diff
@@ -136,6 +140,12 @@ def _component_subtree_is_reactive(
             return True
     if component._get_hooks() is not None or component._get_added_hooks():
         return True
+    # ``addEvents`` no longer hoists a hook here (it's reached via the
+    # module-level import in ``Imports.EVENTS``), so a no-arg
+    # ``on_click=State.ping`` only shows up as ``event_triggers`` — without
+    # this check the boundary skips memoization and the callback leaks.
+    if component.event_triggers:
+        return True
```
masenf: i think it's okay to allow the callback to leak into the main page here, because it's not something that will cause a re-render since `addEvents` is now a statically imported function.

FarhanAliRaza: fixed

packages/reflex-base/src/reflex_base/components/component.py
```diff
@@ -2132,6 +2163,30 @@ def _get_app_wrap_components() -> dict[tuple[int, str], Component]:
...
 
+    def _get_event_app_wraps(self) -> dict[tuple[int, str], Component]:
+        """Return state/event-loop providers required by event triggers.
+
+        Kept separate from :meth:`_get_app_wrap_components` so subclass
+        overrides of that method don't inadvertently strip these out —
+        the providers must be in the React tree for ``StateContexts``,
+        ``EventLoopContext``, and the websocket plumbing to stay alive
+        even though ``addEvents`` itself is reached via module-level
+        import rather than a hook closure.
```
masenf: this is very confusing imo

packages/reflex-base/src/reflex_base/vars/base.py
```diff
@@ -259,6 +269,16 @@ def merge(*all: VarData | None) -> VarData | None:
             component for var_data in all_var_datas for component in var_data.components
         )
 
+        app_wraps_seen: set[tuple[int, str]] = set()
+        app_wraps_list: list[tuple[int, BaseComponent]] = []
+        for var_data in all_var_datas:
+            for priority, wrapper in var_data.app_wraps:
+                key = (priority, wrapper.tag or type(wrapper).__name__)
+                if key in app_wraps_seen:
+                    continue
```
masenf: i think identical tuple values defined later should take precedent or we should just immediately fail if the wrapper is being overwriten and isn't the same. silently taking the first (priority, wrapper) tuple seems like a silent bug waiting to happen

masenf (commented):
haven't gotten through all of the review yet. initial comments

packages/reflex-base/src/reflex_base/components/component.py
FarhanAliRaza: clarified the comment.

packages/reflex-base/src/reflex_base/vars/base.py
FarhanAliRaza: now it raises an error.

packages/reflex-base/src/reflex_base/event/__init__.py
```diff
@@ -2098,11 +2098,14 @@ def _dispatch_mixed_event_var(event_like_var: Var) -> FunctionVar:
...
         )
+        # Lazy import: state_context → component → event (this module).
+        from reflex_base.components.state_context import get_event_app_wraps
+
         add_events = FunctionStringVar.create(
             CompileVars.ADD_EVENTS,
             _var_data=VarData(
                 imports=Imports.EVENTS,
-                hooks={Hooks.EVENTS: None},
+                app_wraps=get_event_app_wraps(),
```
masenf, line 2108: if we're putting the app wraps on the event chains themselves, do we still need Component._get_event_app_wraps? they should be picked up because they're on the event vars i would think

FarhanAliRaza: At walk time the EventChain is not converted to Var yet. So we can not read it. Maybe if we added some way to store on EventChain, then maybe we can delete it.

masenf: Sounds good, let's get a backlog ticket capturing some of that context, but we don't have to prioritize it.

masenf (approved):
Working well for me. I'm thinking let's cut a 0.9.4 before we bring this in, but it's ready to merge.

PR #6448, opened 2026-05-04
Add checks for broken docs urls

carlosabadia:


masenf:
@greptile-apps re-review

masenf (commented):
can the github actions workflow be an add-on step for the existing reflex-docs regression. most of the time taken in this workflow is actually building the app, but we already do that in the other workflow, so we basically get the link checking for free.

i also think the script output could be a little more verbose so you can see all of the links that got checked in the CI instead of "All /docs links resolve against sitemap.xml." and having to trust that it actually checked and didn't just scan the wrong dir and find no markdown files.

finally, love the tests for the test script, very nice.

adhami3310:
you can parse the markdown files instead of doing regexes btw, we reflex-docgen has a transformer that can be used for this

adhami3310 (approved):

PR #6454, opened 2026-05-05
Do not memoize parent component of Foreach

masenf:
Avoid issue where broader-than-needed SNAPSHOT memoization was occuring when these components should have been memoized separately.

This is likely a benchmark regression, since before the entire stateful_page was getting wrapped up in a single memo component, so most of the compiler tree walk was getting skipped.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -46,6 +46,7 @@ docs/                   # documentation site (separate workspace member)
 - No block comments (`# --- Section ---`, `# ============`). Plain inline comments only.
 - Be cautious creating new public APIs — they must be documented and supported long-term.
 - Google-style docstrings on all functions: one-line summary, optional detail sentence(s), then Args/Returns (or Yields)/Raises.
+- Prefer imports at the top of the module in isort order. Only use inline imports when necessary to avoid circular dependencies.
 
 ## Testing
 
--- a/packages/reflex-base/src/reflex_base/components/memoize_helpers.py
+++ b/packages/reflex-base/src/reflex_base/components/memoize_helpers.py
@@ -214,26 +214,6 @@ def _is_structural_memoization_child(component: Component) -> bool:
     return bool(isinstance(component, Foreach))
 
 
-def _has_memoization_snapshot_child(component: Component) -> bool:
-    """Whether ``component`` has a structural child that needs a memo snapshot.
-
-    Component-valued ``Foreach`` is a structural form, not an ordinary user
-    child. It must render its body in the memo module instead of flowing
-    through as a normal ``children`` payload.
-
-    Args:
-        component: The component whose direct children should be inspected.
-
-    Returns:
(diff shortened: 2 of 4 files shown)


FarhanAliRaza (dismissed):
We might want to add a full test. 

```python
def test_foreach_parent_does_not_absorb_sibling_into_snapshot() -> None:                                                                      
      """Foreach owns its own snapshot — its parent stays passthrough.                                                                          
                                                                                                                                                
      Regression for the foreach-parent memoization fix: previously a parent                                                                    
      component holding a Foreach was promoted to SNAPSHOT, and any sibling                                                                     
      reactive content under that parent was absorbed into a single wide memo                                                                   
      body. The parent should now render on the page side, with the Foreach                                                                     
      and any reactive sibling each getting their own independent wrapper.                                                                      
      """                                                                                                                                       
      from reflex_components_core.core.foreach import Foreach                                                                                   
                                                                                                                                                
      ctx, _page_ctx = _compile_single_page(                                                                                                    
          lambda: rx.box(                                                                                                                       
              Bare.create(SpecialFormMemoState.items.length()),  # reactive sibling                                                             
              rx.foreach(                                                                                                                       
                  SpecialFormMemoState.items,
                  lambda item: rx.text(item),                                                                                                   
              ),                                                             
          )
      )

      wrapped_types = {                                                                                                                         
          type(defn.component)
          for defn in ctx.auto_memo_components.values()                                                                                         
          if isinstance(defn, ExperimentalMemoComponentDefinition)                                                                              
      }
                                                                                                                                                
      # Parent rx.box is NOT memoized — it renders directly on the page side.                                                                   
      from reflex_components_core.radix.themes.layout.box import Box                                                                            
      assert Box not in wrapped_types                                                                                                           
                                                                             
      # Foreach owns its own SNAPSHOT memo.                                                                                                     
      assert Foreach in wrapped_types                                        
      foreach_defn = next(                                                                                                                      
          defn
          for defn in ctx.auto_memo_components.values()                                                                                         
          if isinstance(defn, ExperimentalMemoComponentDefinition)           
          and isinstance(defn.component, Foreach)                                                                                               
      )
      assert get_memoization_strategy(foreach_defn.component) is MemoizationStrategy.SNAPSHOT                                                   
                                                                                                                                                
      # Sibling Bare gets its own PASSTHROUGH memo, distinct from the foreach                                                                   
      # wrapper — it is not absorbed into the foreach-parent snapshot.                                                                          
      bare_defn = next(                                                                                                                         
          defn                                                               
          for defn in ctx.auto_memo_components.values()                                                                                         
          if isinstance(defn, ExperimentalMemoComponentDefinition)           
          and isinstance(defn.component, Bare)                                                                                                  
      )
      assert get_memoization_strategy(bare_defn.component) is MemoizationStrategy.PASSTHROUGH                                                   
      assert bare_defn is not foreach_defn                                                                                                      
  ```

adhami3310 (approved):

PR #6456, opened 2026-05-05
feat: propagate VarData.module_code into page module code

### LINKED ISSUE #6219: Compiler: Add `module_code` field to VarData
## Summary

Add a `module_code` field to `VarData` so that Vars can declare custom JavaScript code that needs to be present at the module level of the page (or component file) where the Var is rendered. The compiler will pick up `module_code` from Vars via the components those Vars are attached to.

This is the Var-level equivalent of `Component._get_custom_code()`, but with a clearer name. The long-term goal is to standardize on `module_code` throughout the codebase and eventually deprecate the `custom_code` naming.

### Current state: `custom_code` on Component

Today, custom JavaScript snippets that need to be at the module level are handled via `Component._get_custom_code()`:

```python
class MyComponent(rx.Component):
    def _get_custom_code(self) -> str | None:
        return "console.log('This runs at module level');"
```

These are collected recursively via `_get_all_custom_code()` and passed to the page template as `custom_codes=...`.

### What's missing: Vars can't contribute module code

`VarData` currently has these fields:

```python
@dataclasses.dataclass(eq=True, frozen=True)
class VarData:
    state: str = ""
    field_name: str = ""
    imports: ParsedImportTuple = ()
    hooks: tuple[str, ...] = ()
    deps: tuple[Var, ...] = ()
    position: Hooks.HookPosition | None = None
    components: tuple[BaseComponent, ...] = ()
```

Vars can declare `imports` and `hooks`, but they cannot declare module-level code. This means if a Var needs a module-level helper function or constant, it must be handled indirectly through a Component — even when the Var itself is the thing that needs it.

## Proposed Change

Add a `module_code` field to `VarData`:

```python
@dataclasses.dataclass(eq=True, frozen=True)
class VarData:
    state: str = ""
    field_name: str = ""
    imports: ParsedImportTuple = ()
    hooks: tuple[str, ...] = ()
    deps: tuple[Var, ...] = ()
    position: Hooks.HookPosition | None = None
    components: tuple[BaseComponent, ...] = ()
    module_code: tuple[str, ...] = ()  # NEW
```

### Compiler integration

The compiler (or `ConsolidateModuleCodePlugin` in the new single-pass architecture) should collect `module_code` from Vars during compilation. Since Vars are attached to components (via props, event handlers, children), the component tree walk already visits them. The collection path is:

1. For each component, examine its Var-containing fields (props, event handlers, style, children that are Vars)
2. For each Var, check `_var_data.module_code`
3. Add any non-empty module code to the page's module code set

This naturally integrates with the existing `_get_all_custom_code()` flow (or its plugin replacement). The module code from Vars and from Components should be merged into the same output set — they both end up as top-level JS in the page file.

### VarData.merge() update

The `VarData.merge()` static method needs to handle the new field:

```python
@staticmethod
def merge(*others: VarData | None) -> VarData | None:
    ...
    module_code = tuple({mc for od in others if od for mc in od.module_code})
    ...
```

## Naming: `module_code` vs `custom_code`

The existing codebase uses `custom_code` for the same concept on Components. This name is confusing because "custom code" could mean many things. `module_code` is more precise — it's code that goes at the JS module level (top of the file, outside any component function).

This issue introduces `module_code` on `VarData`. A follow-up task (not in scope here) would be to rename `_get_custom_code()` / `_get_all_custom_code()` on Component to `_get_module_code()` / `_get_all_module_code()` and standardize the naming throughout.

## Acceptance Criteria

- [ ] `VarData` has a `module_code: tuple[str, ...]` field (default empty tuple)
- [ ] `VarData.merge()` correctly combines `module_code` from multiple VarData instances (deduplicating)
- [ ] `VarData.__bool__` / truthiness accounts for the new field
- [ ] The compiler collects `module_code` from Vars encountered during page compilation
- [ ] Module code from Vars appears in the compiled page output alongside Component custom code
- [ ] Unit test: a Var with `module_code` attached to a component results in that code appearing in the compiled page
- [ ] All existing tests pass

## Key Files

- `reflex/vars/base.py`:
  - `VarData` dataclass (~line 99) — add the field
  - `VarData.merge()` — update merging logic
- `reflex/components/component.py`:
  - `_get_all_custom_code()` (~line 1590) — should also collect from Vars (or a new method should)
  - `_get_vars()` / `_get_vars_from_props()` — how Vars are currently extracted from components
- `reflex/compiler/compiler.py`:
  - `_compile_page()` — where `custom_codes` are passed to the template
- `reflex/compiler/templates.py`:
  - `page_template()` — receives `custom_codes` parameter

## Notes

- The `module_code` tuple uses strings (like `hooks`), deduplicated by set membership. Order shouldn't matter for module-level code.
- This is a prerequisite for the app_wraps-on-VarData work, where Vars need to declare providers and context code at the app level.
- The frozen dataclass pattern is maintained — `module_code` is a tuple, not a list.

FarhanAliRaza:
Adds module_code on VarData so Vars can contribute top-of-file JS helpers/constants. The default collector and the legacy _get_all_custom_code path both pick it up, ensuring snippets carried by Vars on memoized stateful components aren't dropped from the memo file.

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

closes #6219
fixes ENG-9151


packages/reflex-base/src/reflex_base/components/component.py
```diff
@@ -1948,10 +1991,15 @@ def _get_ref_hook(self) -> Var | None:
     def _get_vars_hooks(self) -> dict[str, VarData | None]:
         """Get the hooks required by vars referenced in this component.
 
+        The Var module_code is gathered during this same walk and cached on
+        ``_var_module_code_cache`` so :meth:`_iter_var_only_module_code` reuses
+        it instead of iterating the Vars (and their VarData) a second time.
+
         Returns:
             The hooks for the vars.
         """
```
masenf, line 2000: this feels like it should be a "`_get_all_var_data()`" because now that it also collects the module code.

i think ultimately we want to simplify how this stuff is stored, combined, and passed around by leveraging the var system.

during the component walk we could grab all of a component's var's data, cache that, and then reuse it in a Var-wrapped literal component or vice versa.

but more specifically, this function has a yucky side effect of populating just a module cache now. if we're going to cache, we should cache a more reusable and standard VarData

reflex/compiler/plugins/builtin.py
```diff
@@ -329,6 +353,34 @@ def _collect_component_custom_code(
...
+        for hook_var_data in hooks_internal.values():
+            if hook_var_data is None:
+                continue
+            for snippet in hook_var_data.module_code:
+                module_code.setdefault(snippet, None)
+        for hook_var_data in added_hooks.values():
+            if hook_var_data is None:
+                continue
+            for snippet in hook_var_data.module_code:
+                module_code.setdefault(snippet, None)
```
masenf, line 382: very similar code in Component too, i don't think we want this kind of thing in both places. i'd rather have this combining stuff in the compiler and get most of the weird recursive collection stuff out of the Component so it can be a very simple class.

packages/reflex-base/src/reflex_base/components/component.py
```diff
@@ -1965,6 +2013,8 @@ def _get_vars_hooks(self) -> dict[str, VarData | None]:
                 )
```
masenf, line 2013: when is var_data.hooks a dict? it's typed as `tuple[str]` currently? 🤔

masenf (commented):
needs a bit of a rework and beginnings of a larger cleanup in the Component code.

i think we can and should now collapse most of the per-component data gathering functions into a centralized `_get_var_data(self) -> VarData`

we might have to slightly change how `positions` are stored on `VarData` so we can more accurately map individual hooks strings to corresponding positions values. we already use explicit priority ints for the app_wraps in VarData, i wonder if it makes sense to adopt that same principal for module_code and hooks.

i think it would be ideal to be able to arbitrarily and cheaply combine any VarData and then dedupe, validate, sort and render their internals once at compile time. this would get more complex logic out of Component too.

masenf:
drafting for now; will bring this in with some overall Component -> VarData refactoring which is planned.

PR #6457, opened 2026-05-05
feat(compiler): mirror memo output paths to Python source modules

### LINKED ISSUE #6218: Compiler: Track provenance of rx.memo components and output to mirrored Python paths
## Summary

When a `rx.memo` (or `rx._x.memo`) component is created, the compiler should inspect the call stack to determine which file in the user's app defined it. The compiled JS output for that memo component should then be written to a path that mirrors the Python source path, rather than being lumped into a single `utils/components.jsx` or `utils/stateful_components.jsx` file.

For example, a memo component defined in `counter_app/ui/buttons.py` would be rendered to `.web/counter_app/ui/buttons.js`.

## Motivation

Today, all custom/memo components are compiled into a single monolithic file (`.web/utils/components.jsx`), and all shared stateful components go into another (`.web/utils/stateful_components.jsx`). This has two problems:

1. **Hot reload granularity**: When a common imported helper changes (e.g., `counter_app/ui/buttons.py` is edited), the entire `components.jsx` file must be rewritten and vite must rebundle it, even if only one memo component changed. If each Python module's memo components were in their own JS file, only that specific file needs to be rewritten, and vite can do a more targeted incremental rebuild.
2. **Debuggability**: When looking at `.web/` output, there's no way to trace a compiled component back to the Python source that defined it. Mirrored paths make this obvious.

### Stack-frame provenance at creation time

When `rx.memo` (currently aliased to `custom_component()`) is called, the decorator should inspect the call stack to find the most recent frame in the user's app code:

```python
def custom_component(component_fn):
    # Walk up the stack to find the caller's module
    frame = inspect.currentframe()
    caller_module = _find_user_app_frame(frame)
    # e.g., "counter_app.ui.buttons"
    
    @wraps(component_fn)
    def wrapper(*children, **props):
        comp = CustomComponent._create(...)
        comp._source_module = caller_module  # attach provenance
        return comp
    ...
```

"User app code" means any module that isn't part of the `reflex` package itself. The heuristic should walk up the stack skipping `reflex.*` frames until it finds the first external frame. However, the "external" frame could be another package… in which case, it's path should be relative to it's package root (typically subdir in site-packages).

### Output path mirroring

During compilation, instead of writing all memo components to a single `utils/components.jsx`, group them by their `_source_module` and write each group to a mirrored path:

```
Python source                        JS output
counter_app/ui/buttons.py     →      .web/counter_app/ui/buttons.jsx
counter_app/ui/sidebar.py     →      .web/counter_app/ui/sidebar.jsx
counter_app/components.py     →      .web/counter_app/components.jsx
```

Components without provenance (e.g., those defined in reflex internals or in the REPL) fall back to the current `utils/components.jsx` location.

### Import rewriting

Pages that use these memo components currently import from `utils/components`. The import paths need to be updated to point to the new mirrored locations:

```js
// Before
import { MyButton } from "$/utils/components"

// After
import { MyButton } from "$/counter_app/ui/buttons"
```

## Acceptance Criteria

- [ ] `rx.memo` / `custom_component()` captures the calling module's path at decoration time
- [ ] The provenance is stored on the `CustomComponent` and flows through to compilation
- [ ] Heuristic correctly skips `reflex.*` internal frames and finds user app code
- [ ] Compiled memo component JS is written to paths mirroring the Python source module
- [ ] Import statements in page files correctly reference the new paths
- [ ] Components without determinable provenance fall back to the current `utils/components.jsx` path
- [ ] Hot reload of a single Python module only rewrites the corresponding JS file (not the monolithic components file)
- [ ] All existing tests pass

## Key Files

* `reflex/components/component.py`:
  * `custom_component()` decorator (\~line 2270) — where provenance capture should happen
  * `CustomComponent` class (\~line 2008) — needs `_source_module` field
  * `memo = custom_component` alias (\~line 2314)
* `reflex/compiler/compiler.py`:
  * `_compile_memo_components()` — currently writes all to one file, needs grouping by source module
  * `compile_memo_components()` — public API, returns `(path, code, imports)`
  * `get_components_path()` — currently returns single path `utils/components.jsx`
* `reflex/compiler/utils.py`:
  * `get_components_path()` (\~line 504) — needs a per-module variant
* `reflex/compiler/templates.py`:
  * `memo_components_template()` — may need adaptation for per-file output

## Notes

* This pairs well with the single-pass compiler work ([ENG-9142](https://linear.app/reflex-dev/issue/ENG-9142/single-pass-compiler-introduce-compilerplugin-protocol-compilerhooks)-[ENG-9149](https://linear.app/reflex-dev/issue/ENG-9149/single-pass-compiler-remove-legacy-multi-threadingmultiprocessing)) but doesn't depend on it — it can be implemented against the current compiler.
* The `_source_module` heuristic should handle edge cases: lambdas, nested functions, dynamically generated components, and components defined in `__init__.py` files.
* Consider whether `StatefulComponent` wrappers (or their replacement from [ENG-9145](https://linear.app/reflex-dev/issue/ENG-9145/single-pass-compiler-replace-statefulcomponent-auto-memoization-with)) should also track provenance. If a stateful wrapper is generated from a component in `buttons.py`, the wrapper's JS should probably live alongside it.
* This is a stepping stone toward more granular hot reload — if each Python module maps to a specific JS file, the compiler can skip recompiling modules whose source hasn't changed.

FarhanAliRaza:
Memos now compile into a single JSX file per user module at a path that mirrors the module's dotted name, instead of one file per memo under . The page-side import surface matches the source
layout, which makes debugging easier and lets Vite group co-defined memos in the same chunk.

Memos without a captured source module keep the legacy per-name files and  index. A manifest in  records emitted
paths so stale files from previous compiles get pruned.

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

closes #6218
fixes ENG-9150


reflex/utils/memo_paths.py
```diff
@@ -0,0 +1,24 @@
...
+)
+
+__all__ = [
+    "capture_source_module",
+    "library_specifier_for",
+    "mirrored_jsx_path",
+    "mirrored_library_specifier",
+    "module_to_mirrored_segments",
+    "resolve_user_module_from_frame",
+]
```
masenf: i dont think we should do this for net-new code. if there was _never_ a `rx.utils.memo_paths` available, we shouldn't be adding it, especially since these functions do not really seem to be external public API functions.

tests/units/utils/test_memo_paths.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def test_capture_source_module_filters_missing():
+    assert memo_paths.capture_source_module(None) is None
+    fn = type("F", (), {"__module__": ""})
+    assert memo_paths.capture_source_module(fn) is None
+
+
+def test_module_to_mirrored_segments_module():
+    spec = importlib.util.find_spec("reflex.experimental.memo")
```
masenf: would recommend using a module that is less likely to go away at some point. experimental subpackage is definitely not stable and would lead to test churn if/when this reflex.experimental.memo module is removed.

reflex/compiler/utils.py
```diff
@@ -819,6 +835,99 @@ def write_file(path: str | Path, code: str):
...
+    emitted_relative: set[str] = set()
+    for path in emitted_paths:
+        absolute = Path(path)
+        if not absolute.is_absolute():
+            absolute = web_dir / absolute
+        try:
+            relative = absolute.relative_to(web_dir)
+        except ValueError:
+            continue
+        emitted_relative.add(str(relative).replace(os.sep, "/"))
```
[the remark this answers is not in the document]
masenf: yep, noticed this in local testing as well. the pruned paths are not cleaned up when the module is deleted/renamed

masenf (commented):
we need to put the modules under a subdir of `.web` that is not already used, maybe `.web/app_components`.

The problem arises when I have an app called `app` and a module called `app.root` which exports a memo component... this ends up getting overwritten by the `app/root.jsx` that the framework emits and then the page breaks when you try to load it.

The user's file tree hierarchy really shouldn't be able to break the app, hence keeping it under a subdirectory that is _only_ used for emitting memo components would protect the rest of the app from unexpected changes.

packages/reflex-base/src/reflex_base/utils/memo_paths.py
```diff
@@ -20,27 +21,22 @@
...
     "reflex_base",
     "reflex_site_shared",
     "reflex_hosting_cli",
     "reflex_docgen",
-})
+)
+
+# Bare-name prefixes matched against the whole module name (no dot boundary),
+# covering families of framework packages such as ``reflex_components_radix``.
+_FRAMEWORK_NAME_PREFIXES = ("reflex_components_",)
```
masenf: i think this might be overly broad.

i'm also wondering, why not just write the framework packages out to `.web/app_components/reflex/` or `.web/app_components/reflex_components_markdown`?

it seems using the arbitrary name `_internal` could actually shadow legitimate package names that downstream users might have in their libraries or app dirs; so if we use the _real_ package name where possible we reduce that chance.

then as a fallback when the name is bad/weird we can keep using the old memo path and method, like `.web/utils/components` and less of the code has to change.

the main thing though is just trying to avoid arbitrary restrictions on what the developer can name their files/functions.

masenf (commented):
One other weird thing i noticed:

```python
import reflex as rx

@rx.memo
def my_routes_func() -> rx.Component:
    return rx.text("This is a memoized component in root.py that changed again.")


@rx.memo
def depends_on_shadowed_memo() -> rx.Component:
    return rx.el.span(my_routes_func(), rx.text("that depends on the shadowed memo."))

@rx.memo
def my_routes_func() -> rx.Component:
    return rx.text("This is a memoized component in root.py that overrides the same name.")
```

When you have a module like this (even if you force eager eval of `depends_on_shadowed_memo`), it seems to always silently take the overridden memo component; no error, and only a single component emitted in the generated JS.

Another weird one i noticed, was that if i _import_ `my_routes_func` into a module that also defines `my_routes_func`, it complains about a name overlap.

```python
import reflex as rx

from .root import my_routes_func as root_routes_func


@rx.memo
def my_routes_func() -> rx.Component:
    """A memoized component that shares a name with one in root.py.

    This tests that memos with the same name in different modules don't collide
    and that the correct one is used in each place. The one in root.py should
    be used in the memo that depends on it, and this one should be used in the
    index page.
    """
    return rx.text("This is a memoized component in app.py.")
```

I'm assuming this one is because they compile to the same import name, so a single page cannot include both of them?

But the trace doesn't suggest as much, it looks like it caught it way earlier in the process, at definition time

```
  File "/opt/homebrew/Cellar/python@3.14/3.14.5/Frameworks/Python.framework/Versions/3.14/lib/python3.14/multiprocessing/process.py", line 320, in _bootstrap
    self.run()
    ~~~~~~~~^^
  File "/opt/homebrew/Cellar/python@3.14/3.14.5/Frameworks/Python.framework/Versions/3.14/lib/python3.14/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/reflex/.venv/lib/python3.14/site-packages/granian/server/mp.py", line 90, in wrapped
    callback = callback_loader()
  File "/Users/masenf/code/reflex-dev/reflex/.venv/lib/python3.14/site-packages/granian/_internal.py", line 65, in load_target
    module = load_module(path)
  File "/Users/masenf/code/reflex-dev/reflex/.venv/lib/python3.14/site-packages/granian/_internal.py", line 47, in load_module
    __import__(module_name)
    ~~~~~~~~~~^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/repro-memo-separate-modules/app/app/app.py", line 14, in <module>
    @rx.memo
     ^^^^^^^
  File "/Users/masenf/code/reflex-dev/reflex/packages/reflex-base/src/reflex_base/components/memo.py", line 1951, in memo
    _register_memo_definition(definition)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/reflex/packages/reflex-base/src/reflex_base/components/memo.py", line 470, in _register_memo_definition
    raise ValueError(msg)
ValueError: Memo name collision for `MyRoutesFunc`: `app.root.my_routes_func` and `app.app.my_routes_func` both compile to the same memo name.
```

☝️ that error is occuring because the key for checking for duplicate memo definitions does not include the source_file.

if i patch that in, then we hit the compile error that i had in mind

```
  File "/Users/masenf/code/reflex-dev/reflex/reflex/utils/prerequisites.py", line 361, in compile_or_validate_app
    get_compiled_app(
    ~~~~~~~~~~~~~~~~^
        check_if_schema_up_to_date=check_if_schema_up_to_date,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        prerender_routes=prerender_routes,
        trigger=trigger,
        ^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/masenf/code/reflex-dev/reflex/reflex/utils/prerequisites.py", line 273, in get_compiled_app
    app._compile(
    ~~~~~~~~~~~~^
        prerender_routes=prerender_routes,
    ...<2 lines>...
        trigger=trigger,
        ^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/masenf/code/reflex-dev/reflex/reflex/app.py", line 1329, in _compile
    did_real_compile = compiler.compile_app(
        self,
    ...<2 lines>...
        use_rich=use_rich,
    )
  File "/Users/masenf/code/reflex-dev/reflex/reflex/compiler/compiler.py", line 1190, in compile_app
    compile_ctx.compile(
    ~~~~~~~~~~~~~~~~~~~^
        evaluate_progress=lambda: progress.advance(task),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        render_progress=lambda: progress.advance(task),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/masenf/code/reflex-dev/reflex/packages/reflex-base/src/reflex_base/plugins/compiler.py", line 858,
in compile
    compiler.compile_page_from_context(page_ctx)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/reflex/reflex/compiler/compiler.py", line 737, in compile_page_from_context
    imports=utils.compile_imports(imports),
            ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/reflex/reflex/compiler/utils.py", line 127, in compile_imports
    validate_imports(collapsed_import_dict)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masenf/code/reflex-dev/reflex/reflex/compiler/utils.py", line 103, in validate_imports
    raise ValueError(msg)
ValueError: Can not compile, the tag MyRoutesFunc is used multiple time from $/app_components/app/root and 
$/app_components/app/app
```

So I think the fix for this one could be including the module in the import alias so we can _actually_ disambiguate these when they're used in the same page.

I feel like the tests should have caught these. There's definitely intention in the test to catch these kind of edge cases. But the tests are too isolated and these kind of problems are occuring at integration points between the functions. We really should spend some time thinking about if the tests are really good, just because there's a lot of them and they are passing doesn't necessarily mean they are exercising the system in ways that will prevent regressions.

masenf (approved):

PR #6458, opened 2026-05-05
Deepcopy EventSpec args when creating Event.from_event_type

masenf:
Fix regression from moving event loop to backend where mutable values in Event args yielded to the queue were preserved, causing subsequent changes to attempt to modify the original value. Previously all yielded events were serialized and deserialized before processing, so this created new copies automatically. The new behavior is to explicitly deepcopy values that are not known to be immutable.


masenf:
@greptile-apps re-review

adhami3310 (approved):

PR #6462, opened 2026-05-05
feat: support embedding Reflex apps into host pages via mount_target

FarhanAliRaza:
Add  and  config options for embedding a Reflex
app into an arbitrary DOM node on a host page. When  is set, the entry client mounts a memory data router into the target element via
 instead of hydrating the document, the build emits a stable
 shim re-exporting the hashed Vite chunk, and a route
manifest is generated at compile time so the embedded app owns its own URL space.  now sources  from the data router's
location so the backend sees the in-widget URL rather than the host page's.

### Usage 
rxconfig.py
```
import reflex as rx

config = rx.Config(
    app_name="my_app",
    mount_target="#reflex-root",
    # only needed when host page and bundle live on different origins
    embed_origin="http://localhost:3000",
    cors_allowed_origins=["*"],  # tighten for prod
)
```
usage in the HTML file. 
```
<div id="reflex-root">loading reflex...</div>
<script type="module" src="http://localhost:3000/app/entry.client.js"></script>
```

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:


masenf (commented):
My main feedback here is to implement this primarily as a Plugin instead of wiring it through the whole framework. The plugin can take in the mount_target and embed_origin as parameters instead of adding more to the global config.

The only real downside of this approach is the current `interpret_plugin_env` parser for environment var specified plugins does not expose a way to provide arguments to a plugin class.

FarhanAliRaza:
My thought was plugin too, but most of the will still remain outside of plugin i think.

So I implemented this way.. I ll update it.

masenf (commented):
i see some of what you were saying before about how the changes are still somewhat spread out even with the plugin.

at least the changes in `reflex/compiler/compiler.py` seem like they could be moved into the embed plugin module, since they are constructing the embed manifest and only used when embedding is enabled.

the problematic areas are where we're calling `get_embed_plugin()`; this seems like a role reversal. ideally the _rest_ of the code doesn't need to know anything about the internals of the EmbedPlugin for it to work. what it tells me is that maybe we need more hooks into the compilation process to allow plugins like this to be expressed succinctly.

* `update_env_json` - this could return dictionaries that get merged together and become `env.json`.
* maybe `_emit_stable_entry_bootloader` could be moved to a post_compile step, which should run before the production build step.
* the compile_vite_config is slightly more annoying, so we can leave that for now. but I'm wanting to bring in something like @riebecj's https://github.com/riebecj/reflex-vite-config-plugin at some point. So it would be cool if we could bring in something like that which exposed its own hooks, like `modify_vite_config` which would allow plugins to make structural modifications to the vite config before it gets written out.

packages/reflex-base/src/reflex_base/plugins/embed.py
```diff
@@ -0,0 +1,247 @@
...
+    """
+    from reflex_base.config import get_config
+
+    return next(
+        (p for p in get_config().plugins if isinstance(p, EmbedPlugin)),
+        None,
+    )
+
+
+Plugin = EmbedPlugin
```
riebecj: This is a useless assignment. Nowhere in any of your code are you importing `plugins.embed.Plugin`. Which makes [this import statement](https://github.com/reflex-dev/reflex/pull/6462/changes#diff-4c9a74e0d889609115f40b2c5b9480c3472321a9b226cbd442ceabe243592b2bR22) unnecessary. Just `from .base import Plugin` and subclass as `class EmbedPlugin(Plugin):` so we're not creating needlessly confusing module attributes.

masenf (approved):

PR #6474, opened 2026-05-08
Update shared components for new web design

carlosabadia:


masenf:
@greptile-apps re-review

packages/reflex-site-shared/src/reflex_site_shared/views/footer.py
```diff
@@ -228,96 +207,149 @@ def newsletter_input() -> rx.Component:
...
+                                footer_link(
+                                    "Agent Toolkit",
+                                    "/docs/ai/integrations/overview/",
+                                ),
+                                footer_link(
+                                    "Enterprise",
+                                    "/docs/enterprise/overview/",
+                                ),
+                                footer_link("App Management", "/hosting"),
+                                footer_link("Pricing", "/pricing"),
```
masenf, line 269: do we really want these footer links to internal pages all opening in new tabs? is that intentional

carlosabadia: internal shoulnd't open in new tab, I'll update that

Alek99 (approved):

PR #6475, opened 2026-05-08
Move bun.lock under reflex.lock/bun.lock

masenf:
## Motivation

Keep frontend packages as locked as possible across runs while still letting the framework update them when it's upgraded.

Before this PR, the bun lockfile lived inside `.web/` (gitignored along with the rest of the generated frontend). Even users who pinned `reflex==X.Y.Z` could see transitive frontend deps drift between runs, and apps hosting their own JS in the same directory couldn't easily commit their own lockfile. #6289 made the project-root `bun.lock` canonical; this PR finishes the job: the canonical lockfile (and a snapshot of `package.json`) moves under `reflex.lock/`, is honored on every install, and surrounding install logic is reworked so the resolved versions actually stick.

Followup for #6289.

## Changes

**Persisted lockfile & package.json**
- Move `bun.lock` to `reflex.lock/bun.lock` so it can be committed without colliding with user-owned JS at the project root.
- Persist `.web/package.json` alongside it under `reflex.lock/package.json` so the resolved dep set survives `.web/` regeneration.
- Sync `reflex.lock/` ↔ `.web/` on each run; invalidate the install cache when the persisted lockfile changes.
- Exclude `reflex.lock/` from hot-reload watch paths.

**Frontend install flow**
- Consolidate the install path into two phases: recover the persisted lockfile with a frozen install, then mutate via `add`/`remove`.
- Show a friendly "delete `reflex.lock/` and retry" hint when the lockfile is out of sync with `package.json`.
- Skip the initial install on brand-new projects where no lockfile exists yet.
- Drop the package-manager fallback during install — switching from bun to npm mid-flow would silently re-resolve and defeat the lockfile guarantee.
- Remove packages from `package.json` when no component, plugin, or framework constant calls for them anymore.

**npm support**
- Treat a committed `package-lock.json` as an opt-in signal that the project is npm-managed (in addition to `REFLEX_USE_NPM=1`).
- Gate bun-only flags (`--frozen-lockfile`) behind a runtime check so npm doesn't fail on unknown options.

**Section-aware unpinned dep handling**
- Replace bun's `--only-missing` flag with a manager-agnostic check: skip unpinned packages that already appear in the correct section of `package.json`, preserving previously resolved versions across runs.
- Move packages between `dependencies` and `devDependencies` when the framework/plugin/import-graph disagrees with the current placement.
- When the same package is requested as both a regular and a development dependency, install it as a regular dependency.

### DIFF
--- a/reflex/utils/js_runtimes.py
+++ b/reflex/utils/js_runtimes.py
@@ -591,37 +591,42 @@ def _install_frontend_packages(
         show_status_message="Installing base frontend packages",
     )
 
-    framework_deps = _pinned_args_from_constants(constants.PackageJson.DEPENDENCIES)
-    if framework_deps:
-        run_package_manager(
-            [primary_package_manager, "add", "--legacy-peer-deps", *framework_deps],
-            show_status_message="Pinning framework frontend dependencies",
-        )
-
-    framework_dev_deps = _pinned_args_from_constants(
-        constants.PackageJson.DEV_DEPENDENCIES
-    )
-    if framework_dev_deps:
-        run_package_manager(
-            [
-                primary_package_manager,
-                "add",
-                "--legacy-peer-deps",
-                "-d",
-                *framework_dev_deps,
-            ],
-            show_status_message="Pinning framework frontend dev dependencies",
-        )
-
+    pinned_packages, unpinned_packages = _split_by_version_specifier(packages)
     pinned_dev_deps, unpinned_dev_deps = _split_by_version_specifier(development_deps)
-    if pinned_dev_deps:
+
+    all_pinned = (
+        _pinned_args_from_constants(constants.PackageJson.DEPENDENCIES)
+        | pinned_packages
+    )
+    all_pinned_dev = (
+        _pinned_args_from_constants(constants.PackageJson.DEV_DEPENDENCIES)
+        | pinned_dev_deps
+    )
+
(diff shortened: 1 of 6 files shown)


adhami3310 (dismissed):

masenf:
LOL hot reload is broken now, because reflex.lock is not excluded from the reload path 🤦

masenf:
@greptile-apps re-review

reflex/utils/js_runtimes.py
```diff
@@ -415,44 +625,98 @@ def _install_frontend_packages(
...
         )
 
-    # Install custom packages defined in frontend_packages
-    if packages:
+    # Install against the recovered lockfile so its pins are honored
+    # before any further mutation. ``--frozen-lockfile`` ensures bun
+    # refuses to silently rewrite the lockfile here; npm ignores the
+    # flag. Skip on brand-new projects where no lockfile exists yet —
+    # ``--frozen-lockfile`` would error, and the subsequent ``bun add``
+    # calls will generate a fresh lockfile.
```
masenf:
```suggestion
    # Install against the recovered lockfile so its pins are honored
    # before any further mutation.
```

masenf:
@greptile-apps re-review

adhami3310 (approved):

PR #6481, opened 2026-05-11
Fix rx.code_block rendering empty when content is a state Var

### LINKED ISSUE #6480: rx.code_block renders empty when content is a state Var (0.9.2)
## Describe the bug
`rx.code_block(state_var, language="yaml", show_line_numbers=True)` renders an empty `<pre><code class="language-yaml">` element when the first positional argument is a state `Var`. The Var is populated at runtime (other components reading the same Var work — e.g. `rx.set_clipboard(state_var)` copies the full content, and `rx.cond(state_var != "", ...)` evaluates `True`) — but the content never reaches `SyntaxHighlighter`.

The same call with a **literal Python string** in the same position renders correctly.

`rx.markdown(<expression containing state_var>, component_map={"pre": ...})` — the documented v0.9.0 replacement for `codeblock` in `component_map` — also renders no `<pre>` at all in this configuration, so the markdown route is not a workaround.

## Versions
- `reflex` 0.9.2
- `reflex-components-code` 0.9.2
- `reflex-components-markdown` 0.9.1
- Python 3.13 (container) / 3.12 (local) — same result on both
- React 19, react-router 7, Vite 8

## Minimal reproduction

```python
import reflex as rx


class State(rx.State):
    content: str = (
        "version: v3\n"
        "teleport:\n"
        "  nodename: example\n"
        "  auth_token: abcd1234\n"
    )


@rx.page()
def index() -> rx.Component:
    return rx.vstack(
        rx.heading("Literal string (renders correctly):", size="3"),
        rx.code_block(
            "version: v3\nteleport:\n  nodename: example\n  auth_token: abcd1234\n",
            language="yaml",
            show_line_numbers=True,
        ),
        rx.heading("State Var (renders empty):", size="3"),
        rx.code_block(
            State.content,
            language="yaml",
            show_line_numbers=True,
        ),
        spacing="4",
        padding="4",
    )


app = rx.App()
```

## Expected behavior
The second `rx.code_block` renders the same YAML content as the first, syntax-highlighted with line numbers.

## Actual behavior
The second `<pre>` element renders, but its `<code class="language-yaml">` child has zero text content:

```html
<pre class="css-..."><code class="language-yaml" style="..."></code></pre>
```

## Root cause

The auto-memoized wrapper compiled for `rx.code_block(state_var, ...)` (`.web/utils/components/Codeblock_prismasynclight_<hash>.jsx`) reads the state Var via `useContext` and passes it via `props.children` — but also passes the destructured wrapper-children as the **third positional argument** to `@emotion/react`'s `jsx()`:

```jsx
import { memo, useContext } from "react"
import { PrismAsyncLight as SyntaxHighlighter } from "react-syntax-highlighter"
import { StateContexts } from "$/utils/context"
import { jsx } from "@emotion/react"

export const Codeblock_prismasynclight_<hash> = memo(({children}) => {
    const reflex___state__... = useContext(StateContexts.reflex___state__...)
    return(
        jsx(SyntaxHighlighter,
            {children: reflex___state__....content_rx_state_, css: ..., language: "yaml", showLineNumbers: true, style: ...},
            children   // <-- destructured from wrapper props; undefined when wrapper is invoked with no children
        )
    )
});
```

`@emotion/react`'s `jsx` (in the production bundle):

```js
function jsx(e, t) {
  var n = arguments;
  if (t == null || !hasOwnProperty.call(t, "css")) {
    return React.createElement.apply(void 0, n);    // <-- both paths spread all 3 args
  }
  // ...wraps with emotion css component, still spreads all args...
}
```

`React.createElement(type, props, undefined)` has `arguments.length === 3`, so it sets `props.children = undefined`, **overriding** the children value set in the props object. `SyntaxHighlighter` then receives `children: undefined` and renders the language-class wrapper but no highlighted content.

## Source pointer

The unconditional positional spread happens in `reflex_base/components/component.py:2617-2623` (v0.9.2):

```python
return FunctionStringVar.create(
    "jsx",
).call(
    tag_name,
    props,
    *[render_dict_to_var(child) for child in tag["children"]],   # spreads even when empty
)
```

When `tag["children"]` is empty the spread is empty, but the auto-memoized wrapper for a stateful subtree (e.g. `Codeblock_prismasynclight_<hash>` above) still introduces a destructured `children` parameter that gets passed positionally as the third arg, producing `jsx(SyntaxHighlighter, props, undefined)`. The undefined is enough to override `props.children`.

`reflex_components_code/code.py:502-514` is where the state Var ends up in `props["children"]`:

```python
def _render(self):
    out = super()._render()
    theme = self.theme
    return (
        out
        .add_props(style=theme)
        .remove_props("theme", "code")
        .add_props(children=self.code)   # <-- value goes into props.children
    )
```

## Workaround
`rx.box(rx.text(state_var, font_family="mono", style={"white-space": "pre-wrap"}))`. Loses syntax highlighting but content is visible.

## Observed but separate
On every build, Vite 8 logs:
> `Warning: Invalid input options (1 issue found) - For the "jsx". Invalid key: Expected never but received "jsx".`

This is the `rollupOptions: { jsx: {} }` line in `reflex_base/compiler/templates.py:583` — `rollupOptions.jsx` is no longer accepted by Vite 8. This is a build-config compatibility issue separate from the rendering bug above (the rendering bug reproduces regardless of the warning since the .jsx files use explicit `jsx()` calls, not JSX syntax).

jryberg:
### Type of change

### Summary

Fixes #6480.

`rx.code_block(state_var, language=..., show_line_numbers=True)` renders an empty `<pre><code class="language-yaml">` element. The state Var is populated at runtime — `rx.set_clipboard(state_var)` copies the full content, and `rx.cond(state_var != "", ...)` evaluates `True` — but `SyntaxHighlighter` receives nothing.

### Root cause

`CodeBlock._render()` moves the user's first positional argument into `props["children"]`. For a stateful subtree, the compiler wraps the call in an auto-memo:

```jsx
export const Codeblock_prismasynclight_<hash> = memo(({children}) => {
    const reflex___state__... = useContext(...)
    return jsx(SyntaxHighlighter, {children: state_var, ...}, children);
});
```

The third positional `children` is the wrapper's destructured prop, which is `undefined` when the wrapper is invoked with no children (always, for `rx.code_block`). `@emotion/react`'s classic `jsx` delegates to `React.createElement.apply(void 0, args)`, and `React.createElement(type, props, undefined)` has `arguments.length === 3`, so React sets `props.children = undefined`, overriding the legitimate value from the props object.

### Fix

In `_RenderUtils.render_tag` (`packages/reflex-base/src/reflex_base/compiler/templates.py`), when `props` already declares a `children:` key, omit the positional children in the emitted `jsx()` call. Components that don't bind `props.children` keep their existing behaviour.

### Tests

Two new tests in `tests/units/compiler/test_compiler.py`:

- `test_render_tag_skips_positional_children_when_props_have_children` — direct regression for #6480.
- `test_render_tag_preserves_positional_children_for_normal_components` — guards the common path so a future refactor can't re-introduce the bug by over-correcting.

`uv run pytest tests/units/compiler/` — 160 passed.

### Checklist

- [x] Tests pass: `uv run pytest tests/units/compiler/test_compiler.py -k render_tag`
- [x] Reproduced the bug end-to-end before the fix, and confirmed the fix renders syntax-highlighted content for `rx.code_block(state_var, ...)` across three independent call sites in a downstream application (Reflex 0.9.2 + the equivalent patch applied via post-install script).

### DIFF
--- a/packages/reflex-base/src/reflex_base/compiler/templates.py
+++ b/packages/reflex-base/src/reflex_base/compiler/templates.py
@@ -67,12 +67,22 @@ class _RenderUtils:
     @staticmethod
     def render_tag(component: Mapping[str, Any]) -> str:
         name = component.get("name") or "Fragment"
-        props = f"{{{','.join(component['props'])}}}"
+        props_list = component["props"]
+        props = f"{{{','.join(props_list)}}}"
         rendered_children = [
             _RenderUtils.render(child)
             for child in component.get("children", [])
             if child
         ]
+
+        # When `props` already declares `children:` (e.g. `rx.code_block` routes
+        # its content there via `CodeBlock._render`), do NOT also emit positional
+        # children. React's `createElement(type, props, x)` overrides
+        # `props.children` with the positional arg even when it is `undefined`
+        # (which is what the auto-memo wrapper passes when invoked with no
+        # children), silently nulling the legitimate value in props. See #6480.
+        if any(p.startswith("children:") for p in props_list):
+            return f"jsx({name},{props})"
 
         return f"jsx({name},{props},{','.join(rendered_children)})"
 
--- a/tests/units/compiler/test_compiler.py
+++ b/tests/units/compiler/test_compiler.py
@@ -5,6 +5,7 @@ from pathlib import Path
 import pytest
 from pytest_mock import MockerFixture
 from reflex_base import constants
+from reflex_base.compiler.templates import _RenderUtils
 from reflex_base.components.dynamic import bundle_library, reset_bundled_libraries
 from reflex_base.constants.compiler import PageNames
 from reflex_base.utils.imports import ImportVar, ParsedImportDict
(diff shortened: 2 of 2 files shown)


FarhanAliRaza:
I think this was already fixed here https://github.com/reflex-dev/reflex/pull/6466

masenf:
since we already have a more general fix for this in `main` (unreleased as of now), but the fix didn't include any kind of regression test, we probably should add an integration test for code block component that asserts on functionality: theme, copy button, language, content and these props variously coming statically or from state vars.

masenf:
We have released the fix for this in 0.9.2.post1, please upgrade and confirm that the issue you were seeing is resolved.

I filed #6484 as an enhancement to add the true integration tests for code_block component.

Will close this PR now, but we appreciate your involvement in the project 😌

PR #6496, opened 2026-05-12
ENG-9484: update docs for auto setters

adhami3310:


Alek99 (changes requested):
docs/library/tables-and-data-grids/table.md (line 369) still documents auto setters: “Every base var has a built-in event handler ... set_VARNAME.” This PR removes the setters page and moves examples to explicit event handlers, so this remaining paragraph is now misleading, especially since the example immediately below defines set_sort_value and set_search_value explicitly

Alek99 (approved):

PR #6505, opened 2026-05-14
[codex] Loosen enterprise docs alert copy

Alek99:
## Summary
- Shorten the Enterprise overview alert title so the collapsed state feels less crowded.
- Move the badge explanation into the alert body while preserving the Built with Reflex link.

## Why
The previous alert used the full badge explanation as the accordion title, which made the notice feel cramped in its collapsed state.

## Validation
- `uv run --project docs/app pytest tests/test_doc_links.py`
- `uv run --project docs/app pytest tests/test_routes.py`
- Verified locally at `http://localhost:3001/docs/enterprise/overview/` that the collapsed alert uses the shorter title and the expanded body contains the badge details/link.


adhami3310 (approved):
while this technically fixes the issue, i would have expected a fix to the line height

PR #6506, opened 2026-05-14
[codex] Fix memoized component hook prop rendering

Alek99:
## Summary
- Compile auto-memo component hooks before rendering so hook-mutated props are present in memo modules.
- Add regression coverage for auto-memo components whose add_hooks fills a render prop.

## Root Cause
The auto-memo full-component path rendered before calling _get_all_hooks(). DataEditor.add_hooks() generates the getData_* hook and mutates the component with the getCellContent prop, so the memoized docs module included the hook function but rendered without getCellContent. Glide then tried to call the missing callback and the docs page failed with TypeError: x is not a function.

## Validation
- uv run ruff check reflex/compiler/utils.py tests/units/compiler/test_memoize_plugin.py
- uv run pytest tests/units/compiler/test_memoize_plugin.py -q
- uv run pytest tests/units/components/datadisplay/test_dataeditor.py -q
- Ran the docs locally at http://localhost:3030/docs/library/tables-and-data-grids/data-editor/ and verified the page renders, including the first data grid, without the Application Error / TypeError. The only console noise seen locally was unrelated Base UI nativeButton messaging from the docs shell.

### DIFF
--- a/reflex/compiler/utils.py
+++ b/reflex/compiler/utils.py
@@ -460,8 +460,8 @@ def compile_experimental_component_memo(
         rendered = render.render()
     else:
         render = _apply_component_style_for_compile(copy.deepcopy(definition.component))
-        rendered = render.render()
         hooks = render._get_all_hooks()
+        rendered = render.render()
         custom_code = render._get_all_custom_code()
         dynamic_imports = render._get_all_dynamic_imports()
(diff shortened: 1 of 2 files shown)


adhami3310 (commented):
precommit isn't passing

masenf (approved):

PR #6511, opened 2026-05-14
[codex] update README example sections

Alek99:
## Summary

- Simplifies the README "Create your first app" instructions into a shorter uv-based quickstart.
- Removes the Resources, Status, Contributing, and License sections from the README.
- Adds the image generation Reflex/OpenAI example code directly below the example video.

## Impact

This keeps the README focused on the quickstart and example app while removing duplicated or less relevant project-meta sections.

## Validation

- `git diff --cached --check`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Build an image generation app in Python with Reflex: define the UI, manage state
 import reflex as rx
 import openai
 
-client = openai.OpenAI()
+client = openai.AsyncOpenAI()
 
 
 class State(rx.State):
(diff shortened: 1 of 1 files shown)


masenf (commented):
greptile raises a good point about the example code being essentially broken by default. we probably need to get non-async event handlers running in a thread pool sooner rather than later. i'm not 100% sure what fallout effects that could have, but right now, non-async events just block the event loop and kill server performance. it's too easy to write code with bad performance characteristics.

masenf (approved):

PR #6519, opened 2026-05-15
feat(telemetry): track per-feature usage counters on compile events

FarhanAliRaza:
Introduce a typed FeatureName registry and `increment_feature` helper so runtime call sites (uploads, cookies, storage, models, lifespan tasks, shared state, dynamic routes) and the compile-time collector (state-manager mode, CORS, background event handlers) feed a uniform counters map into the compile event. Always emit every known key so zeros are distinguishable from missing detectors.

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

fixes ENG-9480

### DIFF
--- a/reflex/utils/telemetry_accounting.py
+++ b/reflex/utils/telemetry_accounting.py
@@ -96,24 +96,21 @@ def _collect_compile_event_payload(
     """
     config = get_config()
     user_states = list(_walk_states(app._state))
-    component_counts, upload_count = _walk_components(app._pages.values())
     return {
         "plugins_enabled": [p.__class__.__name__ for p in config.plugins],
         "plugins_disabled": [p.__name__ for p in config.disable_plugins],
         "pages_count": len(app._pages),
-        "component_counts": component_counts,
+        "component_counts": _count_components(app._pages.values()),
         "states": [_collect_state_stats(s) for s in user_states],
-        "features_used": _collect_features_used(app, config, user_states, upload_count),
+        "features_used": _collect_features_used(app, config, user_states),
         "duration_ms": ctx.elapsed_ms(),
         "trigger": ctx.trigger,
         "exception": _sanitize_exception(ctx.exception),
     }
 
 
-def _walk_components(
-    pages: Iterable[BaseComponent],
-) -> tuple[dict[str, int], int]:
-    """Walk page trees once and aggregate class-name counts and the upload total.
+def _count_components(pages: Iterable[BaseComponent]) -> dict[str, int]:
+    """Count component types across one or more component trees.
 
     Auto-memoized components live in the tree as dynamic
     ``ExperimentalMemoComponent_<Type>_<tag>_<hash>`` subclasses. Bucketing by
@@ -125,24 +122,19 @@ def _walk_components(
         pages: Component-tree roots to walk.
 
     Returns:
-        ``(counts, upload_count)`` where ``counts`` maps class name to
-        occurrence count and ``upload_count`` is the number of ``Upload``
-        instances (including subclasses).
+        Mapping of component class name to occurrence count.
     """
     counts: dict[str, int] = {}
-    upload_count = 0
     stack: list[BaseComponent] = list(pages)
     while stack:
         node = stack.pop()
-        if isinstance(node, Upload):
-            upload_count += 1
         node_cls = type(node)
         wrapped = getattr(node_cls, "_wrapped_component_type", None)
         name = wrapped.__name__ if wrapped is not None else node_cls.__name__
         counts[name] = counts.get(name, 0) + 1
         if node.children:
             stack.extend(node.children)
-    return counts, upload_count
+    return counts
 
 
 def _walk_states(root: type[BaseState] | None) -> Iterator[type[BaseState]]:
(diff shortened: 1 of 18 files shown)


reflex/utils/telemetry_accounting.py
Alek99, line 1: [P2] reflex/utils/telemetry_accounting.py (line 270) double-counts inherited client-storage vars. _walk_state_features() walks state_cls.get_fields() for every state returned by _walk_states(), but child states include parent fields there. A root state with c: str = Cookie() and child state with ls: str = LocalStorage() reports cookie_count == 2 and local_storage_count == 1; the cookie is counted once on the root and again through the child’s inherited field. This affects cookie_count, local_storage_count, and session_storage_count for normal state hierarchies. It should probably walk non-inherited fields only, e.g. names from state_cls.base_vars, then resolve those names back to fields.

Alek99 (dismissed):

reflex/utils/telemetry_accounting.py
FarhanAliRaza: fixed

tests/integration/test_telemetry_compile.py
```diff
@@ -0,0 +1,218 @@
...
+        app_name="telemetry_compile_app",
+    )
+    harness._initialize_app()
+    try:
+        yield harness
+    finally:
+        if harness._registry_token is not None:
+            from reflex_base.registry import RegistrationContext
+
+            RegistrationContext.reset(harness._registry_token)
```
masenf: we should be using the context manager here? is there a specific reason why not?


```suggestion
    with AppHarness.create(
        root=root,
        app_source=functools.partial(
            TelemetryCompileApp,
            events_log_path=str(telemetry_events_log),
        ),
        app_name="telemetry_compile_app",
    ) as harness:
        yield harness
```

tests/units/utils/test_telemetry_accounting.py
```diff
@@ -203,16 +229,235 @@ def test_memo_wrapper_class_records_wrapped_component_type():
...
+    """``upload_count`` aggregates ``Upload`` and any subclass occurrences."""
+    from reflex_components_core.core.upload import StyledUpload, Upload
+
+    class _PlainStub:
+        children = ()
+
+    upload_inst = object.__new__(Upload)
+    upload_inst.children = []
+    styled_inst = object.__new__(StyledUpload)
+    styled_inst.children = []
```
masenf: why are we doing this and not just `upload_inst = Upload()`?

tests/units/utils/test_telemetry_accounting.py
```diff
@@ -203,16 +229,235 @@ def test_memo_wrapper_class_records_wrapped_component_type():
...
+
+    features = telemetry_accounting._collect_features_used(
+        _fake_app(),  # pyright: ignore[reportArgumentType]
+        _fake_config(),
+        [_StorageState],
+        0,
+    )
+    assert features["cookie_count"] == 2
+    assert features["local_storage_count"] == 1
+    assert features["session_storage_count"] == 1
```
masenf, line 274: should also check the plain field count?

tests/units/utils/test_telemetry_accounting.py
```diff
@@ -203,16 +229,235 @@ def test_memo_wrapper_class_records_wrapped_component_type():
...
 
     assert counts == {"Button": 1}
+    assert upload_count == 0
+
+
+def test_walk_components_counts_upload_subclass_instances():
+    """``upload_count`` aggregates ``Upload`` and any subclass occurrences."""
+    from reflex_components_core.core.upload import StyledUpload, Upload
+
+    class _PlainStub:
```
masenf: is there a particular reason why these internally defined classes all have the underscore prefix? it looks kind of weird IMO

tests/units/utils/test_telemetry_accounting.py
FarhanAliRaza: added ` assert "plain" not in features `

tests/units/utils/test_telemetry_accounting.py
FarhanAliRaza: updated to `Upload.create()`

tests/units/utils/test_telemetry_accounting.py
FarhanAliRaza: renamed

reflex/utils/telemetry_accounting.py
```diff
@@ -78,21 +95,25 @@ def _collect_compile_event_payload(
         The properties dict to send to PostHog.
     """
     config = get_config()
+    user_states = list(_walk_states(app._state))
+    component_counts, upload_count = _walk_components(app._pages.values())
```
masenf: do we need to have another tree walk? i thought the plan was to plug this into the compiler so we didn't have to walk the tree again. was there some issue with that implementation which made that unworkable?

reflex/utils/telemetry_accounting.py
```diff
@@ -104,19 +125,24 @@ def _count_components(pages: Iterable[BaseComponent]) -> dict[str, int]:
...
+        occurrence count and ``upload_count`` is the number of ``Upload``
+        instances (including subclasses).
     """
     counts: dict[str, int] = {}
+    upload_count = 0
     stack: list[BaseComponent] = list(pages)
     while stack:
         node = stack.pop()
+        if isinstance(node, Upload):
+            upload_count += 1
```
masenf: why is upload being special cased here? this is contributing about 20% to the overall telemetry walk overhead. We already track `Upload.is_used` internally? maybe instead of a bool, we could treat this as a counter instead and then just read it directly as a ClassVar?

reflex/utils/telemetry_accounting.py
FarhanAliRaza: I forgot the actual reason but I tried it. I ll check again.

masenf (approved):

masenf:
we can keep the walk for now, overall it's not contributing a significant amount to the runtime; we can optimize it later if it turns out to be more significant

PR #6523, opened 2026-05-18
Update internal drawer component

carlosabadia:


Alek99:
Screenshot 2026-05-18 at 10 10 55 AM

I dont think the chevron down is clear that its should open a drawer

carlosabadia:
> Screenshot 2026-05-18 at 10 10 55 AM
> I dont think the chevron down is clear that its should open a drawer

this doesnt change the drawer in the docs, its just to update the reflex ui drawer base component to use the base ui one

Alek99:
I understand this is a general comment about how the chevron here is not clear it open a drawer

FarhanAliRaza:
image
There is not way to turn off the "x" button. If that is intentional, that is fine. Usually, drawers don't have it.

carlosabadia:
> image There is not way to turn off the "x" button. If that is intentional, that is fine. Usually, drawers don't have it.

yeah, don't mind the high level component

FarhanAliRaza (approved):

PR #6541, opened 2026-05-20
Bump base ui version for internal components

carlosabadia:


Alek99 (approved):

masenf:
we need to be updating the frontend lock files when frontend dependencies are changing.


i was testing this out and finding that it was not installing the new version of the library, so i was looking into that bug. i will keep scoping that out and send a follow up

masenf:
follow on https://github.com/reflex-dev/reflex/pull/6543

PR #6562, opened 2026-05-25
Improve foreach docs examples

Alek99:
## Summary
- clarify when to use `rx.foreach` versus Python comprehensions
- update foreach callback annotations to use `rx.Var[...]`
- improve demo styling so color and dictionary examples render readably
- make todo example buttons visible and wire `finish_item` as an event handler

## Testing
- Ran docs app locally with `CI=1` and whitelisted `/library/dynamic-rendering/foreach`
- Verified the page in the Codex in-app browser at `http://localhost:3000/docs/library/dynamic-rendering/foreach/`
- Confirmed the todo `Done` button removes an item from the rendered list
- Ran `git diff --check`

### DIFF
--- a/docs/library/dynamic-rendering/foreach.md
+++ b/docs/library/dynamic-rendering/foreach.md
@@ -1,8 +1,3 @@
----
-components:
-  - rx.foreach
----
-
 ```python exec
 import reflex as rx
 ```
@@ -17,11 +12,8 @@ This is useful for dynamically rendering a list of items defined in a state.
 ```
 
 ```python demo exec
-from typing import List
-
-
 class ForeachState(rx.State):
-    colors: List[str] = [
+    colors: list[str] = [
         "#E5484D",
         "#12A594",
         "#3E63DD",
@@ -31,7 +23,7 @@ class ForeachState(rx.State):
     ]
 
 
-def color_swatch(label, color: rx.Var[str]):
+def color_swatch(label: rx.Var[str | int], color: rx.Var[str]):
     return rx.box(
         rx.text(label, color="white", weight="medium"),
         bg=color,
@@ -59,7 +51,7 @@ def foreach_example():
 The function can also take an index as a second argument.
 
 ```python demo exec
-def colored_box_index(color: rx.Var[str], index: int):
+def colored_box_index(color: rx.Var[str], index: rx.Var[int]):
     return color_swatch(index, color)
 
 
@@ -78,11 +70,8 @@ When indexing into a nested list, it's important to declare the list's type as R
 This ensures that any potential frontend JS errors are caught before the user can encounter them.
 
 ```python demo exec
-from typing import List
-
-
 class NestedForeachState(rx.State):
-    numbers: List[List[str]] = [["1", "2", "3"], ["4", "5", "6"], ["7", "8", "9"]]
+    numbers: list[list[str]] = [["1", "2", "3"], ["4", "5", "6"], ["7", "8", "9"]]
 
 
 def display_row(row: rx.Var[list[str]]):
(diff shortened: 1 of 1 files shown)


Alek99:
Updated the complex dictionary example layout after local review: it now renders the three secondary colors as three grouped columns with spacing instead of wrapping the third group underneath.

Alek99:
Updated the color examples again after visual review: they now use explicit site-friendly hex colors and a shared rounded `color_swatch` helper. The complex dictionary example renders the parent colors as swatches too, so the top row matches the child swatches.

Alek99:
Simplified the swatch examples after review: the basic color example is back to `List[str]`, and the complex dictionary example is back to `Dict[str, List[str]]`. The strings are now site-friendly hex colors, so the rendered swatches stay polished without introducing nested tuple data structures.

docs/library/dynamic-rendering/foreach.md
```diff
@@ -9,45 +9,64 @@ import reflex as rx
...
+        "#E5484D",
+        "#12A594",
+        "#3E63DD",
+        "#AD5700",
+        "#F76B15",
+        "#8E4EC6",
+    ]
+
+
+def color_swatch(label, color: rx.Var[str]):
```
masenf:
```suggestion
def color_swatch(label: rx.Var[str], color: rx.Var[str]):
```

was there a reason to not annotate both?

docs/library/dynamic-rendering/foreach.md
```diff
@@ -40,8 +44,14 @@ def foreach_example():
 The function can also take an index as a second argument.
 
 ```python demo exec
-def colored_box_index(color: str, index: int):
-    return rx.box(rx.text(index), bg=color)
+def colored_box_index(color: rx.Var[str], index: int):
```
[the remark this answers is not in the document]
masenf: agree; any values passed from a `foreach` will be Var-typed arguments

docs/library/dynamic-rendering/foreach.md
```diff
@@ -66,7 +85,7 @@ class NestedForeachState(rx.State):
     numbers: List[List[str]] = [["1", "2", "3"], ["4", "5", "6"], ["7", "8", "9"]]
```
masenf:
```suggestion
    numbers: list[list[str]] = [["1", "2", "3"], ["4", "5", "6"], ["7", "8", "9"]]
```

and remove the typing import to simplify the example and use modern conventions

docs/library/dynamic-rendering/foreach.md
```diff
@@ -142,20 +153,23 @@ def todo_example():
...
-
-
 class SimpleDictForeachState(rx.State):
     color_chart: dict[int, str] = {1: "blue", 2: "red", 3: "green"}
 
 
-def display_color(color: List):
-    # color is presented as a list key-value pair([1, "blue"],[2, "red"], [3, "green"])
-    return rx.box(rx.text(color[0]), bg=color[1])
+def display_color(color: rx.Var[tuple[int, str]]):
```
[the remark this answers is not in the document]
masenf: greptile is correct. leaving it with the `int` and attempting to adding `1` to it results in the following exception:

```
reflex_base.utils.exceptions.VarTypeError: Unsupported Operand type(s) for +: StringCastedVar, int
```

So we need to make a note that when iterating through a dict, the keys will always be coerced as str

docs/library/dynamic-rendering/foreach.md
masenf:
```suggestion
from dataclasses import dataclass, field
from uuid import UUID, uuid4


@dataclass
class TodoItem:
    text: str
    id: UUID = field(default_factory=uuid4)


class ListState(rx.State):
    items: list[TodoItem] = [
        TodoItem(text="Write Code"),
        TodoItem(text="Sleep"),
        TodoItem(text="Have Fun"),
    ]
    new_item: str

    @rx.event
    def set_new_item(self, new_item: str):
        self.new_item = new_item

    @rx.event
    def add_item(self):
        self.items += [TodoItem(text=self.new_item)]

    @rx.event
    def finish_item(self, item_id: UUID):
        self.items = [i for i in self.items if i.id != item_id]


def get_item(item: rx.Var[TodoItem]):
    return rx.list.item(
        rx.hstack(
            rx.button(
                "Done",
                on_click=lambda: ListState.finish_item(item.id),
                size="1",
                variant="soft",
            ),
            rx.text(item.text, font_size="1.25em"),
        ),
        key=item.id,
    )
```

I suggest making the complex todo list example a bit more complex so we can show off the `key` prop and how it is used to make a stable list when the values can be re-ordered or the visible text renamed.

... idk why my suggestion didn't really work, probably because it's partly gray line

masenf (commented):
The API Reference at the bottom of this page seems weird. It only shows the `iterable` prop, which mismatches with how the foreach component is supposed to be used.

masenf:
@greptile

masenf (approved):

PR #6563, opened 2026-05-25
Tree-shake bundled libraries while preserving dynamic components

FarhanAliRaza:
Automatically bundled plugin dependencies currently expose their entire namespace on `window.__reflex`, preventing unused exports from being removed. Collect the imports used across pages, app wrappers, and memo modules, and expose eligible libraries through named imports so the frontend bundler can tree-shake them.

Preserve complete namespaces for apps using runtime `Component` vars, explicit `bundle_library()` registrations, internal modules, React/Emotion, and default, namespace, or unsupported import shapes. A component returned by a later event can introduce exports absent from the initial page. Explicit registrations survive recompilation; generated aliases also avoid collisions between package paths and export names.

The marketing homepage build reduces compressed modulepreload payload from **372,768 to 326,254 bytes**: **45.4 KiB / 12.5% less**. The generated Radix namespace retains eight exports (`Box`, `Code`, `Dialog`, `Heading`, `Link`, `Skeleton`, `Text`, `Theme`). The global CSS asset is identical in both builds.

Production homepage measurements (three cold loads per build per method, Lighthouse 13.4.1 / Chrome 152, mobile defaults):

| Metric | Main | This PR |
| --- | ---: | ---: |
| First-party JavaScript transfer | 437.9 KiB | 392.5 KiB |
| Applied-throttling LCP, median (range) | 3.888 s (3.879–3.895) | 3.653 s (3.651–3.653) |
| Applied-throttling total blocking time, median | 429 ms | 416 ms |
| Applied-throttling main-thread work, median | 4,049 ms | 4,067 ms |
| Simulated-throttling LCP, median (range) | 5.109 s (3.908–5.110) | 4.959 s (4.885–5.034) |

This is a reproducible payload reduction and a modest local paint improvement; the CPU timings do not establish a reduction. The marketing app was pinned to `db57617a92a47c22df729de31a101deb70049020`, compared against Reflex `4d4437d416cdb917112dae88c7e301f9ab24b1e6`. Both production frontend builds used the same dependencies, frozen public template catalog, backend, assets, and CSS. Third-party scripts remained enabled in timed runs. All baseline samples preceded PR samples, so these small samples are not a field-performance estimate.

Validation:

- Full unit suite: **8,375 passed, 18 skipped**, **76.22% coverage** on Python 3.14.5.
- All repository pre-commit hooks passed, including Pyright, ty, Ruff, generated stubs, and Biome.
- Production browser regression passed: existing dynamic content/counter behavior plus an event that first creates a Radix Card after initial render, then handles an event from that Card.
- Separate compatibility checks with #7016 passed: subpath default/named/namespace serialization, a dynamically rendered Lucide icon, and the late-created Card. The explicit-registration helpers align with #7016; its serializer rewrite remains in that PR.
- Regression coverage includes memo imports, explicit registration across recompiles and context forks, unsupported tags, and generated alias collisions.
- Homepage menus, prompt input, image loading, and carousel behavior passed at 320, 390, and 1440 px. All nine corresponding screenshots were byte-identical; there were no page errors or horizontal overflow. Tracking was blocked only for these interaction/screenshot checks.

The dynamic-component test now uses its actual button/counter events to establish backend readiness. Its previous token-value wait also failed on unmodified main; the dynamic token input is still required to render.

This change addresses JavaScript namespace retention. Radix CSS pruning is outside this PR's final scope.

Fixes ENG-9323.


FarhanAliRaza:
Add more tests for dynamic components.

packages/reflex-base/src/reflex_base/plugins/tailwind_v3.py
```diff
@@ -63,7 +63,7 @@ def compile_root_style(include_radix_themes: bool = True):
     Returns:
         The compiled Tailwind root style.
     """
-    from reflex.compiler.compiler import RADIX_THEMES_STYLESHEET
+    from reflex_components_radix.plugin import RADIX_THEMES_STYLESHEET
```
adhami3310, line 67: can we figure a way such that tailwind doesn't conceptually depend on radix themes?

packages/reflex-base/src/reflex_base/components/dynamic.py
```diff
@@ -102,6 +113,11 @@ def make_component(component: Component) -> str:
         component_imports = component._get_all_imports()
         compiler._apply_common_imports(component_imports)
 
+        for lib, ivs in component_imports.items():
```
adhami3310: please avoid uncommon abbreviations like `ivs` in variable names

packages/reflex-base/src/reflex_base/compiler/templates.py
```diff
@@ -161,12 +162,85 @@ def document_root_template(*, imports: list[_ImportDict], document: dict[str, An
...
+        A tuple of ``(import_block, useEffect_body)``. Both are empty when
+        no dynamic components are in play.
+    """
+    if not window_library_imports:
+        return "", ""
+    import_lines: list[str] = []
+    entries: list[str] = []
+    for lib, names in window_library_imports.items():
+        alias = f"__reflex_{_normalize_window_lib_alias(lib)}"
+        if names is None or any(not _JS_IDENTIFIER_RE.match(n) for n in names):
```
adhami3310: what does `any(not _JS_IDENTIFIER_RE.match(n) for n in names)` mean? can we create a private helper function with a name for it?

PR #6567, opened 2026-05-27
fix(recharts): bundle es-toolkit/compat as ESM (#6561)

### LINKED ISSUE #6561: Recharts components cause TypeError and infinite reload on any route
**Describe the bug**
So, today i tried to work on my side-project but as today(May 25) i ran into in an issue:

Any Reflex page that renders a `rx.recharts` component (`area_chart`, `line_chart`, `composed_chart`, etc.) fails to load at runtime with a `TypeError`. React Router catches the error and retries the route module indefinitely, producing a blank page with an infinite reload loop. The app is otherwise healthy — all non-chart routes work normally.

The root cause is that recharts 3.x pulls `es-toolkit/compat/get.js` (a CommonJS file) into the bundle through its barrel export (`recharts/es6/index.js`). Vite 8's Rolldown bundler places the CJS interop helpers in a shared `rolldown-runtime` chunk that is not guaranteed to be loaded before lazy-loaded route chunks execute.

**To Reproduce**

1. Add any `rx.recharts` chart to a route page, e.g.:
   ```python
   def my_page() -> rx.Component:
       return rx.recharts.area_chart(
           rx.recharts.area(data_key="value"),
           rx.recharts.x_axis(data_key="name"),
           data=[{"name": "A", "value": 1}],
           width="100%",
           height=300,
       )
   ```
2. Run `reflex run --env prod` (or deploy with the default Docker setup).
3. Navigate to the page containing the chart.
4. Observe infinite reload loop in the browser; browser console shows the errors below.

**Expected behavior**

The chart renders. The recharts components work as documented in the reflex-components-recharts package.

**Screenshots**

Browser console — production mode (Rolldown):
```
Error loading route module `/assets/_analytics_._run_._index-DyGOIpBr.js`, reloading page...
TypeError: t is not a function
    Ct  http://localhost:8180/assets/es6-3vbUkV3-.js:1
    o   http://localhost:8180/assets/rolldown-runtime-CATlyzt-.js:1
    wt  http://localhost:8180/assets/es6-3vbUkV3-.js:1
    o   http://localhost:8180/assets/rolldown-runtime-CATlyzt-.js:1
    <anonymous> http://localhost:8180/assets/es6-3vbUkV3-.js:1
```

Browser console — dev mode (Rollup):
```
Error loading route module `/app/routes/[analytics].[run]._index.jsx`, reloading page...
TypeError: require_isUnsafeProperty is not a function
    get.js:1
    __commonJSMin  chunk-C-Qwzh9l.js:8
    import_get     get.js:1
    recharts.js:1654
```

- Python Version: 3.14.5
- Reflex Version: 0.9.3 (reflex-components-recharts 0.9.1, recharts 3.8.1, es-toolkit 1.47.0)
- OS: Linux — Docker (Debian trixie), also reproduced on Windows 11 local dev
- Browser (Optional): Firefox 138, Chrome 136

**Additional context**

**Root cause:** `recharts/es6/SunburstChart.js`, `Sankey.js`, and `Treemap.js` each import `es-toolkit/compat/get.js`, which is a CommonJS shim (`module.exports = require('../dist/compat/object/get.js').get`). Because recharts declares `"sideEffects": false`, tree-shaking *should* eliminate these unused components. In practice they are retained — likely because `reflex-components-recharts` imports recharts through its barrel in a way that prevents the bundler from tracing live exports.

In **Rolldown** (Vite 8 production), all CJS interop helpers are emitted into a shared `rolldown-runtime-*.js` chunk. When React Router 7 lazily loads a route module, `rolldown-runtime` is not declared as a module-level dependency of the lazy chunk, so it may not be initialised before the route code runs → `t is not a function` (minified `__commonJS`).

In **Rollup** (dev mode), each CJS wrapper defines a local `require_isUnsafeProperty` helper. Code-splitting places recharts and es-toolkit in separate chunks; the helper is in one chunk's closure but called from another → `require_isUnsafeProperty is not a function`.

**Things tried that did not fix it:**

| Attempt | Result |
|---|---|
| `area_chart` / `line_chart` instead of `composed_chart` | No effect — barrel still retains es-toolkit |
| `react-router.config.js: unstable_optimizeDeps: false` | New error: MIME type `""` for `.vite/deps/` ×999 |
| `vite.config.js: optimizeDeps: { include: ['recharts'] }` | No effect in prod — `optimizeDeps` is dev-server only |
| `vite.config.js: manualChunks` (recharts + es-toolkit → one chunk) | Rolldown does not honour `manualChunks` the same way as Rollup |

my side project:
- **Code / Link to Repo:** https://github.com/performance-studio/openload  
- **PR where this surfaced:** https://github.com/performance-studio/openload/pull/13

masenf:
## Summary

Fixes #6561. Pages that render a `rx.recharts` chart could crash at runtime with a `TypeError` and an infinite route-reload loop.

recharts default-imports the CommonJS `es-toolkit/compat/*` shims (`get`, `range`, `sortBy`, `throttle`, …). These are reached through recharts' shared chart utilities, so even a single chart pulls a CJS module into the bundle. Vite then emits the CJS interop helper into a chunk that isn't guaranteed to be initialized before a lazily-loaded route chunk runs, which surfaces as `TypeError: … is not a function` and a reload loop.

This adds a small Vite plugin to the generated `vite.config.js` that redirects `es-toolkit/compat/<name>` default-imports to the pure-ESM `es-toolkit/compat` barrel (it re-exports the same functions). The module graph stays ESM-only, so no CJS interop helper is emitted, and the plugin is a no-op for apps that don't import `es-toolkit/compat`.

Thanks @hctilf for the thorough report and reproduction.

## Verification

Building a minimal `import get from 'es-toolkit/compat/get'` with Vite 8:

- **before:** the output includes the `__commonJS` interop helper and throws `require_isUnsafeProperty is not a function` at runtime
- **after:** the output is pure ESM with no interop helper, and `get()` still works

Also adds `tests/integration/tests_playwright/test_recharts.py`, which renders an area chart and asserts the SVG appears with no page errors. It runs in both dev and prod via the existing `app_harness_env` fixture.

## Test plan

- [ ] recharts integration test passes in CI (dev + prod)
- [ ] existing integration tests unaffected

---
_Generated by [Claude Code](https://claude.ai/code/session_01B53hepRQu8LCtZxaxAFPSm)_

### DIFF
--- a/packages/reflex-base/news/6561.bugfix.md
+++ b/packages/reflex-base/news/6561.bugfix.md
@@ -1,0 +1,1 @@
+Fixed recharts charts crashing at runtime with a `TypeError` and an infinite route-reload loop. recharts default-imports CommonJS `es-toolkit/compat/*` shims, whose interop helper could be bundled into a chunk that loaded after a lazy route chunk; the generated `vite.config.js` now redirects those imports to the pure-ESM `es-toolkit/compat` barrel.
--- a/packages/reflex-base/src/reflex_base/compiler/templates.py
+++ b/packages/reflex-base/src/reflex_base/compiler/templates.py
@@ -576,9 +576,10 @@ function fullReload() {{
 }}
 
 function esToolkitCompatEsm() {{
-  // recharts default-imports CJS shims like `es-toolkit/compat/get`; Rolldown can
-  // emit their interop helper into a chunk that loads after a lazy route chunk,
-  // crashing any page with a chart (#6561). Redirect to the pure-ESM compat barrel.
+  // recharts default-imports CJS shims like `es-toolkit/compat/get`; the CJS interop
+  // helper can land in a chunk that runs before it is initialized, crashing any page
+  // with a chart (#6561). Redirect to the pure-ESM compat barrel. Registered for both
+  // the prod build (plugins) and the dev prebundle (optimizeDeps.rolldownOptions).
   const VIRTUAL = "\0es-toolkit-compat:";
   return {{
     name: "vite-plugin-es-toolkit-compat-esm",
@@ -621,6 +622,11 @@ export default defineConfig((config) => ({{
           ],
         }},
       }},
+    }},
+  }},
+  optimizeDeps: {{
+    rolldownOptions: {{
+      plugins: [esToolkitCompatEsm()],
     }},
   }},
   experimental: {{
--- a/packages/reflex-base/src/reflex_base/compiler/templates.py
+++ b/packages/reflex-base/src/reflex_base/compiler/templates.py
@@ -575,10 +575,31 @@ function fullReload() {{
   }};
 }}
 
+function esToolkitCompatEsm() {{
+  // recharts default-imports CJS shims like `es-toolkit/compat/get`; Rolldown can
+  // emit their interop helper into a chunk that loads after a lazy route chunk,
+  // crashing any page with a chart (#6561). Redirect to the pure-ESM compat barrel.
+  const VIRTUAL = "\0es-toolkit-compat:";
+  return {{
+    name: "vite-plugin-es-toolkit-compat-esm",
+    enforce: "pre",
+    resolveId(source) {{
+      const match = /^es-toolkit\/compat\/([A-Za-z0-9_]+)$/.exec(source);
+      return match ? VIRTUAL + match[1] : null;
+    }},
+    load(id) {{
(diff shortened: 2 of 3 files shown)


masenf:
@greptile

masenf:
this is probably the wrong layer to patch this; upstream libraries should be handling this better. a fix like this is nothing more than a compat hack

PR #6568, opened 2026-05-27
ENG-9611: defer tracking/analytics scripts so they don't block render

adhami3310:
## Summary

- External `src` tracker scripts (gtag loader, clearbit) get `async_=True` so they no longer block HTML parsing while downloading.
- Inline tracker bootstraps (posthog, koala, common_room, rb2b, unify, default, gtag init) get `type="module"`, which the browser defers until after HTML parsing.
- All eight telemetry script components switch from `rx.script` to `rx.el.script`. These trackers reach the DOM via `App(head_components=...)`, which `reflex/compiler/utils.py:create_document_root` compiles straight into `<Head>` — so the Helmet wrapper that `rx.script` adds is redundant reconciliation cost.
- `gtag_report_conversion` is intentionally left non-module: it defines a global `function gtag_report_conversion()` that page `onclick` handlers call by name; module scope would hide it from `window`.

Linear: [ENG-9611](https://linear.app/reflex-dev/issue/ENG-9611/defer-trackinganalytics-scripts-so-they-dont-block-page-render)

## Cross-repo follow-up

`get_common_room_trackers` is currently rendered inside `rx.theme(rx.cond(...))` in two consumers (cloud, flexgen) rather than `head_components`. The old `rx.script`/Helmet wrapper hoisted it to `<head>` regardless; switching to `rx.el.script` means it now stays in the body where the theme renders it. The IIFE is idempotent and `type="module"` defers the same way regardless of DOM position, so it still works — but the head/body location changes for those two consumers.

A follow-up PR to **cloud** and **flexgen** should move the `get_common_room_trackers(...)` call into `head_components=` and adapt the `rx.cond(~AuthState.in_session & is_prod_mode(), ...)` gate accordingly.

## Test plan

- [x] `uv run ruff check .` clean
- [x] `uv run ruff format .` clean
- [x] `uv run pyright` clean on the telemetry package
- [ ] After merge/deploy: open DevTools → Network on a site using `get_pixel_website_trackers()` and confirm tracker requests are no longer in the render-blocking section of the waterfall

### DIFF
--- a/packages/reflex-components-internal/src/reflex_components_internal/blocks/telemetry/clearbit.py
+++ b/packages/reflex-components-internal/src/reflex_components_internal/blocks/telemetry/clearbit.py
@@ -19,4 +19,5 @@ def get_clearbit_trackers(public_key: str) -> rx.Component:
     return rx.el.script(
         src=CLEARBIT_SCRIPT_URL_TEMPLATE.format(public_key=public_key),
         referrer_policy="strict-origin-when-cross-origin",
+        async_=True,
     )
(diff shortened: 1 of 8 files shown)


adhami3310:
@greptile

masenf (approved):

PR #6573, opened 2026-05-28
Extend reflex docgen

carlosabadia:


packages/reflex-docgen/src/reflex_docgen/markdown/transformer/__init__.py
```diff
@@ -1,13 +1,49 @@
...
+        AttributeError: If the attribute does not exist.
+    """
+    if name == "ReflexComponentTransformer":
+        from reflex_docgen.markdown.transformer._reflex import (
+            ReflexComponentTransformer,
+        )
+
+        return ReflexComponentTransformer
+    msg = f"module {__name__!r} has no attribute {name!r}"
+    raise AttributeError(msg)
```
adhami3310: i think a compromise could be that the new file is a public module so, like `.reflex` such that imports would have to be `reflex_docgen.markdown.transformer.reflex`, and there you can define public API in `__all__`

adhami3310 (approved):

PR #6577, opened 2026-05-28
strict pin subpackage dependencies of "reflex" package

masenf:
During the `publish.yml` workflow, identify the latest version of subpackages reachable from the commit being published and strictly pin them in pyproject.toml.

This allows for repeatable installation of a pinned top level "reflex" package without worrying about newer, possibly incompatible, subpackages getting installed later. It also makes it possible to test behavior between top level reflex version without subpackage deps floating and changing the behavior.


adhami3310 (changes requested):
this is a bit radical of an approach, can you justify why its downsides are less impactful than its upsides?

masenf:
@adhami3310 we've had 5 releases since 0.9.0 and the following commands produce broken environments:

```
pip install reflex==0.9.0
pip install reflex==0.9.1
```

There's constant subpackage drift. It's not simple to go back and compare behavior between the latest reflex and one from a month ago, because installing the month-ago-version just doesn't work at all today.

A better option obviously would be to have the sub-packages themselves manage their lower bounds better. So `reflex-components-core-0.9.3` should depend on `reflex-base>=0.9.2` instead of currently depending on `reflex-base>=0.9.0`, which is just incorrect. What is a good way to reliably determine correct minimum dependencies prior to release and make sure that package dep declarations remain correct? We could use `uv` to resolve deps as minimal supported set and run the tests to see if they pass; this would be in addition to running with the latest versions (bundled in the repo).

So I suppose this PR is a bit of a sledgehammer -- and still doesn't address the underlying issue of more accurate subpackage dependency declarations, which we need anyway. I just don't want installations to be broken for users. And I've been personally hitting this like every day while testing issues and trying to compare behavior. Maybe I'm an outlier, but if I'm already butting up against this, I would assume others are as well.

adhami3310:
i understand the upsides, but the downsides are also quite drastic, if we would make a breaking change in lucide icons, then reflex needs a breaking change as well since they are tied one to one

i would rather make our sub packages accurately reflex-base version dep

masenf:
> i would rather make our sub packages accurately reflex-base version dep

I agree this would be ideal, but what is the strategy for making sure it happens? and if one slips out, like with reflex-components-core-0.9.2, the mitigation is kind of a nightmare

1. pin package to reflex-base >= 0.9.2
2. publish reflex-components-core-0.9.2.post1
3. yank reflex-components-core-0.9.2 (otherwise, it will still be picked by dep resolution)
4. repeat for 0.9.3

if we catch a bug report in a few weeks about an incompatibility with another package that we didn't notice at release time, do we do the same post-release, yank process for it? or do we just accept that some previous versions of reflex will not resolve to functional environments anymore?

"We only test the latest versions of everything and you should strict pin your working environment and only upgrade forward" is not the best policy IMO. I think the subpackages are great for future customizability of the install, but it's beginning to create a testing matrix that will be hard for us to manage.

Above all, we need to not be broken when installing previous versions of the framework, so maybe we just fix our pins and do the post-release/yank as soon as we notice breakage and just tell users that they need to pin transitives via `uv` or similar if they want a stable environment... it just doesn't feel right 🤷

adhami3310:
we can have a github action detecting if you made a change to X and Y, where X depend on Y, without bumping the dep in pyproject

masenf:
will handle this properly by making sure subpackage dependency min versions are specified correctly.

PR #6592, opened 2026-06-01
ENG-9585: Remove os templates

carlosabadia:


FarhanAliRaza (dismissed):
`packages/reflex-site-shared/.../templates/gallery_app_page.py`
`packages/reflex-site-shared/.../components/code_card.py`
They become orphaned if that is intentional.

masenf (dismissed):
there's more code that needs to be removed here, about to push a commit

FarhanAliRaza (commented):
https://github.com/reflex-dev/reflex/blob/d9640b16d74cd8c54d292d4f4459cf685b6883fb/packages/reflex-base/src/reflex_base/constants/base.py#L104

There is no user of this constant. 

in `validate_and_create_app_using_remote_template` 

```
  After simplification, template_url is always a non-None f-string on the only
  non-raising path, so:
  if template_url is None:
      return
```

FarhanAliRaza (approved):

PR #6595, opened 2026-06-02
fix rx.set_clipboard from a backend event handler on Safari/iOS

### LINKED ISSUE #6583: rx.set_clipboard from a backend event handler fails on Safari/iOS with NotAllowedError
### Describe the bug
On Safari (desktop) and iOS Safari, `rx.set_clipboard(...)` fails with
`NotAllowedError: The request is not allowed by the user agent or the
platform in the current context` whenever it is returned from a backend
`@rx.event` handler (or otherwise reached after a WebSocket round-trip).

The same code works on Chromium and Firefox.

### Root cause
Safari/WebKit requires `navigator.clipboard.writeText()` to be invoked
synchronously inside the user-activation window of the originating user
gesture (click/keydown). When `rx.set_clipboard` is returned from a backend
event handler, the actual `writeText` call runs from the socket.io message
handler, *after* the round-trip — so user activation is no longer in
effect and WebKit rejects the write.

### Minimal repro
```python
import reflex as rx

class State(rx.State):
    secret: str = "hello"

    @rx.event
    def copy(self):
        return rx.set_clipboard(self.secret)

def index():
    return rx.button("Copy", on_click=State.copy)

app = rx.App()
app.add_page(index)

carlosabadia:
closes https://github.com/reflex-dev/reflex/issues/6583


masenf (commented):
i'm quite hesitant on this one, because i don't like the idea of entering that clipboard.write promise for every event on a webkit browser.

is it possible to refactor the places where we're returning `set_clipboard` from the backend to just not do that. we can say `set_clipboard` has to be directly attached to an event trigger and maybe even raise an error/warning if someone tries to use it from the backend?

the other option i was considering, but @adhami3310 didn't really like, is exposing this clipboard arming behavior as an event action (like debounce or throttle). then backend events that wanted clipboard permissions could set this as a kwarg on the `rx.event` decorator or pass it at the trigger site to opt-in to the clipboard arming. this _kind of_ falls apart for chained events on the backend though because the clipboard opt-in event stuff would only work if that event handler is directly attached to a component's trigger.

i still think just warning when set_clipboard is used from the backend and maybe updating the docs to show examples of directly attaching it to a trigger is the simplest solution so we don't have to maintain hacks or try to break safari's security model for questionable gains.

carlosabadia:
> i'm quite hesitant on this one, because i don't like the idea of entering that clipboard.write promise for every event on a webkit browser.
> 
> is it possible to refactor the places where we're returning `set_clipboard` from the backend to just not do that. we can say `set_clipboard` has to be directly attached to an event trigger and maybe even raise an error/warning if someone tries to use it from the backend?
> 
> the other option i was considering, but @adhami3310 didn't really like, is exposing this clipboard arming behavior as an event action (like debounce or throttle). then backend events that wanted clipboard permissions could set this as a kwarg on the `rx.event` decorator or pass it at the trigger site to opt-in to the clipboard arming. this _kind of_ falls apart for chained events on the backend though because the clipboard opt-in event stuff would only work if that event handler is directly attached to a component's trigger.
> 
> i still think just warning when set_clipboard is used from the backend and maybe updating the docs to show examples of directly attaching it to a trigger is the simplest solution so we don't have to maintain hacks or try to break safari's security model for questionable gains.

sounds good, ill close this one and lets put up one for the warning

PR #6604, opened 2026-06-04
Update docs dark mode palette

carlosabadia:


masenf (approved):
looks the same to me running locally

PR #6613, opened 2026-06-05
[claude-code] Add SessionStart hook to provision web environment

masenf:
For running properly in Claude Code web containers, which are baked with tooling that predates this repo's current requirements.

### `.claude/hooks/session-start.sh` (registered via `.claude/settings.json`)

Runs only in web containers (`CLAUDE_CODE_REMOTE=true`) and reconciles the environment with the repo on every session start. It does **not** sync dependencies — that cost is paid lazily on the first `uv run` / `uv sync` (e.g. running tests or an app), so sessions that don't touch the code start fast.

- **uv**: if the `uv` on `PATH` is older than `tool.uv.required-version`, pip-install a satisfying one for `--user` (lands in `~/.local/bin`, assumed already on `PATH`). The astral.sh installer and `uv self update` are blocked / GitHub-rate-limited here, so PyPI is the only reliable source.
- **Python**: an older uv's bundled release manifest only knows 3.14 prereleases, so a later `uv sync` would resolve `.python-version=3.14` to a release candidate whose `typing._eval_type` signature is incompatible with the pinned pydantic (breaks every import). Upgrading uv fixes the manifest; `uv python install` (which excludes prereleases) then guards against reusing a prerelease a stale-uv run may have fetched.
- **git tags**: web clones are shallow with no tags, so Reflex's VCS-derived version resolves to `0.0.0` and trips downstream version gates (e.g. reflex-hosting-cli's compatibility check). Fetching tags lets the version compute.

### `.gitignore`

Narrow the previous blanket `.claude` ignore to just `.claude/.worktrees` and `.claude/settings.local.json` so the shared hook and `settings.json` are tracked, ignore user-local `CLAUDE.local.md`, and ignore tool scratch dirs (`__pycache__`, `.pytest_cache`, `.ruff_cache`, `.mypy_cache`).

> https://claude.ai/code/session_014Bp9oVCfiq8DSnTK8VfozK

### DIFF
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -34,9 +34,14 @@ cd "$CLAUDE_PROJECT_DIR"
 req="$(sed -n 's/^[[:space:]]*required-version[[:space:]]*=[[:space:]]*"[^0-9]*\([0-9][0-9.]*\).*/\1/p' pyproject.toml | head -n1)"
 python3 -m pip install --user --quiet --root-user-action=ignore "uv${req:+>=$req}"
 
-# pip --user installs land in ~/.local/bin; make sure they win over the baked uv.
+# pip --user installs land in ~/.local/bin; make sure they win over the baked uv,
+# both for the rest of this script and -- via CLAUDE_ENV_FILE -- for the session's
+# later commands, which otherwise wouldn't see this PATH change once we exit.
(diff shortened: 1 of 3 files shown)


masenf:
@greptile

masenf:
@greptile

adhami3310 (approved):

PR #6619, opened 2026-06-05
Support hybrid_property on ObjectVar for nested dataclasses

### LINKED ISSUE #6617: @rx._x.hybrid_property should work on a model/dataclass

we want to support the following code

```python
from dataclasses import dataclass

import reflex as rx
from reflex.experimental.hybrid_property import hybrid_property


@dataclass
class Info:
    a: str
    b: str

    @hybrid_property
    def a_b(self) -> str:
        return f"{self.a} - {self.b}"


class State(rx.State):
    info: Info = Info(a="a", b="b")


def index() -> rx.Component:
    return rx.vstack(
        rx.text(
            State.info.a,
        ),
        rx.text(
            State.info.b,
        ),
        rx.text(
            State.info.a_b,
        ),
        spacing="5",
    )


app = rx.App()
app.add_page(index)
```

### discussion on discord

**Bert — 10:27**
If you do decide to support dataclasses, then please consider also supporting the SQLAchemy "dataclass" equivalent i.e..:

class Base(MappedAsDataclass, DeclarativeBase, kw_only=True)
**masenf — 10:31**
i'm assuming it works as expected in the backend? you're just needing it to render as a var when used in the frontend
**Bert — 10:40**
Correct but since the dataclass @property doesn't work on the client side in any version it has limited use on the server side. To get around this I typically use BaseModel (Pydantic) with:
    @computed_field
    @property

But that will create, cache and send the combined fields. A typical example is I have several fields that together are used to create a unique id (e.g. manufacturer + product id)

With the new @hybrid property, we could save the extra baggage especially if we have multiple different representations based on the other attributes in the class.

So that brings up that it would also be nice to support it in pydantic BaseModel classes since @hybrid property would be better there too...

**masenf — 10:42**
if we support it anywhere, it should be supported everywhere, because under the hood we treat BaseModel, dataclass, declarative base, etc all as ObjectVar

**Additional context**
When the property `State.info.a_b` is accessed in the frontend, it should use Var-access semantics as if the hybrid_property was directly accessed on the state

masenf:
### Type of change

### Description

This PR extends `hybrid_property` support to work seamlessly with `ObjectVar` attribute access on nested dataclasses, Pydantic models, and SQLAlchemy models—not just `State` classes.

**Key changes:**

1. **Moved `HybridProperty` to `reflex_base`:** Relocated `hybrid_property.py` from `reflex/experimental/` to `reflex_base/vars/` to make it available at the var level where it's needed for `ObjectVar` resolution.

2. **`ObjectVar` descriptor resolution:** Modified `ObjectVar.__getattr__` to detect and resolve `HybridProperty` descriptors on the underlying type. When a hybrid property is accessed (e.g., `State.info.a_b`), the property's frontend logic is evaluated with the object var substituted as `self`, enabling consistent var-level semantics.

3. **`HybridProperty._get_var` extraction:** Refactored the var-resolution logic into a dedicated `_get_var` method that accepts either a class or an `ObjectVar` as the owner, allowing the property to work uniformly across both direct class access and nested object var access.

4. **Comprehensive test coverage:** Added unit tests verifying hybrid property resolution on bare classes, Pydantic models, SQLAlchemy models, and dataclasses. Included integration tests demonstrating re-rendering when nested dataclass fields are updated.

**Example usage:**
```python
@dataclass
class Info:
    a: str
    b: str
    
    @hybrid_property
    def a_b(self) -> str:
        return f"{self.a} - {self.b}"

class State(rx.State):
    info: Info = Info(a="a", b="b")

# Frontend renders as: "a - b"
rx.text(State.info.a_b)
```

Closes #6617

### Testing

- [x] Added unit tests in `tests/units/vars/test_object.py` covering hybrid property resolution on all object types
- [x] Added integration tests in `tests/integration/test_hybrid_properties.py` verifying frontend rendering and re-rendering on state updates
- [x] Existing tests pass with new functionality

https://claude.ai/code/session_01DKFiYGnWRQG8wMNKFW7obm


masenf:
@greptile

benedikt-bartscher:
This is great, thanks @masenf !

FarhanAliRaza (approved):

PR #6625, opened 2026-06-08
Fix router.url.query going stale after history.replaceState

### LINKED ISSUE #6603: Reflex 0.9: self.router.url.query not refreshed when URL is updated with window.history.replaceState
**Describe the bug**
I have a select filter that updates the URL query parameters whenever the selected filter value changes. The URL is updated using rx.call_script with window.history.replaceState.

Before upgrading to Reflex 0.9, updating the URL this way would also update the route query values accessible through self.router.url.query.

After upgrading to Reflex 0.9, the browser URL changes correctly, but self.router.url.query no longer reflects the updated query parameters.

I run with reflex run --env prod

**To Reproduce**
1. Create a select filter component, for instance with query parameter: name.
2. On filter change, update the URL query parameters using:
    rx.call_script(
        "window.history.replaceState({}, '', '?name=test')"
    )
3. Read the current query parameters from: self.router.url.query
4. Observe that after the URL is updated, self.router.url.query still contains the old values.

- Code/Link to Repo:

**Expected behavior**
When the URL query parameters are modified via window.history.replaceState, the router state should stay in sync and self.router.url.query should reflect the updated query parameters.

**Screenshots**

- Python Version: 3.11
- Reflex Version: 0.9.4
- OS: Linux
- Browser (Optional): Chrome

**Additional context**

masenf:
### Type of change

### Description

Fixes #6603: `State.router.url.query` and other `router_data` fields were going stale when the URL was modified via `window.history.replaceState` or `pushState` (e.g., through `rx.call_script`). 

**Root cause:** React Router's `useLocation` hook does not observe direct history manipulation. The frontend was reading the mirrored location from React Router instead of the live `window.location`, causing the backend to receive stale query parameters on subsequent events.

**Solution:** When populating `router_data` in `applyEvent`, read the live `window.location.search` and `window.location.hash` instead of relying on React Router's location object. The pathname remains basename-relative (from React Router) to avoid double-applying the frontend path prefix. In embed mode, the mirrored location is used since the host page's `window.location` is unrelated to the in-widget memory router.

### Changes

1. **`packages/reflex-base/src/reflex_base/.templates/web/utils/state.js`**: Modified `applyEvent` to read live query string and hash from `window.location` when `router_data` is empty, ensuring synchronization with direct history manipulation.

2. **`tests/integration/tests_playwright/test_router_query.py`**: Added comprehensive integration test covering both dev and prod modes. The test verifies that:
   - Query parameters set via `history.replaceState` are correctly reflected in `State.router.url.query`
   - Multiple sequential `replaceState` calls keep the router in sync
   - The `query_parameters` dict is properly populated

3. **`news/6603.bugfix.md`**: Added changelog entry.

### Test Plan

The new integration test (`test_replace_state_syncs_router_query`) covers the fix in both dev and prod modes via `app_harness_env` parametrization. It verifies that:
- Initial state has no query parameters
- After `history.replaceState("?name=test")`, reading the query returns `"name=test"`
- After a second `replaceState("?name=other&page=2")`, the query is updated correctly

All existing tests continue to pass.

https://claude.ai/code/session_01JwZCEa2bkfK9QDp4PP84d8


tests/integration/tests_playwright/test_router_query.py
```diff
@@ -0,0 +1,148 @@
...
+            Returns:
+                A call_script event that runs window.history.replaceState.
+            """
+            return rx.call_script(f"window.history.replaceState({{}}, '', {query!r})")
+
+        @rx.event
+        def read_query(self):
+            """Read the current router query into state vars."""
+            self.query_str = self.router.url.query
+            self.name_param = self.router.url.query_parameters.get("name", "")
```
masenf: add 2 `rx.var` computed vars that return these values and render those in the test app. add assertions that these values change automatically due to the computed var _before_ the user explicitly interacts with the app by sending the `read_query` event.

changing the url via replaceState should induce a reactive change that is reflected without further interaction.

FarhanAliRaza (approved):

PR #6634, opened 2026-06-09
ENG-9533 feat: render concise, readable types in docgen class documentation

FarhanAliRaza:
Rework reflex-docgen's type/signature rendering so generated class docs read cleanly instead of leaking implementation noise.

- Add a public format_type() that strips module qualifiers, preserves type-alias names, and renders unions (Optional[...]), literals, and callables recursively.
- Define the ASGI aliases (Scope, Message, Receive, Send, ASGIApp) via TypeAliasType so their short names survive introspection and docs show e.g. ASGIApp rather than a fully expanded MutableMapping/Awaitable blob.
- Render method signatures without self/cls, with forward-ref strings kept verbatim (each annotation independently, so one TYPE_CHECKING-only name doesn't poison the rest) and optionality normalized to match format_type.
- Clean field/parameter defaults: show function names or empty-collection literals instead of volatile <function ... at 0x...> reprs, and exclude function-valued field defaults from the methods table.
- Polish App attribute docstrings with doc links and a fuller api_transformer description.

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

Before

image

After
image


masenf:
the App, Event, EventHandler, EventSpec,  page still has all of the attributes listed as body text in addition to appearing in the table

image

The `Model` page should be removed from auto-gen reference pages, since it is now deprecated.

FarhanAliRaza:
Fixed. 
image

masenf (approved):

PR #6641, opened 2026-06-10
Deflake test_memo: gate on hydration and capture page diagnostics

masenf:
test_memo interacted with the page immediately after goto, racing the
websocket connect/hydrate startup; the typed input event could be lost,
leaving #memo-last-value empty through every rerun of an affected CI
job. Gate all memo tests on the standard #token hydration marker used
by the rest of the playwright suite before interacting.

Also add a tests_playwright conftest that records console messages,
page errors, and websocket frames per test and attaches them to failed
test reports, so any future flake of this kind is diagnosable directly
from CI job logs instead of reproducing blind.

https://claude.ai/code/session_01Xvaqxi4LJeWu2fo4Fk7qMD

### DIFF
--- a/tests/integration/tests_playwright/conftest.py
+++ b/tests/integration/tests_playwright/conftest.py
@@ -1,0 +1,70 @@
+"""Conftest for Playwright integration tests.
+
+Records browser-side activity (console messages, page errors, websocket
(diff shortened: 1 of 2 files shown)


masenf:
this actually doesn't seem like it would fix the problem

PR #6658, opened 2026-06-12
ENG-9801: sync `reflex.lock/package.json` to `.web/package.json`

masenf:
Every time the framework requests installation of packages, sync the `package.json` along with the lock files, so out of band changes are picked up.

After init, sync any package.json or lock files to reflex.lock so subsequent installs keep the initialized hooks and overrides.

The reflex.lock/package.json is interesting because it's checked in, and it's the canonical copy of `package.json` that will replace what exists in `.web`... but it's also a deletable lock-ish file that will itself be recreated from the compiled app. After this change though, the reflex.lock/package.json is now an appropriate place to keep local changes like package version overrides.

_PR written without AI assistance 😅_


masenf (commented):
ideally we would consolidate the package.json combining code in frontend_skeleton.py with that of templates.py... but i'm holding off on that change until 0.10.0 when we combine the compiler into reflex-base, otherwise we'll be introducing cross-package incompatibility for no real gain other than hygiene.

FarhanAliRaza (dismissed):
Maybe we want to add tests for these cases 
(a) prune=False preserving .web/package.json when
  the root copy is absent
(b) initialize_web_directory persisting package.json back to root.

FarhanAliRaza (approved):

PR #6659, opened 2026-06-12
fix(memo): preserve runtime typing for unannotated params

### LINKED ISSUE #6631: `@rx.memo` with unannotated params silently degrades to `Var[Any]` in 0.9.4 → `UntypedVarError` at compile (regression from 0.9.3)
**Describe the bug**

A legacy `@rx.memo` (the old custom-component decorator) whose parameters are **unannotated** compiled fine on 0.9.3 and raises `UntypedVarError` at compile time on 0.9.4.

The memo refactor in 0.9.4 (`de6393e9` / #6517, softened by `baf10e65` / #6598) changed how a memo parameter's `Var` **type** is derived:

- **0.9.3** typed each prop placeholder from the **call-site value** — `CustomComponent.get_prop_vars()`:
  ```python
  Var(
      _js_expr=name + CAMEL_CASE_MEMO_MARKER,
      _var_type=(prop._var_type if isinstance(prop, Var) else type(prop)),
  ).guess_type()
  ```
- **0.9.4** types it from the **parameter annotation** — `_var_placeholder()` → `_annotation_inner_type()`, and in `_analyze_params()` a **missing annotation** becomes `Var[Any]`:
  ```python
  if is_missing or is_legacy:
      ...
      elif is_missing:
          annotation = Var[Any]      # <- untyped placeholder
  ```

So an unannotated param that used to be typed from what the caller passed is now an untyped `Var[Any]`, and any operation on it fails. (#6598 added bare-type coercion + a deprecation warning, so `param: dict` works — only the *missing*-annotation case still degrades to `Var[Any]`.)

**To Reproduce**

```python
import reflex as rx

@rx.memo
def user_card(user):                 # legacy unannotated param
    return rx.box(
        rx.heading(user["name"]),
        rx.text(user["email"]),
    )

def index():
    return user_card(user={"name": "Ada", "email": "ada@example.com"})

app = rx.App()
app.add_page(index)
```

`reflex run` (or even `index().render()`) raises:

```
UntypedVarError: Cannot access the item 'name' on untyped var 'userRxMemo'
of type 'typing.Any'. You can call the var's .to(desired_type) method to
convert it to the desired type.
```

The surface error depends on what the body does with the param (all compile on 0.9.3):

| memo body on an unannotated param | 0.9.4 result |
|---|---|
| `user["x"]` / `user.x` | `UntypedVarError` |
| `name.upper()`, `.lower()`, `.split()`, `.startswith()` | `UntypedVarError` |
| `rx.foreach(items, ...)` | `ForeachVarError` |
| `"Hello " + name` | `TypeError: can only concatenate str (not "Var") to str` |
| `rx.input(on_change=handler_param)` | `ValueError: Invalid event chain ... of type typing.Any` |

**Expected behavior**

Either keep the 0.9.3 behavior (type an unannotated / `Var[Any]` param from the bound call-site value via `.guess_type()`), or treat a missing annotation as the loud, eager deprecation/error case rather than a silent `Var[Any]` that explodes downstream with a confusing message.

- Python Version: reproduced on 3.10 and 3.14
- Reflex Version: regression introduced in **0.9.4** (works on 0.9.3)
- OS: any (reproduced on Linux; not OS-specific — see below)

**Additional context**

Verified head-to-head on clean PyPI installs:

| | Python 3.10 | Python 3.14 |
|---|---|---|
| 0.9.3 | ✅ compiles | ✅ compiles |
| 0.9.4 | ❌ `UntypedVarError` | ❌ `UntypedVarError` |

Workaround for users: annotate the param, e.g. `user: rx.Var[dict]` (or `rx.RestProp`, `rx.Var[rx.Component]` for children).

This was investigated alongside a telemetry spike in `UntypedVarError`, `EventHandlerArgTypeMismatchError`, and `AttributeError` on 0.9.4:
- `EventHandlerArgTypeMismatchError` fires **identically** on 0.9.3 and 0.9.4 in every path tested (`call_event_handler` / `_check_event_args_subclass_of_callback` are unchanged between the tags) — not reproducible as a regression; likely secondary.
- `AttributeError`: `VarAttributeError` is unchanged across versions; the only new `AttributeError` path is `HybridProperty.__get__` (#3806) — unconfirmed as a regression.
- The "Linux" skew in telemetry is most likely a deployment/volume artifact (apps recompile at scale in Linux CI/Docker/prod, where memo bodies first evaluate), not an OS-specific code path. Suggest re-slicing telemetry by Python version and `run` vs `export`.

Distinct from #6599 (RestProp CSS props). A related secondary 0.9.4 regression: base props (`class_name`, `style`, `id`) on a `@rx.memo` without `rx.RestProp` now raise `TypeError` where 0.9.3 forwarded them.

harsh21234i:
## Summary

  Fixes #6631.

  `@rx.memo` component bodies with unannotated parameters were regressing to `Var[Any]` in 0.9.4, which caused downstream `UntypedVarError` failures when the memo body
  tried to access object fields or call var methods.

  This change preserves the legacy behavior for missing annotations by seeding the memo body with the first call’s runtime value type before the component is
  materialized. That keeps the deprecation warning in place, but avoids the confusing compile-time failure for existing call sites.

  ## Testing

  - `uv run pytest -o addopts='' tests/units/components/test_memo.py -q`

<h3>Greptile Summary</h3>

This PR fixes a regression introduced in 0.9.4 where `@rx.memo` component bodies with unannotated parameters would see `Var[Any]` placeholders, causing `UntypedVarError` failures when the body accessed object fields or called var methods. The fix seeds the lazy body's first evaluation with the first call's runtime value type for truly-unannotated params, while leaving explicitly-typed and bare-type params unchanged.

- Adds `_runtime_inferred_params: frozenset[str]` to `MemoComponentDefinition` to track which params need runtime type seeding, populated by the new `missing_params` output of `_analyze_params`.
- Extends `_LazyBody.get()` with an optional one-shot `thunk` override and exposes `is_ready`, so `_MemoComponentWrapper.__call__` can inject a `runtime_values`-aware body thunk on the first call without touching the eagerly-evaluated passthrough path.
- New tests cover the plain-Python-object path, the `Var`-typed-runtime-value path, and confirm that explicit `Var[Any]` annotations are not overridden.

<h3>Confidence Score: 5/5</h3>

Safe to merge — the change is narrowly scoped to unannotated-param type seeding on the first call, the lazy/eager body paths are cleanly separated, and existing memo behaviour is unaffected.

The core mechanism (one-shot thunk override guarded by `is_ready`) is correct and the three new tests cover the plain-value path, the Var-typed path, and the explicit-Any preservation case. The only non-blocking note is that `LiteralVar.create` on an unusual Python object could raise before falling back to `Var[Any]`, but this edge case does not affect normal usage.

**Files Needing Attention:** No files require special attention.

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| packages/reflex-base/src/reflex_base/components/memo.py | Core fix: adds `_runtime_inferred_params` tracking, a `thunk`-override to `_LazyBody.get()`, and an `is_ready` property to drive first-call type seeding for unannotated memo params. Logic is sound; lazy/eager body paths are cleanly separated. |
| tests/units/components/test_memo.py | Three new tests cover plain-Python-value path, Var-runtime-value path (addressing the previously untested branch), and explicit Var[Any] preservation. A unit test for the new `_LazyBody.get(thunk)` override is also added. |
| packages/reflex-base/news/6659.bugfix.md | New changelog entry describing the bugfix. |
| pyi_hashes.json | Hash update for the regenerated memo.pyi stub, consistent with the signature change to `_LazyBody.get`. |

</details>

<sub>Reviews (17): Last reviewed commit: ["pyi hashes"](https://github.com/reflex-dev/reflex/commit/e50cd963539b7b573ba6ad34528de6f41a8017f2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37185587)</sub>


tests/units/components/test_memo.py
```diff
@@ -520,6 +520,40 @@ def soft_missing(value) -> rx.Component:
...
 
+def test_memo_uses_first_call_value_type_for_missing_param_annotation():
+    """Component memos should infer missing parameter types from the first call."""
+
+    @rx.memo
+    def user_card(user) -> rx.Component:
+        return rx.box(
+            rx.heading(user["name"]),
+            rx.text(user["email"]),
+        )
```
masenf, line 547: some assertion on the runtime type of the `user` var is needed here

harsh21234i: done!

harsh21234i:
any update? @masenf

masenf:
for the runtime type check, i meant, inside the memo function body, there should be an assertion that the `user` arg is passed as the correct typed Var, so an ObjectVar with _var_type == dict in this case.

overall i like the compatibility with the previous implementation, but it would be nice if this path raised some kind of deprecation warning indicating to users that they should be typing their memo function args for best results.

harsh21234i:

harsh21234i:
can you look into this @masenf ?

FarhanAliRaza (commented):
I dont like the solution how it is dependent on  dict that we have to clear manually. And it repeats the work at every run.

FarhanAliRaza:
Follow-up in `c66ed977`:

- Runtime type inference is now limited to parameters with no annotation; an explicit `rx.Var[Any]` remains intentionally untyped.
- Removed the shared mutable runtime-value dictionary. The first call supplies a contextual `_LazyBody` thunk, whose result is cached for subsequent calls.
- Restored eager evaluation for strict internal memo definitions.
- Added regressions covering explicit `Var[Any]` and one-time contextual lazy evaluation.

Validated with 96 memo unit tests, 6 Playwright memo tests, and a live Reflex app compile/browser run that confirmed the generated `memo(...)` components and rendered output.

FarhanAliRaza (approved):

masenf (approved):

PR #6662, opened 2026-06-15
ENG-9830 fix: avoid config re-entry when a State is defined in rxconfig.py

### LINKED ISSUE #6661: [regression] circular import when rxconfig.py attempts to import State
**Describe the bug**
The State `__init_subclass__` path depends on `get_config`. So the config itself is unable to actually import any State classes, which makes it impossible to express any dynamic state vars passed to plugins, like `RadixThemesPlugin`.

**To Reproduce**
In `rxconfig.py`:

```
class State(rx.State):
    ...
```

**Expected behavior**
User should be able to define and/or import State classes in the rxconfig.py for the purposes of dynamic configuration.

Alternatively, if there was a way to define config and plugins outside of an importable module, that might be better and more forward looking.

For example, if the config defined some kind of `__post_init__` callback, then dynamic configuration could be deferred until the config object actually existed. And it would allow mostly static (toml) config, while still providing a hook for running python code after the static config is loaded.

- Python Version: 3.10+
- Reflex Version: 0.9.3+
- OS: any

FarhanAliRaza:
State-class creation read get_config().state_auto_setters at class-creation time, which re-entered config loading while rxconfig.py was still importing and raised AttributeError. Cache the flag when the Config is built and read it via get_state_auto_setters(), falling back to the env var before any Config exists.

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

closes #6661

### DIFF
--- a/news/6662.bugfix.md
+++ b/news/6662.bugfix.md
@@ -1,0 +1,1 @@
+Avoid re-entering config loading when a `State` subclass is defined in `rxconfig.py`.
--- a/packages/reflex-base/news/6662.bugfix.md
+++ b/packages/reflex-base/news/6662.bugfix.md
@@ -1,0 +1,1 @@
+Avoid re-entering config loading when a `State` subclass is defined in `rxconfig.py`.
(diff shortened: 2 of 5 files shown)


masenf (approved):
i know we need and want this functionality, but this implementation just seems like such a band-aid fix...

i'm going to take it for now, but i want to remove it as soon as auto_state_setters goes away

PR #6672, opened 2026-06-19
ENG-8753 docs: add enterprise auth documentation section

FarhanAliRaza:
Add docs for the OIDC AuthPlugin covering the secure-by-default model, providers, custom auth pages, and testing guarded code. Register the new pages in the enterprise sidebar, add an Authentication category to the enterprise overview, and whitelist the section for preview.

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:


docs/enterprise/auth/secure-by-default.md
```diff
@@ -0,0 +1,581 @@
...
+A check may be `async` — return an awaitable and the framework awaits it. Whether
+a check ran sync or async is decided by its **result**, not by inspecting the
+function, so a sync check is run inline while an async one is awaited at the
+right point (the per-event gate for handlers, the delta resolution for
+fields/vars).
+
+An async check can reach **any other state**, a database, or a remote
+authorization service (e.g. OpenFGA / a ReBAC backend) via
+`await ctx.auth_user_state.get_state(...)` — the round-trip a sync check can't
+make:
```
masenf: this sentence is kind of confusing; sounds like you make openfga checks with `get_state`... but that function is only for accessing other states

docs/enterprise/auth/secure-by-default.md
```diff
@@ -0,0 +1,581 @@
...
+```python
+import reflex as rx
+import reflex_enterprise as rxe
+from reflex_enterprise.auth import AuthContext
+
+
+class PolicyState(rx.State):
+    """A sibling state (or a stand-in for a DB / authz service)."""
+
+    admin_group: rx.Field[str] = rxe.field("admins", auth=False)
```
masenf: i don't love this example, but we should at the least be using a backend var for anything related to auth checks

masenf (commented):
i dont really like the overall tone of the docs, it reads as _very_ LLMish.

Stuff like this is particularly bad:

* "The plugin auto-installs middleware that blocks cross-site GET navigations to /logout — using the browser-set, JS-unspoofable Sec-Fetch-Site header — so an attacker's <iframe> can't silently log the user out (it is redirected to the frontend root before the SPA boots)"
* "User is an alias of AuthUserState; you can read claims off either name."
  * the part after the semicolon is obvious if it's an alias, we don't need to patronize our readers. The sentence could be improved by saying "User is an alias of reflex_enterprise.auth.AuthUserState and may be used interchangeably"
* "This page covers the default provider, naming your own, the environment variables each one reads, registering providers with the plugin, scopes and refresh tokens, running several providers at once, the claims they return, and the advanced hooks you can override."
  * No one wants to read a long running inline list. Either bullet or remove. The headings are visible on the right anyway, so the user knows what to expect on this page.
* "Render a login button for the provider with its get_login_button() classmethod (pass children to customize the clickable element):"
  * This seems like it should maybe be under the customization section, feels kind of out of place on the OIDC provider page, especially because we would prefer them to redirect to `/login` and display the palette rather than rendering provider-specific login buttons themselves.
* The dash then rephrasing what was obviously stated is an annoying LLMism. These should be rephrased more like reference material, not like an LLM explaining/justifying a PR.
  * "When the app detects it is embedded in an iframe, login and logout automatically switch from a top-level redirect to a popup window with a postMessage token hand-off — so a Reflex auth app embeds inside another product with no extra configuration. "
  * "This auto-switch only works through the provider's get_login_button() — it mounts the message listener that receives the tokens posted from the popup"

idk i could keep going, but my recommendation is to have your agent reword a lot of this material to be less conversational and LLMish and moreso fit the tone of reference documentation.

docs/enterprise/auth/overview.md
```diff
@@ -0,0 +1,215 @@
...
+
+```python
+import reflex as rx
+from reflex_enterprise.auth import GenericOIDCAuthState
+
+# Either link to the logout route.
+rx.link("Sign out", href="/logout")
+
+# Or bind the provider's logout event.
+rx.button("Sign out", on_click=GenericOIDCAuthState.redirect_to_logout)
```
masenf: this isn't really the same as linking to `/logout`, because it only affects the default provider. if the app is using multi-provider, then it wont work.

we should either remove this, or add a dispatching `redirect_to_logout` event to `AuthUserState`

docs/enterprise/auth/providers.md
```diff
@@ -0,0 +1,427 @@
...
+
+### Reading granted scopes
+
+Every provider exposes a `granted_scopes` Var containing the space-delimited
+scopes the IdP actually granted. Use it to gate optional-scope features at
+runtime:
+
+```python
+rx.cond(
+    DatabricksAuthState.granted_scopes.contains("offline_access"),
```
masenf: Unclear how to know whether `DatabricksAuthState` is the active one. I'd consider an example inside an event handler that does

```
"offline-access" in await User.current_provider().granted_scopes
```

docs/enterprise/auth/providers.md
```diff
@@ -0,0 +1,427 @@
...
+    rx.text("Long-lived session enabled"),
+    rx.text("Session ends at token expiry"),
+)
+```
+
+## Multiple providers
+
+With two or more providers, `/login` shows one login button per provider instead
+of redirecting automatically to a single IdP. The callback resolves the
```
masenf: i dont think we redir automatically anymore right?

masenf (commented):
wherever we say "org wide", we should say "app wide"

docs/enterprise/auth/providers.md
```diff
@@ -0,0 +1,427 @@
...
+    ):
+        await super()._set_tokens(
+            access_token, id_token, refresh_token, granted_scopes, **kwargs
+        )
+        self._org_id = org_id
+```
+
+The popup/iframe login path sets tokens directly and does **not** route through
+`_set_tokens_payload_from_exchange`. Give the extra `_set_tokens` keyword a
+default (as `org_id=""` above) to keep that path working.
```
masenf: not sure this is exactly true. the popup path _does_ hit this function, but saving it as a state var wont work due to the popup being a separate client token. i wonder if we should mention the iframe auth payload? might be a little excessive for now, but wanted to call it out here.

masenf (commented):
aside from the outstanding comments, this looks good to go

masenf (approved):

PR #6676, opened 2026-06-23
fix(match): bind useContext for state vars used in rx.match case conditions

### LINKED ISSUE #6675: rx.match: state vars in case conditions (component branches) emit unbound useContext → ReferenceError: Can't find variable: <substate>
## Description

When `rx.match(...)` returns **components** in its cases and a **state var is used in a case _condition_** (e.g. `(MyState.x > 0, some_component)`), the compiled page references the substate's React context variable but never emits the matching `useContext(...)` binding. At render the browser throws:

```
ReferenceError: Can't find variable: reflex___state____state__<app>___states___<module>____<state>
```

This happens whenever a state var appears only in the match **conditions** (not in the subject) with component return values — most commonly via the idiomatic "match on a literal subject" pattern:

```python
rx.match(True, (State.x > 0, comp_a), (State.y > 0, comp_b), default_comp)
```

## Minimal reproduction

`rxconfig.py`:

```python
import reflex as rx

config = rx.Config(app_name="rx_match_repro")
```

`rx_match_repro/rx_match_repro.py`:

```python
import reflex as rx


class MatchState(rx.State):
    a: int = 0
    b: int = 0


def index() -> rx.Component:
    return rx.match(
        True,
        (MatchState.a > 0, rx.text("a positive")),
        (MatchState.b > 0, rx.text("b positive")),
        rx.text("neither"),
    )


app = rx.App()
app.add_page(index)
```

`reflex run` and open `/` → blank page, and the browser console shows:

```
ReferenceError: Can't find variable: reflex___state____state__rx_match_repro___rx_match_repro____match_state
```

## Compiled output (the bug)

`.web/app/routes/_index.jsx` (reflex `0.9.5.post2`):

```jsx
  return (
    jsx(Fragment,{},jsx(Fragment,{},(() => {
  switch (JSON.stringify(true)) {
    case JSON.stringify((reflex___state____state__rx_match_repro___rx_match_repro____match_state.a_rx_state_ > 0)):
      return jsx(RadixThemesText,{as:"p"},"a positive");
    case JSON.stringify((reflex___state____state__rx_match_repro___rx_match_repro____match_state.b_rx_state_ > 0)):
      return jsx(RadixThemesText,{as:"p"},"b positive");
    default:
      return jsx(RadixThemesText,{as:"p"},"neither");
  }
})()), ... )
  )
}
```

`Component()` references `reflex___state____state__...____match_state` in the `switch` cases but contains **no** `const ... = useContext(StateContexts....)` binding (`grep -c useContext _index.jsx` → `0`). Hence the `ReferenceError` on first render.

For contrast, `rx.cond`/`rx.foreach` are extracted into memoized child components that each emit their own `useContext` binding, so they work.

## Root cause

`reflex_components_core/core/match.py`, `Match._create_match_cond_var_or_component`:

The **Var-return** branch merges the case-condition var-data correctly:

```python
return Var(
    _js_expr=format.format_match(...),
    _var_type=default._var_type,
    _var_data=VarData.merge(
        match_cond_var._get_all_var_data(),
        *[el._get_all_var_data() for case in match_cases for el in case[0]],  # <-- conditions
        *[case[1]._get_all_var_data() for case in match_cases],
        default._get_all_var_data(),
    ),
)
```

But the **BaseComponent-return** branch does **not** propagate the subject/condition var-data to the rendered Match component — it only forwards the branch components as `children`:

```python
if isinstance(match_cases[0][1], BaseComponent):
    if default is None:
        default = Fragment.create()
    return Fragment.create(
        cls._create(
            cond=match_cond_var,
            match_cases=match_cases,
            default=default,
            children=[case[1] for case in match_cases] + [default],  # only branch components
        )
    )
```

So for component branches, the var-data (hooks/imports — including the `useContext` for any substate referenced in the `case[0]` conditions, and the subject) is never hoisted into the page `Component`. With a literal subject the subject contributes no var-data either, so nothing is emitted at all and the condition references are left unbound.

Likely related to the earlier #2393 ("Need to recursively enumerate VarData for all rx.match components"), which addressed return values; the **condition** var-data in the component branch still appears to be dropped.

## Expected behavior

The compiler should emit the `useContext` binding for any substate referenced in `rx.match` case conditions when the branches are components (e.g. include the condition vars in the Match component's `_get_vars()` / var-data), so the page renders instead of throwing a `ReferenceError`. Alternatively, reject a constant subject combined with state-var case conditions at compile time.

## Workaround

Replace the `rx.match` with chained `rx.cond` (which compiles to child components that each bind `useContext`):

```python
rx.cond(
    MatchState.a > 0,
    rx.text("a positive"),
    rx.cond(MatchState.b > 0, rx.text("b positive"), rx.text("neither")),
)
```

## Environment

- Reflex `0.9.5.post2` (`reflex-components-core` — `core/match.py`)
- Python 3.13
- Reproduces in a local `reflex run` and on Reflex Cloud deploys.

picklelo:
## Problem

`rx.match(...)` with **component** branches and a **state Var used in a case _condition_** raises a frontend `ReferenceError: Can't find variable: reflex___state____state__...` at render. The idiomatic "match on a literal subject" form triggers it:

```python
rx.match(True, (State.x > 0, comp_a), (State.y > 0, comp_b), default)
```

Reported in #6675 (with a compiled-output trace).

## Root cause

`Match` rendered its `switch` **inline into the page component**, and the per-case condition Vars live in `match_cases` (which is not a JS property), so they were never surfaced to the compiler. The page route emits state `useContext` bindings only through memoized child components, never inline — so the inlined switch referenced the substate context variable without ever binding it.

The `Var`-return branch of `_create_match_cond_var_or_component` already merges the case-condition var-data; the component-return branch did not, and `Match` had no `_get_vars` / memoization handling for the conditions.

## Fix

- `_memoization_mode = MemoizationMode(recursive=False)` — render `Match` inside a snapshot memo wrapper (mirrors `Foreach`, whose stateful render logic likewise belongs in the memo body), so the switch + conditions get an in-scope `useContext` binding instead of being inlined into the page component.
- Override `_get_vars` to also yield the per-case condition Vars, so their hooks (the `useContext` bindings) are emitted.
- `add_imports` now merges the case-condition var-data (mirrors the `Var`-return branch), so the `StateContexts` / `useContext` imports are present.

## Verification

Compiled a minimal repro (`rx.match(True, (S.a > 0, ...), (S.b > 0, ...), default)`) with the patch:

- The page route (`_index.jsx`) no longer references the substate var (previously the unbound reference).
- A `Match_comp_*.jsx` memo wrapper is now generated that declares `const ...state = useContext(StateContexts....)` and contains the `switch` referencing the now-bound var.
- Verified for both "state only in the conditions" and "state in conditions + branches (incl. a nested `rx.cond`)".
- `ruff check` and `ruff format --check` pass.

Fixes #6675.

### DIFF
--- a/packages/reflex-components-core/news/6675.bugfix.md
+++ b/packages/reflex-components-core/news/6675.bugfix.md
@@ -1,0 +1,1 @@
+Fix `rx.match` raising `ReferenceError: Can't find variable` at render when a state Var is used in a case *condition* with component branches. The match is now rendered inside a memo wrapper that binds the required state context, and the case-condition Vars are surfaced to the compiler so their `useContext` hooks/imports are emitted.
--- a/packages/reflex-components-core/src/reflex_components_core/core/match.py
+++ b/packages/reflex-components-core/src/reflex_components_core/core/match.py
@@ -1,6 +1,7 @@
 """rx.match."""
 
 import textwrap
+from collections.abc import Iterator
 from typing import Any, cast
 
 from reflex_base.components.component import BaseComponent, Component, field
@@ -310,13 +311,46 @@ class Match(Component):
         """
         return dict(self._render())
 
+    def _get_vars(
+        self, include_children: bool = False, ignore_ids: set[int] | None = None
+    ) -> Iterator[Var]:
+        """Walk all Vars used in this component, including the case conditions.
+
+        The per-case condition Vars live in ``match_cases``, which is not a
+        JavaScript property, so the base implementation does not surface them.
+        Yield them here so the hooks they require are emitted -- in particular
+        the ``useContext`` binding for a state Var referenced only in a case
+        condition. Without this, the compiled ``switch`` references the
+        substate context variable without ever binding it, raising
+        ``ReferenceError: Can't find variable`` at render time.
+
+        Args:
+            include_children: Whether to include Vars from children.
+            ignore_ids: The ids to ignore.
+
+        Yields:
+            Each Var referenced by the component, plus the case conditions.
+        """
+        yield from super()._get_vars(
+            include_children=include_children, ignore_ids=ignore_ids
+        )
+        for conditions, _ in self.match_cases:
+            yield from conditions
+
     def add_imports(self) -> ImportDict:
         """Add imports for the Match component.
 
         Returns:
             The import dict.
         """
-        var_data = VarData.merge(self.cond._get_all_var_data())
+        var_data = VarData.merge(
+            self.cond._get_all_var_data(),
+            *[
+                condition._get_all_var_data()
+                for conditions, _ in self.match_cases
+                for condition in conditions
+            ],
+        )
         return var_data.old_school_imports() if var_data else {}
 
 
--- a/tests/units/compiler/test_memoize_plugin.py
+++ b/tests/units/compiler/test_memoize_plugin.py
@@ -1087,6 +1087,40 @@ def test_match_stateful_condition_memoizes_whole_match_and_stateful_branch() -> 
     assert any("withprop" in tag.lower() for tag in wrapper_tags)
 
 
+def test_match_literal_subject_stateful_condition_is_memoized() -> None:
+    """Match with a literal subject and a state Var in a case condition is memoized.
+
+    The case-condition Vars (not just the subject) must count toward Match's
+    statefulness. Otherwise the compiled ``switch`` references the substate
+    context variable without a ``useContext`` binding, raising
+    ``ReferenceError: Can't find variable`` at render. Regression test.
+    """
+
+    def page() -> Component:
+        comp = rx.match(
+            True,
+            (
+                SpecialFormMemoState.value == "a",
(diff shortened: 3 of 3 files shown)


masenf:
This does seem to work, but now that memoization mode is `MemoizationMode(recursive=False)`, the compiler will never memoize the stateful values returned by the match. I think what we need is to update the statefulness detection for rx.match so if any conditions are stateful, then it gets memoized while still allowing its children to be separately memoized if they are themselves stateful.

harsh21234i:
The approach looks correct. Match was hiding its case-condition vars from the compiler, so _should_memoize() could  miss reactive state in conditions when the subject is a literal. Surfacing those vars via _get_vars() is the right layer because it keeps memoization and hook/import collection aligned with the existing component contract. The only improvement I’d suggest is factoring the repeated condition-var traversal into a helper so _get_vars() and add_imports() stay in sync.

PR #6691, opened 2026-06-30
support async context for mutable proxies

### LINKED ISSUE #6689: (Immutable)MutableProxy should support `__aenter__` protocol
If all i'm holding if an MutableProxy object, I should be able to directly `async with some_obj` to take the lock, refresh the object, and make it modifiable. This allows helper methods to work with background tasks/non-mutable state without also passing along a reference to the state.

harsh21234i:
Fixes #6689 
## Summary
  - add async context-manager support to MutableProxy and
  ImmutableMutableProxy
  - refresh the proxy from its bound state field after entering the state
  context
  - cover ImmutableMutableProxy mutation through `async with proxy`

  ## Testing
  - uv run pytest tests/units/
  test_state.py::test_immutable_mutable_proxy_async_context_manager tests/
  units/test_state.py::test_rebind_mutable_proxy tests/units/istate/
  test_proxy.py -q
  - uv run ruff check reflex/istate/proxy.py tests/units/test_state.py
  - uv run towncrier check --config pyproject.toml --dir . --compare-with
  upstream/main


reflex/istate/proxy.py
```diff
@@ -592,10 +655,11 @@ def __getattr__(self, __name: str) -> Any:
         if is_mutable_type(type(value)) and __name not in (
             "__wrapped__",
             "_self_state",
+            "_self_path",
```
masenf, line 746:
```suggestion
            "_self_path",
            "_self_actx_state",
```

for completeness

reflex/istate/proxy.py
```diff
@@ -513,11 +572,14 @@ def _is_called_from_dataclasses_internal() -> bool:
                 return True
         return False
 
-    def _wrap_recursive(self, value: Any) -> Any:
+    def _wrap_recursive(
+        self, value: Any, path: tuple[tuple[str, Any], ...] | None = None
```
masenf: since this function is already designed for recursive wrapping of the current proxy, it should just take `new_path_segment: tuple[str, Any] | None` and append it to `self._self_path` inside the function.

this simplifies the call sites and helps avoid potential future mistakes/bugs when this code is refactored.

reflex/istate/proxy.py
```diff
@@ -467,6 +476,56 @@ def __repr__(self) -> str:
...
+        )
+        raise RuntimeError(msg)
+
+    async def __aexit__(self, *exc_info: Any) -> None:
+        """Exit the async context manager protocol through the bound state.
+
+        Args:
+            exc_info: The exception info tuple.
+        """
+        context_state = self._self_actx_state or self._self_state
```
masenf: when `self._self_actx_state is None`, we should return early, because that indicates we have already exited the context manager

reflex/istate/proxy.py
```diff
@@ -467,6 +476,56 @@ def __repr__(self) -> str:
...
         return f"{type(self).__name__}({self.__wrapped__})"
 
+    async def __aenter__(self) -> Self:
+        """Enter the async context manager protocol through the bound state.
+
+        Returns:
+            This proxy refreshed from the current state field.
+        """
+        context_state = self._self_state
+        self._self_actx_state = context_state
```
masenf, line 527: if `self._self_actx_state is not None`, then we should raise an exception; nesting the context managers here can cause subtle issues so we should fail loud and fast as this is a programming error

reflex/istate/proxy.py
```diff
@@ -446,18 +446,27 @@ def __new__(cls, wrapped: Any, *args, **kwargs) -> MutableProxy:
             cls = cls.__dataclass_proxies__[wrapper_cls_name]
         return super().__new__(cls)  # pyright: ignore[reportArgumentType]
 
-    def __init__(self, wrapped: Any, state: BaseState, field_name: str):
+    def __init__(
+        self,
+        wrapped: Any,
+        state: BaseState,
+        field_name: str,
+        path: tuple[tuple[str, Any], ...] | None = None,
```
masenf: don't use a regular `str` for the `access_kind`. define `LiteralAccessKind = Literal["attr", "item", "iter"]` and use it instead, that way we get better static type checking.

reflex/istate/proxy.py
```diff
@@ -467,6 +476,56 @@ def __repr__(self) -> str:
...
+        self._self_actx_state = context_state
+        state = await context_state.__aenter__()
+        try:
+            refreshed_value = getattr(state, self._self_field_name)
+            for access_kind, access_key in self._self_path:
+                refreshed_value = (
+                    getattr(refreshed_value, access_key)
+                    if access_kind == "attr"
+                    else refreshed_value[access_key]
+                )
```
masenf: be defensive here. if the `access_kind != "attr"` it could literally be any other value, we can't assume it is `"item"` and should `_raise_refresh_error` instead of potentially getting a weird KeyError or IndexError that the user won't understand.

reflex/istate/proxy.py
masenf: greptile calls out a legitimate gap in the implementation here. i don't know of a _great_ way to solve it, but I hinted above about having an access_type of "iter", so I think you could pass that here as `new_path_segment=("iter", None)` that way we can catch it in `__aenter__` and tell the user that you it's not possible to `async with X` when `X` was obtained through iteration.

reflex/istate/proxy.py
masenf: If you wanted to take it a step further, you could replace the inner tuple type with something like

```
AccessSpec = tuple["attr", str] | tuple["item", int | str] | tuple["iter", NoneType]
```

Then reuse that same definition for `new_path_segment` in the `_wrap_recursive` signature.

Use the type system to our advantage to catch easy errors during development and provide good signal to agents working in the code base.

reflex/istate/proxy.py
```diff
@@ -467,6 +476,56 @@ def __repr__(self) -> str:
...
+            if isinstance(refreshed_value, MutableProxy):
+                super().__setattr__("__wrapped__", refreshed_value.__wrapped__)
+                self._self_state = refreshed_value._self_state
+                self._self_field_name = refreshed_value._self_field_name
```
masenf: instead of overwriting `_self_field_name`, add to the isinstance check: `and self._self_field_name == refreshed_value._self_field_name`, so if it's NOT equal to the then that's an error and we'll fall through and raise.

this kind of defensive style can catch subtle bugs.

harsh21234i:
Addressed the review feedback in d8f12f3a.

  Changes made:
  - Added typed `AccessSpec` path entries for `attr`, `item`, and `iter`.
  - Updated `_wrap_recursive` to append a single new path segment internally.
  - Added `_self_actx_state` to internal proxy attrs.
  - Rejected nested `async with proxy` usage explicitly.
  - Made proxy `__aexit__` idempotent when already exited.
  - Made async refresh path replay defensive and fail clearly for iteration-sourced proxies.
  - Added coverage for nested context rejection, idempotent exit, and iteration-derived proxy rejection.

  Checks passed:
  - `uv run pytest tests/units/test_state.py::test_immutable_mutable_proxy_async_context_manager tests/units/
  test_state.py::test_immutable_mutable_proxy_async_context_rejects_iter_proxy tests/units/test_state.py::test_rebind_mutable_proxy tests/units/istate/
  test_proxy.py -q`
  - `uv run ruff check reflex/istate/proxy.py tests/units/test_state.py`
  - `uv run ruff format --check reflex/istate/proxy.py tests/units/test_state.py`
  - `uv run pyright reflex/istate/proxy.py tests/units/test_state.py tests/units/istate/test_proxy.py`

FarhanAliRaza:
@codspeedbot fix this regression

harsh21234i:
hey can you go through this @masenf @FarhanAliRaza

reflex/istate/proxy.py
```diff
@@ -467,6 +483,73 @@ def __repr__(self) -> str:
...
+                        self._raise_refresh_error()
+            if (
+                isinstance(refreshed_value, MutableProxy)
+                and self._self_field_name == refreshed_value._self_field_name
+            ):
+                super().__setattr__("__wrapped__", refreshed_value.__wrapped__)
+                self._self_state = refreshed_value._self_state
+                self._self_path = refreshed_value._self_path
+            else:
+                self._raise_refresh_error()
```
FarhanAliRaza, line 560: **Confirmed bug (reproduced):** proxies returned by `dict.get()` / `setdefault()` silently rebind to the *parent* container here.

Those methods are in `__wrap_mutable_attrs__` and route through `_wrap_recursive_decorator`, which calls `self._wrap_recursive(wrapped(*args, **kwargs))` with **no** `new_path_segment` — so the child proxy inherits the parent's `_self_path`. This guard can't catch the mismatch because `_self_field_name` is the top-level field name for *every* descendant proxy:

```python
data_proxy = state_proxy.data          # dict[str, list[int]] = {"a": [1], "b": [2]}
child = data_proxy.get("a")            # wraps [1], but path=() like its parent
async with child:
    # child.__wrapped__ is now {"a": [1], "b": [2]} — the whole dict!
    child.append(3)                    # AttributeError
```

`child.append(...)` at least raises `AttributeError`, but with homogeneous nesting (`dict[str, dict]`) writes like `child["x"] = 5` land **silently in the wrong container** — data corruption with no error. Note `data_proxy["a"]` (via `__getitem__`) works correctly; only the two decorator-routed methods are affected.

Suggested fix: either pass `("item", args[0])` as the segment in `_wrap_recursive_decorator` (always correct for `setdefault`, correct for `get` whenever the value actually came from the dict), or — safer and minimal — tag decorator-returned proxies with the unresolvable sentinel (like the iter marker) so `__aenter__` raises the existing "Unable to refresh" error instead of silently rebinding. The sentinel handles `get(key, default)` returning the default, which can't be represented as a path at all.

reflex/istate/proxy.py
```diff
@@ -467,6 +483,73 @@ def __repr__(self) -> str:
...
 
+    async def __aenter__(self) -> Self:
+        """Enter the async context manager protocol through the bound state.
+
+        Returns:
+            This proxy refreshed from the current state field.
+        """
+        if self._self_actx_state is not None:
+            msg = (
+                "Mutable proxy is already mutable. Do not nest `async with proxy` "
```
FarhanAliRaza: Minor: if the same proxy object is shared between two asyncio tasks, task B hits this immediately with "Do not nest" — which is misleading for the cross-task case — whereas `async with state_proxy` in task B would have *blocked* on `_self_actx_lock` and proceeded. Uncommon in practice (each `StateProxy.__getattr__` access mints a fresh proxy), but worth either clarifying the message or keying the guard to the current task.

reflex/istate/proxy.py
```diff
@@ -467,6 +483,73 @@ def __repr__(self) -> str:
         """
         return f"{type(self).__name__}({self.__wrapped__})"
 
+    async def __aenter__(self) -> Self:
+        """Enter the async context manager protocol through the bound state.
+
+        Returns:
+            This proxy refreshed from the current state field.
```
FarhanAliRaza, line 506: Nit: this raises `RuntimeError` (nesting guard, refresh failure) but the docstring has no `Raises:` section.

reflex/istate/proxy.py
```diff
@@ -467,6 +483,73 @@ def __repr__(self) -> str:
...
+            msg = (
+                "Mutable proxy is already mutable. Do not nest `async with proxy` "
+                "blocks."
+            )
+            raise RuntimeError(msg)
+        context_state = self._self_state
+        self._self_actx_state = context_state
+        aenter_ok = False
+        try:
+            state = await context_state.__aenter__()
```
FarhanAliRaza, line 530: Worth noting: for a `MutableProxy` from a regular (non-background) handler, `_self_state` is a raw `BaseState`, whose `__aenter__` is a no-op returning `self` — so `async with proxy` succeeds but does nothing meaningful (no lock, no refresh). Functionally harmless since mutations persist through the normal handler flow, but it reads as if it locks something. Might deserve at least a comment here, or a docs note.

reflex/istate/proxy.py
```diff
@@ -532,10 +618,18 @@ def _wrap_recursive(self, value: Any) -> Any:
         # Recursively wrap mutable types.
         if is_mutable_type(type(value)):
             base_cls = globals()[self.__base_proxy__]
+            path = self._self_path
+            if new_path_segment is not None:
+                path = (
+                    _ROOT_ITER_PATH
+                    if new_path_segment is _ITER_ACCESS_SPEC and not path
+                    else (*path, new_path_segment)
```
FarhanAliRaza: Perf note: every nested mutable attribute/item access now allocates a path tuple (here plus the `("attr", name)` / `("item", key)` tuples at the call sites), paid by *all* event handlers even though only background-task `async with proxy` ever consumes the path. It's marginal next to the proxy allocation that already happens per access, but flagging since this is the state read hot path — fine if you've considered it acceptable.

reflex/istate/proxy.py
```diff
@@ -622,7 +722,7 @@ def __iter__(self) -> Any:
         """
         for value in super().__iter__():  # pyright: ignore[reportAttributeAccessIssue]
             # Recursively wrap mutable items retrieved through this proxy.
-            yield self._wrap_recursive(value)
+            yield self._wrap_recursive(value, _ITER_ACCESS_SPEC)
```
FarhanAliRaza: Not a blocker — the poison marker is a sound choice — but for lists specifically you could record `("item", i)` via `enumerate` instead, which would make iteration-sourced list proxies enterable (`for row in self.data: async with row: ...`). Could be a follow-up.

FarhanAliRaza (changes requested):
Thanks — the overall design here is solid: delegating to the bound `StateProxy`, replaying a recorded access path on the refreshed state, and poisoning iteration-sourced proxies so they fail loudly is the right shape, and the exception paths are carefully handled and tested. Tests, ruff, and pyright all pass locally.

Requesting changes for one confirmed correctness bug (reproduced locally): proxies returned by `dict.get()`/`setdefault()` carry their **parent's** access path, so `async with` on them silently rebinds the proxy to the parent container — in the worst case writes land in the wrong container with no error. Details inline, along with a few smaller comments.

harsh21234i:
hey @FarhanAliRaza  
Thanks for the detailed review. I addressed the confirmed `dict.get()` / `setdefault()` bug in `040ae098`.

  Changes made:
  - `dict.get(key)` now records an item path when the key exists, so returned mutable proxies refresh to the actual
  child value.
  - `dict.setdefault(key, value)` now records the inserted/existing item path.
  - `dict.get(missing_key, mutable_default)` now uses the unrefreshable marker, so `async with` fails with the existing
  `Unable to refresh mutable proxy` error instead of rebinding to the parent container.
  - Added a regression test covering the nested `dict[str, dict]` corruption case.
  - Added the missing `Raises:` docstring section and clarified the overlapping async context error message.

reflex/istate/proxy.py
masenf: i think we leave this one unhandled unless it comes up down the road as a separate feature request

reflex/istate/proxy.py
```diff
@@ -610,9 +720,13 @@ def __getitem__(self, key: Any) -> Any:
         """
         value = super().__getitem__(key)  # pyright: ignore[reportAttributeAccessIssue]
         if isinstance(key, slice) and isinstance(value, list):
-            return [self._wrap_recursive(item) for item in value]
+            indices = range(len(self.__wrapped__))[key]
+            return [
+                self._wrap_recursive(item, ("item", index))
+                for item, index in zip(value, indices, strict=False)
```
masenf: shouldn't we have `strict=True` here?

harsh21234i: by changing the slice zip to `strict=True` the sliced values and computed index range should
  always stay aligned, so this is the better assertion here.

harsh21234i:
any update @FarhanAliRaza @masenf ?

reflex/istate/proxy.py
```diff
@@ -467,6 +495,83 @@ def __repr__(self) -> str:
...
+        """
+        if self._self_actx_state is not None:
+            msg = (
+                "Mutable proxy is already in an async context. Do not reuse the "
+                "same proxy in overlapping `async with proxy` blocks."
+            )
+            raise RuntimeError(msg)
+        if _UNREFRESHABLE_ACCESS_SPEC in self._self_path:
+            self._raise_refresh_error()
+        context_state = self._self_state
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in 110767a4. `MutableProxy.__aenter__` now raises `ImmutableStateError` when the bound state is a `ReadOnlyStateProxy`, before entering it (previously entering set `_self_mutable=True` on the read-only proxy, letting `_mark_dirty` accept mutations). Regression test: `test_read_only_state_proxy_field_rejects_async_context`.

reflex/istate/proxy.py
```diff
@@ -467,6 +495,83 @@ def __repr__(self) -> str:
...
+                        case _:
+                            self._raise_refresh_error()
+            except (LookupError, AttributeError, TypeError):
+                # The container changed shape since this proxy was captured.
+                self._raise_refresh_error()
+            if (
+                isinstance(refreshed_value, MutableProxy)
+                and self._self_field_name == refreshed_value._self_field_name
+            ):
+                super().__setattr__("__wrapped__", refreshed_value.__wrapped__)
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in 110767a4. The refresh acceptance check now also requires the wrapped type to be unchanged when the current wrapped value is a dataclass (the proxy class is specialized per dataclass type via `__dataclass_proxies__`, so accepting a different type would leave stale `__dataclass_fields__` on the instance's class). A type-changed refresh now raises the existing "Unable to refresh mutable proxy" error; same-type replacement still refreshes normally. Regression test: `test_immutable_mutable_proxy_async_context_rejects_type_change`.

reflex/istate/proxy.py
```diff
@@ -555,7 +677,19 @@ def _wrap_recursive_decorator(
...
+        if args:
+            key = args[0]
+            if method_name == "get":
+                new_path_segment = (
+                    ("item", key)
+                    if key in self.__wrapped__
+                    else _UNREFRESHABLE_ACCESS_SPEC
+                )
+            elif method_name == "setdefault":
+                new_path_segment = ("item", key)
```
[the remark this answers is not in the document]
FarhanAliRaza: Not reproducible — the flagged `key in self.__wrapped__` line is unreachable for dataclass/pydantic objects with a custom `get` method. Pure-Python methods have `__func__`, so `__getattr__` returns `functools.partial(func, self)` (the method-rebind branch) *instead of* the `FunctionWrapper` that routes through `_wrap_recursive_decorator`. The decorator only ever fires for builtin `dict.get`/`dict.setdefault` (no `__func__`), where the wrapped object is always a dict supporting containment.

Verified empirically: `proxy.get("a")` on a proxied dataclass with a custom `get` runs the method fine, and the inner accesses in its body even produce a correctly path-tracked proxy (`(('attr', 'entries'), ('item', 'a'))`). Added `test_mutable_proxy_custom_get_method_path_tracking` in 110767a4 to lock this behavior in.

FarhanAliRaza (approved):

PR #6693, opened 2026-07-01
Cancel stale on_load event chains

### LINKED ISSUE #6593: Long running `on_load` events continue running after page navigation
**Describe the bug**
When a page has an on_load which loads a lot of data, but the user navigates to a different page while it's still loading, but new page's on_load is queued and blocked behind the previous event (whose results are going to be thrown away anyway)

harsh21234i:
Fixes #6593
 

## Summary
  - Track the active internal `on_load` event chain per client token.
  - Cancel the previous `on_load` chain when a newer page navigation starts for the same token.
  - Add regression coverage for stale page-load work being cancelled while the newer load completes.

  ## Testing
  - `uv run pytest tests/units/reflex_base/event/processor/test_event_processor.py -q`
  - `uv run --no-sync ruff check packages/reflex-base/src/reflex_base/event/processor/event_processor.py tests/units/
  reflex_base/event/processor/test_event_processor.py`
  - `uv run --no-sync ruff format --check packages/reflex-base/src/reflex_base/event/processor/event_processor.py tests/
  units/reflex_base/event/processor/test_event_processor.py`
  - `uv run --no-sync pyright packages/reflex-base/src/reflex_base/event/processor/event_processor.py tests/units/
  reflex_base/event/processor/test_event_processor.py`

### DIFF
--- a/news/6593.bugfix.md
+++ b/news/6593.bugfix.md
@@ -1,0 +1,1 @@
+Cancel stale page load events when a newer page navigation starts.
--- a/packages/reflex-base/src/reflex_base/event/processor/event_processor.py
+++ b/packages/reflex-base/src/reflex_base/event/processor/event_processor.py
@@ -20,6 +20,7 @@ from typing_extensions import Self
 from reflex.app_mixins.middleware import MiddlewareMixin
 from reflex.istate.manager import StateManager
 from reflex.utils import console
+from reflex_base import constants
 from reflex_base.event.context import EventContext
 from reflex_base.event.processor.compat import as_completed
 from reflex_base.event.processor.future import EventFuture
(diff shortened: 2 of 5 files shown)


masenf (commented):
i don't really like _where_ in the code this is implemented. EventProcessor is supposed to be a generic mechanism for queuing the events, and it's starting to grow legs with this change.

I do respect the effort though, and I looked around the code base and don't really see a better integration point that is currently exposed.

Let me think about this for a bit and see if we can maybe extend some other part of the API to provide a clean place for this kind of logic to run.

FarhanAliRaza:
Thank you for trying. 
Superceded by https://github.com/reflex-dev/reflex/pull/6713

PR #6708, opened 2026-07-06
Wrap Recharts Sankey chart

### LINKED ISSUE #6558: Wrap recharts SankeyChart
Attempted here https://github.com/reflex-dev/reflex/pull/5179

PR should fully implement SankeyChart component and update docs to include a basic example, a stateful example, and an example with custom node types and styles.

harsh21234i:
Closes reflex-dev/reflex#6558

  ## Summary
  - Add `rx.recharts.SankeyChart` / `rx.recharts.sankey_chart`
  - Export Sankey through the Recharts package API
  - Add docs with simple, stateful, and custom typed/styled node examples
  - Add unit coverage for responsive-container wrapping and render alias

  ## Tests
  - `uv run pytest tests/units/components/graphing/test_recharts.py -q`
  - `uv run ruff check ...`
  - `uv run ruff format --check ...`
  - `uv run pyright ...`


masenf (commented):
needs a changelog entry.

doesn't build

```
TypeError: Invalid var passed for prop SankeyChart.data, expected type <class 
'reflex_components_recharts.charts.SankeyData'>, got value 
reflex___state____state___docgen_exec___docs_library_graphing_charts_sankeychart_md____sankey_state
.data_rx_state_ of type collections.abc.Mapping[str, 
collections.abc.Sequence[collections.abc.Mapping[str, int]] | 
collections.abc.Sequence[collections.abc.Mapping[str, str]]].
While rendering demo block in docs/library/graphing/charts/sankeychart.md:
class SankeyState(rx.State):
    data = {
        "nodes": [
            {"name": "Marketing"},
            {"name": "Trial"},
            {"name": "Sales"},
            {"name": "Support"},
         
Happened while evaluating page 'library/graphing/charts/sankeychart'
```

harsh21234i:
Removed `link_width` because Recharts does not expose a top-level `Sankey` prop for link width. Link width is
  computed from link values and exposed only to custom link renderers.
 Widened `SankeyChart.data` to accept mapping-shaped state vars as well as the explicit `SankeyData` TypedDict, so
  the stateful docs example with unannotated state data works at component construction.
 Added regression coverage for both the removed dead prop and unannotated state data.

  The earlier `SankeyNode.name` and `SankeyLink` optional styling comments were already addressed in the current branch.

harsh21234i:
hey @masenf  @FarhanAliRaza  can someone look into this ?

harsh21234i:
Pushed `ad5de4fa` with two fixes:

- Added the missing `reflex-components-core` dependency for `reflex-components-recharts` and its pyi build hook dependencies, which fixes the package build failure caused by `reflex_base.components.memo` importing `reflex_components_core`.
- Updated the Sankey custom link payload types to match the Recharts runtime shape: `payload.source` / `payload.target` are resolved node payloads, link height uses `dy`, and `index` remains on the outer link props. Added a regression test for this shape.

harsh21234i:
Hey @masenf  The failing check is test_call_script[inline], where external4 is not loaded. This PR does not modify the call- script or integration harness code, and the external-script variant passes. Could this be rerun or compared against main to confirm whether it is a baseline/flaky CI failure?

harsh21234i:
hey @masenf  please take a look  when free :)

packages/reflex-base/src/reflex_base/vars/special.py
```diff
@@ -0,0 +1,62 @@
...
+        A Var representing the React hook.
+    """
+    var_name = get_unique_variable_name()
+    hook_alias = f"{hook}_{var_name}"
+    return Var(
+        var_name,
+        _var_type=cast(GenericType, _var_type),
+        _var_data=VarData(
+            imports={library: ImportVar(tag=hook, alias=hook_alias)},
+            hooks=(f"const {var_name} = {hook_alias}();",),
```
FarhanAliRaza, line 53: The hook lives in the Var's `VarData`, so every component that renders the var hoists its own `useId()` call. Sibling elements with Var props compile into separate memoized components, and each one gets a different id.

The `gradient_rect` example in `docs/api-reference/var_system.md` compiles to a `Lineargradient_...` component and a `Rect_...` component. Both declare `const udaxihhe = useId_udaxihhe();`. In the browser the gradients get ids `_r_0_` and `_r_2_`, the rect fills reference `_r_1_` and `_r_3_`, and the rects render invisible.

The same function wrapped in `@rx.memo` works, because the memo body owns the single hook call. The custom link renderer in the Sankey docs works for the same reason.

Requested: either hoist the hook at a common owner, or document that `use_id()` / `use_hook_var()` values may only be consumed inside one `@rx.memo` or renderer body, and change the `gradient_rect` example. Add a regression test where the var is referenced from two sibling components; `test_hook_var_hoisted_into_component` only covers one.

packages/reflex-components-recharts/src/reflex_components_recharts/charts.py
```diff
@@ -534,6 +538,238 @@ class FunnelChart(ChartBase):
...
+    node = staticmethod(sankey_node)
+    link = staticmethod(sankey_link)
+    __call__ = staticmethod(SankeyChart.create)
+
+    # For type checking
+    SankeyNode = SankeyNode
+    SankeyLink = SankeyLink
+    SankeyData = SankeyData
+    SankeyNodePayload = SankeyNodePayload
+    SankeyNodeProps = SankeyNodeProps
```
FarhanAliRaza: In the generated `charts.pyi` these assignments become self-referential type aliases. Pyright reports `Type alias "SankeyNodeProps" cannot use itself in its definition` for all seven names.

For a user file that copies the docs signature `node: rx.Var[rx.recharts.sankey_chart.SankeyNodeProps]`, pyright reports `Variable not allowed in type expression`, and `node.x` and `node.payload` resolve to `Var[Unknown]`.

Requested: bind the names through module-level aliases with distinct names, or expose the TypedDicts from the package module instead of the namespace, so the stub does not alias a name to itself.

packages/reflex-components-recharts/pyproject.toml
```diff
@@ -7,7 +7,10 @@ readme = "README.md"
 authors = [{ name = "Khaleel Al-Adhami", email = "khaleel@reflex.dev" }]
 maintainers = [{ name = "Khaleel Al-Adhami", email = "khaleel@reflex.dev" }]
 requires-python = ">=3.10"
-dependencies = ["reflex-base >= 0.9.0"]
+dependencies = [
+    "reflex-base >= 0.9.7",
```
FarhanAliRaza: `general.py` imports `reflex_base.vars.special` unconditionally. That module does not exist in any published reflex-base. Latest on PyPI is 0.9.10.post2, and `import reflex_base.vars.special` against 0.9.10 raises `ModuleNotFoundError`.

Requested: use a `*.dev` pin for the unreleased reflex-base, per the repo convention, so `scripts/check_min_deps.py` resolves it from the workspace. The check currently passes only because both the minimum and latest environments lack the module, so the error diff is empty.

packages/reflex-components-recharts/src/reflex_components_recharts/general.py
```diff
@@ -299,6 +303,21 @@ class Cell(Recharts):
...
+def use_chart_width() -> Var[int | None]:
+    """Get the chart width as a var.
+
+    Outside of a chart context, this will be None/undefined.
+
+    Returns:
+        The chart width var.
+    """
+    return use_hook_var(
+        library=Recharts.library or "",
```
FarhanAliRaza: `Recharts.library` is the literal `"recharts@3.8.1"`, so the `or ""` branch is dead. Requested: drop the fallback and fix the type of `library` if pyright needs it.

FarhanAliRaza (changes requested):
Tested in a real app on top of this branch, with `main` as the baseline for the unchanged chart types.

The app renders a literal-data Sankey chart with a tooltip, a state-bound chart with node and link config dicts and `sort=False`, a randomize event, an empty nodes/links toggle, custom node and link renderers on two chart instances, `rx.vars.use_id()` in a plain function and inside `@rx.memo`, `rx.recharts.use_chart_width()` outside a chart, and an area chart as a regression check. Results were asserted from the DOM and the compiled JSX, not from screenshots.

The Sankey wrapper works. The state-bound chart updates on events, empty data does not crash, custom renderers resolve their gradients, and `use_chart_width()` flips the label anchor at the right edge. The regression chart is unchanged.

Two faults block: `use_id()` produces different ids for sibling elements, so the documented gradient example renders nothing, and the generated stubs make the documented renderer signature a pyright error for users. The reflex-base lower bound also points at published versions that lack the new module.

Unit tests for the touched test files pass, ruff is clean, and pyright on the package source is clean apart from the stub aliases.

packages/reflex-base/src/reflex_base/vars/special.py
harsh21234i: Addressed. Hook vars are now documented as scoped to the component or renderer body that consumes them. The shared gradient example is wrapped in `@rx.memo` so the gradient definition and reference share one hook owner, and `test_hook_var_is_scoped_to_each_consuming_component` covers the sibling-consumer behavior.

packages/reflex-components-recharts/pyproject.toml
```diff
@@ -7,7 +7,10 @@ readme = "README.md"
 authors = [{ name = "Khaleel Al-Adhami", email = "khaleel@reflex.dev" }]
 maintainers = [{ name = "Khaleel Al-Adhami", email = "khaleel@reflex.dev" }]
 requires-python = ">=3.10"
-dependencies = ["reflex-base >= 0.9.0"]
+dependencies = [
+    "reflex-base >= 0.9.7.post31.dev0",
```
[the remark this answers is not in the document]
harsh21234i: This development pin is intentional and follows the repository release convention. The publish workflow runs `pin-lockstep` and then rewrites/checks development pins before package metadata is built; `reflex-otel` uses the same unreleased `reflex-base` floor pattern. No released artifact receives the `.dev` requirement.

harsh21234i:
Thanks for identifying this. The failure was caused by stale generated .pyi hashes in pyi_hashes.json, not by the implementation. I updated the charts.pyi and recharts.pyi hashes from the CI-generated values and pushed commit 561e4031. The branch is clean, and CI has been triggered again.

packages/reflex-components-recharts/src/reflex_components_recharts/charts.py
FarhanAliRaza: Fixed in c1939a2d9, kept at e4f46c3b8. The TypedDicts are exported from the package module, so the docs signature is `rx.Var[rx.recharts.SankeyNodeProps]` and pyright resolves the parameter as `Var[SankeyNodeProps]`. The self-referential aliases in the stub are gone. Attribute access such as `node.x` on a `Var[TypedDict]` is still untyped. That is a framework-wide gap in the var system, not this PR.

packages/reflex-components-recharts/pyproject.toml
FarhanAliRaza: Fixed in 5cfa5c2b3. The pin is now `reflex-base >= 0.9.7.post31.dev0`, so `check_min_deps.py` resolves it from the workspace and passes. The pin gets replaced at release, like the other dev pins.

packages/reflex-components-recharts/src/reflex_components_recharts/general.py
FarhanAliRaza: Fixed in 9c1bc20b1. `general.py` now passes the `_RECHARTS_LIBRARY` constant and the fallback branch is gone.

tests/units/components/graphing/test_recharts.py
```diff
@@ -50,3 +66,87 @@ def test_scatter_chart():
...
+
+def test_sankey_renderer_decorator_rejects_positional_only_parameter():
+    def custom_node(node: rx.Var[SankeyNodeProps], /) -> rx.Component:
+        return rx.fragment()
+
+    with pytest.raises(TypeError, match="keyword"):
+        sankey_chart.node(custom_node)
+
+
+def test_use_chart_width():
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid. Fixed in e4f46c3b8. The package-level exports are back under test. Regression test: `test_sankey_typed_dicts_exported_from_package`.

docs/wrapping-react/custom-code-and-hooks.md
```diff
@@ -114,3 +114,20 @@ export function Div_7178f430b7b371af8a12d8265d65ab9b() {
...
+
+
+def use_chart_width() -> rx.Var[int | None]:
+    """Get the width of the enclosing recharts chart as a var."""
+    return rx.vars.use_hook_var(
+        library="recharts@3.8.1", hook="useChartWidth", _var_type=int | None
+    )
+```
+
+For React's built-in [`useId`](https://react.dev/reference/react/useId), `rx.vars.use_id()` returns a `Var[str]` with a stable unique id for the component being rendered, e.g. for linking SVG elements to gradient or filter definitions.
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid. Fixed in e4f46c3b8. The page now states that a hook var is evaluated once per compiled component and names the scopes that satisfy this. The suggested warning text is out of date: since 5cfa5c2b3 an `rx.el.svg` root compiles with its children as one component, so sibling elements under it share one `use_id()` value. The test it cites was replaced by `test_svg_boundary_shares_hook_var_between_children`.

FarhanAliRaza (approved):

PR #6713, opened 2026-07-06
ENG-9661 fix: cancel stale on_load chains via handler-declared supersession

FarhanAliRaza:
Navigating away while a page's on_load chain is still running left the stale chain executing, blocking the new page's events behind it and applying its late deltas (#6593).

Add a SUPERSEDES_MARKER for event handlers with latest-wins semantics: enqueuing a new chain-root invocation cancels the previous unfinished event chain rooted at the same handler for the same client token. Mark on_load_internal with it so a newer navigation supersedes the previous page's unfinished load.

The EventProcessor tracks the active chain root per (event name, token), drops events chained from an already-cancelled parent before they enter the queue, and defers future cleanup while the handler task is still unwinding so late-chained events can find their cancelled parent.

### All Submissions:

- [ ] Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/reflex-dev/reflex/pulls) for the desired changed?

### Type of change

### New Feature Submission:

### Changes To Core Features:


packages/reflex-base/src/reflex_base/event/processor/event_processor.py
```diff
@@ -395,7 +403,14 @@ async def enqueue(
         tracked.add_done_callback(self._on_future_done)
         # If this context has a parent, register as a child of the parent's future.
         if parent_future is not None:
+            if parent_future.cancelled():
+                # The chain this event belongs to was cancelled; the event is
+                # stillborn and never enters the queue.
```
masenf:
```suggestion
                # The chain this event belongs to was cancelled, so cancel the
                # tracker since this event will never enter the queue.
```

lets get rid of the potentially trama-linked terminology. "stillborn" could be triggering for some

packages/reflex-base/src/reflex_base/event/processor/event_processor.py
```diff
@@ -395,7 +403,14 @@ async def enqueue(
...
+                tracked.cancel()
+                return tracked
             parent_future.add_child(tracked)
```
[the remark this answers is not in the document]
masenf: i think this is fine actually. if the parent_future reached a done state by some other means including exception, that doesn't affect the scheduling of chained events.

packages/reflex-base/src/reflex_base/event/processor/event_processor.py
```diff
@@ -625,10 +677,13 @@ def _dispatch_next_for_token(self, token: str) -> None:
...
-        # Skip cancelled futures.
+        # Skip cancelled futures. Before a task exists, the only way a future
+        # can be done (and thus already cleaned up) is cancellation, so a
+        # missing future also means the entry was cancelled.
         future = self._futures.get(entry.ctx.txid)
-        if future is not None and future.cancelled():
-            self._try_clean_future(future)
+        if future is None or future.cancelled():
+            if future is not None:
+                self._try_clean_future(future)
```
masenf: shouldn't this already hit since we add `_try_clean_future` as a done callback when we create the future?

packages/reflex-base/src/reflex_base/event/processor/event_processor.py
```diff
@@ -625,10 +677,13 @@ def _dispatch_next_for_token(self, token: str) -> None:
...
         entry, registered_handler = token_queue[0]
-        # Skip cancelled futures.
+        # Skip cancelled futures. Before a task exists, the only way a future
+        # can be done (and thus already cleaned up) is cancellation, so a
+        # missing future also means the entry was cancelled.
         future = self._futures.get(entry.ctx.txid)
-        if future is not None and future.cancelled():
-            self._try_clean_future(future)
+        if future is None or future.cancelled():
+            if future is not None:
```
masenf: why did this conditional change structure like this? are these two nested conditionals not equivalent to what we had before?

reflex/state.py
```diff
@@ -2482,6 +2484,15 @@ def on_load_internal(self) -> list[Event | EventSpec | event.EventCallback] | No
         ]
 
 
+# A newer navigation supersedes the previous unfinished on_load chain for the
+# same client token, cancelling its stale work (#6593).
+setattr(
+    OnLoadInternalState.event_handlers["on_load_internal"].fn,
+    SUPERSEDES_MARKER,
+    True,
+)
```
masenf: probably would be nice to expose this as a kwarg on `rx.event` like `background`

packages/reflex-base/src/reflex_base/event/processor/event_processor.py
FarhanAliRaza: Applied the suggested wording.

reflex/state.py
FarhanAliRaza: Done — added `supersedes` as a kwarg on `rx.event` (mirroring `background`), and `on_load_internal` is now decorated with `@event(supersedes=True)` instead of the post-class `setattr`.

packages/reflex-base/src/reflex_base/event/processor/event_processor.py
FarhanAliRaza: They're not equivalent: the old code treated a missing future as "proceed and dispatch". Pre-supersession that was unreachable in practice (nothing cancelled a queued root event), but now supersession cancels entries that are still sitting in the queue, and the cancel's `_try_clean_future` done callback removes the future from `_futures` before the entry is dequeued. So `future is None` has to mean "cancelled, skip" — otherwise a superseded entry whose cleanup callback already ran would execute anyway. The nested conditional is gone now (see next comment).

packages/reflex-base/src/reflex_base/event/processor/event_processor.py
FarhanAliRaza: You're right — entries examined here never had a task created, so cleanup is never deferred for them and the done callback always removes the future. Dropped the eager `_try_clean_future` calls in both dispatch paths.

benedikt-bartscher:
should we add a debug log when an event is cancelled?

tim-haselhoff:
I did some testing and if I understand it correctly this would not work with reflex background events which are executed by other reflex background events via returning.

packages/reflex-base/src/reflex_base/event/processor/event_processor.py
```diff
@@ -491,14 +506,51 @@ def _try_clean_future(self, future: EventFuture) -> None:  # type: ignore[overri
         """
         if not future.done():
             return
+        if future.txid in self._tasks:
+            # The handler task is still running or unwinding; keep the future
+            # so late-chained events can find their (possibly cancelled) parent.
+            return
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in d57a254bc. `_try_clean_future` now retains a done future only when it is *cancelled* and its task is still unwinding (`future.cancelled() and future.txid in self._tasks`); failed futures are released immediately, so a backend exception handler task reusing the txid can chain recovery events normally. Regression test: `test_exception_handler_can_chain_recovery_events`.

tests/units/reflex_base/event/processor/test_event_processor.py
```diff
@@ -735,3 +810,108 @@ async def _watcher():  # noqa: RUF029
...
+async def _drain_superseded(ep: EventProcessor) -> None:
+    """Give done callbacks a few ticks to clean the supersession tracking.
+
+    Args:
+        ep: The event processor to wait on.
+    """
+    for _ in range(20):
+        if not ep._superseded:
+            return
+        await asyncio.sleep(0)
```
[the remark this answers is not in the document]
FarhanAliRaza: Addressed in d57a254bc: the helper now raises `AssertionError` with the leftover dict contents instead of silently falling through, and the bound was raised to 100 ticks. The CI-load flake scenario doesn't apply, though: the loop awaits `asyncio.sleep(0)`, which yields event-loop *scheduler ticks*, not wall-clock time — done-callback cleanup completes in a fixed number of ticks regardless of machine load, so exceeding the bound would indicate a real cleanup bug, not a timing flake (this is now documented in the helper's docstring).

packages/reflex-base/src/reflex_base/event/processor/event_processor.py
```diff
@@ -491,14 +506,53 @@ def _try_clean_future(self, future: EventFuture) -> None:  # type: ignore[overri
...
+        try:
+            registered = RegistrationContext.get().event_handlers.get(event.name)
+        except LookupError:
+            return
+        if registered is None or not registered.handler.supersedes:
+            return
+        key = (event.name, token)
+        previous = self._superseded.get(key)
+        if previous is not None and not previous.all_done():
+            previous.cancel()
```
masenf, line 571: i agree with benedikt, we should debug log here at least; especially since this will be an invisible behavior change for users after upgrading. on_load never cancelled in flight before, so users might see new tracebacks if their async on_load is not properly shielded.

the other thing with just cancelling a future like this is if the event is a sync function and doesn't yield, we essentially have no way to stop it; we can mark it cancelled, but if we can't interrupt the event loop then there's nothing we can do.

masenf (commented):
for the comment on this approach not working for background tasks, i believe you are correct. if a background event is passed to `on_load`, then it _would_ be inhibiting from start if it hadn't already, but once it gets fired off, then it becomes its own task and wouldn't be cancelled along with the outer on_load_internal (by design i think). i need to test some more, but in theory it should be possible to mark the background event itself with `supersedes` and then you could avoid duplicate background tasks that way.

benedikt-bartscher:
@FarhanAliRaza do you plan finishing this PR?

FarhanAliRaza:
Yes I am gone finish it.

packages/reflex-base/src/reflex_base/event/processor/event_processor.py
FarhanAliRaza: Valid — fixed in 28d2125d5.

**Debug log:** `_supersede_previous` now emits `console.debug("Cancelling the previous unfinished <event.name> chain for token <token>, superseded by a newer invocation.")` right before `previous.cancel()`. It fires only when a chain is actually cancelled, not on the first enqueue. Regression test: `test_superseding_event_logs_debug_on_cancel`.

**Sync handlers that never yield:** agreed, and nothing can be done about that — it is the standard asyncio constraint. I documented the guarantee instead of pretending to one we cannot make. `EventHandler.supersedes` and the `supersedes=` arg on `@rx.event` now both state that cancellation is cooperative: a handler that never yields to the event loop runs to completion, and only its not-yet-started chained events are skipped.

news/6593.bugfix.md
```diff
@@ -0,0 +1 @@
+Cancel the previous unfinished `on_load` event chain when a newer page navigation arrives for the same client, instead of letting stale page-load work block and outlive the navigation.
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in 66585761a. The note now leads with the user-visible symptom (stale on_load work blocking/outliving navigation) and states the fix second.

packages/reflex-base/src/reflex_base/event/processor/event_processor.py
```diff
@@ -738,40 +798,63 @@ def _finish_task(self, task: asyncio.Task):
...
-                if future is not None and not future.done():
-                    future.set_result(result)
+                if sys.version_info < (3, 12):
+                    t._event_ctx = task_ctx  # pyright: ignore[reportAttributeAccessIssue]
+                t.add_done_callback(self._finish_task)
+                return True
+            logger.exception(
+                rich.markup.escape(
+                    f"Error in {task.get_name()} [txid={task_ctx.txid}]:"
+                )
```
masenf:
```suggestion
                f"Error in {task.get_name()} [txid={task_ctx.txid}]:",
```

i think we can get rid of the escape with the new logging system

FarhanAliRaza: Valid — fixed in 53acffd3d. Dropped `rich.markup.escape` from both log calls this PR added (the sink only parses markup for records with `extra={"rich": True}`). Left the pre-existing escape at the queue-entry error log untouched.

masenf (approved):

PR #6718, opened 2026-07-08
docs: add missing chained event handler

### LINKED ISSUE #6694: Issue with reflex.dev/events/chaining-events
The example given for chaining events has the definition of the set_progress() missing.
it should be:
`    @rx.event
    def set_progress(self, value:int):
        self.progress=value`

harsh21234i:
Closes reflex-dev/reflex#6694

  ## Summary
  - Add the missing `set_progress` event handler to the chaining events docs example.
  - Keep the example focused on calling one event handler from another.

  ## Tests
  - `git diff --check`

### DIFF
--- a/docs/events/chaining_events.md
+++ b/docs/events/chaining_events.md
@@ -17,6 +17,10 @@ class CallHandlerState(rx.State):
     progress: int = 0
 
     @rx.event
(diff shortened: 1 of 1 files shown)


masenf (approved):
thank you

PR #6722, opened 2026-07-08
some typing nits

benedikt-bartscher:


packages/reflex-base/src/reflex_base/plugins/shared_tailwind.py
```diff
@@ -108,7 +107,7 @@ def tailwind_config_js_template(
     imports = [
         plugin["import"]
         for plugin in plugins
-        if isinstance(plugin, Mapping) and "import" in plugin
+        if not isinstance(plugin, str) and "import" in plugin
```
masenf, line 110: i dont like this one because it's a runtime check. despite the static typing, if this `plugin` object is not a str, it could be anything else and indexing into it with a str would be invalid. hence, checking if it's explicitly a `Mapping` makes the `in` check and subsequent `__getitem__` call safe.

what was the motivation for this change? i might be misreading it.

PR #6725, opened 2026-07-09
ENG-9859 fix(state): preserve source function __dict__ and kwdefaults in _copy…

FarhanAliRaza:
…_fn (ENG-9859)

When the State metaclass pulls a computed-var getter or event handler in from a mixin, _copy_fn rebuilds the function via FunctionType, which drops the function's __dict__. Previously it hand-copied only the background-task and event-actions markers back on, silently losing any other attribute a framework or downstream user set on the getter/handler.

Copy the full __dict__ onto the copy, which subsumes the per-marker copies and lets arbitrary markers (e.g. Reflex Enterprise's _auth tag) survive mixin inheritance. Also carry __kwdefaults__, which FunctionType likewise drops — without it a mixin handler with a keyword-only default raised TypeError when called through the subclass.

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:

### DIFF
--- a/tests/units/test_state.py
+++ b/tests/units/test_state.py
@@ -4306,7 +4306,7 @@ def test_copy_fn_preserves_kwonly_defaults() -> None:
     """Test that _copy_fn preserves keyword-only default arguments."""
     handler_fn = UsesMarkerMixin.event_handlers["kwonly_default_handler"].fn
     assert handler_fn.__kwdefaults__ == {"count": 1}
-    instance = UsesMarkerMixin(_reflex_internal_init=True)  # pyright: ignore [reportCallIssue]
+    instance = UsesMarkerMixin()
     assert handler_fn(instance) == 1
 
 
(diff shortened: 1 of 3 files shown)


tests/units/test_state.py
```diff
@@ -4243,6 +4243,73 @@ def test_bare_mixin_state() -> None:
...
+    assert copied_handler_fn.__dict__ is not orig_handler_fn.__dict__
+    copied_handler_fn.__dict__["_leaked"] = True
+    assert "_leaked" not in orig_handler_fn.__dict__
+
+
+def test_copy_fn_preserves_kwonly_defaults() -> None:
+    """Test that _copy_fn preserves keyword-only default arguments."""
+    handler_fn = UsesMarkerMixin.event_handlers["kwonly_default_handler"].fn
+    assert handler_fn.__kwdefaults__ == {"count": 1}
+    instance = UsesMarkerMixin(_reflex_internal_init=True)  # pyright: ignore [reportCallIssue]
```
masenf: `_reflex_internal_init` shouldn't be needed inside a pytest case

masenf (commented):
looks good, needs a changelog entry

masenf (approved):

PR #6726, opened 2026-07-09
ENG-9860 fix(vars): preserve custom Field attributes in state metaclass

FarhanAliRaza:
BaseStateMeta rebuilt Field instances for state annotations, discarding any custom attributes callers had set. Copy over attrs the rebuilt field did not recompute via Field._copy_custom_attrs_from so subclassed or customized fields keep their extra state.

### All Submissions:

### Type of change

### New Feature Submission:

### Changes To Core Features:


masenf (commented):
i think this copying functionality should be part of `__init__`; you can add another optional param `source_field` and apply the copying of extra values from there.

what greptile says about deepcopying the values makes sense to me, we don't want to have shared references to mutable types across different Field instances.

having a list of reserved attributes also makes sense to avoid misidentification with pydantic fields.

add changelog entry

masenf (approved):

PR #6733, opened 2026-07-10
ENG-9917 fix: run plugin post_compile hooks at most once per app instance

FarhanAliRaza:
Fixes [ENG-9917](https://linear.app/reflex-dev/issue/ENG-9917) at the framework level.

## Problem

`App.__call__` runs every plugin's `post_compile(app=self)` each time the ASGI app is constructed. `post_compile` hooks mutate the app — `add_middleware`, mounting routes — so constructing the ASGI app twice on the same app instance (a test harness rebuilding it, repeated `app()` calls) re-applies those mutations. The concrete instance that surfaced this: enterprise `AuthPlugin` installing a second `AuthMiddleware`, gating every event twice. But the bug class belongs to the lifecycle, not to any one plugin — every app-mutating `post_compile` hook is affected.

## Fix

Latch the hook loop with a per-app-instance flag (`_plugins_post_compiled`), so `post_compile` runs at most once per app instance. The flag is set **after** the loop completes, so a hook that raises (e.g. a plugin rejecting a misconfigured app) is retried on the next construction instead of being silently skipped. The at-most-once guarantee is documented on `Plugin.post_compile` so plugin authors don't hand-roll idempotency guards.

Backend-only processes (`REFLEX_SKIP_COMPILE`) are unaffected: they construct a fresh `App` per process, and the flag starts false. Dev hot reload restarts the worker process (fresh `App`), so hooks still re-run there as before.

## Tests

- `test_call_app_runs_plugin_post_compile_once` — regression test, failed before the fix (hook ran twice).
- `test_call_app_retries_post_compile_after_failure` — pins the not-latched-on-failure semantics.

Full unit suite passes (6619, coverage 77.97%), ruff + pyright clean.


masenf (commented):
It's potentially problematic that we allow a recompile on the same instance, but don't re-run the hooks on the same instance.

If the second compile adds a page, and a plugin's post_compile does something with pages, then the newly added page will be compiled into the app, but won't be affected by the plugin.

Is there a legit use case where this hits? or does it just happen in testing?

My thought would be if we run `_compile`, then we should run `post_compile` hooks.

Maybe instead we just need to reset things on the app instance, like the pages/routes, middleware, lifespan tasks, etc to make it safe to re-`__call__` the app instance and essentially rebuild it. (I think properly speaking we _should_ be attaching these to the returned asgi and maybe not mutating the App at all, but that can come later)

PR #6735, opened 2026-07-10
Dedup auto-memoize passthrough definitions during compile

Alek99:
### All Submissions:

### Type of change

- [x] Performance improvement (non-breaking change, identical compile output)

### Changes To Core Features:

## What this does

The auto-memoize plugin (`MemoizeStatefulPlugin`) built a full memo definition for **every** stateful call site and only deduped by `(tag, source_module)` afterwards. In profiling of a 500-duplicated-row page this was ~56% of compile CPU: 1,001 definitions were built and then collapsed to 3 unique ones. Per call site it:

1. ran `_analyze_params` twice on a constant signature — with `from __future__ import annotations`, `get_type_hints` eval-compiles the stringized annotations each time (2,472 `builtins.compile` calls in the profile), and
2. evaluated the memo body twice (once for the tag preview, once inside `_create_component_definition`).

### Fix (behavior-preserving)

- **Hoist the params analysis to an import-time constant** (`_PASSTHROUGH_PARAMS`): the passthrough wrapper signature is always `(children: Var[Component]) -> Component`.
- **Reuse the evaluated preview** as the definition body (`_LazyBody.ready(_lift_rest_props(preview))`) instead of re-running the memo function.
- **Check `CompileContext.auto_memo_components` for the computed `(tag, source_module)` key first** (new optional `existing_definitions` parameter): repeated identical subtrees reuse the first definition instead of rebuilding an equivalent one per call site. The cache stays per-`CompileContext`, so definitions are still never shared across compiles (covered by `test_passthrough_memo_definitions_are_not_shared_globally`).

Compile output is unchanged: the surviving definition is structurally identical to the one that previously overwrote the registry entry last.

## Measured results (CodSpeed instrumentation, this PR vs `main` @ 9a5c4d3)

**Overall: +21.96% compile performance.**

| Benchmark | `BASE` | `HEAD` | Efficiency |
| --------- | ------ | ------ | ---------- |
| `test_compile_all_artifacts[_stateful_page]` | 27.2 ms | 21.7 ms | **+25.74%** |
| `test_compile_page[_stateful_page]` | 30.8 ms | 25.3 ms | **+21.94%** |
| `test_compile_page_full_context[_stateful_page]` | 35.8 ms | 30.3 ms | **+18.3%** |
| `test_compile_page_duplicated_rows` (new) | N/A | 828.1 ms | baseline for follow-ups |

## Verification

- New unit tests: cache-hit returns the cached definition object (and a different `source_module` still misses); the memo body is evaluated exactly once per definition build.
- New CodSpeed benchmark `test_compile_page_duplicated_rows` (200 identical stateful rows) stresses this path so follow-up compiler-perf PRs are measured on every run.

Part of a compiler-performance series; tracked in Linear as ENG-10101.


FarhanAliRaza (dismissed):

masenf:
supersede by #7123

PR #6738, opened 2026-07-10
Skip deep type-validation on state var hot paths

Alek99:
Linear: [ENG-10093](https://linear.app/reflex-dev/issue/ENG-10093/deep-type-validation-on-assignment-and-cached-reads-never-gates)

### Description

Two hot paths ran `_isinstance(value, type, nested=1)`, which walks every element of list/dict values, only to gate a diagnostic log that never changes behavior:

- `BaseState.__setattr__` validated the full container on every assignment.
- `ComputedVar.__get__` / `AsyncComputedVar.__get__` validated the return type on **every** access, including cache hits.

Changes:

- Computed var return types are now validated only when the value is actually recomputed (sync and async). Cache hits skip the check entirely.
- Element-wise validation depth is now `1` in dev and `0` in prod (new cached `_validation_depth()` helper in `reflex_base.utils.types`), so production only checks the outer type. Dev keeps the exact same diagnostics as before.

### Benchmarks (GitHub Actions runner, [run](https://github.com/reflex-dev/reflex/actions/runs/29118932603), 2 passes each)

| Case | main | this PR | speedup |
|---|---|---|---|
| read cached computed var returning 10k dicts (dev) | 13.81 / 14.05 ms | **0.0015 / 0.0015 ms** | ~9,200x |
| read cached computed var (prod) | 13.66 / 13.93 ms | **0.0013 / 0.0015 ms** | ~9,300x |
| assign 100k-int list (prod) | 90.89 / 90.96 ms | **0.020 / 0.024 ms** | ~4,000x |
| assign 100k-int list (dev) | 90.74 / 92.67 ms | 91.76 / 94.74 ms | unchanged (by design) |

cProfile (3 assigns + 20 cached reads, dev): main spends 3.06s in 10.5M calls dominated by `_isinstance` (510k calls); with this PR the cached-read side disappears (`_check_deprecated_return_type` no longer on the cache-hit path) and in prod the whole workload is 538 function calls / 0.003s.

### Type of change

### Changes To Core Features:

  - `tests/units/reflex_base/vars/test_base.py`: return type checked on recompute, not on cache hits (sync, async, and `cache=False`).
  - `tests/units/reflex_base/utils/test_types.py`: `_validation_depth()` is 0 in prod / 1 in dev.
  - `tests/units/test_state.py`: wrong-typed assignment still logs an error in dev.

Note: behavior-wise, a wrong-typed computed var now logs once per recompute instead of once per access, and prod no longer walks container elements for the log-only check. No exceptions or control flow depend on these checks.

https://claude.ai/code/session_01G8cXh3TUbNtbjm2ERqE62X

---
_Generated by [Claude Code](https://claude.ai/code/session_01G8cXh3TUbNtbjm2ERqE62X)_

### DIFF
--- a/packages/reflex-base/src/reflex_base/vars/base.py
+++ b/packages/reflex-base/src/reflex_base/vars/base.py
@@ -65,6 +65,7 @@ from reflex_base.utils.types import (
     GenericType,
     Self,
     _isinstance,
+    _validation_depth,
     get_origin,
     has_args,
     safe_issubclass,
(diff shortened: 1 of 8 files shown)


packages/reflex-base/src/reflex_base/vars/base.py
```diff
@@ -2502,23 +2503,28 @@ def __get__(self, instance: BaseState | None, owner: type):
...
-            if not hasattr(instance, self._cache_attr) or self.needs_update(instance):
-                # Set cache attr on state instance.
-                setattr(instance, self._cache_attr, self.fget(instance))
-                # Ensure the computed var gets serialized to redis.
-                instance._was_touched = True
-                # Set the last updated timestamp on the state instance.
-                setattr(instance, self._last_updated_attr, datetime.datetime.now())
-            value = getattr(instance, self._cache_attr)
+            self._check_deprecated_return_type(instance, value)
+            return value
```
masenf, line 2592: if we return early, how does the computed value get cached?

Alek99:
Folded into #6743, which now spans the full state hot-path set — this PR, #6740, and the dirty-propagation frontier all edit the same functions (`__setattr__`, `_get_attribute`, `ComputedVar.__get__`, `_mark_dirty*`), so they're easier to review as one final shape. Commits are preserved there per original PR.

Re: the open question on the early return — the value is cached by the `setattr(instance, self._cache_attr, ...)` on the line immediately before that return; the early return only skips re-reading and re-validating the just-computed value. Also answered in #6743's description.

PR #6740, opened 2026-07-10
Reduce MutableProxy per-element overhead and cache per-field proxies

Alek99:
Linear: [ENG-10094](https://linear.app/reflex-dev/issue/ENG-10094/mutableproxy-per-element-overhead-on-iteration-and-reads)

### Description

- `MutableProxy._wrap_recursive` walked 5 stack frames (`_is_called_from_dataclasses_internal`) for **every** element retrieved through a proxy, before the cheap `is_mutable_type` check that usually decides nothing needs wrapping. The checks are reordered so immutable elements (the common case: ints, strings in a proxied list) skip the frame walk entirely. The dataclasses-internal check still runs for mutable values, preserving `dataclasses.asdict`/`astuple` behavior.
- `BaseState._get_attribute` constructed a fresh `MutableProxy` on every read of a mutable state var. The proxy is now cached per instance and field name, invalidated by identity when the underlying value is reassigned. The cache is excluded from pickling (`__getstate__`), and `copy`/`deepcopy` already round-trip through `__getstate__`, so copies never share or carry proxies.

One deliberate nuance: when `_wrap_recursive` receives an already-proxied value *and* is called from dataclasses internals, it now returns the unwrapped value (previously the still-wrapped proxy), matching the documented intent of that path.

### Benchmarks (GitHub Actions runner, [run](https://github.com/reflex-dev/reflex/actions/runs/29119575765), 2 passes each)

| Case | main | this PR | speedup |
|---|---|---|---|
| iterate 1000-int list via proxy | 0.977 / 0.973 ms | **0.219 / 0.214 ms** | ~4.5x |
| index 1000 elements via proxy | 1.204 / 1.171 ms | **0.396 / 0.381 ms** | ~3.0x |
| read mutable var attribute | 2.82 / 2.87 us | **1.17 / 1.15 us** | ~2.4x |

cProfile (20 iterations + 2000 attr reads): `_is_called_from_dataclasses_internal` (20,000 calls, 0.028s cumulative) disappears from the profile; total function calls drop from 168k to 72k.

### Type of change

### Changes To Core Features:

  - `tests/units/istate/test_proxy.py`: proxy reused across reads and invalidated on reassignment; cache never serialized (pickle round-trip rebinds proxies to the restored state); iteration yields plain immutables while mutable elements stay wrapped.

https://claude.ai/code/session_01G8cXh3TUbNtbjm2ERqE62X

---
_Generated by [Claude Code](https://claude.ai/code/session_01G8cXh3TUbNtbjm2ERqE62X)_

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -1519,6 +1519,9 @@ class BaseState(EvenMoreBasicBaseState):
 
         if name in self.backend_vars:
             self._backend_vars.__setitem__(name, value)
+            if (cache := self.__dict__.get("_mutable_proxy_cache")) is not None:
+                # Drop the proxy wrapping the replaced value.
+                cache.pop(name, None)
             self.dirty_vars.add(name)
             self._mark_dirty()
             return
@@ -1551,6 +1554,10 @@ class BaseState(EvenMoreBasicBaseState):
 
         # Set the attribute.
         object.__setattr__(self, name, value)
+
+        if (cache := self.__dict__.get("_mutable_proxy_cache")) is not None:
+            # Drop the proxy wrapping the replaced value.
+            cache.pop(name, None)
 
         # Add the var to the dirty list.
         if name in self.base_vars:
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -437,6 +437,12 @@ class BaseState(EvenMoreBasicBaseState):
     # Whether the state has ever been touched since instantiation.
     _was_touched: bool = field(default=False, is_var=False)
 
+    # Per-field cache of the MutableProxy wrapping each mutable var, so
+    # repeated reads don't rebuild the proxy. Never pickled.
+    _mutable_proxy_cache: builtins.dict[str, MutableProxy] = field(
+        default_factory=builtins.dict, is_var=False
+    )
+
     # A special event handler for setting base vars.
     setvar: ClassVar[EventHandler]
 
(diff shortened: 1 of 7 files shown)


reflex/state.py
```diff
@@ -1478,7 +1478,17 @@ def _get_attribute(self, name: str) -> Any:
...
+            cache = super().__getattribute__("__dict__").get("_mutable_proxy_cache")
+            if cache is None:
+                cache = {}
+                object.__setattr__(self, "_mutable_proxy_cache", cache)
+            proxy = cache.get(name)
+            # isinstance also rejects entries degraded by deepcopy, which
+            # copies a MutableProxy as its unwrapped value.
+            if not isinstance(proxy, MutableProxy) or proxy.__wrapped__ is not value:
+                proxy = MutableProxy(wrapped=value, state=self, field_name=name)
+                cache[name] = proxy
```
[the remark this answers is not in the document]
masenf: looks like this did get fixed

reflex/state.py
```diff
@@ -1478,7 +1478,17 @@ def _get_attribute(self, name: str) -> Any:
             name in super().__getattribute__("base_vars") or name in backend_vars
         ):
             # track changes in mutable containers (list, dict, set, etc)
-            return MutableProxy(wrapped=value, state=self, field_name=name)
+            cache = super().__getattribute__("__dict__").get("_mutable_proxy_cache")
```
masenf: similar to the last PR, unclear on why these cache flags are being directly added into the `__dict__` and using `object.__setattr__` and the like instead of just defining them as reserved fields on the state.

if we define them as real fields on the state, then users will get an exception if they try to reuse the name, instead of internal tracking machinery just getting broken or corrupting user data.

masenf (changes requested):

reflex/state.py
Alek99: Good call — converted in d6ab47c: `_mutable_proxy_cache` is now a declared `field(default_factory=dict, is_var=False)` like `_backend_vars`/`_was_touched`, added to `RESERVED_BACKEND_VAR_NAMES` so a user var with that name can't silently collide. It stays out of pickles (`__getstate__` pop) and is recreated empty in `__setstate__`. I'll apply the same treatment to the propagation-frontier fields in #6743.

---

Alek99:
Folded into #6743, which now spans the full state hot-path set — this PR, #6738, and the dirty-propagation frontier all edit the same functions, so they're easier to review as one final shape. Commits are preserved there per original PR.

The feedback from this thread is already in those commits: `_mutable_proxy_cache` is a declared reserved state field (per the review), and the cache entry is dropped on reassignment so a replaced value can't be kept alive through a stale proxy.

PR #6743, opened 2026-07-10
perf: cut state hot-path overhead — deep-validation skip, proxy caching, dirty-propagation frontier

Alek99:
Consolidates the state hot-path perf work from #6738 (ENG-10093), #6740 (ENG-10094), and the original scope of this PR (ENG-10095) into one review unit. All three edit the same few functions — `BaseState.__setattr__`, `_get_attribute`, `_mark_dirty*`, `ComputedVar.__get__` — so reviewing the final shape of those functions once beats reviewing three interleaved rewrites. Commits are preserved per original PR for per-commit review. Stacked on #6735; #6744 continues on top of this branch.

Motivation: profiling the event loop on current main, a trivial event (`counter += 1` from a button click) spends ~30% of framework CPU in `_get_attribute` (129 attribute reads, 121 `isinstance` calls per event) and ~25% in dirty-tracking/delta bookkeeping. This PR targets both buckets.

## 1. Skip deep type-validation on state var hot paths (from #6738)

`__setattr__` and `ComputedVar.__get__` ran `_isinstance(value, field_type, nested=1)` — walking every element of assigned/returned containers — only to gate a `console.error` diagnostic. Computed-var return types are now validated only on actual recompute (cache hits skip entirely), and element-wise validation depth becomes 1 in dev / 0 in prod via a cached `_validation_depth()` helper.

- read cached computed var returning 10k dicts: 13.8 ms → 0.0015 ms
- assign 100k-int list (prod): 90.9 ms → 0.02 ms
- cProfile of the same scenario: 10.5M calls / 3.06 s → 538 calls / 0.003 s

## 2. Cut MutableProxy per-element overhead; cache per-field proxies (from #6740)

The recursive wrap path walked 5 stack frames (dataclasses-internal check) for every element read through a proxy; immutable elements now skip wrapping and the frame walk entirely. `_get_attribute` built a fresh `MutableProxy` on every read of a mutable var; proxies are now cached per instance+field, invalidated by identity on reassignment, and excluded from pickling.

- iterate a 1000-int list via proxy: 0.98 ms → 0.22 ms (~4.5x)
- index 1000 elements: ~3.0x
- read a mutable var attribute: 2.8 us → 1.2 us (~2.4x)
- proxy micro-suite: 168k → 72k function calls

## 3. Only propagate newly-dirty vars in `_mark_dirty_computed_vars` (original scope)

Every setattr/proxied mutation re-scanned all computed vars for interval expiry and re-walked the dependency closure of the entire accumulated `dirty_vars` set. This adds per-class precomputed `_interval_computed_vars` (the expiry scan is skipped when none exist) and a per-instance propagated frontier so only newly-dirty vars are walked; a process-wide generation counter bumped on recompute resets the frontier to preserve mid-cycle invalidation correctness.

- proxied list append (state with 2 cached computed vars): 16.7 us → 8.6 us (~1.9x)
- int var setattr: 21.0 us → 9.3 us (~2.2x)
- `_mark_dirty_computed_vars` cumulative over 2000 appends: 0.102 s → 0.019 s

## Review feedback from the folded threads, already incorporated

- Internal tracking state is declared as **reserved state fields** rather than raw `__dict__` entries, per masenf's request on both #6740 and #6743: `_mutable_proxy_cache` and the propagation-frontier fields are declared fields (excluded from pickles, safe against user-var name collisions).
- The per-field proxy cache entry is dropped/updated when a mutable var is reassigned, so a replaced value cannot be kept alive through a stale proxy (flagged and confirmed fixed on #6740).
- Answering the open question on #6738 ("if we return early, how does the computed value get cached?"): the value is cached by the `setattr(instance, self._cache_attr, ...)` on the line immediately before that return — the early return only skips re-reading and re-validating the value that was just computed and stored.

Benchmark numbers are from the original PRs, measured on GitHub Actions runners over two passes.

### DIFF
--- a/reflex/state.py
+++ b/reflex/state.py
@@ -55,6 +55,7 @@ from reflex_base.utils.exceptions import ImmutableStateError as ImmutableStateEr
 from reflex_base.utils.serializers import serializer
 from reflex_base.utils.types import _isinstance, _validation_depth
 from reflex_base.vars import Field, VarData, field
+from reflex_base.vars import base as reflex_base_vars_base
 from reflex_base.vars.base import (
     ComputedVar,
     DynamicRouteVar,
@@ -368,6 +369,7 @@ CLASS_VAR_NAMES = frozenset({
     "_always_dirty_computed_vars",
     "_always_dirty_substates",
     "_potentially_dirty_states",
+    "_interval_computed_vars",
 })
 
 
@@ -406,6 +408,11 @@ class BaseState(EvenMoreBasicBaseState):
 
     # Set of states which might need to be recomputed if vars in this state change.
     _potentially_dirty_states: ClassVar[set[str]] = set()
+
+    # Names of computed vars with an update interval, refreshed whenever
+    # computed_vars changes. Empty for most classes, which lets dirty
+    # propagation skip the per-mutation expiry scan.
+    _interval_computed_vars: ClassVar[tuple[str, ...]] = ()
 
     # The parent state.
     parent_state: BaseState | None = field(default=None, is_var=False)
@@ -723,8 +730,21 @@ class BaseState(EvenMoreBasicBaseState):
         # Initialize per-class var dependency tracking.
         cls._var_dependencies = {}
         cls._init_var_dependency_dicts()
+        cls._refresh_interval_computed_vars()
 
         all_base_state_classes[cls.get_full_name()] = None
+
+    @classmethod
+    def _refresh_interval_computed_vars(cls) -> None:
+        """Recompute the names of computed vars that have an update interval.
+
+        Must be called whenever cls.computed_vars is modified.
+        """
+        cls._interval_computed_vars = tuple(
+            name
+            for name, cvar in cls.computed_vars.items()
+            if cvar._update_interval is not None
+        )
 
     @classmethod
     def _add_event_handler(
@@ -830,6 +850,7 @@ class BaseState(EvenMoreBasicBaseState):
 
         setattr(cls, unique_var_name, computed_var_func_arg)
         cls.computed_vars[unique_var_name] = computed_var_func_arg
+        cls._refresh_interval_computed_vars()
         cls.vars[unique_var_name] = computed_var_func_arg
         cls._update_substate_inherited_vars({unique_var_name: computed_var_func_arg})
         cls._always_dirty_computed_vars.add(unique_var_name)
@@ -1403,6 +1424,7 @@ class BaseState(EvenMoreBasicBaseState):
 
         # Update tracking dicts.
         cls.computed_vars.update(dynamic_vars)
+        cls._refresh_interval_computed_vars()
         cls.vars.update(dynamic_vars)
         cls._update_substate_inherited_vars(dynamic_vars)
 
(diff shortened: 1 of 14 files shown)


reflex/state.py
```diff
@@ -1798,14 +1820,31 @@ async def get_var_value(self, var: Var[VAR_TYPE]) -> VAR_TYPE:
...
+        # propagated, so repeated mutations only process newly-dirty vars.
+        # Recomputing any cached var re-materializes a cache that a later
+        # mutation must invalidate again, so the frontier is only valid for
+        # the recompute generation it was built in.
+        instance_dict = self.__dict__
+        propagated = instance_dict.get("_propagated_dirty_vars")
+        current_gen = reflex_base_vars_base._computed_var_recompute_generation
+        if propagated is None:
+            propagated = set()
+            object.__setattr__(self, "_propagated_dirty_vars", propagated)
```
masenf: if we already have the `instance_dict`, why not just shove the value in there?

reflex/state.py
```diff
@@ -1952,6 +1993,8 @@ def _clean(self):
         # Clean this state.
         self.dirty_vars = set()
         self.dirty_substates = set()
+        # Discard the propagation frontier along with the dirty vars it tracked.
+        self.__dict__.pop("_propagated_dirty_vars", None)
```
masenf: why are we treating these new fields like this and not defining them as normal fields in a BaseState, like all the other fields

reflex/state.py
Alek99: Addressed in the propagation-frontier commits: the generation counter is written via the instance `__dict__` directly (56d4cf57, for StateProxy compatibility through wrapt), and the frontier fields are declared reserved state fields rather than ad-hoc `__dict__` entries (ad41b4d0). Same treatment as `_mutable_proxy_cache` in the folded #6740.

reflex/state.py
Alek99: Addressed in ad41b4d0: `_propagated_dirty_vars` / `_propagated_generation` are declared reserved state fields (excluded from pickles, name-collision-safe), matching the treatment requested for `_mutable_proxy_cache` on #6740.

reflex/state.py
```diff
@@ -1481,7 +1515,12 @@ def _get_attribute(self, name: str) -> Any:
             name in super().__getattribute__("base_vars") or name in backend_vars
         ):
             # track changes in mutable containers (list, dict, set, etc)
-            return MutableProxy(wrapped=value, state=self, field_name=name)
+            cache = super().__getattribute__("_mutable_proxy_cache")
+            proxy = cache.get(name)
+            if proxy is None or proxy.__wrapped__ is not value:
+                proxy = MutableProxy(wrapped=value, state=self, field_name=name)
+                cache[name] = proxy
+            return proxy
```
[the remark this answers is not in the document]
Alek99: The StateProxy path does run the eviction: `StateProxy.__setattr__` delegates through wrapt to the wrapped state's `BaseState.__setattr__`, which pops the cache entry on every reassignment (base and backend vars). Added a regression test proving reassignment through a StateProxy inside `async with self` evicts the cached proxy: 7d1e9dcd.

PR #6749, opened 2026-07-11
Add content hashes to asset URLs

### LINKED ISSUE #6550: rx.asset: add content-hash cache-busting to returned URL
Currently rx.asset() returns a plain path like /my-image.png. When you redeploy with an updated asset, browsers may serve the stale cached version since the URL hasn't changed.

The fix is straightforward: append a short SHA-256 hash of the file's contents as a query parameter (?v=<8-char-hash>), so the URL only changes when the file actually changes. This gives permanent cacheability without manual cache invalidation.

# current
rx.asset("logo.svg")  # → /logo.svg

# proposed
rx.asset("logo.svg")  # → /logo.svg?v=a1b2c3d4
We've been running this in production with a local wrapper and it works well.

harsh21234i:
Closes #6550.

  Adds content-hash cache busting to `rx.asset` URLs by appending an 8-character SHA-256 query parameter based on the
  asset file contents.

  Example:
  `rx.asset("logo.svg")` now returns `/logo.svg?v=<hash>`.

  The hash changes only when the file content changes, allowing browsers to cache assets safely across deploys.

  Implementation notes:
  - Applies to both local and shared assets.
  - Keeps `AssetPathStr.importable_path` unversioned so existing build-time import behavior is preserved.
  - Preserves backend-only behavior for missing local assets.
  - Adds regression coverage for hash changes and pickle/copy round trips.

### DIFF
--- a/reflex/assets.py
+++ b/reflex/assets.py
@@ -9,6 +9,9 @@ from reflex_base import constants
 from reflex_base.config import get_config
 from reflex_base.environment import EnvironmentVariables
 
+_HASH_CHUNK_SIZE = 1024 * 1024
+_MAX_HASH_ATTEMPTS = 3
+
 if TYPE_CHECKING:
     from typing_extensions import Buffer
 
@@ -123,7 +126,20 @@ def _short_content_hash(path: Path) -> str:
     Returns:
         The first 8 hex characters of the file's SHA-256 hash.
     """
-    return hashlib.sha256(path.read_bytes()).hexdigest()[:8]
+    digest = hashlib.sha256()
+    for _ in range(_MAX_HASH_ATTEMPTS):
+        digest = hashlib.sha256()
+        start_stat = path.stat()
+        with path.open("rb") as file:
+            while chunk := file.read(_HASH_CHUNK_SIZE):
+                digest.update(chunk)
+        end_stat = path.stat()
+        if (
+            start_stat.st_size == end_stat.st_size
+            and start_stat.st_mtime_ns == end_stat.st_mtime_ns
+        ):
+            break
+    return digest.hexdigest()[:8]
 
 
 def _versioned_asset_path(relative_path: str, source_file: Path) -> AssetPathStr:
--- a/reflex/assets.py
+++ b/reflex/assets.py
@@ -2,7 +2,6 @@
 
 import hashlib
 import inspect
-import os
 import time
 from pathlib import Path
 from typing import TYPE_CHECKING, overload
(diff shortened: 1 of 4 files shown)


reflex/assets.py
```diff
@@ -87,7 +103,57 @@ def __getnewargs__(self) -> tuple[str]:
...
+def _short_content_hash(path: Path) -> str:
+    """Get a short content hash for an asset file.
+
+    Args:
+        path: The file to hash.
+
+    Returns:
+        The first 8 hex characters of the file's SHA-256 hash.
+    """
+    digest = hashlib.sha256()
```
masenf: seems like a useless extra initialization; just keep the one in the loop

reflex/assets.py
```diff
@@ -87,7 +103,57 @@ def __getnewargs__(self) -> tuple[str]:
...
+        with path.open("rb") as file:
+            while chunk := file.read(_HASH_CHUNK_SIZE):
+                digest.update(chunk)
+        end_stat = path.stat()
+        if (
+            start_stat.st_size == end_stat.st_size
+            and start_stat.st_mtime_ns == end_stat.st_mtime_ns
+        ):
+            break
+    return digest.hexdigest()[:8]
```
masenf:
```suggestion
            return digest.hexdigest()[:8]
    console.warning(f"{path} was modified {_MAX_HASH_ATTEMPTS} times while calculating hash")
    return str(time.time())
```

we should consider doing something reasonable when we can't get a stable hash, like using the compile timestamp as the hash rather than always returning the last digest, which might cause the browser to load a previous cached version. at least if we take the timestamp, the browser will reload the asset once per compile, which is probably good enough for most uses.

reflex/assets.py
```diff
@@ -87,7 +103,57 @@ def __getnewargs__(self) -> tuple[str]:
...
+    """Create an asset URL with a content-hash query parameter.
+
+    Args:
+        relative_path: The unprefixed public asset path.
+        source_file: The source file used to calculate the content hash.
+
+    Returns:
+        A versioned asset URL that preserves an unversioned import path.
+    """
+    versioned_path = f"{relative_path}?v={_short_content_hash(source_file)}"
```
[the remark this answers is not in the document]
masenf: this is fine enough i think

harsh21234i:
hey @masenf  @Alek99  any update?

harsh21234i:

masenf (approved):

PR #6753, opened 2026-07-13
Sanitize buffered upload filenames

### LINKED ISSUE #6752: Buffered uploads preserve path segments in UploadFile.path while streamed uploads sanitize to basename
### Description

  Buffered and streamed uploads currently normalize uploaded filenames differently.

  The streamed upload parser sanitizes the multipart filename to a basename:

  ```python
  filename = Path(filename.lstrip("/")).name

  But the buffered upload path keeps the original path segments:

  path=Path(file.filename.lstrip("/")) if file.filename else None

  This means a filename such as ../secret.txt becomes:

  - streamed upload: secret.txt
  - buffered upload: ../secret.txt in UploadFile.path

  UploadFile.name still returns the basename, but UploadFile.path is exposed to user handlers, so app code can observe
  inconsistent behavior depending on whether the handler uses rx.upload_files or rx.upload_files_chunk.

  ### Reproduction

  import io
  from pathlib import Path

  from starlette.datastructures import UploadFile as StarletteUploadFile
  from reflex_components_core.core._upload import UploadFile

  file = StarletteUploadFile(file=io.BytesIO(b"x"), filename="../secret.txt")

  upload = UploadFile(
      file=file.file,
      path=Path(file.filename.lstrip("/")),
      size=file.size,
      headers=file.headers,
  )

  print(upload.path)
  print(upload.name)

  On Windows this prints:

  ..\secret.txt
  secret.txt

  The streamed parser path for the same filename emits:

  secret.txt

  ### Expected behavior

  Buffered uploads should apply the same filename normalization as streamed uploads, so UploadFile.path does not
  preserve directory or traversal segments from the client-provided multipart filename.

  ### Suggested fix

  Use the same basename normalization in the buffered upload path:

  path=Path(Path(file.filename.lstrip("/")).name) if file.filename else None

  or factor this into a shared helper so both upload modes always normalize filenames identically.

  ### Why this matters

  This avoids inconsistent behavior between rx.upload_files and rx.upload_files_chunk, and prevents user handlers from
  accidentally trusting path segments supplied by the client.


  I searched existing issues with upload/path/filename traversal terms and did not find an existing report for this
  specific buffered-vs-streamed inconsistency.

harsh21234i:
### Description
close #6752 
  Buffered uploads previously preserved client-supplied path segments in `UploadFile.path`, while streamed uploads
  normalized filenames to a basename.

  This made `rx.upload_files` and `rx.upload_files_chunk` expose different filename/path behavior for the same multipart
  filename, for example `../secret.txt`.

  This PR adds one shared filename sanitizer and uses it for both upload paths.

  ### Changes

  - Add `_sanitize_upload_filename` to strip path segments from upload filenames.
  - Add `_upload_file_from_starlette` so buffered upload file construction is directly testable.
  - Use the shared sanitizer in both buffered and streamed upload handling.
  - Add tests for POSIX-style and Windows-style path segment filenames.

  ### Checks

  - `uv run pytest tests/units/components/core/test_upload.py -q`
  - `uv run ruff check packages/reflex-components-core/src/reflex_components_core/core/_upload.py tests/units/
  components/core/test_upload.py`
  - `uv run ruff format --check packages/reflex-components-core/src/reflex_components_core/core/_upload.py tests/units/
  components/core/test_upload.py`
  - `uv run pyright packages/reflex-components-core/src/reflex_components_core/core/_upload.py tests/units/components/
  core/test_upload.py`


masenf (commented):
it seems like this breaks directory upload.

if you make an upload component, then drag a whole directory onto it and upload it, i would expect the path segments to be retained.

harsh21234i:
Good catch. I adjusted the sanitizer so directory uploads keep safe relative paths instead of flattening everything to
  a basename.

  Current behavior:
  - `photos/2026/image.png` is preserved for directory uploads.
  - `../secret.txt` is normalized to `secret.txt`.
  - absolute / drive paths still collapse to basename.

  I added coverage for both buffered uploads and streamed upload chunks to make sure safe directory paths are retained
  consistently.

harsh21234i:

masenf (approved):

PR #6771, opened 2026-07-15
fix(tokens): claim redis socket tokens atomically

### LINKED ISSUE #6879: RedisTokenManager can assign the same client token to two sockets under concurrency
## Description

`RedisTokenManager.link_token_to_sid` used a non-atomic check-then-set against Redis (`EXISTS` then `SET`). Under concurrent workers, two connections can both observe an absent token and then store different socket records for it, so two sockets share the same client token/state.

## Expected Behavior

Only one worker should successfully claim a given token. A losing worker should get a new token (or a safe local fallback if Redis is unavailable).

## Actual Behavior

Two workers can both link the same token to different SIDs when the claims race.

## Steps to Reproduce

1. Run multiple app workers backed by the same Redis instance.
2. Have two clients connect around the same time with a token that is not yet present in Redis.
3. Observe both workers treating the token as newly claimable.

## Proposed Fix

Claim tokens with an atomic Redis `SET` using `NX` (and the existing TTL). Retry with a new UUID if the claim loses; keep the local token-manager fallback on Redis errors.

Tracked in https://github.com/reflex-dev/reflex/pull/6771

anxkhn:
### All Submissions:

### Type of change

### Description

closes #6879

`RedisTokenManager.link_token_to_sid` previously checked whether a token existed before storing its socket record in a separate Redis command. Concurrent workers could both observe an absent token and then associate it with different sockets, allowing two connections to share the same client state.

This change claims each token with an atomic Redis `SET` using `NX` and the existing expiration. A worker that loses the claim retries with a new UUID (capped at 3 attempts), while Redis errors continue to use the local token-manager fallback.

Greptile feedback is addressed in the follow-up commit: retry cap + cleaned concurrency test that asserts the NX-wins-first-caller contract.

### Testing

- `uv run pytest tests/units/utils/test_token_manager.py -q` (29 passed, 4 skipped)
- `uv run ruff check .`
- `uv run ruff format . --check`

### Changes To Core Features:


Alek99:
Hi appreciate the contributions but can you prioritize Github issue also please address the AI review before marking it out of draft

anxkhn:
@alek99 opened https://github.com/reflex-dev/reflex/issues/6881 and converted this pr back to draft as requested.

PR #6776, opened 2026-07-15
fix(init): declare RadixThemesPlugin in default rxconfig template

### LINKED ISSUE #6483: DeprecationWarning: Implicit Radix Themes enablement has been deprecated in version 0.9.0.
**Describe the bug**
The default template (and other templates) use radix themes, but do not list the radix themes plugin in the rxconfig.py file, so this warning is raised.

**To Reproduce**
```
reflex init
reflex run
```

**Expected behavior**
A new app should not have deprecation warnings.

- Python Version: 3.10+
- Reflex Version: 0.9.1+

gaurav0107:
## Type of change

Bug fix (non-breaking change which fixes an issue).

## Summary

`reflex init` followed by `reflex run` emitted a spurious deprecation warning on a brand-new app:

> Implicit Radix Themes enablement has been deprecated in version 0.9.0.

The `blank` app template renders Radix Themes components (`rx.heading(size=...)`, `rx.text`, `rx.code`, `rx.button`), which auto-enable the Radix plugin during compilation. The generated `rxconfig.py`, however, only declared `SitemapPlugin` and `TailwindV4Plugin` — not `RadixThemesPlugin`. With no explicit Radix plugin in the config, the compiler falls back to `RadixThemesPlugin.create_implicit()` (`_explicit=False`), whose `enter_component` fires the deprecation warning the first time a Radix component compiles.

This declares `rx.plugins.RadixThemesPlugin()` in the config template that `reflex init` writes, which is exactly the remedy the deprecation message tells users to apply. The plugin is added **last** in the `plugins` list to preserve the existing plugin/stylesheet ordering the implicit path already used (the implicit plugin was appended after the others in the compiler chain), so scaffolded apps behave identically apart from the warning disappearing. The default `RadixThemesPlugin()` carries the same default blue theme the implicit path used.

A freshly scaffolded app is now warning-free.

closes #6483

## Testing

Extended the existing `test_create_config_e2e` in `tests/units/utils/test_utils.py`, which already execs the generated `rxconfig.py` into a real `rx.Config`. It now asserts the config declares a `RadixThemesPlugin`:

```python
assert any(isinstance(plugin, RadixThemesPlugin) for plugin in config.plugins)
```

This assertion fails on `main` (the template omits the plugin) and passes with the fix — a genuine fail-first regression test.

- `uv run pytest tests/units/utils/test_utils.py tests/units/test_config.py` — 320 passed
- `uv run pytest tests/units/compiler/test_plugins.py tests/units/compiler/test_memoize_plugin.py` — 142 passed
- `uv run ruff check .` (changed files) — clean
- `uv run ruff format --check .` (changed files) — clean
- `uv run pyright` (changed files) — 0 errors


FarhanAliRaza:
Tracked down the CI failures — the `test_appearance` ones are actually caused by this PR. `AppHarness` scaffolds its test apps with `reflex init`, so they now get the explicit `RadixThemesPlugin()` from the template, and with an explicit plugin `apply_app_theme` warns but then throws the deprecated `App(theme=...)` away:

```python
if self._explicit:
    return          # App(theme=...) silently ignored

self.enabled = True
self.theme = theme
```

Same thing hits real users: a fresh app doing `rx.App(theme=rx.theme(appearance="dark"))` — still what the theming docs teach — now compiles with the default blue theme instead. The fix is just deleting that early return so the deprecated param keeps working until its 1.0 removal. Pushed it with a fail-first regression test and the news fragments the `changelog` check wants: FarhanAliRaza/reflex@770b5db1e2aef4412dc8d02a47ceeed14e7ab14e — feel free to cherry-pick onto this branch.

The min-deps `check` jobs fail on `main` too (missing lower-bound bumps from other commits), so those aren't on this PR.

FarhanAliRaza (approved):

PR #6780, opened 2026-07-15
docs: database efficiency and table recipes

amsraman:
Fifth batch of upstreaming Reflex Build's AI-builder knowledge base. Database/table techniques, translated into the docs' own ORM idiom:

- **database/queries**: new "Writing Efficient Queries" section — aggregate in SQL not Python (`func.count`/`group_by`, conditional aggregation with `case`), avoid N+1 with `.in_()` (plus a raw-SQL `expanding=True` bindparam alternative), return only what the UI needs (column selection, LIMIT, load-once options, don't cache whole tables in per-user state); new "Handling NULL Values" section (`func.coalesce`, Python-side guards)
- **database/relationships**: "Querying Related Rows on Demand" — master-detail selection querying children by FK instead of eager-loading everything
- **tables-and-data-grids/table**: `overflow_x` container guidance, per-table pagination state (reset offset on filter change), debounced search inputs (`.debounce(500)`) with a per-keystroke-query warning

Flexgen-environment rules (raw-SQL-only mandates, schema conventions) were deliberately not upstreamed. Pre-commit clean.

### DIFF
--- a/docs/database/queries.md
+++ b/docs/database/queries.md
@@ -227,9 +227,12 @@ class OrderStats(rx.State):
             self.order_count = session.exec(select(func.count(Order.id))).one()
 
             # One row per group, ready for a chart or summary table.
-            self.orders_by_status = session.exec(
-                select(Order.status, func.count(Order.id)).group_by(Order.status)
-            ).all()
+            self.orders_by_status = [
+                tuple(row)
+                for row in session.exec(
+                    select(Order.status, func.count(Order.id)).group_by(Order.status)
+                ).all()
+            ]
 ```
 
 Related metrics can be computed in a single query with conditional
--- a/docs/database/queries.md
+++ b/docs/database/queries.md
@@ -293,7 +293,7 @@ import sqlalchemy
 
 with rx.session() as session:
     rows = session.execute(
-        sqlalchemy.text("SELECT * FROM user WHERE id IN :user_ids").bindparams(
+        sqlalchemy.text("SELECT * FROM users WHERE id IN :user_ids").bindparams(
             sqlalchemy.bindparam("user_ids", expanding=True)
         ),
         {"user_ids": [1, 2, 3]},
--- a/docs/database/queries.md
+++ b/docs/database/queries.md
@@ -190,6 +190,175 @@ class State(rx.State):
             return [list(row) for row in session.execute("SELECT * FROM user").all()]
 ```
 
+## Writing Efficient Queries
+
+The database is much better at filtering, joining, and aggregating than Python.
+Every row returned by a query is sent over the network and held in memory as
+part of the state — for every user with an open session. Do the work in the
+query itself and return only display-ready results.
+
+The examples below use a simple `Order` model.
+
+```python
+class Order(rx.Model, table=True):
+    user_id: int
+    status: str
+    amount: float | None = None
+```
+
+### Aggregate in SQL, not Python
+
+Avoid fetching a whole table just to compute a summary in Python. Use the
+database's aggregate functions (`COUNT`, `SUM`, `AVG`, `MIN`, `MAX`) with
+`GROUP BY` so only the summary rows come back.
+
+```python
+from sqlmodel import select, func
+
+
+class OrderStats(rx.State):
+    order_count: int = 0
+    orders_by_status: list[tuple[str, int]] = []
+
+    @rx.event
+    def load_stats(self):
+        with rx.session() as session:
+            # Count rows in the database instead of len() on a full fetch.
+            self.order_count = session.exec(select(func.count(Order.id))).one()
+
+            # One row per group, ready for a chart or summary table.
+            self.orders_by_status = session.exec(
+                select(Order.status, func.count(Order.id)).group_by(Order.status)
+            ).all()
+```
+
+Related metrics can be computed in a single query with conditional
+aggregation, instead of issuing one query per metric:
+
+```python
+import sqlalchemy
+from sqlmodel import select, func
+
+
+class OrderKPIs(rx.State):
+    total_orders: int = 0
+    completed_orders: int = 0
+    total_amount: float = 0.0
+
+    @rx.event
+    def load_kpis(self):
+        with rx.session() as session:
+            total, completed, amount = session.exec(
+                select(
+                    func.count(Order.id),
+                    func.sum(
+                        sqlalchemy.case((Order.status == "completed", 1), else_=0)
+                    ),
+                    func.coalesce(func.sum(Order.amount), 0.0),
+                )
+            ).one()
+        self.total_orders = total
+        self.completed_orders = completed or 0
+        self.total_amount = float(amount)
+```
+
+### Avoid Queries Inside Loops
+
+Issuing a query per item — the "N+1" pattern — multiplies network round trips.
+Fetch everything in one statement with a `JOIN`, an `IN (...)` list, or
+`GROUP BY`.
+
+```python
+# Inefficient: one query per user (N+1).
+with rx.session() as session:
+    for user in users:
+        orders = session.exec(Order.select().where(Order.user_id == user.id)).all()
+
+# Efficient: one query for all users at once.
+with rx.session() as session:
+    orders = session.exec(
+        Order.select().where(Order.user_id.in_([user.id for user in users]))
+    ).all()
+```
+
+The `.in_()` method binds each value as a separate query parameter, so it is
+safe to use with user-provided values. In raw SQL, use an *expanding* bind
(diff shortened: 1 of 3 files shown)


FarhanAliRaza (commented):
rx.Model is deprecated. do we want to add new docs using deprecated feature ? We can use new code example maybe.

docs/database/relationships.md
```diff
@@ -160,3 +160,42 @@ class Post(rx.Model, table=True):
...
+For master-detail interfaces — a table of parent rows where selecting a row
+reveals its related rows — it is not necessary to eager load every
+relationship up front. Store the selected id in state and query only the
+related rows for that selection, filtering on the foreign key.
+
+```python
+class PostFlagsState(rx.State):
+    posts: List[Post] = []
+    selected_post_id: Optional[int] = None
+    selected_flags: List[Flag] = []
```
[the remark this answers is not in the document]
amsraman: Good catch. Added `from typing import List, Optional` (and `import reflex as rx`) to the block so it's copy-paste safe in isolation, keeping the `List`/`Optional` style consistent with the rest of the page. Fixed in bc3e976.

docs/database/queries.md
```diff
@@ -190,6 +190,175 @@ class State(rx.State):
...
+    @rx.event
+    def load_stats(self):
+        with rx.session() as session:
+            # Count rows in the database instead of len() on a full fetch.
+            self.order_count = session.exec(select(func.count(Order.id))).one()
+
+            # One row per group, ready for a chart or summary table.
+            self.orders_by_status = session.exec(
+                select(Order.status, func.count(Order.id)).group_by(Order.status)
+            ).all()
```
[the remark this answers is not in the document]
amsraman: Fixed in 0c93fcc — each row is now wrapped with `tuple(row)` so the returned value matches the `list[tuple[str, int]]` annotation and avoids relying on coercion of SQLAlchemy `Row` objects.

amsraman:
Thanks @FarhanAliRaza — addressed this round.

**`rx.Model` deprecation (your comment):** You're right that it's deprecated — in 0.9.6 `rx.Model.__init_subclass__` emits a deprecation warning (`deprecation_version="0.9.2"`, `removal_version="1.0.0"`, reason "Directly use database ORM layer, like sqlalchemy or SQLModel"). But I'd rather not migrate just these two examples. Every database docs page uses `rx.Model` today: the canonical [Tables](/docs/database/tables) page literally instructs "make a class that inherits from `rx.Model`", `queries.md` opens with `class User(rx.Model, table=True)`, and the `Post`/`Flag` models these new examples build on are pre-existing `rx.Model` classes. Switching one section to bare `SQLModel` + explicit primary key (and `select(Order)` over `Order.select()`) would leave the page internally inconsistent — `User(rx.Model)` at the top, `Order(SQLModel)` halfway down — and confuse anyone reading top-to-bottom. I think the accurate fix is a separate, section-wide migration of all of `docs/database/*` off `rx.Model` in one pass, rather than piecemeal in this efficiency-notes PR. Happy to open that as a follow-up ticket if you agree.

While in here I also fixed two accuracy issues in the new content:

**"single indexed query" claim (relationships.md, 8b41c43):** `Flag.post_id` was only `foreign_key="post.id"`, which does **not** create an index, so the per-selection lookup was actually a full table scan — the "indexed" claim was false. Added `index=True` to `Flag.post_id` (verified it emits the `ix_flag_post_id` index) and noted that adding an index to an existing table requires a migration, so the claim now holds.

**Reserved keyword in raw SQL (queries.md, 2e6bb28):** the expanding `IN (...)` example used `SELECT * FROM user`, but `user` is a reserved word in PostgreSQL and fails unquoted. Renamed the table to `users` in that snippet so it runs as written.

docs/database/queries.md
```diff
@@ -190,6 +190,178 @@ class State(rx.State):
             return [list(row) for row in session.execute("SELECT * FROM user").all()]
 ```
 
+## Writing Efficient Queries
+
+The database is much better at filtering, joining, and aggregating than Python.
```
amsraman: "Databases are generally"

docs/database/queries.md
```diff
@@ -190,6 +190,178 @@ class State(rx.State):
...
+Fetch everything in one statement with a `JOIN`, an `IN (...)` list, or
+`GROUP BY`.
+
+```python
+# Inefficient: one query per user (N+1).
+with rx.session() as session:
+    for user in users:
+        orders = session.exec(Order.select().where(Order.user_id == user.id)).all()
+
+# Efficient: one query for all users at once.
```
amsraman, line 275: Efficient should go above inefficient

docs/database/queries.md
amsraman: Done — reworded to "Databases are generally much better at filtering, joining, and aggregating than Python."

docs/database/queries.md
amsraman: Done — reordered so the efficient (single-query) example is shown above the inefficient N+1 one.

docs/database/queries.md
```diff
@@ -190,6 +190,178 @@ class State(rx.State):
...
+
+Issuing a query per item — the "N+1" pattern — multiplies network round trips.
+Fetch everything in one statement with a `JOIN`, an `IN (...)` list, or
+`GROUP BY`.
+
+```python
+# Efficient: one query for all users at once.
+with rx.session() as session:
+    orders = session.exec(
+        Order.select().where(Order.user_id.in_([user.id for user in users]))
```
masenf:
```suggestion
        select(Order).where(Order.user_id.in_([user.id for user in users]))
```

masenf: do this throughout

docs/database/queries.md
```diff
@@ -190,6 +190,178 @@ class State(rx.State):
...
+        sqlalchemy.text("SELECT * FROM users WHERE id IN :user_ids").bindparams(
+            sqlalchemy.bindparam("user_ids", expanding=True)
+        ),
+        {"user_ids": [1, 2, 3]},
+    ).all()
+```
+
+If the related objects are linked with foreign keys, the
+[relationship loading techniques](/docs/database/relationships) can also fetch
+linked objects without extra queries.
```
masenf: put the raw sql pointer inside an `md alert info` block so it's not just rendered on the page but the agent will still see

docs/database/relationships.md
```diff
@@ -160,3 +160,49 @@ class Post(rx.Model, table=True):
...
+off a full table scan (adding the index to an existing table requires a
+migration). Avoid the inverse pattern of querying children in a Python loop
+over parents (the "N+1" pattern) — when the children of many parents are
+needed at once, use one query with `.in_()` on the foreign key, or one of the
+eager loading options above.
+
+```python
+post_ids = [post.id for post in self.posts]
+with rx.session() as session:
+    flags = session.exec(Flag.select().where(Flag.post_id.in_(post_ids))).all()
```
masenf: avoid `Model.select`

docs/library/tables-and-data-grids/table.md
```diff
@@ -1044,6 +1050,11 @@ def loading_data_table_example3():
     )
 ```
 
+```md alert info
+# Combining pagination with search and sorting
+Give each paginated table its own state vars (`offset`, `limit`, `total_items`) so multiple tables on a page paginate independently, and reset `offset` to `0` whenever the search or sort value changes — otherwise the user may be left on a page number beyond the end of the newly filtered results.
```
masenf, line 1055: recommend and link to ComponentState for creating a paginated table

docs/database/queries.md
amsraman: Done — replaced `Model.select()` with `select(Model)` throughout queries.md (User/Order) and relationships.md (Post/Flag).

docs/database/queries.md
amsraman: Done — moved the raw-SQL guidance (use an expanding bind parameter for `IN` lists; never string-format values) into an `md alert info` block so the agent picks it up. Let me know if you meant the relationship-loading pointer just below instead.

docs/database/relationships.md
amsraman: Done — switched to `select(Flag)`/`select(Post)` and added the `from sqlmodel import select` import to the example.

docs/library/tables-and-data-grids/table.md
amsraman: Done — added a recommendation to build reusable paginated tables with [ComponentState](/docs/state-structure/component-state/) right after the pagination alert.

masenf (approved):

PR #6781, opened 2026-07-15
docs: enterprise component field notes (ag-grid, map, react flow)

amsraman:
Sixth batch of upstreaming Reflex Build's AI-builder knowledge base. Enterprise component gaps:

**AG Grid**
- Two **new pages** for previously-undocumented features: **Tree Data** (`tree_data`/`get_data_path` FunctionVar pattern, conditional paths, aggregation) and **Master-Detail** (`detail_cell_renderer_params` with `successCallback`) — both registered in the sidebar
- `index.md`: selection-changed handler signature (list[dict], and the exact mismatch error a dict signature raises), floating filters, enterprise multi-column filter
- `cell-selection.md`: fill-handle modes/direction, per-column suppression, nonexistent-prop pitfalls
- `model-wrapper.md`: **bug fix** — docs said `class_model=`, the actual kwarg is `model_class=` (verified against reflex_enterprise); fixed-height requirement, `_is_authorized` example
- `value-transformers.md`: formatter patterns and cell renderers (incl. the `@rx.memo`-via-lambda requirement for interactive renderers)

**Map**: Leaflet icon-options dicts (no `rxe.map.icon` exists), map-API pitfalls (function not class, don't store in state, positional args), zoom debounce.

**React Flow**: controlled vs uncontrolled flows, the where-changes-get-applied rule (`apply_node_changes` in component wiring, never state code), provider takes no handlers, `screen_to_flow_position` usage.

Memories already covered by existing pages (pivot mode, basic usage, value getters, editing) were skipped; one memory's unverifiable event-payload claim was rejected. Pre-commit incl. pyright clean.

### DIFF
--- a/docs/enterprise/ag_grid/master-detail.md
+++ b/docs/enterprise/ag_grid/master-detail.md
@@ -93,20 +93,27 @@ def master_detail_grid():
 
 The detail grid is a full AG Grid instance with its own `column_defs`, independent from the master grid's columns.
 
-## Static vs State Configuration
+## Static vs Stateful Configuration
 
-Both the column definitions and the detail renderer params can be plain module-level objects or state vars, depending on whether they need to change at runtime:
+`row_data` and `column_defs` are plain serializable data, so they can live in state and change at runtime:
 
 ```python
-STATIC_DETAIL_PARAMS = {
+class MasterDetailState(rx.State):
+    master_data: list[dict] = []  # fetch/replace at runtime
+    column_defs: list[dict] = []
+```
+
+`detail_cell_renderer_params` is different because it holds a callback (`get_detail_row_data`). A callable cannot be stored in a state var: Reflex serializes state to the client as JSON, and a Python lambda (or `FunctionStringVar`) has no JSON representation, so syncing that state raises a serialization error at runtime.
+
+Keep the renderer params as a module-level object and pass it to the grid directly — it is compiled into the app once and does not need to change per request:
+
+```python
+DETAIL_PARAMS = {
     "detail_grid_options": {"column_defs": [{"field": "count"}]},
     "get_detail_row_data": lambda params: rx.vars.function.FunctionStringVar(
         "params.successCallback"
     ).call(params.data.counts),
 }
-
-
-class State(rx.State):
-    # Dynamic variant: update these vars to reconfigure the grid
-    detail_cell_renderer_params: dict = STATIC_DETAIL_PARAMS
 ```
+
+Reserve state vars for the serializable pieces (row data, column defs) and leave the callback-bearing renderer params at module level.
--- a/docs/enterprise/ag_grid/master-detail.md
+++ b/docs/enterprise/ag_grid/master-detail.md
@@ -1,0 +1,112 @@
+---
+meta_description: "Enable master-detail rows in AG Grid with Reflex. Expand rows to reveal a nested detail grid backed by per-row data, configured entirely in Python."
+title: Master Detail
+---
+
+# Master Detail
+
+Master-detail lets rows expand to show detailed information in a nested grid. Each master row carries its detail rows as a nested list, and an expandable column reveals them.
+
+Three pieces are required:
+
+1. `master_detail=True` on the grid.
+2. A column with `"cell_renderer": "agGroupCellRenderer"`, which renders the expand/collapse arrows.
+3. `detail_cell_renderer_params` describing the detail grid's columns and how to extract the detail rows from the master row.
+
+```python
+import reflex as rx
+import reflex_enterprise as rxe
+
+
+class MasterDetailState(rx.State):
+    master_data: list[dict] = [
+        {
+            "id": 1,
+            "name": "Product A",
+            "category": "Electronics",
+            "price": 299.99,
+            "counts": [  # Detail rows for this master row
+                {"count": 10, "value": "Stock Level"},
+                {"count": 5, "value": "Orders Today"},
+                {"count": 25, "value": "Total Sales"},
+            ],
+        },
+        {
+            "id": 2,
+            "name": "Product B",
+            "category": "Clothing",
+            "price": 49.99,
+            "counts": [
+                {"count": 50, "value": "Stock Level"},
+                {"count": 12, "value": "Orders Today"},
+                {"count": 78, "value": "Total Sales"},
+            ],
+        },
+    ]
+
+
+column_defs = [
+    {
+        "field": "id",
+        "header_name": "ID",
+        "width": 80,
+        "cell_renderer": "agGroupCellRenderer",  # Required for expand/collapse
+    },
+    {"field": "name", "header_name": "Product Name", "width": 150},
+    {"field": "category", "header_name": "Category", "width": 120},
+    {
+        "field": "price",
+        "header_name": "Price",
+        "width": 100,
+        "value_formatter": "params.value ? '$' + params.value.toFixed(2) : ''",
+    },
+]
+
+detail_cell_renderer_params = {
+    "detail_grid_options": {
+        "column_defs": [
+            {"field": "count", "header_name": "Count"},
+            {"field": "value", "header_name": "Description"},
+        ]
+    },
+    "get_detail_row_data": lambda params: rx.vars.function.FunctionStringVar(
+        "params.successCallback"
+    ).call(params.data.counts),
+}
+
+
+def master_detail_grid():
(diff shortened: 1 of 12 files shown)


docs/enterprise/ag_grid/master-detail.md
```diff
@@ -0,0 +1,112 @@
...
+    "detail_grid_options": {"column_defs": [{"field": "count"}]},
+    "get_detail_row_data": lambda params: rx.vars.function.FunctionStringVar(
+        "params.successCallback"
+    ).call(params.data.counts),
+}
+
+
+class State(rx.State):
+    # Dynamic variant: update these vars to reconfigure the grid
+    detail_cell_renderer_params: dict = STATIC_DETAIL_PARAMS
```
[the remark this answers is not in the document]
amsraman: Good catch — fixed. `detail_cell_renderer_params` is a compile-time component prop (`Var[... | Callable]`), so the module-level pattern works because the lambda is compiled into JS. But assigning a dict that contains the lambda to a state var breaks: state is serialized to the client as JSON and a Python callable has no JSON representation.

Reworked the "Static vs Stateful Configuration" section so it no longer presents the renderer params as a state var. Now: `row_data`/`column_defs` (serializable) can live in state, while the callback-bearing `detail_cell_renderer_params` stays a module-level object passed directly to the grid.

docs/enterprise/ag_grid/master-detail.md
```diff
@@ -0,0 +1,119 @@
...
+
+Master-detail lets rows expand to show detailed information in a nested grid. Each master row carries its detail rows as a nested list, and an expandable column reveals them.
+
+Three pieces are required:
+
+1. `master_detail=True` on the grid.
+2. A column with `"cell_renderer": "agGroupCellRenderer"`, which renders the expand/collapse arrows.
+3. `detail_cell_renderer_params` describing the detail grid's columns and how to extract the detail rows from the master row.
+
+```python
```
masenf: add `python demo exec` and lets see if this code actually works

docs/enterprise/ag_grid/tree-data.md
```diff
@@ -0,0 +1,157 @@
...
+
+- `auto_group_column_def` configures the tree column (the one with the expand/collapse arrows).
+- `group_default_expanded` controls the initial expansion: `0` for collapsed, `-1` for fully expanded, or a positive number for the depth to expand to.
+- `aggFunc` on value columns aggregates values at the group level (e.g. total size of a folder).
+
+## Custom Paths: `get_data_path`
+
+For anything beyond reading a single field, pass a JavaScript function as `get_data_path`. It must be built with `rx.vars.function.ArgsFunctionOperation` and cast to `rx.vars.FunctionVar`:
+
+```python
```
masenf: make all of the python examples that do not have external dependencies into `python demo exec` blocks so we can a) verify the code works and b) user can see what the code does

docs/enterprise/ag_grid/tree-data.md
```diff
@@ -0,0 +1,157 @@
...
+def tree_grid_custom():
+    return rxe.ag_grid(
+        id="tree_grid_custom",
+        row_data=TreeDataState.data,
+        tree_data=True,
+        get_data_path=rx.vars.function.ArgsFunctionOperation.create(
+            ["data"],
+            rx.Var("data.path"),
+        ).to(rx.vars.FunctionVar),
```
masenf: don't think this cast is necessary since ArgsFunctionOperation inherits from FunctionVar

docs/enterprise/ag_grid/value-transformers.md
```diff
@@ -94,4 +96,93 @@ def ag_grid_value_formatter():
...
+]
+```
+
+### Interactive Cell Renderers
+
+Renderers that use state or event handlers must be defined as an `@rx.memo` component. Pass it to `cell_renderer` through a lambda, with all arguments given by keyword — never pass the memoized function itself directly:
+
+```python
+@rx.memo
+def row_action_button(rowid: str) -> rx.Component:
```
masenf:
```suggestion
def row_action_button(rowid: rx.Var[str]) -> rx.Component:
```

new memo requires args to be Var-typed to avoid warnings

docs/enterprise/ag_grid/master-detail.md
amsraman: Done — converted to `python demo exec`. Verified the example builds against reflex_enterprise (the master-detail `detail_cell_renderer_params` lambda compiles fine into the grid prop).

docs/enterprise/ag_grid/tree-data.md
amsraman: Done — the three renderable examples (simple path, custom get_data_path, dynamic switching) are now `python demo exec` and verified to build. Left the trailing aggregate value-formatter block as plain `python` since it's a fragment (just a column_defs list, no component to render).

docs/enterprise/ag_grid/tree-data.md
amsraman: Done — removed the cast; verified it builds without it. (Kept the `.to(rx.vars.FunctionVar)` on the dynamic example's `rx.cond(...)` result, since that one isn't a direct ArgsFunctionOperation.)

docs/enterprise/ag_grid/value-transformers.md
amsraman: Done — `rowid: rx.Var[str]`.

docs/enterprise/ag_grid/tree-data.md
```diff
@@ -101,7 +101,7 @@ Returning the function from an `@rx.var` (or storing it in state) raises `Invali
 
 `get_data_path` can be selected at render time with `rx.cond`. Give the grid a `key` derived from the condition so it re-initializes when the path logic changes:
 
-```python
+```python demo exec
 class TreeDisplayState(rx.State):
     combine_hosts: bool = True
```
masenf, line 106: this var is never toggled, so this example feels incomplete

docs/enterprise/ag_grid/value-transformers.md
```diff
@@ -15,6 +15,8 @@ AgGrid allow you to apply transformers based on the column of your grid. This al
 TOC:
 - [Value Getter](#value-getter)
 - [Value Formatter](#value-formatter)
+- [Formatter Patterns](#formatter-patterns)
+- [Cell Renderer](#cell-renderer)
```
masenf: i'd say get rid of this whole section, we have this on the right sidebar

docs/enterprise/ag_grid/tree-data.md
amsraman: Done — added a switch bound to `combine_hosts` so the example is interactive: toggling it flips `get_data_path` (and the grid `key`), and the tree regroups between one combined tree and per-host trees. Verified it builds.

docs/enterprise/ag_grid/value-transformers.md
amsraman: Done — removed the manual TOC list; the on-page sidebar covers it.

docs/enterprise/ag_grid/value-transformers.md
```diff
@@ -94,4 +90,93 @@ def ag_grid_value_formatter():
...
+
+Python lambdas are useful for basic type conversions — the lambda receives a `params` var and operates on it symbolically:
+
+```python
+column_defs = [
+    {
+        "field": "number",
+        "value_formatter": lambda params: round(params.value.to(float), 2),
+    },
+    {"field": "status", "value_formatter": lambda params: params.value.to(str).title()},
```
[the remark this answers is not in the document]
amsraman: This is a false positive — the formatter here is a Reflex **Var lambda**, not a raw JS string, so `.title()` is `StringVar.title()` (reflex_base/vars/sequence.py:837), which compiles to valid JS:

```js
params.value.split(' ').map(word => word.charAt(0).toUpperCase() + word.slice(1).toLowerCase()).join(' ')
```

not `params.value.title()`. Same for `round(params.value.to(float), 2)` → `(+params.value.toFixed(2))`. I verified the column def builds in `rxe.ag_grid(...)` and the compiled output above. The raw-JS-string caveat only applies to the string formatters earlier in the section (e.g. `"params.value.toUpperCase()"`); the lambda form exists precisely so you get title-casing without hand-writing JS. Leaving the example as-is.

masenf (approved):

PR #6784, opened 2026-07-16
fix: make Moment a MemoizationLeaf

benedikt-bartscher:


masenf (approved):
good catch

masenf:
i was trying to figure out why i wasn't seeing this in my apps, and everywhere i was using moment was either in a foreach or explicitly memo'd outer component, both of which would mask the bug.

PR #6785, opened 2026-07-16
Extract docs shared componets into shared package

carlosabadia:


Alek99 (dismissed):

FarhanAliRaza (approved):
Approving, but three things worth a quick fix or follow-up:

1. **Stale font variables in `tailwind-theme.css` (lines 164–166).** It still defines `--font-instrument-sans: 'Instrument Sans'` and `--font-jetbrains: 'JetBrains Mono'` — families whose `@font-face`/woff2 files this PR deletes. The new `fonts.css` redefines both vars with the Fontsource names and currently wins only because it's imported one line later in the compiled root stylesheet. With `SharedSiteStylesPlugin(include_fonts=False)` (a shipped constructor flag) or any import reordering, every page silently falls back to the browser default sans — verified live: a canvas probe confirms `'Instrument Sans'` no longer resolves to a real font. Suggest deleting the stale definitions so `fonts.css` is the single source of truth, and documenting that `include_fonts=False` means "define these two vars yourself."

2. **`e.add_note()` breaks the declared Python 3.10 support** — `reflex_site_shared/docs/markdown.py:439` and `:474`. `add_note` is 3.11+, the package declares `requires-python = ">=3.10"`, and this PR's own 3.10 fixes (`StrEnum`, `datetime.UTC`) suggest 3.10 is a goal. On 3.10, any failing `python demo` block (631 of them across `docs/`) raises `AttributeError` and masks the real error plus its context note. Pyright doesn't catch it even at `--pythonversion 3.10`, so it needs the one-line `hasattr` guard.



Minor: `es-toolkit@1.46.1` is added to `frontend_packages`/lockfile but nothing imports it — if it's an `@inkeep/cxkit-react` peer dep, a comment saying so would stop someone from cleaning it up later.

tests/units/reflex_site_shared/docs/test_markdown.py
```diff
@@ -0,0 +1,280 @@
...
+    assert "Top Title" in rendered
+    assert "Subsection" in rendered
+    assert "Detail" in rendered
+    assert "inline_code" in rendered
+    assert "/docs/target/" in rendered
+    assert "first item" in rendered
+    assert "ordered two" in rendered
+    assert "A quoted insight." in rendered
+    assert "mode" in rendered
+    assert "plain code block" in rendered
```
masenf, line 79: these assertions and others in this file are not the best. just because the string appears in the rendered output doesn't mean it was rendered correctly. just sayin.

carlosabadia: right, ill remove / update in a follow up

PR #6786, opened 2026-07-16
make pydantic optional

benedikt-bartscher:


masenf (commented):
interesting; i thought we had pydantic as optional for some time now, but i guess not...

benedikt-bartscher:
> interesting; i thought we had pydantic as optional for some time now, but i guess not...

I remember reading some git commits or changelogs which mentioned this as well - however somehow it currently isn't optional.

I guess this is ready for review now.

FarhanAliRaza:
Validated this end to end in a real app: installed the branch's packages into a clean venv (confirmed nothing pulls pydantic transitively — 40 packages, no pydantic), ran a counter app, and exercised state events in the browser — compile, hydration, and event round-trips all work without pydantic. Also confirmed the `[pydantic]` extra resolves and re-activates model serialization / `MUTABLE_TYPES` / `ObjectVar` support.

One perf issue surfaced and is fixed in 02f71522c: `_transform_event_arg` called `find_spec("pydantic")` per dict-payload event argument. That was harmless while pydantic was a hard dep (`sys.modules` hit, ~125ns), but with pydantic absent every call did a full `sys.path` scan — measured 85.3µs per argument, and 0.64µs after hoisting the check to import time like the other guarded sites.

FarhanAliRaza (approved):

PR #6792, opened 2026-07-17
Process persisted package json before web sync

### LINKED ISSUE #6765: package.json is copied from reflex.lock/ to .web/ without processing
**Describe the bug**
If the user removes the scripts or other sections that reflex puts into package.json during `reflex init`, a subsequent hot reload can result in the dev server not being able to start.

**To Reproduce**
1. `reflex run`
2. `truncate -s 0 reflex.lock/package.json` then touch a python file -> trigger hot reload
3. get the lockfile out of sync warning
4. exit reflex, delete `reflex.lock`
5. `reflex run` -> dev server cannot start since package.json is missing the scripts section

**Expected behavior**
Reflex shouldn't blindly copy package.json without ensuring the necessary dev scripts are in place.

harsh21234i:
## Summary

  Fixes #6765.

  This updates the root lockfile sync path so `package.json` is processed before being mirrored into `.web/`.

  Previously, a stale or corrupted `reflex.lock/package.json` could be copied directly into `.web/package.json`, which
  could drop Reflex-managed scripts like `dev` and `export`.

  ## Changes

  - Add a dedicated package.json sync helper instead of treating it like a raw lockfile.
  - Re-render persisted package.json content through Reflex’s package JSON compiler.
  - Preserve the existing `.web/package.json` when no root persisted package.json exists.
  - Add regression coverage for both direct sync and frontend install repair behavior.

### DIFF
--- a/news/6765.bugfix.md
+++ b/news/6765.bugfix.md
@@ -1,0 +1,1 @@
+Process persisted package.json files before mirroring them into the web directory.
--- a/reflex/utils/frontend_skeleton.py
+++ b/reflex/utils/frontend_skeleton.py
@@ -368,6 +368,7 @@ def sync_root_package_json_to_web() -> bool:
         return False
 
     changed = output_path.exists()
+    path_ops.mkdir(output_path.parent)
     output_path.write_text(rendered)
     return changed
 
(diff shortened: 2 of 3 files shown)


harsh21234i:
hey any update @claude @masenf ??

masenf (approved):
thanks for the contribution!

PR #6804, opened 2026-07-21
improve compile perf

benedikt-bartscher:


masenf:
seems i haev some overlap here with #6947

benedikt-bartscher:
> seems i haev some overlap here with https://github.com/reflex-dev/reflex/pull/6947

yeah, maybe we can try to get best of both worlds?

masenf:
rode in with #6947

PR #6807, opened 2026-07-23
Fix disk state manager debounced writes

### LINKED ISSUE #6806: StateManagerDisk drops debounced set_state updates for non-BaseState tokens
## Problem

  `StateManagerDisk.set_state()` can lose updates when disk write debouncing is enabled.

  For non-`BaseState` tokens, `set_state()` queues only the first pending write for a token:

  ```python

  If the same token is written again before the debounce flush runs, the queued item is not updated. This causes two
  related problems:

  1. get_state() immediately after set_state() can return the default value instead of the value just written.
  2. Multiple writes before flush persist the first value, not the latest value.

  ## Reproduction

  import asyncio
  import tempfile
  from pathlib import Path

  async def main():
      with tempfile.TemporaryDirectory() as td:
          manager = StateManagerDisk(_write_debounce_seconds=60)
          manager.__dict__["states_directory"] = Path(td)

          print("after first set, get_state:", await manager.get_state(token))

          await manager.close()

          fresh = StateManagerDisk(_write_debounce_seconds=0)
          fresh.__dict__["states_directory"] = Path(td)
          print("after close/reload:", await fresh.get_state(token))
          await fresh.close()


  asyncio.run(main())

  ## Actual Behavior

  after first set, get_state: 0
  after close/reload: 1

  ## Expected Behavior

  after first set, get_state: 1
  after close/reload: 2

  ## Why This Matters

  StateManagerDisk should preserve normal state-manager semantics:

  - set_state() should update the current in-memory value.
  - The debounced disk write should persist the latest value for a token, not the first queued value.

  This is especially relevant for non-BaseState StateTokens and any code path that replaces a state object rather than
  mutating a cached BaseState instance in place.

  ## Suspected Cause

  Likely fix direction:

  - Update self.states[token.cache_key] = state inside set_state().
  - Replace or refresh the existing _write_queue[token] item when a newer state arrives before flush.
  - Add regression coverage for debounced non-BaseState tokens.

  ## Relevant Code

  - reflex/istate/manager/disk.py
  - StateManagerDisk.get_state()
  - StateManagerDisk.set_state()

harsh21234i:
Fixes #6806

## Summary

Fix debounced `StateManagerDisk.set_state()` writes so repeated updates for the same token persist the latest value instead of the first queued value.

## Problem

When disk write debouncing was enabled, only the first write for a token was retained in `_write_queue`. A later write before the debounce flush was silently discarded, so background processing or `close()` could persist stale state.

## Changes

- Replace an existing queued payload with the latest state while preserving the original queue timestamp.
- Keep the in-memory cache aligned when `set_state()` receives an arbitrary instance instead of the one returned by `get_state()`.
- Explicitly persist supplied `BaseState` instances, including untouched instances, so the cache never points to a state that was not written to disk.
- Preserve the existing touched-state optimization for the normal same-instance path.

## Testing

- `uv run pytest tests/units/test_state.py::test_state_manager_disk_debounced_set_state_flushes_latest_non_base_state -q`
- `uv run pytest tests/units/test_state.py::test_state_manager_disk_set_state_updates_cache_for_arbitrary_instance -q`
- `uv run pytest tests/units/test_state.py::test_state_manager_disk_set_state_persists_untouched_base_state -q`
- `uv run ruff check reflex/istate/manager/disk.py tests/units/test_state.py`
- `uv run ruff format --check reflex/istate/manager/disk.py tests/units/test_state.py`
- `git diff --check`


harsh21234i:
hey @masenf @FarhanAliRaza  can you go through this?

reflex/istate/manager/disk.py
```diff
@@ -332,17 +332,22 @@ async def set_state(
...
         if self._write_debounce_seconds > 0:
             # Deferred write to reduce disk IO overhead.
-            if token not in self._write_queue:
-                self._write_queue[token] = QueueItem(
-                    token=token,
-                    state=state,
-                    timestamp=time.time(),
-                )
+            if not is_base_state_token:
+                self.states[token.cache_key] = state
```
masenf: if we're actually going to fix this, it should be fixed for both supported token types, similar to `get_state`.

in most cases this is just going to be assigning the same value to the same key due to how the module works, but i could see an edge case where someone just calls `set_state` on maybe a synthetic instance that wasn't retrieved from `get_state`, or a plugin is trying to copy a state retrieved from one `get_state` call by posting it to another token in `set_state`... and then it doesn't actually set it in the cache, which would be unexpected.

reflex/istate/manager/disk.py
```diff
@@ -332,17 +332,22 @@ async def set_state(
...
+            queued_item = self._write_queue.get(token)
+            self._write_queue[token] = QueueItem[TOKEN_TYPE](
+                token=token,
+                state=state,
+                timestamp=queued_item.timestamp if queued_item else time.time(),
+            )
```
masenf: throughout this class, the states in self.states are assumed to be the mutable in memory primary store for state instances.

whenever the queue item gets picked out to be written, it is assumed to reference the current live and latest values for that state.

this patch now creates a new QueueItem for each set_state, which is a hot path. creating a frozen dataclass isn't a ton of overhead, but it does add up and i don't think it really buys us a lot. frozen states are not currently supported and i don't know that we will go that direction.

masenf (commented):
appreciate the contribution, but i'm inclined to close this one.   if it's rescoped just to persisting the state in `set_state`, we could probably take that.

harsh21234i:
Hi @masenf, I rescoped PR #6807 to persistence-only as requested.

  I removed the in-memory cache changes and related test, kept the latest queued-value disk persistence fix, and updated
  the PR description and news fragment. The focused regression test and Ruff checks pass.

masenf:
this leaves open the hole that i mentioned in my [first comment](https://github.com/reflex-dev/reflex/pull/6807#discussion_r3754590433)

i'm okay with updating the state reference in the QueueItem (rather than recreating), but both of the fixes are necessary for handling the "calling `set_state` with some arbitrary instance instead of the instance returned by `get_state`" case.

i'm not convinced how _important_ that case is to handle, but it's potentially a correctness issue, so I can see a case being made for supporting it -- but if we're going to support it, it needs to be consistent from both sides and not significantly impact performance

harsh21234i:
I’ve addressed both correctness cases: queued items are updated in place, and explicitly supplied BaseState instances are force-persisted when they differ from the cached instance. The immediate and debounced paths now keep persistence and cache behavior consistent. CI is passing; could you please take another look?

FarhanAliRaza:
Closing might reopen the git diff is messed up because of maybe some merge issue.

PR #6813, opened 2026-07-26
fix(format): recognize quote and backtick wrapping in is_wrapped

eeshsaxena:
`is_wrapped()` never recognizes quote- or backtick-wrapped text, so `wrap()` with the default `check_first=True` double-wraps it:

```python
from reflex.utils.format import wrap
wrap("`x`", "`")   # "``x``"  (expected "`x`")
wrap('"x"', '"')    # '""x""' (expected '"x"')
```

When `open == close` (the quote and backtick entries in `WRAP_MAP`), the opening delimiter matched both `if ch == open` and `if ch == close` on the same character, so `depth` dropped straight back to 0 and the function returned `False` for any such string. `wrap(text, open)` then never noticed the input was already wrapped and wrapped it again.

Identical delimiters can't nest, so this handles them directly: the text counts as wrapped when the delimiter doesn't reappear inside the content. Distinct-delimiter behavior is unchanged.

`test_is_wrapped` only exercised `{` and `(`, so the quote/backtick cases are added (including `` `a`b` `` which should stay `False`).

I couldn't run the full unit suite locally: its conftest imports `reflex_components_core`, and pydantic v1 doesn't load on my Python 3.14. I verified instead by importing `reflex_base.utils.format` directly and running the parametrized cases plus the distinct-delimiter regressions, which all pass, and confirming `wrap("`x`", "`")` is now idempotent.

### DIFF
--- a/news/6813.bugfix.md
+++ b/news/6813.bugfix.md
@@ -1,0 +1,1 @@
+Fixed `is_wrapped` to recognize quote- and backtick-wrapped text so `wrap` no longer double-wraps it.
--- a/packages/reflex-base/src/reflex_base/utils/format.py
+++ b/packages/reflex-base/src/reflex_base/utils/format.py
@@ -95,8 +95,13 @@ def is_wrapped(text: str, open: str, close: str | None = None) -> bool:
         Whether the text is wrapped.
     """
     close = get_close_char(open, close)
-    if not (text.startswith(open) and text.endswith(close)):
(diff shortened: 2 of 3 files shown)


eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

PR #6835, opened 2026-08-03
test: fix flaky memo test by waiting for hydration before interacting

FarhanAliRaza:
## Summary

`test_memo_event_handler_raw_pass_through` failed intermittently in CI:

```
AssertionError: Locator expected to have text 'typed_value'
Actual value:
```

The test filled `#memo-input` immediately after `page.goto()`, racing React hydration: if the `fill()` fired before the `on_change` handler was attached (and before the websocket connected), the input event was silently lost, so `#memo-last-value` stayed empty until the 5s expect timed out.

## Fix

Use the same pattern as `test_debounce_input.py`:
- Render `MemoState.router.session.client_token` in a read-only `#token` input — it only becomes non-empty once the client is hydrated and connected to the backend.
- Add a `_load_page()` helper (`goto` + wait for `#token` to be non-empty) and use it in all six tests of the module, since they all interact right after navigation and share the same latent race.

## Testing

- `uv run pytest tests/integration/tests_playwright/test_memo.py` — 6 passed
- ruff check/format clean

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/6835?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/integration/tests_playwright/test_memo.py
+++ b/tests/integration/tests_playwright/test_memo.py
@@ -89,6 +89,11 @@ def MemoApp():
 
     def index() -> rx.Component:
         return rx.vstack(
+            rx.input(
+                value=MemoState.router.session.client_token,
+                read_only=True,
+                id="token",
+            ),
             rx.text(MemoState.last_value, id="memo-last-value"),
             my_memoed_component(
                 some_value="memod_some_value", event=MemoState.set_last_value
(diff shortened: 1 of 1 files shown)


masenf (approved):
thanks for fixing, this race gets much worse in vite 8.2.0 for some reason, which is why i held it back when updating the deps

PR #6840, opened 2026-08-04
fix: replace queued write with latest state in StateManagerDisk

### LINKED ISSUE #6839: StateManagerDisk debounced writes can persist stale state after repeated set_state calls
### Description

`StateManagerDisk.set_state` only adds a token to `_write_queue` if that token is not already queued. When disk write debouncing is enabled, a second `set_state` call for the same token before the queued item is flushed does not replace the queued state.

That means the disk backend can persist an older state snapshot even though a newer `set_state` call completed successfully.

Relevant code path:

```py
        self._write_queue[token] = QueueItem(
            token=token,
            state=state,
            timestamp=time.time(),
        )
```

Because the existing `QueueItem` is left unchanged, `_flush_write_queue` / `_process_write_queue` later writes the first queued value, not the latest value.

### Reproduction

```py
import asyncio
import tempfile
from pathlib import Path

async def main():
    with tempfile.TemporaryDirectory() as tmp:
        manager = StateManagerDisk()
        manager.__dict__["states_directory"] = Path(tmp)
        manager._write_debounce_seconds = 60

        print("queued_state", manager._write_queue[token].state)
        await manager.close()

        reloaded = StateManagerDisk()
        reloaded.__dict__["states_directory"] = Path(tmp)
        print("persisted", await reloaded.load_state(token))
        await reloaded.close()

asyncio.run(main())
```

Current output:

```text
queued_state 1
persisted 1
```

### Expected behavior

The latest completed `set_state` call should win. The queued and persisted value should be `2`.

### Actual behavior

The first queued value is retained and persisted. Later `set_state` calls for the same token during the debounce window are ignored by the write queue.

### Impact

This is backend state persistence behavior specific to the disk state manager. It can cause stale state to be written to disk when callers use direct `set_state` with replacement state objects while debounced writes are pending.

The common `modify_state` path may hide this when the queued object is mutated in place, but direct replacement through `set_state` can lose the newer value.

### Possible fix

When a token is already queued, update the queued item to reference the latest state. Depending on intended debounce semantics, either preserve the original timestamp to flush at the first scheduled time or refresh the timestamp to debounce from the latest update. The key requirement is that the queued state object must be the latest completed `set_state` value.

### Duplicate check

I searched for existing open/closed issues and PRs using terms around `StateManagerDisk`, `_write_queue`, `debounce`, `set_state`, stale state, and lost disk persistence, and did not find an existing report for this behavior.

honma89:
### All Submissions:

### Type of change

### Changes To Core Features:

## What does this PR do and why?

`StateManagerDisk.set_state` only enqueued a token for a debounced write if that token wasn't already in `_write_queue`:

```python
```

If `set_state` was called again for the same token before the queued item flushed, the existing `QueueItem` was left untouched, so `_flush_write_queue` / `_process_write_queue` would later write the **first** queued value to disk instead of the **latest** one. Direct `set_state` calls with a replacement state object (as opposed to `modify_state`, which mutates the object in place) could silently lose data during the debounce window.

This PR changes `set_state` so that when a token is already queued, the queued `QueueItem` is replaced with one carrying the latest state via `dataclasses.replace` (since `QueueItem` is a frozen dataclass), while preserving the original `timestamp` so the flush still happens at its originally scheduled time — only the value that gets written changes.

A regression test (`test_state_manager_disk_set_state_updates_queued_write`) was added that mirrors the reproduction from the linked issue: it calls `set_state` twice for the same token with debouncing enabled, asserts the queued state reflects the latest call, then verifies the value persisted to disk after `close()` also reflects the latest call. Verified this test fails on the pre-fix code (`1 == 2`) and passes with the fix.

Also added a towncrier news fragment (`news/+disk-state-manager-stale-write-queue.bugfix.md`) per repo convention.

closes #6839

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/6840?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/reflex/istate/manager/disk.py
+++ b/reflex/istate/manager/disk.py
@@ -340,6 +340,15 @@ class StateManagerDisk(StateManager):
                     state=state,
                     timestamp=time.time(),
                 )
+            else:
+                # A write for this token is already queued; replace the queued
+                # state with the latest value so a stale snapshot is not
+                # flushed to disk. Preserve the original timestamp so the
+                # item still flushes at its originally scheduled time.
+                self._write_queue[token] = dataclasses.replace(
+                    self._write_queue[token],
+                    state=state,
+                )
         else:
             # Immediate write to disk.
             await self.set_state_for_substate(token, state)
--- a/tests/units/test_state.py
+++ b/tests/units/test_state.py
@@ -4441,7 +4441,7 @@ async def test_state_manager_disk_set_state_updates_queued_write(tmp_path, token
         token: A token.
     """
     state_manager = StateManagerDisk()
-    state_manager.__dict__["states_directory"] = tmp_path
+    object.__setattr__(state_manager, "states_directory", tmp_path)
     state_manager._write_debounce_seconds = 60
 
     state_token = StateToken(ident=token, cls=int)
@@ -4454,7 +4454,7 @@ async def test_state_manager_disk_set_state_updates_queued_write(tmp_path, token
     await state_manager.close()
 
     reloaded_state_manager = StateManagerDisk()
-    reloaded_state_manager.__dict__["states_directory"] = tmp_path
+    object.__setattr__(reloaded_state_manager, "states_directory", tmp_path)
     assert await reloaded_state_manager.load_state(state_token) == 2
     await reloaded_state_manager.close()
 
--- a/tests/units/test_state.py
+++ b/tests/units/test_state.py
@@ -46,7 +46,7 @@ from reflex.istate.manager import StateManager
 from reflex.istate.manager.disk import StateManagerDisk
 from reflex.istate.manager.memory import StateManagerMemory
 from reflex.istate.manager.redis import StateManagerRedis
-from reflex.istate.manager.token import BaseStateToken
+from reflex.istate.manager.token import BaseStateToken, StateToken
 from reflex.istate.proxy import StateProxy
 from reflex.state import (
     BaseState,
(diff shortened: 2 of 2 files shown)


harsh21234i:
The implementation looks correct to me. Replacing the queued `QueueItem` state while preserving the original timestamp
  fixes the stale flush without changing debounce timing.

  One CI issue remains: pyright flags the test because it assigns through `state_manager.__dict__`. Replacing those two
  assignments with `object.__setattr__(..., "states_directory", tmp_path)` fixes the type error locally.

  Also, the PR body mentions a news fragment, but I don’t see one in the diff. Could you add the fragment .

honma89:
Thanks for the thorough review, really appreciate it.

On the pyright issue you're right, reaching into __dict__ was the wrong move even though it happened to work at runtime (since states_directory is a cached_property and that's basically how it stores its value anyway). object.__setattr__ is more explicit about what's actually happening and it's a pattern pyright understands. Swapped both occurrences and reran the disk state manager tests locally — all still passing.

I did write it locally as part of the original change, but never actually git added it before generating the diff, so it just sat there untracked and invisible. Staged it properly now and confirmed with git status that it's showing up as expected.

PR #6853, opened 2026-08-06
[fix] Only check the redis session lock once per set_state fan-out

### LINKED ISSUE #6848: Redundant per-substate lock checks in `set_state`
I had Opus 5 do a deep dive analysis on some redis state lock performance issues we are having and it reported that this issue in the reflex framework might be a contributing factor

## Summary

`StateManagerRedis.set_state` re-validates the session Redis lock once per
substate in the in-memory tree, even when nothing is dirty and no key is
written. For a root-token persist over N state classes this is **2N** Redis
round trips (`GET` + `PTTL` of the same lock key) that can be done once.

## Evidence

Against a 35-class application state tree with opportunistic locking disabled
and nothing marked dirty:

| Token | Redis commands in `set_state` | Writes (`SET`) |
| --- | --- | --- |
| root (`rx.State`) | 70 (`35× GET` + `35× PTTL`) | 0 |
| typical leaf substate | 6 (`3× GET` + `3× PTTL`) | 0 |

The recursion in `reflex/istate/manager/redis.py` `set_state` is:

1. If `lock_id` is set: `await self.redis.get(self._lock_key(token))` and compare.
2. If `lock_id` is set and the key is not in `_local_leases`: `await self.redis.pttl(...)`.
3. `asyncio.create_task(self.set_state(...))` for **every** `base_state.substates` value.
4. Serialize/write only when `_get_was_touched()`.
5. `await` every child task.

Steps 1–2 therefore run once per visited substate, always against the same
`token.lock_key` (session ident). Payload size and pickle cost are negligible;
lock-held wall time scales with Redis RTT × substate count.

Production lock-timing logs showed the same signature:
`persist_ms` hundreds of ms with `redis_sets=0` / `persist_bytes=0` while a
neighboring cycle wrote ~5–7 KB in ~2 ms — Redis latency alone does not explain
exit-phase cost when the command fan-out is this large under event-loop load.

## Proposed fix (upstream)

Hold the lock check (and optional `PTTL` warning sample) **once** at the top of
a `set_state` fan-out, then recurse with a flag such as `lock_checked=True` /
`skip_lock_check=True` so children only persist touched nodes.

Alternatively, collect touched states in one walk and issue a single pipeline
of `SET`s after one lock verification.

## Why it matters

Reflex uses one Redis lock per client session (`StateToken.lock_key` ==
`ident`). Background handlers that open many short `async with state` cycles
(auth helpers, project load resets) each pay a full `set_state` walk. Cutting
the redundant checks removes O(substates) Redis RTT from every lock exit path.

qubydev:
### All Submissions:

### Type of change

### Changes To Core Features:

closes #6848

## Summary

`StateManagerRedis.set_state` was rechecking the session lock (`GET` + `PTTL`) once per substate while walking the in memory tree. That's the same lock key every time, so for a large state tree it was a lot of useless Redis round-trips, even when nothing was dirty.

This validates the lock once at the root of the fan-out and passes `lock_checked=True` into the recursive calls. Dirty substates still get written the same way as before.

## Test plan

- [x] Added `test_set_state_validates_lock_once_per_fanout`
- [x] Updated `test_state_manager_lock_warning_threshold_contend` (one warning per fan-out, not one per substate)
- [x] `uv run pytest tests/units --cov --no-cov-on-fail --cov-report=`
- [x] `uv run ruff check .`
- [x] `uv run ruff format .`
- [x] `uv run pyright reflex tests`

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/6853?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/reflex/istate/manager/redis.py
+++ b/reflex/istate/manager/redis.py
@@ -380,6 +380,7 @@ class StateManagerRedis(StateManager):
         state: TOKEN_TYPE,
         *,
         lock_id: bytes | None = None,
+        lock_checked: bool = False,
         **context: Unpack[StateModificationContext],
     ):
         """Set the state for a token.
@@ -388,6 +389,9 @@ class StateManagerRedis(StateManager):
             token: The token to set the state for.
             state: The state to set.
             lock_id: If provided, the lock must be held with this value to set the state.
+            lock_checked: If true, skip Redis lock revalidation for this fan-out. Parent
+                calls validate once, then recurse with this set so each substate does not
+                repeat GET/PTTL on the same session lock key.
             context: The event context.
 
         Raises:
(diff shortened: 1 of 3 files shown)


qubydev:
Would love to know your thoughts on this @harsh21234i

harsh21234i:
Hey @qubydev  Thanks for reaching out. The reduction in Redis round-trips makes sense, but the current lock_checked=True propagation appears unsafe because the  lock may expire or be reacquired by another worker during the fan-out. Recursive writes should retain an ownership check at the point of persistence, or use an equivalent atomic mechanism. Please add regression coverage for lock expiry and ownership replacement during fan-out. Once that is addressed, I’d be happy to re-review.

qubydev:
> Hey @qubydev Thanks for reaching out. The reduction in Redis round-trips makes sense, but the current lock_checked=True propagation appears unsafe because the lock may expire or be reacquired by another worker during the fan-out. Recursive writes should retain an ownership check at the point of persistence, or use an equivalent atomic mechanism. Please add regression coverage for lock expiry and ownership replacement during fan-out. Once that is addressed, I’d be happy to re-review.

Yeah that's a valid concern, I'll take a deeper look and apply changes. Thanks.

PR #6857, opened 2026-08-07
[vite 8.2.0] Fix REFLEX_ENV_MODE leak between AppHarness instances

masenf:
### Type of change

### Description

Fixes a critical bug where `AppHarnessProd` leaked `REFLEX_ENV_MODE=prod` into subsequent dev `AppHarness` instances in the same process. When `AppHarnessProd` runs `export()`, it sets `REFLEX_ENV_MODE=prod` process-wide and never restores it. A dev `AppHarness` compiling later would then write `prerender: true` into its dev react-router config, causing the dev server to serve prerendered page HTML. On pages with hydration hazards, this failed hydration on every load and dropped events dispatched during the hydration-recovery window.

**Root cause:** The environment variable leak was masked by a vite 8.0.16 pin that was added to work around memo component re-rendering failures in the full integration suite. Those failures were actually caused by this `REFLEX_ENV_MODE` leak, not by vite itself.

### Changes

1. **`reflex/testing.py`:** Pin `REFLEX_ENV_MODE` to `dev` in `AppHarness._initialize_app()`, mirroring what `reflex run` does. This ensures dev harnesses always compile in dev mode regardless of prior `AppHarnessProd` instances.

2. **`packages/reflex-base/src/reflex_base/constants/installer.py`:** Bump vite from 8.0.16 to 8.2.0 now that the root cause of the memo regression is fixed.

3. **Tests:** Added `test_app_harness_initialize_resets_leaked_prod_env_mode()` to verify that dev `AppHarness` correctly resets `REFLEX_ENV_MODE` to dev mode and disables route prerendering, even when the environment variable was previously set to prod.

4. **News fragments:** Added changelog entries documenting the bugfix and vite bump.

### Update (2026-08-07): react-router 8.3.0 merge + docker-test fixes

Merged main (react-router 8.3.0 upgrade, #6854) into this branch so CI exercises the shipping combination, and fixed the two `reflex-init-in-docker-test` failures that upgrade introduced on main:

5. **`packages/reflex-base/src/reflex_base/compiler/templates.py`:** Pin `preview.host` to `127.0.0.1` in the generated vite config. react-router 8 prerenders by fetching pages from a `vite preview` server started with the project's config; with no pinned host, the bound socket and the fetched `http://localhost:<port>/` URL can resolve to different loopback address families (as in docker containers), failing `reflex export` with `Prerender: Request failed for /: ECONNREFUSED`. Same fix TanStack Router shipped for the identical failure (TanStack/router#6305). Full diagnosis in [this comment](https://github.com/reflex-dev/reflex/pull/6857#issuecomment-5221869634).

6. **`reflex/utils/exec.py`:** Pre-enable the `development` export condition (`NODE_OPTIONS`/`BUN_OPTIONS`) when starting the dev server. react-router 8's dev CLI otherwise re-executes itself to enable the condition, which bun does not honor on installs without node, tripping the CLI's restart guard and exiting the dev server. Details in [this comment](https://github.com/reflex-dev/reflex/pull/6857#issuecomment-5222063518).

### Test Plan

- New regression test `test_app_harness_initialize_resets_leaked_prod_env_mode()` verifies the env-mode fix
- New unit tests assert the generated vite config pins the preview host, and that the development condition is merged into `NODE_OPTIONS`/`BUN_OPTIONS`
- Existing unit tests pass; memo playwright suites pass 10/10 on the merged react-router 8.3.0 + vite 8.2.0 tree
- Node-free verification (reflex-managed bun 1.3.14, mirroring the docker image): fresh `blank` app `reflex export` succeeds, and `reflex run --env dev` serves 200 with zero restart errors

https://claude.ai/code/session_01AzeRM83L6J367wmwk9Gezk

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/6857?utm_source=github" rel="nofollow noreferrer noopener" target="_blank">``&lt;img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"&gt;``</a>


masenf:
**The `reflex-install-and-init` failures on this PR are not caused by this diff — and the root cause is now fixed here.**

**What broke:** `reflex-init-in-docker-test` fails on every run whose checkout contains the react-router 8.3.0 upgrade (4116ff20 / #6854), starting with [its own push run on main at 16:51Z](https://github.com/reflex-dev/reflex/actions/runs/31199574220) — identical `Prerender: Request failed for /: ECONNREFUSED` at `vite.js:887`. PR runs check out the merge with current main, so this PR inherited it on both attempts, as did `claude/auth-overview-redirect-rsj733` (19:53Z, 20:06Z). Re-runs of workflow runs created *before* the upgrade merged still pass on their pre-upgrade merge commits (e.g. [fix/splat-catchall at 18:08Z](https://github.com/reflex-dev/reflex/actions/runs/31193935899)) — that's why the failure looked flaky. Note #6854's head ran no Actions workflows at all (only cubic + Greptile), so this path was first exercised post-merge.

**Mechanism:** react-router 8 prerenders through the Vite Environment API path: `startPreviewServer` boots a `vite preview` server with the project's vite config (inline options set only `port: 0`) and fetches each path from `resolvedUrls.local[0]`, i.e. `http://localhost:<port>/`. In the docker container `localhost` maps to both `127.0.0.1` and `::1`, and the bound socket vs. the prerender fetch can resolve to different address families under bun → connection refused. Dev machines with IPv4-only `localhost` never hit it, which is why #6854's local verification (and my sandbox) passed. react-router 7 was immune because its build doesn't use the HTTP prerender path (the passing logs have no "Using Vite Environment API"). TanStack Router hit the identical bug in their preview-server prerender — builds fine locally, `ECONNREFUSED` in docker — and fixed it by pinning the preview host: TanStack/router#6275 → TanStack/router#6305.

**Fix (d5cc49a6):** pin `preview: { host: "127.0.0.1" }` in the generated vite config. react-router's inline preview options don't set `host`, so the config-file value survives the merge; the server binds the literal IPv4 loopback and `resolvedUrls.local[0]` becomes `http://127.0.0.1:<port>/` — no hostname resolution left in the loop. Unit test added asserting the pin.

**Verification on this branch merged with main (react-router 8.3.0 + vite 8.2.0):** fresh `blank` app `reflex export` with node and bun stripped from PATH (reflex downloads its pinned bun 1.3.14, mirroring the docker image) succeeds — the preview server binds `127.0.0.1` and serves 200; the memo playwright suites pass 10/10 on the merged combination. My sandbox has no docker daemon and no IPv6 stack, so the in-container repro is validated by this PR's own CI run — watching it now.

Also merged main into this branch (417889f5) so CI exercises the real shipping combination. If you'd rather land the prerender fix on main independently of this PR, d5cc49a6 cherry-picks cleanly.

---

masenf:
**Second react-router 8.3.0 fallout found and fixed (84adcbf0).** The `reflex-install-and-init` run on d5cc49a6 got **past the prerender failure** — confirming the `preview.host` fix in the actual docker environment — and then crashed in the dev phase: react-router 8's dev CLI requires the `development` export condition (probed via its `#development-condition-enabled` import map) and re-executes itself with `NODE_OPTIONS=--conditions=development` to enable it. bun does not apply that option when it executes the CLI — and with no node in the container, bun is what runs it — so the restarted process probes false again and trips the CLI's restart guard: `restartWithMergedOptions() was called, but the process has already been restarted`. The dev server exited, and the workflow hung waiting on the frontend port until cancelled (that's the wedged 55-minute run — I cancelled it after it blew past the 30-minute job timeout). Environments with node installed never see this because the CLI's `#!/usr/bin/env node` shebang wins, which is why the 24-job integration matrix stayed green while the docker test died.

**Fix:** `run_frontend` now appends `--conditions=development` to `NODE_OPTIONS` **and** `BUN_OPTIONS` (merged with any existing value, idempotent) before spawning the dev server. Under bun, the CLI's restarted process picks `BUN_OPTIONS` back up from the environment and comes up with the condition enabled; under node, the condition is active up front and react-router's self-restart is skipped entirely. Prod is untouched — `run_frontend_prod` serves compiled static files with no react-router CLI, and vite's bundle resolution conditions are mode-driven.

**Verified node-free** (reflex-managed bun 1.3.14, mirroring the docker image): plain `bun run dev` reproduces the CI crash exactly; with the fix, `reflex run --env dev` on a fresh blank app serves 200 with zero restart errors. Unit tests added.

Heads-up for main: after cherry-picking the prerender fix, main will also need this one — the docker workflow runs `reflex run --env dev` right after the export. Both fixes are self-contained commits (d5cc49a6, 84adcbf0).

---

FarhanAliRaza (approved):

PR #6863, opened 2026-08-10
ENG-10963 feat(log): add stdlib logging pipeline in reflex_base.utils.log (1/5)

FarhanAliRaza:
Adds `reflex_base.utils.log`: a standard python `logging` pipeline with a rich-rendering console handler (legacy colors preserved), a JSON-lines handler behind `REFLEX_LOG_JSON`, record deduplication, and file logging.

- `LogLevel` gains a correct total ordering (the `str` mixin compared alphabetically, so `--loglevel critical` printed the system-info banner) and `to_logging_level()`.
- `console.set_log_level` delegates into the pipeline; `print`/`rule`/`status`/`progress` respect JSON mode; `print_table` added.
- `import reflex` bootstraps the reflex-owned loggers (sinks attach lazily on the first record).
- No call sites are migrated yet — that happens in PRs 2-4.

## Stack (ENG-10963)
this → #6864 → #6865 → #6866 → #6867.
Merge in order; each PR is based on the previous branch.


packages/reflex-base/src/reflex_base/utils/log.py
```diff
@@ -0,0 +1,679 @@
...
+        return msg
+
+
+class DedupeFilter(logging.Filter):
+    """Drop repeat records that opted into deduplication."""
+
+    def __init__(self):
+        """Initialize the filter with an empty seen-set."""
+        super().__init__()
+        self.seen: set = set()
```
masenf: probably just want to save the hash of the message, not the whole value. otherwise we leak all log messages that had dedupe set

packages/reflex-base/src/reflex_base/utils/log.py
```diff
@@ -0,0 +1,679 @@
...
+        try:
+            style, prefix = _style_for(record)
+            console = _console_stderr if record.levelno >= logging.ERROR else _console
+            # Records may carry a rich Progress to print through, so the
+            # message lands above an active progress bar.
+            progress = getattr(record, "progress", None)
+            if progress is not None:
+                console = progress.console
+            end = getattr(record, "end", "\n")
+            console.print(f"{prefix}{record.getMessage()}", style=style, end=end)
```
masenf: does the rich markup in the message need to be escaped here to avoid getting formatting when we expected something with square brackets?

packages/reflex-base/src/reflex_base/utils/log.py
```diff
@@ -0,0 +1,679 @@
...
+    Args:
+        msg: The message, possibly containing rich markup.
+
+    Returns:
+        The plain-text message.
+    """
+    if "[" not in msg:
+        return msg
+    try:
+        return Text.from_markup(msg).plain
```
masenf, line 142: not convinced that stripping markup is what we want for the JSON or file sinks:

```
>>> msg = "foo[bar]"
>>> Text.from_markup(msg).plain
'foo'
```

what we need to do is escape markdown in the log messages that we're printing via rich APIs and otherwise leave the log message alone for non-rich outputs

packages/reflex-base/src/reflex_base/utils/log.py
```diff
@@ -0,0 +1,679 @@
...
+def _file_handler() -> logging.FileHandler:
+    """Create the full-logging file handler.
+
+    Returns:
+        A file handler writing every record with markup stripped.
+    """
+    handler = logging.FileHandler(_log_file_path(), mode="w", encoding="utf-8")
+    handler.setFormatter(
+        _StripMarkupFormatter("[{asctime}] {levelname}: {message}", style="{")
+    )
```
masenf, line 344: the file output should not be stripping out rich markup, because the framework shouldn't be putting rich markup directly into log records.

if log messages want to use colors, i think that should be a separate opt-in kwarg like `rich=True` and if you pass that, then you as the caller are responsible for calling `rich.markup.escape` on any non-markup data.

when we get a log record with `rich=True` (and only then), we should call `strip_markup` on it. otherwise everything else should come through unfiltered (with the whole message being passed through `escape` when writing to actual rich sink)

Then we can strip markup in a sane way while preserving already-escaped composed strings:

```python
>>> msg = f"[b]Status: {rich.markup.escape('foo[bar]')}[/b]: Fizzed"
>>> Text.from_markup(msg).plain
'Status: foo[bar]: Fizzed'
```

When we strip markup from a string that has already escaped rich markup, we get that original string back without the unescaped rich markup.

tests/units/reflex_base/utils/test_log.py
```diff
@@ -0,0 +1,347 @@
...
+    logger.warning("w")
+    logger.error("e")
+    out, err = capsys.readouterr()
+    assert out == "Debug: d\nInfo: i\nSuccess: s\nWarning: w\n"
+    assert err == "e\n"
+
+
+def test_markup_rendered_in_rich_mode(capsys):
+    """Rich markup in messages is rendered, not shown literally."""
+    logger.info("hello [bold]world[/bold]")
```
masenf: this really needs to be opt-in

```suggestion
    logger.info("hello [bold]world[/bold]", rich=True)
```

tests/units/reflex_base/utils/test_log.py
```diff
@@ -0,0 +1,347 @@
...
+            if handler in logger.handlers:
+                logger.removeHandler(handler)
+
+
+def test_set_log_level_env_propagation(monkeypatch):
+    """Changing the level exports REFLEX_LOGLEVEL for subprocesses."""
+    log.set_log_level(LogLevel.DEBUG)
+    import os
+
+    assert os.environ.get("REFLEX_LOGLEVEL") == "debug"
```
masenf, line 203: does the `monkeypatch` reset this? or does calling `set_log_level` leak an environment variable out of the test case?

tests/units/reflex_base/utils/test_log.py
```diff
@@ -0,0 +1,347 @@
...
+
+def test_set_log_level_rejects_strings():
+    """Passing a raw string raises a TypeError."""
+    with pytest.raises(TypeError):
+        log.set_log_level("debug")  # pyright: ignore[reportArgumentType]
+
+
+def test_set_log_level_none_is_noop():
+    """Passing None keeps the current level."""
+    log.set_log_level(None)
```
masenf, line 216: same story here. if the tests were invoked at debug level, it seems this test case would reset that back to default for the remaining cases

tests/units/reflex_base/utils/test_log.py
```diff
@@ -0,0 +1,347 @@
...
+        "[yellow]DeprecationWarning: OldFeature has been deprecated in version "
+        "0.9.9. Use NewFeature. It will be completely removed in 1.0.[/yellow]",
+        markup=False,
+    )
+
+
+def test_console_print_json_mode(monkeypatch, capsys):
+    """console.print stays machine-readable in JSON mode."""
+    monkeypatch.setenv("REFLEX_LOG_JSON", "true")
+    console.print("plain [bold]message[/bold]")
```
masenf, line 326:
```suggestion
    console.print("AssertionErr: foo[bar] != 'baz'")
```

an example of why rich needs to be opt in and stripping markup not the default for all log messages

tests/units/conftest.py
```diff
@@ -29,6 +31,24 @@
 from .states.upload import SubUploadState, UploadState
 
 
+@pytest.fixture(autouse=True)
+def _capture_reflex_logs(caplog):
+    """Attach pytest's capture handler to Reflex package loggers.
+
+    The logging pipeline deliberately disables propagation at package roots,
+    while pytest normally captures records from the process root logger.
```
masenf: 🤔

masenf (commented):
some legit reviewbot feedback here.

i'm going to skim the rest of the PRs to see if there's maybe some overlap i'm not seeing that will come together

tests/units/conftest.py
masenf: to expand

i think we need a top level `reflex` logger that we do attach/propagate to root, and then attach the package level loggers to that such that we can keep a fairly standard setup, allow a single logger for programmatic tuning downstream, and when running through the `reflex` cli we should configure either the rich or json output; but if we're not in the cli, we should consider just NOT attaching a handler at all.

if we want `--json` mode to be useful, then we need to be careful about only emitting JSON; if we don't attach our handlers to the root logger when running in the cli, then other app code could break that contract by calling `basicConfig` or similar.

for apps that want to customize logging, they're probably wanting to attach systems to the root logger. i can see an advantage in having that intermediate logger below root, but having a changing list of potential non-propagating loggers that would need to be instrumented just seems annoying. this function is the first smell.

packages/reflex-base/src/reflex_base/utils/console.py
```diff
@@ -417,6 +419,27 @@ def ask(
...
+    """
+    table = Table()
+
+    for column in headers:
+        table.add_column(column)
+
+    for row in tabular_data:
+        table.add_row(*row)
+
+    _console.print(table)
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in e71cb591. `print_table` now emits one JSON record carrying `table: {headers, rows}` when `REFLEX_LOG_JSON` is set, instead of Rich text. It is not suppressed like `rule`, because a table is requested output (the hosting CLI's `apps list`, `secrets`, …), not decoration — dropping it would lose what the user asked for. Regression test: `test_console_print_table_json_mode`.

---

packages/reflex-base/src/reflex_base/utils/log.py
```diff
@@ -0,0 +1,679 @@
...
+
+
+@once
+def _file_handler() -> logging.FileHandler:
+    """Create the full-logging file handler.
+
+    Returns:
+        A file handler writing every record with markup stripped.
+    """
+    handler = logging.FileHandler(_log_file_path(), mode="w", encoding="utf-8")
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in e71cb591. `console.log_file_console()` now renders through `log.log_file_stream()`, i.e. the already-open stream of the pipeline's `FileHandler`, so full logging opens exactly one file and the legacy writer can no longer unlink a file the handler still holds open. Regression test: `test_console_and_pipeline_share_one_log_file`.

---

packages/reflex-base/src/reflex_base/utils/log.py
```diff
@@ -0,0 +1,679 @@
...
+        msg: The message.
+        dedupe: If True, suppress repeats of the same message.
+        stderr: Whether the message targets stderr.
+    """
+    if dedupe and not dedupe_once(("print", msg)):
+        return
+    _write_json(
+        {
+            "timestamp": datetime.datetime.now(tz=datetime.timezone.utc).isoformat(),
+            "level": "info",
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in e71cb591. `emit_json_print` now takes a `level`, and `console.print`/`_print_stderr` pass the caller's severity through, so `warn` serializes as `warning`, `error` as `error`, `success` as `success` and deprecations as `warning`. Regression test: `test_console_json_mode_preserves_severity`.

---

packages/reflex-base/src/reflex_base/utils/console.py
```diff
@@ -91,6 +85,9 @@ def print(msg: str, *, dedupe: bool = False, **kwargs):
         dedupe: If True, suppress multiple console logs of print message.
         kwargs: Keyword arguments to pass to the print function.
     """
+    if _log.is_json_mode():
+        _log.emit_json_print(msg, dedupe=dedupe)
+        return
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in e71cb591. Every direct console writer is now JSON-aware: `console.log` emits a JSON record instead of `_console.log`, `print_table` serializes its rows, and `PoorProgress.advance` goes through `console.print` (it previously wrote `Progress: N/M` straight to the terminal). `rule` stays suppressed and `status`/`progress` render nowhere. Regression tests: `test_console_log_json_mode`, `test_console_print_table_json_mode`, `test_poor_progress_json_mode`.

---

packages/reflex-base/src/reflex_base/utils/log.py
```diff
@@ -0,0 +1,679 @@
...
+def _log_file_path() -> Path:
+    """Resolve the path of the full-logging file.
+
+    Returns:
+        The log file path (parent directory created).
+    """
+    from reflex_base.environment import environment
+
+    if env_log_file := environment.REFLEX_LOG_FILE.get():
+        return env_log_file
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in e71cb591. `_log_file_path()` now creates the parent directory for a configured `REFLEX_LOG_FILE` too, not only for the default timestamped path. Regression test: `test_log_file_path_creates_parent_of_configured_path`.

---

packages/reflex-base/src/reflex_base/utils/console.py
```diff
@@ -417,6 +419,27 @@ def ask(
...
+    tabular_data: list[list[str]],
+    headers: Sequence[str] = (),
+) -> None:
+    """Print a table to the console.
+
+    Args:
+        tabular_data: The data to print in tabular format.
+        headers: The headers for the table.
+    """
+    table = Table()
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in e71cb591, but not by suppressing it. A table is requested output (the hosting CLI's `apps list`, `secrets list`, …) rather than decoration like `rule`, so in JSON mode it now emits one record carrying `table: {headers, rows}`; the stream stays parseable and the data is not silently dropped. Regression test: `test_console_print_table_json_mode`.

---

reflex/__init__.py
```diff
@@ -87,14 +87,21 @@
...
+# on the first record, keeping this import cheap.
+_log.bootstrap()
+del _log
 
 if sys.version_info < (3, 11):
-    from reflex_base.utils import console
+    import logging
 
-    console.warn(
+    logging.getLogger(__name__).warning(
```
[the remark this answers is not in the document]
FarhanAliRaza: Doesn't apply to the current head. `_BootstrapHandler` was removed in 75c64fb: `bootstrap()` now only reparents the package loggers under `reflex` and calls `configure()` when the CLI-managed marker (`REFLEX_MANAGED_LOGGING`) is set. Nothing routes a record into `configure()`, so the 3.10 warning cannot pull in `reflex_base.environment` (and with it the config/plugin stack) at `import reflex` time — on 3.10 the record goes to the `reflex` logger, which has no handler outside the CLI.

---

packages/reflex-base/src/reflex_base/environment.py
```diff
@@ -781,13 +787,11 @@ def _load_dotenv_from_files(files: list[Path]):
...
         files: A list of Path objects representing the environment variable files.
     """
-    from reflex_base.utils import console
-
     if not files:
         return
 
     if load_dotenv is None:
-        console.error(
+        logger.error(
```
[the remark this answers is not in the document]
FarhanAliRaza: Working as designed, so not changing it. After 75c64fb the pipeline deliberately attaches **no** sink outside the reflex CLI: records propagate to the root logger and the application's own logging configuration applies (see the `bootstrap`/module docstring in `reflex_base/utils/log.py`). The JSON and full-logging sinks only exist in CLI-managed processes, and those import `reflex` — hence `log.bootstrap()` — before any command body runs, so this dotenv error is captured there. Having `reflex_base.environment` claim handlers on its own import is exactly the library-mode behavior this design avoids.

---

packages/reflex-base/src/reflex_base/utils/log.py
```diff
@@ -0,0 +1,679 @@
...
+    subseconds = int((time.time() % 1) * 1000)
+    timestamp = time.strftime("%Y-%m-%d_%H-%M-%S") + f"_{subseconds:03d}"
+    log_file = Reflex.DIR / "logs" / (timestamp + ".log")
+    log_file.parent.mkdir(parents=True, exist_ok=True)
+    return log_file
+
+
+@once
+def _file_handler() -> logging.FileHandler:
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in e71cb591. `console.log_file_console()` now renders into `log.log_file_stream()` (the pipeline `FileHandler`'s already-open stream), so full logging opens a single file and the legacy writer no longer unlinks a file the handler holds open. Regression test: `test_console_and_pipeline_share_one_log_file`.

---

packages/reflex-base/src/reflex_base/utils/log.py
```diff
@@ -0,0 +1,725 @@
...
+def is_managed_mode() -> bool:
+    """Check whether this process runs under the reflex CLI.
+
+    Returns:
+        True if the reflex CLI (or a worker it spawned) owns log rendering.
+    """
+    return os.environ.get(_MANAGED_ENV_VAR) == "true"
+
+
+def enable_managed_logging():
```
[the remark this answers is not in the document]
FarhanAliRaza: Doesn't apply to this PR. This one only adds the pipeline and migrates no call site; the CLI entry point is migrated in #6865, where the `cli` group callback calls `log.enable_managed_logging()` (`reflex/reflex.py:35`) before any subcommand body runs, and worker subprocesses inherit the marker through `REFLEX_MANAGED_LOGGING`. The stack must merge in order, as noted in the PR description.

---

packages/reflex-base/src/reflex_base/utils/log.py
```diff
@@ -0,0 +1,725 @@
...
+    In managed mode (marker inherited from the CLI) the sinks attach
+    immediately so worker records render from the first line.
+    """
+    for name in PACKAGE_LOGGER_NAMES:
+        child = logging.getLogger(name)
+        child.parent = _REFLEX_LOGGER
+        child.propagate = True
+        # Inherit the effective level from "reflex"; setLevel also clears the
+        # isEnabledFor caches, which reparenting alone does not.
+        child.setLevel(logging.NOTSET)
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in e71cb591. `bootstrap()` now re-sets each package logger's *existing* level (`child.setLevel(child.level)`), which clears the `isEnabledFor` caches that reparenting alone leaves stale without overwriting an application's configuration, and it no longer forces `propagate`. Regression test: `test_bootstrap_preserves_application_logger_config`.

---

packages/reflex-base/src/reflex_base/utils/log.py
```diff
@@ -0,0 +1,725 @@
...
+    def register(self, key: Hashable) -> bool:
+        """Record a dedupe key, reporting whether it is new.
+
+        Args:
+            key: A hashable dedupe key.
+
+        Returns:
+            True the first time the key is seen, False afterwards.
+        """
+        hashed = hash(key)
```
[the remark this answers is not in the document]
FarhanAliRaza: Declining — storing hashes is deliberate. It came out of review feedback on this PR (https://github.com/reflex-dev/reflex/pull/6863#discussion_r3753923439): keeping the keys themselves means every deduped log message is retained for the process lifetime. The cost is a theoretical `hash()` collision between two dedupe keys within one process, whose only effect is that one repeat-suppressible line is not printed; that trade-off is the intended one. Covered by `test_dedupe_filter_stores_hashes_not_messages`.

---

packages/reflex-base/src/reflex_base/utils/log.py
FarhanAliRaza: Done in 75c64fb — `DedupeFilter.seen` is a `set[int]` holding `hash(key)` only, so no deduped message is retained. Covered by `test_dedupe_filter_stores_hashes_not_messages`.

---

packages/reflex-base/src/reflex_base/utils/log.py
FarhanAliRaza: Done in 75c64fb — Rich markup is now opt-in per record (`extra={"rich": True}`); everything else is printed with `markup=False`, so square brackets in a message stay literal. Covered by `test_markup_literal_by_default` / `test_markup_rendered_when_opted_in`.

---

packages/reflex-base/src/reflex_base/utils/log.py
FarhanAliRaza: Agreed, and changed in 75c64fb. Markup is now opt-in per record (`extra={"rich": True}`) and stripping only happens for those records: the JSON handler and the file formatter pass everything else through untouched, so `foo[bar]` survives verbatim. Covered by `test_json_mode_preserves_brackets_in_plain_records` and `test_file_formatter_strips_markup_only_when_opted_in`.

---

packages/reflex-base/src/reflex_base/utils/log.py
FarhanAliRaza: Implemented as described in 75c64fb: `rich` is an opt-in per-record flag (`logger.info(msg, extra={"rich": True})`), the caller owns escaping its interpolated data, and only those records get `strip_markup` in the file/JSON sinks. Records without the flag are written verbatim to the file and rendered with `markup=False` on the console. See `_FileFormatter` and `test_file_formatter_strips_markup_only_when_opted_in`.

---

tests/units/reflex_base/utils/test_log.py
FarhanAliRaza: Done in 75c64fb — the test now opts in explicitly: `logger.info("hello [bold]world[/bold]", extra={"rich": True})`, with `test_markup_literal_by_default` covering the non-opted-in case.

---

tests/units/reflex_base/utils/test_log.py
FarhanAliRaza: Nothing leaks: the autouse `clean_pipeline` fixture does `monkeypatch.delenv("REFLEX_LOGLEVEL", raising=False)` and `monkeypatch.setattr(log, "_log_level", LogLevel.INFO)`, so both the env var that `set_log_level` exports and the module global are restored after each test (and `log._reset()` detaches the sinks).

---

tests/units/reflex_base/utils/test_log.py
FarhanAliRaza: Same as above — `clean_pipeline` monkeypatches both `REFLEX_LOGLEVEL` and `log._log_level`, so a test that runs at debug level cannot change the level seen by later tests; teardown also calls `log._reset()` to detach the sinks and restore propagation.

---

tests/units/reflex_base/utils/test_log.py
FarhanAliRaza: Adopted in 75c64fb — that exact case is now a test: `test_markup_literal_by_default` asserts `logger.info("AssertionErr: foo[bar] != 'baz'")` renders literally, since markup is opt-in.

---

tests/units/conftest.py
FarhanAliRaza: Reworked along exactly these lines in 75c64fb, and the conftest helper that prompted this is gone (`tests/units/conftest.py` no longer differs from main):

- One top-level `reflex` logger owns the sinks. `bootstrap()` reparents the package roots (`reflex_base`, `reflex_cli`, `reflex_components_*`) under it and they propagate, so `logging.getLogger("reflex")` tunes everything in one place and per-package tuning still works. There is no changing list of non-propagating loggers to instrument.
- Handlers attach **only** in CLI-managed mode: `log.enable_managed_logging()` is called from the `reflex` CLI group callback (in #6865) and sets `REFLEX_MANAGED_LOGGING`, which worker subprocesses inherit. Outside the CLI no handler is attached at all — records propagate to root and the app's own configuration applies (`test_library_mode_records_propagate_to_root`, `test_set_log_level_in_library_mode_attaches_nothing`).
- In managed mode `reflex.propagate` is set to `False` so an app-side `basicConfig` cannot double-emit or break the `--json` only-JSON contract (`test_configure_cuts_propagation_to_root`); nothing is ever attached to the root logger.

---

packages/reflex-base/src/reflex_base/utils/log.py
```diff
@@ -0,0 +1,747 @@
...
+    return _dedupe_filter().register(key)
+
+
+def emit_json_print(
+    msg: str,
+    *,
+    level: str = "info",
+    dedupe: bool = False,
+    stderr: bool = False,
+    **fields,
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in 22e6e5be. `**fields` is now expanded first, so the canonical `timestamp`/`level`/`logger`/`message`/`pid` entries always win rather than being overwritten by an extra.

---

masenf:
1. `import reflex` is twice as slow on my machine now (claude claims 35x slower, but i don't see it)

[reflex/__init__.py](https://github.com/reflex-dev/reflex/blob/main/reflex/__init__.py) now eagerly imports reflex_base.utils.log to call bootstrap(), which drags in rich.console (26 ms), reflex_base.constants (29 ms), logging, json, shutil, pathlib, typing:

main:  0.0014 s   58 modules   rich_loaded=False
PR:    0.0882 s  231 modules   rich_loaded=True     (94 of 96 ms is reflex_base.utils.log)

That defeats the lazy_loader design in [reflex/__init__.py](https://github.com/reflex-dev/reflex/blob/main/reflex/__init__.py). codspeed missed it because there's no import-time benchmark.

I prototyped two fixes and measured both. The cheap one: since reparenting a package root fixes all descendants whenever it happens, bootstrap() doesn't need to run at import reflex at all — move the call into [reflex_base/utils/log.py](https://github.com/reflex-dev/reflex/blob/main/reflex_base/utils/log.py)'s module body (console.py already imports log at module scope, so it fires the moment reflex does anything) and drop the import from __init__.py:

self-bootstrapping log.py:   0.0026 s   95 modules   (parenting correct once console loads)
import-light bootstrap mod:  0.0265 s   96 modules   (~24 ms of that is `import logging` itself)

2. Tracebacks get word-wrapped by rich. `RichConsoleHandler.emit` prints the formatted traceback through `console.print`, so at 80 cols file paths break mid-path across lines. soft_wrap=True (or crop=False) fixes it.

3. Deprecation JSON records report the wrong location — [.../utils/log.py:740](https://github.com/reflex-dev/reflex/blob/main/.../utils/log.py#L740) (the framework frame) rather than the user frame, which is only available inside the message string. Worth overriding for kind="deprecation" since machine consumers are the whole point of that sink.

FarhanAliRaza:
@masenf  this is fixed.

reflex/__init__.py
```diff
@@ -89,12 +89,16 @@
 from reflex_base.utils import lazy_loader
 
 if sys.version_info < (3, 11):
-    from reflex_base.utils import console
+    import logging
 
-    console.warn(
+    # Loading the pipeline parents the package loggers and, under the CLI,
+    # attaches the sinks so this warning renders like every other record.
+    import reflex_base.utils.log
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in cf0dfa18e. The 3.10 branch now imports the pipeline as `from reflex_base.utils import log as _log` and `del`s it with `logging`, so `vars(rx)` is identical across Python versions. Regression assertion added to `test_import_reflex_stays_light` (`'reflex_base' not in vars(reflex)`), verified failing-then-passing under `uv run -p 3.10`.

tests/benchmarks/test_import.py
```diff
@@ -0,0 +1,27 @@
...
+def test_import_reflex(benchmark: BenchmarkFixture):
+    saved = sys.modules.copy()
+    try:
+
+        def cold_import():
+            _wipe_reflex_modules()
+            importlib.import_module("reflex")
+
+        benchmark(cold_import)
+    finally:
```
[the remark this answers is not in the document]
FarhanAliRaza: Declined — the premise is inverted. After this commit `import reflex` does **not** load `reflex_base.utils.log` on 3.11+ (that is the point of the change; `test_import_reflex_stays_light` asserts it), so `bootstrap()` never runs inside a benchmark iteration. Empirical check: 5 wipe-and-reimport rounds, snapshotting id/level/propagate/handlers/parent of the `reflex`, `reflex_base`, `reflex_cli`, `reflex_components_core` loggers plus root handlers, before vs after — all identical, both with and without `REFLEX_MANAGED_LOGGING=true`, and the restored `sys.modules` entries are the original module objects. Nothing is printed either. `bootstrap()` is also idempotent by design (`child.parent = <same singleton>`, `setLevel(child.level)`), and the `logging` module itself is never wiped, so logger objects stay the same singletons throughout.

reflex/__init__.py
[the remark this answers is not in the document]
FarhanAliRaza: Superseded by 85e40c8fb: the pipeline import is removed from the 3.10 branch entirely (it only added a prefix for worker subprocesses on a deprecated Python; the CLI parent has not enabled managed mode at import time anyway, so the warning goes via logging.lastResort regardless). `reflex/__init__.py` no longer references the pipeline on any version, which is the real fix for the leak.

masenf (dismissed):

Alek99 (approved):

PR #6888, opened 2026-08-14
reflex-release: add a post-release workflow hook

masenf:
### All Submissions:

### Type of change

### What this does

Adds `post-release-workflow` to `[tool.reflex-release]`. It names a workflow of the consuming repository's own, and `publish.yml` dispatches it **once per published tag**, after the upload, the tag and the GitHub release all exist — the hook for whatever has to follow a release: publishing docs, refreshing an image, notifying a downstream repository.

```toml
[tool.reflex-release]
post-release-workflow = "docs_publish.yml"
```

The dispatch runs **on the tag**, so the workflow sees exactly the tree that was published, and it is handed the three facts about the release:

```yaml
# .github/workflows/docs_publish.yml
on:
  workflow_dispatch:
    inputs:
      tag: { description: "The published tag", required: true, type: string }
      package: { description: "The published package", required: true, type: string }
      version: { description: "The published version", required: true, type: string }
```

All three inputs are required because GitHub rejects a dispatch carrying inputs the target workflow does not declare.

### Design notes

- **The step is only scaffolded when the setting is present.** So is the `actions: write` grant it needs, in `publish.yml` and in both workflows that call it (`release_from_changelog.yml`, `auto_release_internal.yml`). A repository that runs nothing after a release keeps `actions: read` and gets no dead step. `render()` gained a small general rule for this: a placeholder alone on a line is dropped with its line when its value is empty.
- **Naming a generated workflow is rejected by `sync`.** Handing a published tag back to the release pipeline would either re-enter it or fail on inputs it does not declare.
- **A failed dispatch fails the run, loudly.** It is the last thing a release does, so it never leaves a half-published version — but a typo in the workflow name is something you want to hear about, and the error names the tag whose follow-up did not start.
- **Adding or removing the setting changes the generated workflows**, so it needs a `reflex-release sync`. The existing PR drift check catches a repository that forgets.

New `reflex-release post-release` command, wired through `env:` like every other step.

### Changes To Core Features:

`uv run pytest tests/units/reflex_release` — 256 passed (10 new, covering config parsing, both rendering paths, the permission widening in all three workflows, the `sync` guard, and the dispatch/failure behavior). `ruff check`, `ruff format --check` and `pyright` clean.

Split out of a combined branch alongside #6889; the two features are independent, and whichever lands second will need a trivial rebase on `README.md`, `commands.py` and `test_commands.py`.


masenf:
All five review points were valid and are addressed in f1fc2a56.

**Guard bypasses (P1 display names, P2 `OPTIONAL_WORKFLOWS`)** — both real, and the same root cause: the guard compared against the wrong set. It now compares against `generated_workflow_names()`, which is every workflow this tool *can* generate (`CORE_WORKFLOWS + OPTIONAL_WORKFLOWS`, not the currently-managed subset) under both the file name and the `name:` the template declares, read from the templates so the two cannot drift. The `auto_release_internal.yml` case was the worse of the two: it slipped through precisely when `sync` deletes the file, so the dispatch would have had no receiver at all.

**Empty dispatch payload (P3)** — real. `--field package=` is accepted, `gh` exits 0, and the run goes green having told the downstream workflow nothing, which contradicts the loud-failure design. All three values are now required and the package is checked with `require_known`, as `cmd_prepare_publish` does.

**Duplicated identifiers (P2)** — taken in the production code: `POST_RELEASE_WORKFLOW_KEY` and `POST_RELEASE_INPUTS` in `config.py` now feed the config parsing, the dispatch payload, the scaffolded step's comment and its failure message. That last one was the actual drift risk — the generated comment told users to declare inputs named in prose, unconnected to the list the payload was built from.

One deliberate exception: the **tests still spell the key and the input names out literally**. `test_post_release_dispatches_the_workflow_on_the_tag` asserts the exact `gh workflow run` argv, and the conftest helper writes the literal TOML key. Deriving those from the same constants would make them pass whatever the constants said — the duplication there *is* the check.

**Duplicated test helper (P3)** — hoisted to `conftest.py` as `set_post_release_workflow`, next to the `write_lockstep` helper it mirrors.

Tests are now 265 passing (19 new, including the four bypass names as a parametrized case and the three empty-input cases); `ruff check`, `ruff format --check` and `pyright` clean.

---

Alek99 (approved):

PR #6889, opened 2026-08-14
reflex-release: lift unshippable dependency pins when materializing a release

masenf:
### All Submissions:

### Type of change

### The problem

A package that depends on a sibling it is waiting for pins the unreleased version — `widget-core >= 0.2.0.dev1` — so the workspace resolves while the sibling is still unpublished. That pin cannot be published: `*.dev` versions never reach PyPI, so the metadata would be uninstallable. `check-dev-pins` rejects it at build time, which means someone has to remember to lift it once the sibling is out — and finds out at the publish gate, with the changelog already bumped, when they forget.

### What this does

Materialization lifts the pin. Every dependency floor the release cannot ship is rewritten to the **earliest published version that satisfies the whole requirement**, `uv.lock` is re-resolved, and both land in the release commit alongside the changelog bump — through the same review.

| Floor | Materializing a prerelease | Materializing a final version |
| --- | --- | --- |
| `>= 0.2.0.dev1` | earliest published `0.2.0a1`, `0.2.0`, … | earliest published *final* `0.2.0`, … |
| `>= 0.2.0a1` | left alone — an alpha may ship it | lifted to the earliest published final |
| `>= 0.2.0` | left alone | left alone |

A floor no published version satisfies has nowhere to go, so the package is **held back at plan time** rather than materialized into a version that could never be published: auto-selected packages are dropped from the batch (a lockstep group whole, since its members only release together) and listed in the run summary, while an explicit selection fails the dispatch. Release the depended-on package first and the next release lifts the pin by itself.

### Design notes worth reviewing

1. **"Published" means tagged.** Resolution reads the repository's own git tags, not PyPI — tags are created only after a successful upload, which is already this pipeline's record of what shipped. Offline, deterministic, no new network dependency in the plan step; the cost is that it only works for siblings, and that the release workflows must check out full history and tags (they do).
2. **Dependencies outside the repository.** A `*.dev` floor on one holds the release back (it is unpublishable regardless, and `check-dev-pins` would fail the build anyway); a *prerelease* floor on one is left alone, since its releases are not recorded here and pinning a beta is somebody's deliberate choice.
3. **Lifting prerelease floors on a final release is beyond the literal problem, and closes it.** Without it the feature leaks: an alpha release rewrites `>= 0.2.0.dev1` to `>= 0.2.0a1`, `check-dev-pins` is satisfied, and the subsequent *final* release ships a floor that resolves users onto an alpha. Happy to scope this back to dev floors only if you would rather.
4. **`uv lock` is the lock refresh**, auto-detected from a root `uv.lock` and skipped when there is none. Not configurable — the tool already assumes uv everywhere (`cli-command`, `setup-uv` in every generated job).
5. **Lockstep siblings under `pin-exact` are skipped**, since `pin-lockstep` rewrites those to `== <version>` at build time and nothing declared here is shipped.
6. **The upgrade is atomic and narrowly staged.** Pin rewrites and the lock refresh succeed together or are rolled back, so a re-run never finds the pins already lifted and skips re-locking. `materialize` reports the paths it rewrote as a `repinned` output — it runs as a separate process from the delivery step — and `_commit_changelogs` (now `_commit_materialized`) stages exactly those beside the changelogs, so unrelated worktree edits stay out of the release commit.

**`dispatch_release.yml` changes** (the materialize step gains an `id`, the two delivery steps a `REPINNED_JSON` env), so adopting repos need a `reflex-release sync`; the PR drift check enforces it.

### Changes To Core Features:

`uv run pytest tests/units/reflex_release` — 299 passed (34 new here): bound resolution across both prerelease modes, the whole-specifier-set constraint, requirement rewriting that preserves extras/markers/other specifiers and relaxes a strict floor, TOML basic- and literal-string rewriting, the lockstep and outside-dependency exemptions, rollback on a failed lock and on a mid-batch failure, and the plan-time hold-back in explicit, auto and lockstep-group forms. `ruff check`, `ruff format --check` and `pyright` clean.

Rebased onto `main` now that #6888 has merged; the two features compose cleanly (verified the rendered workflows carry both the post-release dispatch step and the `repinned` threading).


packages/reflex-release/src/reflex_release/devpins.py
```diff
@@ -100,3 +154,344 @@ def check_dev_pins(config: Config, packages: list[str]) -> None:
...
+                    None,
+                    f"{parsed.name} is not a package in this repository, so its "
+                    "published versions are not known here; re-pin it by hand",
+                )
+            )
+            continue
+        candidates = [
+            version
+            for version in tag_versions(config, sibling)
+            if allow_prereleases or is_final(version)
```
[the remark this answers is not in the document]
masenf: this is fine and it's by design. the CI jobs should not be performing shallow clones anyway. if a proper publish tag isn't in the repo, we don't materialize the change for the target package. that's how this feature is supposed to work.

packages/reflex-release/src/reflex_release/devpins.py
```diff
@@ -100,3 +154,344 @@ def check_dev_pins(config: Config, packages: list[str]) -> None:
...
+                    bounds,
+                    None,
+                    f"{parsed.name} is not a package in this repository, so its "
+                    "published versions are not known here; re-pin it by hand",
+                )
+            )
+            continue
+        candidates = [
+            version
+            for version in tag_versions(config, sibling)
```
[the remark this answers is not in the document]
masenf: by design. other reviewbot also raised this.

masenf:
All four outstanding points addressed in 75e98b57. (The two shallow-clone threads @masenf already closed as by-design — agreed, and the README now says so where it explains that "published" means tagged.)

**Strict floor rewritten wrong (P1) — a real bug, good catch.** `widget-core > 0.2.0.dev1` admits `0.2.0`, so `0.2.0` is what it resolves to, and rewriting the operator verbatim produced `> 0.2.0` — excluding the very release the requirement had just been lifted onto. Under an upper bound (`> 0.2.0.dev1, < 0.3` with only `0.2.0` in range) it becomes unsatisfiable outright. A strict floor over an unreleased version now becomes an inclusive floor over the release above it.

I also added the invariant behind it: `rewritten()` now checks the result against the version it resolved to and fails rather than returning it. Reverting just the operator mapping, that check fires on its own — `lifting 'widget-core > 0.2.0.dev1' produced 'widget-core > 0.2.0', which 0.2.0 does not satisfy` — so a future gap in the operator handling can't publish metadata that resolves to nothing.

**TOML escaping (P2) — real, and fail-safe rather than corrupting, but it aborts a release for a legal input.** Confirmed the mechanism: for `"widget-core >= 0.2.0.dev1; python_version > \"3.10\""`, tomllib hands back the *unescaped* value, so the quoted search found 0 occurrences. Both spellings a value can have are now tried — an escaped basic string, and a literal string, which cannot escape anything. Not worth a round-tripping TOML dependency on the release path for this.

**Non-atomic upgrade (P2) — real, though it needs a hand-run to bite.** In CI a failed `materialize` fails the job on a throwaway checkout, so nothing is committed; locally the tree keeps lifted pins beside the old lock, and a re-run finds nothing left to lift, skips the refresh, and could commit exactly that pairing. The pyproject rewrites are now rolled back when `uv lock` cannot follow them, so a re-run has the same work to do.

**Over-broad staging (P2) — you're right that the comment overclaimed.** The obstacle was that `materialize` and the delivery step are separate processes, so the committing side could not know what had been rewritten. Fixed by making it known: `materialize` now emits a `repinned` output listing the paths, threaded to `open-release-pr`/`push-prerelease` the way `releases` already flows from `plan`. Only those paths are staged, so an unrelated uncommitted edit to a released package's `pyproject.toml` stays out of the release commit — there's a test for exactly that.

279 passing (6 new); `ruff check`, `ruff format --check` and `pyright` clean. Note `dispatch_release.yml` changed (the materialize step gained an `id` and the two delivery steps a `REPINNED_JSON` env), so this needs a `reflex-release sync` in any adopting repo — the PR drift check enforces it.

---

packages/reflex-release/src/reflex_release/devpins.py
```diff
@@ -100,3 +154,399 @@ def check_dev_pins(config: Config, packages: list[str]) -> None:
...
+        }
+    }
+    changed = apply_pin_upgrades(config, upgrades)
+    try:
+        lock = refresh_lock_file(config)
+    except ReleaseError:
+        for path, text in snapshot.items():
+            path.write_text(text, encoding="utf-8")
+        echo(f"restored {len(snapshot)} pyproject.toml file(s); no pin was lifted")
+        raise
```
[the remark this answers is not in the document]
masenf: isn't the `write_text` the part that's doing the restore? i think this is fine how it is

packages/reflex-release/src/reflex_release/devpins.py
```diff
@@ -100,3 +154,399 @@ def check_dev_pins(config: Config, packages: list[str]) -> None:
...
+    # re-lock and could commit exactly that.
+    snapshot = {
+        path: path.read_text(encoding="utf-8")
+        for path in {
+            config.package_path(upgrade.package) / "pyproject.toml"
+            for upgrade in upgrades
+        }
+    }
+    changed = apply_pin_upgrades(config, upgrades)
+    try:
```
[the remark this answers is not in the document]
masenf: Valid — fixed in fc0dc7b7. The guard started one line too late: `apply_pin_upgrades` rewrites package by package, so a requirement the second package cannot be given (one appearing twice, or a rewrite its resolved version would not satisfy) left the first package's pins lifted while the command failed. The rewrite is now inside the guarded block, so the whole batch and the lock file move together or not at all.

Added a regression test with two packages where the second one's requirement is ambiguous; it fails on the previous arrangement (the root package's pin stays lifted) and passes now. 280 passing.

---

packages/reflex-release/src/reflex_release/devpins.py
```diff
@@ -100,3 +154,400 @@ def check_dev_pins(config: Config, packages: list[str]) -> None:
...
+
+    Returns:
+        The updated file content.
+    """
+    spellings = [
+        (_toml_basic(original), _toml_basic(replacement)),
+        (f"'{original}'", f"'{replacement}'"),
+    ]
+    counts = [text.count(needle) for needle, _ in spellings]
+    if sum(counts) != 1:
```
FarhanAliRaza: `text.count()` scans the whole file, so a copy of the requirement that never gets published counts too. This aborts a release that is otherwise fine:

```toml
dependencies = ["widget-core >= 0.2.0.dev1"]

[dependency-groups]
dev = ["widget-core >= 0.2.0.dev1"]
```
```
Error: expected exactly one quoted 'widget-core >= 0.2.0.dev1' requirement to upgrade, found 2
restored 1 pyproject.toml file(s); no pin was lifted
```

`check_dev_pins` deliberately ignores `[dependency-groups]` as development-only; the rewriter should agree with it. The `dependencies` + extra case fails the same way, and there both copies genuinely need lifting — you already hold both `PinUpgrade` entries. `plan` sees neither, so it lands mid-dispatch. It fails safe, but someone has to re-pin by hand to get the release out.

packages/reflex-release/src/reflex_release/commands.py
```diff
@@ -721,35 +819,58 @@ def _release_summary(releases: list[dict[str, str]]) -> str:
...
+    """Parse the ``repinned`` output of :func:`cmd_materialize`.
+
+    Args:
+        repinned_json: The JSON array of rewritten paths, or an empty string
+            when the pin upgrade rewrote nothing.
+
+    Returns:
+        The repo-relative paths.
+    """
+    return json.loads(repinned_json) if repinned_json.strip() else []
```
FarhanAliRaza, line 847: When `REPINNED_JSON` is unset — a consumer whose workflow has not been re-synced — this returns `[]` without a word, and the pins that were just lifted never reach the commit:

```
mypkg: widget-core >= 0.2.0.dev1 -> widget-core >= 0.2.0a1
commit contents: CHANGELOG.md, news/+thing.feature.md
git show HEAD:pyproject.toml -> widget-core >= 0.2.0.dev1
```

The branch then dies at the publish-time `check-dev-pins` gate with the changelog already bumped and the prerelease branch pushed — the exact failure this PR exists to remove, now with more state to unwind. `sync --check` catches the drift only on the next PR, and the default `cli_command` is an unpinned `uvx reflex-release`, so materialize can gain the output before the workflow does.

Failing in `_commit_materialized` when a `pyproject.toml` or `uv.lock` is still dirty after staging would close it cheaply.

packages/reflex-release/src/reflex_release/devpins.py
```diff
@@ -100,3 +154,400 @@ def check_dev_pins(config: Config, packages: list[str]) -> None:
...
+        for entry in entries:
+            rewritten = entry.rewritten()
+            text = _replace_requirement(text, entry.requirement, rewritten)
+            echo(f"{package}: {entry.requirement} -> {rewritten}")
+        pyproject.write_text(text, encoding="utf-8")
+        changed.append(pyproject.relative_to(config.root).as_posix())
+    return changed
+
+
+def refresh_lock_file(config: Config) -> str | None:
```
FarhanAliRaza, line 598: Two things I noticed running this.

In a uv workspace — this repo's shape — `uv.lock` records siblings with no specifier at all (`{ name = "reflex-base", editable = "packages/reflex-base" }`), so lifting a sibling pin cannot move the lock. The pins that *would* move it are outside deps, and those get held back rather than lifted. So the re-lock is inert exactly where the feature is aimed.

Where it isn't inert, it does more than the pins. `uv lock` re-resolves from PyPI, so a lock that was already stale on `main` brings its churn into the release commit — my run added a whole new transitive package to the diff. Either bail early when the lock is already stale, or drop the README's "nothing else is staged" framing for the lock file.

packages/reflex-release/src/reflex_release/devpins.py
```diff
@@ -100,3 +154,400 @@ def check_dev_pins(config: Config, packages: list[str]) -> None:
...
+    upgrades = _upgrades_for(config, packages, allow_prereleases)
+    blocked: dict[str, list[PinUpgrade]] = {}
+    for upgrade in upgrades:
+        if upgrade.resolved is None:
+            blocked.setdefault(upgrade.package, []).append(upgrade)
+    if blocked:
+        listing = "\n".join(f"  {line}" for line in describe_blockers(blocked))
+        fail(
+            "dependency pins that no published version satisfies cannot be "
+            f"materialized into a release:\n{listing}\n\nRelease the depended-on "
```
FarhanAliRaza: `widget-core == 0.2.0.dev1` reaches here and gets told the next release lifts it automatically:

```
mypkg: 'widget-core == 0.2.0.dev1' - no final release of widget-core satisfies it (newest tagged: 0.3.0)
Release the depended-on package(s) first; the next release lifts these pins automatically.
```

No published version can ever satisfy an exact dev pin, so that package is stuck permanently and the advice sends you in a circle. Worth splitting the `==`/`===` case out and saying re-pin by hand. Same wording at commands.py:277.

FarhanAliRaza (commented):
Ran this for real rather than reading it: built a sandbox repo with a root package, a sibling, and tags at `0.1.0 / 0.2.0a1 / 0.2.0 / 0.3.0`, then drove `plan` -> `materialize` -> `push-prerelease` through a dozen shapes.

The behaviour matches the README table. `>= 0.2.0.dev1` lifts to `>= 0.2.0` for a final and `>= 0.2.0a1` for a prerelease; `> 0.2.0.dev1` becomes `>= 0.2.0`; extras and a `python_version > "3.10"` marker survive untouched; a prerelease floor on an outside dep is left alone while a dev floor on one is held back; an unsatisfiable floor drops an auto-selected package (with a `### Held back` summary) and fails an explicit one; the two-package rollback from the last commit really does restore both pyprojects and leave the fragments on disk; and the commit carries exactly the changelog, the fragment deletion and `pyproject.toml`, with an unrelated dirty file left out.

Two things I'd fix before merging, plus a few caveats — inline.

Two more that didn't have a good line to hang on:

- The pin's target version and the release action can disagree. `>= 0.4.0.dev1` on a sibling at `0.3.0` plus `release-patch` produces `0.3.1`, which still doesn't satisfy the floor, so the dependent stays held back until someone picks a minor. Worth a line in the README next to "the next release lifts the pin by itself".
- The rollback guard covers apply + lock but not the towncrier step after it. I made towncrier fail and `pyproject.toml` was left lifted. CI throws the worktree away so this only bites local runs — noting it rather than asking for a change.

Tests, ruff and pyright are clean (299 passed).

packages/reflex-release/src/reflex_release/devpins.py
masenf: Both halves valid — fixed in 15e2f8d1.

The rewrite now counts and replaces only inside the `[project]` table and its subtables, which is exactly the region `published_dependencies` reads. A `[dependency-groups]` copy is neither counted nor rewritten, so the rewriter and `check_dev_pins` now agree on what "published" means.

And you're right that the `dependencies` + extra case was the same failure for the wrong reason — I had it backwards, with a test asserting the abort as intended behavior. Both copies are published and both need lifting, and I already hold a `PinUpgrade` per copy, so `_replace_requirement` is now told how many to expect and rewrites them all. That test is replaced by one asserting both copies move, plus one asserting the dependency-group copy stays put.

The count mismatch is still a failure, for the case that really is unrewritable (a requirement spelled so the parsed value cannot be found in the text) — that path is what the mid-batch rollback test now triggers.

---

packages/reflex-release/src/reflex_release/commands.py
masenf: Valid, and worse than I'd weighed it. I knew about the un-synced-workflow gap and argued `sync --check` covered it — but that's a *next-PR* check, and with the default unpinned `uvx reflex-release` the tool can gain the output before the workflow does, so the two can disagree on a real release. Silently.

Took your suggestion in 15e2f8d1, scoped to the files materialization owns (the released packages' `pyproject.toml` plus the lock file) rather than the whole tree: after staging, if one of those is still modified, the commit fails and names the re-sync. It can't false-positive on the normal path, since anything the pin upgrade rewrote is in `repinned` and therefore staged.

That does mean an unrelated uncommitted edit to a released package's `pyproject.toml` now fails the command instead of being quietly left out. I think that's right for a release command — the previous behaviour was the silent one — and in CI the checkout is clean, so it only bites a hand-run.

---

packages/reflex-release/src/reflex_release/devpins.py
masenf: Confirmed your first point against this repo's own lock, and it's the more interesting of the two. `uv.lock` records workspace members with no specifier at all:

```
{ name = "reflex-base", editable = "packages/reflex-base" },   # sibling
{ name = "packaging", specifier = ">=24.2,<27" },              # index dep
```

So a lifted sibling pin has nothing in the lock to show up in, and siblings are exactly what this feature lifts. I'd assumed the re-lock was load-bearing here; it isn't.

Partly fixed in 15e2f8d1: the lock is reported and staged **only when the re-resolution actually changes it**. In a uv workspace that means it is not staged at all, so the inert case is now visibly inert instead of quietly staging whatever `uv lock` did. It still earns its place in the other layout — a `packages-dir` repo that is not a uv workspace, where the sibling resolves from an index and the lock does carry its specifier.

On the churn: I took the second of your two options and dropped the README's "nothing else is staged" framing for the lock file, replacing it with a section that says plainly that `uv lock` re-resolves against the index and that catching up an already-stale lock lands in the release commit. I did **not** add the bail-early-on-stale-lock check — it needs a `uv lock --check` pre-step and a policy call about failing a release over staleness that predates it, which is a decision for you rather than something to slip into this PR. Filed as follow-up below.

**Follow-up (not this PR):**
1. Bail out of materialization when `uv.lock` is already stale against the main branch, so a release commit never carries a pre-existing re-resolution. Needs a `uv lock --check` pre-step and a decision on whether that should fail the release or just warn.
2. Reconsider whether the re-lock should run at all for repositories that are uv workspaces, given it provably cannot move the lock there.

---

packages/reflex-release/src/reflex_release/devpins.py
masenf: Valid — fixed in 15e2f8d1. I'd reasoned about exact pins when writing the resolver (they can never resolve, so they fall through to the blocker path) and then let them inherit the generic wait-for-a-release message, which is exactly the circle you describe.

An offending bound under `==`/`===` is now recognised as a dead end rather than a wait: it short-circuits before the tags are consulted (they have nothing to say) and reports "an exact pin on an unpublished version can never be satisfied by a release; re-pin it by hand".

The closing advice is now derived from the blockers rather than fixed, so it matches at both sites, including commands.py: only-waitable pins get "release the depended-on package(s) first", only-exact pins get "re-pin it by hand", and a mixed batch gets both sentences.

---

FarhanAliRaza (approved):

PR #6905, opened 2026-08-18
perf: disable React dev-build owner-stack capture in dev mode

Alek99:
## Why dev mode feels slow to click around

Profiling real navigation clicks in a large Reflex app (the reflex.dev docs site, 1,023 pages) under `reflex run` vs `--env prod`, interleaved click-by-click on the same machine:

| | busy CPU per navigation click |
|---|---|
| `reflex run` (dev) | **345.7 ms** |
| `--env prod` | **62.6 ms** |

**71% of the dev CPU (246.7 ms) is a single function: React's development-build `createElement`.** React 19's dev runtime constructs an `Error("react-stack-top-frame")` plus a `console.createTask()` for *every element created*, to capture "owner stacks" for DevTools. Two things make this brutal in practice:

1. **The cost scales with render depth.** `Error()` walks the live JS stack: measured 2.1 µs/element at stack depth 2, but 12–17 µs at the depth React actually renders at. A docs page creates ~10,300 elements per client-side navigation → ~247 ms.
2. **Humans pay full price on every click.** React budgets the capture to 10k elements, but `react-dom` resets the counter 1 s after the last render. Tight automated loops exhaust the budget once and look fast; a person clicking at a normal pace gets a fresh budget — and the full cost — every single click. This is exactly the "dev feels sluggish but benchmarks look fine" signature.

(The "outdated JSX transform" warning is a red herring here — the modern `jsx`/`jsxs` runtime executes the identical per-element capture.)

## The fix

Emit a small dev-only snippet in the generated `context.js` that pins React's budget counter past its cap, so element creation takes the runtime's own cheap branch (`unknownOwnerDebugStack`). A pinned accessor is required — a plain write would be undone by react-dom's once-per-second reset (this is also why naive attempts to disable it appear to do nothing).

Everything else about the dev build is unchanged: dev warnings, component stacks in errors, StrictMode behavior, HMR/Fast Refresh (all re-verified against a live app, including in-place HMR). The trade-off: elements carry React's shared unknown-owner placeholder, so the public `React.captureOwnerStack()` returns no owner frames in dev — affecting React DevTools' owner-stack view *and* anything built on that API, such as custom error overlays and the `onCaughtError`/`onUncaughtError` root handlers (verified against the pinned React 19.2.8). `REFLEX_REACT_OWNER_STACKS=1` restores full owner stacks; the guard is browser-only, so SSR diagnostics are untouched. The snippet is defensive (existence-checked, try/catch) so a future React that drops the internal is a no-op, and it is not emitted in prod builds at all.

## Measured result

Interleaved A/B on one server boot (docs app, human-paced navigation clicks, CDP CPU profiles, median of n=10 per arm):

| | busy CPU / nav | `createElement` self time | settle time |
|---|---|---|---|
| unpinned (today's dev) | 353.1 ms | 250.3 ms | 348.4 ms |
| pinned (this PR) | **82.8 ms** | **1.0 ms** | **81.8 ms** |

**−270 ms per click (−77%), taking dev from 5.6× prod to ~1.3× prod.** Verified no console errors, navigation, state events, and hot reload working on both a minimal app and the docs app; full unit suite passes (7,460).

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/6905?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/reflex-base/src/reflex_base/compiler/templates.py
```diff
@@ -358,10 +362,40 @@ def context_template(
...
+// REFLEX_REACT_OWNER_STACKS=1 to restore full owner stacks (e.g. for React
+// DevTools' owner-stack view).
+try {
+  const reactInternals =
+    React.__CLIENT_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE;
+  if (
+    reactInternals &&
+    typeof reactInternals.recentlyCreatedOwnerStacks === "number"
+  ) {
+    Object.defineProperty(reactInternals, "recentlyCreatedOwnerStacks", {
```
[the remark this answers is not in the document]
Alek99: Applied — extracted `recentlyCreatedOwnerStacks` into an `ownerStackCounterKey` constant used by both the guard and `Object.defineProperty`, and re-verified the pin still takes effect at runtime in a live dev app.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Alek99:
### Side-by-side: navigating a content-heavy app in dev mode

Recorded a click-through of the same app with and without this change. Left pane is dev mode as it ships today, right pane is with the pin. Same app, same four navigation clicks, same machine; the header bar in each pane shows that click's render time, measured in-page from the click to the last DOM mutation.

| click | before | after |
|---|---|---|
| 1 | 50 ms | 18 ms |
| 2 | 46 ms | 17 ms |
| 3 | 64 ms | 18 ms |
| 4 | 52 ms | 17 ms |

Method notes, since a side-by-side is easy to make lie:

- The two arms were recorded **separately**, not simultaneously — running both at once would have had them competing for CPU and inflated both sides. The recordings are aligned on the first click afterwards, so the same wall-clock moment appears in both panes.
- Both arms run against the **same server build**; the "before" arm restores React's default counter in-page, so the only difference between the panes is the owner-stack capture itself.
- Each click is preceded by ~1.4 s of idle, because `react-dom` resets the 10k-element owner-stack budget one second after the last render. That is the human-paced case this PR is about: a tight automated loop exhausts the budget once and shows almost no difference, which is why this never showed up in benchmarks.
- This is a synthetic content-heavy app (~5k elements per navigation), not the docs site, so the absolute numbers are smaller than the ones in the description. The effect size is the same: **~75% less main-thread CPU per click**. On the reflex.dev docs app the same measurement is 353 ms → 83 ms.

packages/reflex-base/src/reflex_base/compiler/templates.py
```diff
@@ -358,10 +362,54 @@ def context_template(
...
+// so `React.captureOwnerStack()` returns no owner frames. That affects React
+// DevTools' owner-stack view *and* anything built on the public API, such as
+// custom error overlays and the `onCaughtError` / `onUncaughtError` root
+// handlers. Component stacks in React's own error messages, dev warnings and
+// Fast Refresh are unaffected. Set REFLEX_REACT_OWNER_STACKS=1 to turn this
+// off and get full owner stacks back.
+//
+// Browser only: this module is also evaluated by the server renderer, and
+// mutating the shared React internals there would disable owner stacks for
+// SSR diagnostics in the same process.
```
masenf: i don't think we need this novel of a comment in every generated context.js; a short comment referencing this PR for more context is sufficient.

masenf (commented):
i wonder if it makes more sense to switch the flexgen sandbox to use the new `--env preview` mode that basically makes an unoptimized prod build. would probably need to benchmark it to see if the compile time is acceptable.

this feels pretty hacky, but it does make a noticable difference and has a chicken switch, so it's okay with me.

packages/reflex-base/src/reflex_base/compiler/templates.py
Alek99: Trimmed in e8b97a9c: four lines — what it does, what it costs (`captureOwnerStack` owner frames), the `REFLEX_REACT_OWNER_STACKS=1` escape hatch, and a link to this PR for the full context. Re-verified the pin still engages at runtime after the trim (counter reads pinned, clicks work).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Alek99:
Agreed it's a hack — it reaches into a versioned React internal, which is why it's guarded to a no-op if the internal moves, browser-only, and behind the chicken switch.

On `--env preview` for the flexgen sandbox: worth benchmarking, and the trade would look roughly like this based on the numbers gathered for this PR. An unoptimized prod build removes the *entire* dev-React overhead (the docs app measured 62.6 ms/navigation on a prod build vs 82.8 ms with this pin — the pin recovers most but not all of the gap), but you pay a Vite build per iteration instead of HMR: the docs app's production frontend build step was ~47 s, and even an unminified build keeps the Rollup bundle+chunk pipeline, whereas the dev server's incremental rebuild after an edit is effectively instant. So preview mode likely wins for flexgen's "generate → screenshot" flow if generations are infrequent relative to views, while interactive editing keeps wanting dev mode + this pin. Happy to run that benchmark properly (build time at various app sizes vs per-navigation savings) if useful — the harness from this PR mostly does it already.

Alek99:
### The `integration-app-harness` "scripts not loaded" failures were real — root-caused and fixed in `aecdaa62`

I stopped treating this as a rerun-able flake when it failed twice in a row here. Summary of the investigation:

**Symptom** (captured from a live failing state): the app hydrates and is interactive, but the DOM contains *no* `rx.script` tags at all — no helmet-attributed tags, both the inline and external script globals undefined. So it was never a fetch/timing problem with the external file; the tags were simply never injected.

**Cause**: `rx.script` wraps its tag in react-helmet, whose client-side flush is deferred and batched via `requestAnimationFrame` by default. That flush can be lost around hydration, dropping the page's head updates entirely.

**Attribution** (interleaved runs of `test_call_script[external]` on one machine):

| configuration | failures |
|---|---|
| deferred helmet, pin disabled (= main's code path) | 1/5 |
| deferred helmet, pin enabled | 3/5 |
| `defer={false}` helmet, pin enabled | **0/5** |

So the race predates this PR — it also failed on main at `7e57be74` in CI — but this PR's faster dev renders shift the timing and make it fire noticeably more often. That's an interaction I didn't catch in the earlier regression pass, and it's why I'm folding the fix in here rather than leaving the amplification behind.

**Fix**: pass `defer={false}` on `rx.script`'s Helmet wrapper only, so the head update is part of the synchronous commit. Other Helmet usage keeps the default batching. Unit test pins the rendered prop; news entries added for the bugfix.

FarhanAliRaza (approved):

PR #6916, opened 2026-08-19
ci(changelog): re-run the check when PR labels change

FarhanAliRaza:
The \`changelog\` workflow gates on the \`skip-changelog\` / \`changelog-version-edit\` labels, but it only ran on the default \`pull_request\` types (\`opened\`, \`synchronize\`, \`reopened\`). Adding a label after the last push never starts a new run, and re-running the failed run replays the original event payload where the label does not exist yet, so the check keeps failing until someone pushes again (see run 32273787477 on #6866).

Add \`labeled\` and \`unlabeled\` to the trigger types so toggling a label re-evaluates the check with the current label set.

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/6916?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -11,6 +11,16 @@ concurrency:
 on:
   pull_request:
     branches: ["main"]
+    # labeled/unlabeled: skip-changelog and changelog-version-edit waive parts
+    # of this check, so the verdict is only valid for the label set it was
+    # computed under. Without these types, removing a label leaves the green
+    # run that the label produced standing, and the gate silently stays open.
+    #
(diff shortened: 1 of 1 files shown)


masenf:
Reviewbot comment seems legit. Can you also apply this change to the template CI workflow that's part of reflex-release

masenf:
ported into #6941

PR #6933, opened 2026-08-23
fix(config): load rxconfig with cwd prepended

benedikt-bartscher:
<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/6933?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/reflex-base/src/reflex_base/config.py
```diff
@@ -874,28 +904,27 @@ def get_state_auto_setters() -> bool:
 def _load_config() -> Config:
     """Load the config from rxconfig.py with cwd on sys.path.
 
+    The cwd is prepended (not swapped in) so rxconfig resolves from the app
+    directory first while other threads keep a working import path: clearing
+    sys.path here made concurrent first-time imports elsewhere fail with
+    ModuleNotFoundError for the duration of the rxconfig import.
```
abulvenz: This comment describes the historic version that is now gone.

abulvenz (approved):
Thanks, nice finding. I would improve on the in-code-comment as annotated. In the tests those comments can help to understand why that test is needed.

masenf:
Thanks for digging into this @benedikt-bartscher — the `sys.path` half is a clear win. Clearing `sys.path` out from under every other thread was always going to bite us eventually, and prepend + remove-by-identity is the right shape. Nice bonus that it lets the `except Exception: retry` fallback go away; that only ever existed because the path was nuked in the first place.

Where I'm less sure is `_ImportRecorder`. Leaving a finder in `sys.meta_path[0]` permanently means every not-yet-cached import in the process pays a Python-level call for the rest of its life, in any process that ever loads a Reflex config — including someone who just did `import reflex` in a notebook. Your reasoning for why it can't be removed is right (`_find_spec` iterates `sys.meta_path` unlocked, so pulling it out mid-iteration can skip a real finder), but to me that reads more as an argument against the hook than for making it permanent.

And what it buys us seems fairly small? Worst case with the plain `sys.modules` diff is that a module another thread happened to import concurrently gets misattributed and evicted from `sys.modules` on the next load. The thread holding the reference is unaffected — it just means a redundant re-import later. That doesn't feel like it's worth a process-wide import hook.

My inclination would be to land the `sys.path` fix on its own, since that's where the actual user-visible bug is, and either drop the recorder or handle the misattribution some cheaper way. Happy to be talked out of it if you've hit a case where the eviction actually breaks something — you've clearly spent more time in here than I have.

Couple of smaller things while I'm looking:

- Popping `rxconfig` from `sys.modules` before the `find_spec` probe: the motivation makes sense, but it's asymmetric now. `_config_module_deps` is still only cleared *after* the probe, so leaving a project directory drops `rxconfig` but strands the old project's deps in both `sys.modules` and the dep set until some later load. I'd rather both happened on the same side of the probe.
- The new tests pop `colorsys` / `side_module` out of `sys.modules` and never put them back, and the cleanup at the end of `test_concurrent_import_not_recorded_as_rxconfig_dep` isn't in a `finally`, so a failing assert leaks state into whatever runs next. `monkeypatch.delitem(..., raising=False)` would cover both.
- If the recorder does stay: check `self._thread is not None` before calling `threading.get_ident()`, so the inactive path is a single attribute load rather than a syscall-ish call on every import.

---

benedikt-bartscher:
@masenf good push — you're right that "make it permanent" is a bad trade, but I think there's a third option I missed.

The hazard is in-place *mutation*, not removal. `_find_spec` does `meta_path = sys.meta_path` and iterates *that object*, so `.remove()` shifts the list under a concurrent iterator — but rebinding `sys.meta_path = [f for f in sys.meta_path if f is not recorder]` leaves the concurrent iterator holding the old list, where the recorder is still present and still returns `None`. Nothing gets skipped. CPython only started copying `meta_path` in 3.14 (gh-130094); I checked 3.10–3.13 and they all iterate the live list, so this matters for most of our users. Filtering instead of `remove()` also covers the rxconfig-rebuilt-`meta_path` case that the `ValueError` note was about.

So the recorder is now created per load and installed/removed by a pair of rebinds — nothing is left in `sys.meta_path` afterwards. No process-wide cost for someone who just did `import reflex` in a notebook, and your third point about `get_ident()` goes away with it.

On what it buys: I'd gently push back on "redundant re-import". Eviction is filtered to project-local modules, so what gets misattributed is an *app* module, and re-importing one of those isn't free:

```
StateValueError: The substate class 'appmod____my_state' has been defined multiple times.
Shadowing substate classes is not allowed.
```

(the shadowing check in `State.__init_subclass__` fires on the second class object). I haven't caught this in the wild and the window is narrow, so I'm not claiming it's common — but the failure mode is a crash rather than a redundant import, which is why I wanted attribution to be right. With the hook now temporary the cost side is small enough that I'd rather keep it.

Other two are fixed: dep eviction moved above the `find_spec` probe so both happen on the same side of it, and the tests now use `monkeypatch.delitem(..., raising=False)` plus a fixture whose teardown clears `rxconfig` / `side_module` / `_config_module_deps`, so a failing assert can't leak state.

benedikt-bartscher:
Valid, and it invalidates what I said in my last comment — @masenf, flagging so you don't act on the wrong version.

I claimed rebinding `sys.meta_path` made removal safe. It fixes the *skip* hazard, but it introduces a lost-update: anything another thread inserts in place between the comprehension and the assignment is dropped. That is not hypothetical here — `reflex/components/__init__.py` does `sys.meta_path.insert(0, _ComponentsRedirect())` at import time, `reflex.components` is lazily imported (a plain `import reflex` does not pull it in), and the install is guarded by a one-shot `if not any(isinstance(f, _ComponentsRedirect) ...)` that never re-runs once the module is in `sys.modules`. So losing that insertion is permanent:

```
>>> # drop the redirect, then re-import reflex.components (already cached, guard never re-runs)
ModuleNotFoundError: No module named 'reflex.components.radix'
```

A permanent ModuleNotFoundError for every `reflex.components.*` import is strictly worse than the misattributed eviction the recorder exists to prevent. Both ways out of `sys.meta_path` are unsafe — in-place removal can make a concurrent lookup skip a real finder on 3.10–3.13, rebinding drops concurrent insertions — so the recorder goes back to being installed in place and left there, which is the one option that mutates nothing others can observe as missing.

Worth noting the cost is the same shape as something already shipped: `_ComponentsRedirect` is itself a permanent `sys.meta_path[0]` finder doing a `split(".")` on every import miss. Our recorder is one attribute load plus an `is not None` when inactive, and it is only installed the first time a config actually loads, so a notebook doing `import reflex` never gets it.

Also took @masenf's third point while I was in there — `self._thread is not None` short-circuits ahead of `threading.get_ident()`, so the inactive path makes no call at all.

Regression test pins the invariant directly (`sys.meta_path` object identity is preserved across a recording window); it fails against the rebind version.

packages/reflex-base/src/reflex_base/config.py
```diff
@@ -808,40 +809,101 @@ def _set_persistent(self, **kwargs):
...
+    if _import_recorder not in sys.meta_path:
+        sys.meta_path.insert(0, _import_recorder)
+    _import_recorder.start()
+    try:
+        yield _import_recorder
+    finally:
+        _import_recorder.stop()
+
+
 def _get_config() -> Config:
```
masenf: take `project_root` as an argument to eliminate `Path.cwd()` drift between `_load_config` adding the entry and `_get_config` clean up -- `rxconfig.py` could run code that changes the cwd

(pushing a fix)

masenf (approved):

PR #6935, opened 2026-08-24
Fix client state setter event var args

### LINKED ISSUE #5741: Invalid JavaScript Code Generation with EventHandler args_spec Using rx.Var Expressions

wantosure:
Fixes #5741.

`ClientStateVar.set_value()` was using the rendered value expression to build the wrapper function argument. That works for plain event vars, but breaks for optional-chain expressions like `_event?.clientX` because the generated handler ends up with `_event?.clientX` as the argument name:

```js
((_event?.clientX) => ...)
```

This patch keeps the setter call value as-is, but derives the wrapper argument from the leading event identifier instead, so the generated function is valid:

```js
((_event) => ...(_event?.clientX))
```

Added a regression test for the optional-chain client-state setter case.

Tested with:

```text
python -m pytest tests\units\test_event.py -q
python -m pytest tests\units\test_event.py::test_call_function_client_state_setter_accepts_optional_chain_var tests\units\compiler\test_memoize_plugin.py::test_client_state_setter_in_call_function_event_imports_refs -q
python -m ruff check reflex\experimental\client_state.py tests\units\test_event.py
```

Pytest warned that it could not write `.pytest_cache` in my Windows workspace, but the tests passed.

### DIFF
--- a/reflex/experimental/client_state.py
+++ b/reflex/experimental/client_state.py
@@ -22,6 +22,12 @@ _refs_import = {
 }
 
 
+def _event_arg_name(value_str: str) -> tuple[str, ...]:
+    """Return the event argument name for an event-derived expression."""
+    match = re.match(r"^([A-Za-z_$][0-9A-Za-z_$]*)", value_str)
+    return (match.group(1),) if match else ()
+
+
 def _client_state_ref(var_name: str) -> Var:
     """Get the ref accessor Var for a ClientStateVar.
 
@@ -194,7 +200,7 @@ class ClientStateVar(Var):
 
     @property
     def value(self) -> Var:
-        """Get a placeholder for the Var.
+        """A placeholder for the Var.
 
         This property can only be rendered on the frontend.
 
@@ -235,8 +241,7 @@ class ClientStateVar(Var):
             value_str = str(value_var)
 
             setter = ArgsFunctionOperationBuilder.create(
-                # remove patterns of ["*"] from the value_str using regex
-                args_names=(re.sub(r"(\?\.)?\[\".*\"\]", "", value_str),)
+                args_names=_event_arg_name(value_str)
                 if value_str.startswith("_")
                 else (),
                 return_expr=setter.call(value_var),
--- a/reflex/experimental/client_state.py
+++ b/reflex/experimental/client_state.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import dataclasses
-import re
+import unicodedata
 from collections.abc import Callable
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


wantosure:
Thanks, this was a valid catch.

The old regex only handled ASCII identifier characters, so an event expression such as `_événement?.clientX` would bind `_` as the callback arg and then reference `_événement` inside the body.

I replaced that with a small Unicode-aware JS identifier scanner and added a regression test for the non-ASCII event arg case.

Verified with:
- `python -m pytest tests\units\test_event.py -q`
- `python -m ruff check reflex\experimental\client_state.py tests\units\test_event.py`

PR #6945, opened 2026-08-25
Add displayName to React components for DevTools visibility

masenf:
### Type of change

### Description

This PR adds `displayName` properties to all compiled React components and contexts, improving visibility in React DevTools. Without explicit display names, DevTools shows generic labels like `Context.Provider`, `Anonymous`, or `Component` for every instance in the tree, making it difficult to debug component hierarchies.

**Changes:**

1. **Context template** (`context_template`): Added `displayName` assignments for:
   - Built-in contexts: `ColorModeContext`, `UploadFilesContext`, `DispatchContext`, `EventLoopContext`
   - State contexts: Named as `StateContext(python.state.name)` to identify which Python state they carry
   - `ClientSide` wrapper: Now a named function component with display name reflecting the wrapped component

2. **Page template** (`page_template`): Every page compiles to a component named `Component`, so the route is now carried in its `displayName` (e.g., `Component(about)`, `Component(test/[dynamic])`) to distinguish pages in DevTools

3. **Memo components** (`_render_memo_component`): Added `displayName` assignment to exported memoized components, using either:
   - The Python function name for `@rx.memo` decorated components
   - The wrapped Python class name for auto-memoized components (instead of the hash-suffixed tag)

4. **MemoComponentDefinition**: Added optional `display_name` field to carry the name React DevTools should show

5. **Dynamic imports**: Updated `ClientSide()` calls to pass the component name as a second argument, enabling the wrapper to set an appropriate display name

6. **Theme context** (`react-theme.js`): Added `displayName` to the theme context

### Test coverage

- Added 5 new unit tests in `test_compiler.py` covering context display names, page display names, and the `ClientSide` wrapper naming
- Added 3 new unit tests in `test_memo.py` covering memo display names, custom wrappers, and string escaping
- Added 1 new unit test in `test_memoize_plugin.py` covering auto-memo display names
- All existing tests pass

### Checklist

- [x] Tests pass with adequate coverage
- [x] `uv run ruff check .` and `uv run ruff format .` clean
- [x] `uv run pyright reflex tests` passes
- [x] `pyi_hashes.json` updated
- [x] News fragment added

https://claude.ai/code/session_01W4Psp6qZsEpbuRePuRVkYX

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/6945?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/reflex-base/src/reflex_base/compiler/templates.py
```diff
@@ -808,7 +841,12 @@ def _render_memo_component(component: dict[str, Any]) -> str:
     if wrapper and not _MEMO_WRAPPER_CALLEE_RE.fullmatch(wrapper):
         wrapper = f"({wrapper})"
     export_expr = f"{wrapper}{function_expr}" if wrapper else function_expr
-    return f"\nexport const {component['name']} = {export_expr};\n"
+    name = component["name"]
+    display_name = json.dumps(component.get("display_name") or name)
+    return (
+        f"\nexport const {name} = {export_expr};\n"
+        f"{name}.displayName = {display_name};\n"
```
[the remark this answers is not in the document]
masenf: Not changing this. The failure mode isn't reachable on any wrapper the compiler emits, and I'd rather not put a defensive branch in every generated memo module for it.

The three cases:

- **default `memo`** — React returns a plain extensible object. In dev builds it goes further and installs a `displayName` accessor *with a setter*, precisely so this assignment works (it also forwards the name to the inner function). Checked against react 19.2.8 under both `NODE_ENV=development` and `production`: `Object.isExtensible(memo(f)) === true` and the assignment succeeds in both.
- **`wrapper=None`** — a function, always extensible.
- **`wrapper=<custom>`** — arbitrary in principle, but assigning `displayName` to a component type is the documented React convention; `forwardRef` results behave the same as `memo`, and I'm not aware of a library that freezes a component type.

If it ever does come up, the right fix isn't a guard at the assignment — it's to name the inner arrow before wrapping it (`getComponentNameFromType` falls back to `type.displayName` for a memo whose outer name is unset), which mutates only an object we created.

---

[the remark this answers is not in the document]
masenf: Declining this suggestion — the guard doesn't actually make the assignment safe, and the case it targets isn't reachable from the wrappers the compiler emits.

`Object.isExtensible` only rules out frozen/sealed objects. An extensible object carrying a non-writable `displayName` passes the guard and still throws:

```js
const wrapped = {};
Object.defineProperty(wrapped, "displayName", { value: "locked", writable: false });
Object.isExtensible(wrapped);                                  // true
if (Object.isExtensible(wrapped)) wrapped.displayName = "Foo"; // TypeError: Cannot assign to read only property
```

So it would add a branch to every generated memo module while leaving a hole. On the underlying concern: with the default `memo` wrapper React returns an extensible object, and in dev builds it installs a `displayName` accessor *with a setter* specifically for this assignment (verified on react 19.2.8 in both dev and prod). `wrapper=None` emits a function, which is always extensible. A custom `wrapper=` is arbitrary, but naming a component type via `displayName` is the documented React convention and no library I know of freezes one.

If a real wrapper ever trips this, the fix is to name the inner arrow before wrapping it — React falls back to `type.displayName` for a memo with no outer name — so we'd only ever mutate an object we created.

---

packages/reflex-base/src/reflex_base/compiler/templates.py
```diff
@@ -512,15 +531,30 @@ def page_template(
     custom_codes: Iterable[str],
     hooks: dict[str, VarData | None],
     render: dict[str, Any],
+    route: str,
```
FarhanAliRaza: **Breaking for out-of-tree callers.** `page_template` is a public symbol in the shipped `reflex-base` package, and `route` is required, so any downstream compiler plugin that calls it now raises `TypeError`. Every in-repo caller is updated, so this costs nothing to fix:

```suggestion
    route: str = "",
```

With an empty default the page simply gets `Component()` as its display name, which is no worse than the pre-PR behaviour of no name at all.

reflex/compiler/utils.py
```diff
@@ -457,6 +457,7 @@ def compile_experimental_component_memo(
             "name": memo_paths.library_and_symbol(
                 definition.source_module, definition.export_name
             )[1],
+            "display_name": definition.display_name or definition.export_name,
```
FarhanAliRaza, line 460: The `or definition.export_name` fallback is applied twice. `_render_memo_component` already computes `component.get("display_name") or name`, so the template can never observe a missing or empty value from here. Pick one place to own the fallback — CLAUDE.md asks not to repeat validation the upstream layer already did.

Either pass `definition.display_name` straight through and let the template resolve it, or keep this one and simplify the template side to a plain `component["display_name"]`.

packages/reflex-base/src/reflex_base/components/component.py
```diff
@@ -2339,11 +2339,12 @@ def _get_dynamic_imports(self) -> str:
...
             else ".then((mod) => mod.default.default ?? mod.default)"
         )
+        name = self.alias or self.tag
         return (
-            f"const {self.alias or self.tag} = ClientSide(() => "
+            f"const {name} = ClientSide(() => "
             + library_import
             + mod_import
-            + ")"
+            + f', "{name}")'
```
FarhanAliRaza: The name is interpolated into a JS string literal without escaping, while the memo path uses `json.dumps` and even carries a dedicated test for it (`test_component_memo_display_name_is_escaped`). `alias`/`tag` are identifier-shaped in practice, so this is safe today, but the asymmetry is free to remove:

```suggestion
            + f", {json.dumps(name)})"
```

packages/reflex-components-plotly/src/reflex_components_plotly/plotly.py
```diff
@@ -376,7 +376,7 @@ def dynamic_plotly_import(name: str, package: str) -> str:
     return f"""
 const {name} = ClientSide(() =>
     {library_import}{mod_import}
-)
+, "{name}")
```
FarhanAliRaza: Same as the `component.py` case: raw interpolation into a JS string literal where the memo path uses `json.dumps`. Worth making the two paths consistent.

FarhanAliRaza (changes requested):
## Reviewed by running it, not just reading it

I checked this PR out and exercised it in a real app under `examples/` — three routes (`/`, `/about`, `/blog/[slug]`), a root state plus a substate, an `@rx.memo` component, auto-memoized components, and `rx.plotly` for the `ClientSide` path. I then read the live React fiber tree, so the names below are what React DevTools actually resolves, not what the template emits.

### Verified working

- **Pages:** `Component(index)`, `Component(about)`, `Component(blog/[slug])`, `Component(404)`, all present *under* `WithComponentProps`. This is the exact failure mode the new docstring warns about, and it holds in dev **and** in a production `reflex export` build that I served and loaded.
- **Contexts:** `ColorModeContext`, `DispatchContext`, `EventLoopContext`, `ThemeContext`, and one `StateContext(...)` per state, substates and internal states included. Every context still unnamed in the tree comes from react-router or Radix, not from Reflex.
- **Memos:** `memo(NamedWidget)` for `@rx.memo`; `memo(Div)`, `memo(Button)`, `memo(Plotly)`, `memo(ErrorBoundary)` for auto-memos. The hash-suffixed tags no longer surface.
- **Client-only:** `ClientSide(Plot)`, and the lazily loaded plotly component still mounts and renders.
- **Regressions:** event round-trip works, the dynamic route arg resolves, the production build succeeds, the prod route module loads with no `ReferenceError`, and the display-name strings survive minification. No console errors or warnings on any page. `ruff`, `pyright`, and the 280 tests in the three touched test files pass.

### Why changes are requested

Nothing here breaks the feature. The blocking item is (1): `page_template` is a shipped symbol in `reflex-base`, and the new parameter is required, so an out-of-tree compiler plugin that calls it raises `TypeError`. Items (2) and (3) are consistency cleanups worth folding into the same pass.

One informational note, no change needed: the index page reads `Component(index)` and the error page reads `Component(404)`, because the value passed is the internal route key rather than the URL. That matches `get_page_path`, so it is self-consistent — just be aware the label is not always the URL.

packages/reflex-base/src/reflex_base/compiler/templates.py
masenf: Good catch — fixed in cd0963c6. `route` defaults to `""` again.

One deviation from the suggestion: with an empty route the `displayName` assignment is skipped entirely rather than emitting `Component.displayName = "Component()"`. An out-of-tree caller then gets byte-for-byte the pre-PR output (the function name `Component` alone) instead of a contentless label in their DevTools tree. Covered by `test_page_template_without_a_route_omits_the_display_name`.

---

reflex/compiler/utils.py
masenf: Fixed in cd0963c6, taking your second option — `compile_experimental_component_memo` keeps the fallback, the template now reads `component["display_name"]` plainly.

Worth recording why it has to be that direction: the two fallbacks were not actually equivalent. This one resolves to `definition.export_name` (`NamedWidget`), while the template's resolved to `name`, the JS symbol — which carries the per-module hash (`NamedWidget_fe2c0c45`). Every `@rx.memo` component has `display_name is None` (only auto-memos set it), so the template's fallback was dead code that would have emitted a hashed label if it ever fired. Deleting the wrong one and keeping the template's turned `memo(NamedWidget)` into `memo(NamedWidget_fe2c0c45)` — two existing tests caught it.

---

packages/reflex-base/src/reflex_base/components/component.py
masenf: Applied in cd0963c6 — `json.dumps(name)` here, with `import json` added to the module. Agreed the asymmetry was free to remove.

---

packages/reflex-components-plotly/src/reflex_components_plotly/plotly.py
masenf: Same fix applied here in cd0963c6 — `json.dumps(name)` in `dynamic_plotly_import`, so both `ClientSide` call sites now match the memo path.

---

FarhanAliRaza (approved):

PR #6968, opened 2026-08-28
Add click_element utility for robust element clicking in tests

masenf:
### Type of change

### Description

Fixes a `StaleElementReferenceException` flake in `test_on_load_navigate_non_dynamic[prod]`, then applies the same fix to the other Selenium integration tests with the same shape.

**Problem:** `poll_for_navigation` returns as soon as `driver.current_url` changes, which happens when the router pushes history — before React swaps the route component. An element located in that window belongs to the outgoing tree and is unmounted before the click is dispatched. In the observed failure, `/` and `/static/x` render the same component, so `link_page_x` existed on both pages and was located on the page being navigated away from. Prod mode widens the window because the route chunk is fetched lazily.

**Solution:** `click_element(driver, by, value, timeout)` re-locates the element on every attempt, so it clicks whichever node is currently rendered.

Retries are confined to `NoSuchElementException` and `StaleElementReferenceException`. Both are raised while locating or validating the element reference, never after the click reaches the browser, so a click cannot be dispatched twice — which matters because these tests assert exact event counts. Everything else propagates on the first attempt, including `InvalidSelectorException` (it derives from `WebDriverException`, not `NoSuchElementException`) and `ElementClickInterceptedException`.

### Changes

1. **`tests/integration/utils.py`** — add `click_element()`: a deadline loop that re-locates and clicks, raising `TimeoutError` with the locator and the last error chained as `__cause__`.
2. **`tests/integration/test_dynamic_routes.py`** — all link clicks go through the helper: the `link_page_x`/`link_index` clicks in `test_on_load_navigate_non_dynamic` (the flaking test), the `link_page_next` loop and post-redirect click plus `link_missing` in `test_on_load_navigate`, and both `next-page` clicks in `test_render_dynamic_arg`. Cached element variables and their presence polls drop out, since the helper re-locates.
3. **`tests/integration/test_navigation.py`** — `external`/`external2`, located immediately after `driver.back()` and clicked later with a window switch in between; `internal` for consistency.
4. **`tests/integration/test_login_flow.py`** — `doit`, located as soon as the URL became `/login`; `login` and `logout` in the same flow.
5. **`tests/integration/test_event_chain.py`** — `unmount`, located right after `driver.get()` and clicked once `assert_token()` has waited out hydration (where the StrictMode remount lands).

Deliberately unchanged: single-expression `find_element(...).click()` calls, which have no window between locating and clicking (`test_auto_memo.py`, `test_deploy_url.py`), and clicks that never cross a navigation (`test_client_storage.py`, `test_connection_banner.py`, `test_upload.py` — those already re-locate after each refresh).

### Follow-up work (not in this PR)

`test_dynamic_routes.py` carries two `TODO: drop after flakiness is resolved` band-aids — `driver.implicitly_wait(30)` and an `await asyncio.sleep(3)` in `test_render_dynamic_arg`. Removing the implicit wait means converting every remaining bare `find_element` in that module to a polling lookup, and neither removal can be demonstrated except by repeated CI runs, so both are left for a later change.

### Testing

- `uv run ruff check .`, `uv run ruff format .` and `uv run pyright reflex tests` are clean.
- No new tests: this is a fix to test infrastructure, and the flake it removes is timing-dependent, so CI is the verification — `integration-app-harness` (Selenium, dev + prod) and `integration-tests` (Playwright) are green on the head commit.

### Checklist

- [x] Changes follow the guidelines in CONTRIBUTING.md
- [x] Code is linted and formatted
- [x] Docstring follows Google style with Args/Raises sections
- [x] Existing tests pass with the changes

https://claude.ai/code/session_01W7hxgQtxcFzZC795xyftVS

### DIFF
--- a/tests/integration/test_dynamic_routes.py
+++ b/tests/integration/test_dynamic_routes.py
@@ -235,24 +235,18 @@ def test_on_load_navigate(
         token: The token visible in the driver browser.
     """
     assert dynamic_route.app_instance is not None
-    link = driver.find_element(By.ID, "link_page_next")
-    assert link
 
     exp_order = [f"/page/[page_id]-{ix}" for ix in range(10)]
     # click the link a few times
     for ix in range(10):
         # wait for navigation, then assert on url
         with poll_for_navigation(driver):
-            link.click()
+            click_element(driver, By.ID, "link_page_next")
         assert urlsplit(driver.current_url).path == f"/page/{ix}"
 
-        link = AppHarness.poll_for_or_raise_timeout(
-            lambda: driver.find_element(By.ID, "link_page_next")
-        )
         page_id_input = driver.find_element(By.ID, "page_id")
         raw_path_input = driver.find_element(By.ID, "raw_path")
 
-        assert link
         assert page_id_input
 
         assert dynamic_route.poll_for_value(
@@ -273,9 +267,8 @@ def test_on_load_navigate(
 
     # make sure internal nav still hydrates after redirect
     exp_order += ["/page/[page_id]-11"]
-    link = driver.find_element(By.ID, "link_page_next")
     with poll_for_navigation(driver):
-        link.click()
+        click_element(driver, By.ID, "link_page_next")
     poll_assert_event_order(driver, exp_order)
 
     # load same page with a query param and make sure it passes through
@@ -393,19 +386,15 @@ async def test_render_dynamic_arg(
             )
 
     assert_content("0", "")
-    next_page_link = driver.find_element(By.ID, "next-page")
-    assert next_page_link
     with poll_for_navigation(driver):
-        next_page_link.click()
+        click_element(driver, By.ID, "next-page")
     assert (
         driver.current_url.removesuffix("/")
         == f"{frontend_url.removesuffix('/')}/arg/1"
     )
     assert_content("1", "0")
-    next_page_link = driver.find_element(By.ID, "next-page")
-    assert next_page_link
     with poll_for_navigation(driver):
-        next_page_link.click()
+        click_element(driver, By.ID, "next-page")
     assert (
         driver.current_url.removesuffix("/")
         == f"{frontend_url.removesuffix('/')}/arg/2"
--- a/tests/integration/test_dynamic_routes.py
+++ b/tests/integration/test_dynamic_routes.py
@@ -12,7 +12,7 @@ from selenium.webdriver.common.by import By
 
 from reflex.testing import AppHarness, WebDriver
 
-from .utils import poll_assert_event_order, poll_for_navigation
+from .utils import click_element, poll_assert_event_order, poll_for_navigation
 
 
 def DynamicRoute():
@@ -300,9 +300,8 @@ def test_on_load_navigate(
 
     # next/link to a 404 and ensure we still hydrate
     exp_order += ["/404-no page id"]
-    link = driver.find_element(By.ID, "link_missing")
     with poll_for_navigation(driver):
-        link.click()
+        click_element(driver, By.ID, "link_missing")
 
     # hit a page that redirects back to dynamic page
     exp_order += ["on_load_redir-{'foo': 'bar', 'page_id': '0'}", "/page/[page_id]-0"]
@@ -330,29 +329,24 @@ def test_on_load_navigate_non_dynamic(
         driver: WebDriver instance.
     """
     assert dynamic_route.app_instance is not None
-    link = driver.find_element(By.ID, "link_page_x")
-    assert link
 
     with poll_for_navigation(driver):
-        link.click()
+        click_element(driver, By.ID, "link_page_x")
     assert urlsplit(driver.current_url).path.removesuffix("/") == "/static/x"
     poll_assert_event_order(driver, ["/static/x-no page id"])
 
     # go back to the index and navigate back to the static route
-    link = driver.find_element(By.ID, "link_index")
     with poll_for_navigation(driver):
-        link.click()
+        click_element(driver, By.ID, "link_index")
     assert urlsplit(driver.current_url).path.removesuffix("/") == ""
 
-    link = driver.find_element(By.ID, "link_page_x")
     with poll_for_navigation(driver):
-        link.click()
+        click_element(driver, By.ID, "link_page_x")
     assert urlsplit(driver.current_url).path.removesuffix("/") == "/static/x"
     poll_assert_event_order(driver, ["/static/x-no page id", "/static/x-no page id"])
 
     for _ in range(3):
-        link = driver.find_element(By.ID, "link_page_x")
-        link.click()
+        click_element(driver, By.ID, "link_page_x")
         assert urlsplit(driver.current_url).path.removesuffix("/") == "/static/x"
     poll_assert_event_order(driver, ["/static/x-no page id"] * 5)
 
--- a/tests/integration/utils.py
+++ b/tests/integration/utils.py
@@ -8,7 +8,10 @@ from contextlib import contextmanager
 from http.client import HTTPConnection
 from urllib.parse import urlsplit
 
-from selenium.common.exceptions import StaleElementReferenceException
+from selenium.common.exceptions import (
+    NoSuchElementException,
+    StaleElementReferenceException,
+)
 from selenium.webdriver.common.by import By
 from selenium.webdriver.remote.webdriver import WebDriver
 
@@ -72,21 +75,23 @@ def click_element(
     value: str,
     timeout: TimeoutType = None,
 ) -> None:
-    """Locate an element and click it, re-locating it if it goes stale.
+    """Locate an element and click it, tolerating the churn of a navigation.
 
     Client-side navigation swaps the DOM after the URL changes, so an element
-    located right after navigating can be unmounted before the click is
-    dispatched. Only a stale reference is retried: it is raised before the
-    click reaches the browser, so the click is never dispatched twice.
+    located right after navigating may not be rendered yet, or may be unmounted
+    before the click is dispatched. Both are retried until `timeout`, since both
+    are raised while locating or validating the element reference, never after
+    the click reaches the browser. Every other error, including an invalid
+    selector or an intercepted click, propagates on the first attempt.
 
     Args:
         driver: WebDriver instance.
         by: Locator strategy, one of the `By` constants.
         value: Locator value.
-        timeout: How long to keep re-locating a stale element.
+        timeout: How long to keep re-locating the element.
 
     Raises:
-        TimeoutError: if the element remained stale for the whole timeout.
+        TimeoutError: if the element could not be clicked within the timeout.
     """
     deadline = time.monotonic() + (
         DEFAULT_TIMEOUT if timeout is None else float(timeout)
@@ -94,9 +99,9 @@ def click_element(
     while True:
         try:
             driver.find_element(by, value).click()
-        except StaleElementReferenceException as exc:
+        except (NoSuchElementException, StaleElementReferenceException) as exc:
             if time.monotonic() >= deadline:
-                msg = f"Element {by}={value!r} remained stale while polling."
+                msg = f"Could not click element {by}={value!r} while polling."
                 raise TimeoutError(msg) from exc
             time.sleep(POLL_INTERVAL)
         else:
--- a/tests/integration/utils.py
+++ b/tests/integration/utils.py
@@ -2,15 +2,17 @@
 
 from __future__ import annotations
 
+import time
 from collections.abc import Generator, Iterator, Sequence
 from contextlib import contextmanager
 from http.client import HTTPConnection
 from urllib.parse import urlsplit
 
+from selenium.common.exceptions import StaleElementReferenceException
 from selenium.webdriver.common.by import By
 from selenium.webdriver.remote.webdriver import WebDriver
 
-from reflex.testing import AppHarness, TimeoutType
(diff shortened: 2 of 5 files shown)


tests/integration/utils.py
```diff
@@ -64,6 +64,43 @@ def poll_for_navigation(
...
+    Raises:
+        TimeoutError: if the element could not be clicked within the timeout.
+    """
+    last_exc: Exception | None = None
+
+    def _click() -> bool:
+        nonlocal last_exc
+        try:
+            driver.find_element(by, value).click()
+        except Exception as exc:
```
FarhanAliRaza: Narrow this to the exceptions that are actually retryable, for example `(StaleElementReferenceException, NoSuchElementException)`.

`except Exception` makes every non-retryable failure wait the full `DEFAULT_TIMEOUT` and then report `TimeoutError`. A wrong locator, a wrong `By` constant, or a dead driver no longer fails fast, and the original exception type is lost. Selenium also offers `WebDriverWait(..., ignored_exceptions=...)` for this.

tests/integration/utils.py
```diff
@@ -64,6 +64,43 @@ def poll_for_navigation(
...
+
+    Raises:
+        TimeoutError: if the element could not be clicked within the timeout.
+    """
+    last_exc: Exception | None = None
+
+    def _click() -> bool:
+        nonlocal last_exc
+        try:
+            driver.find_element(by, value).click()
```
FarhanAliRaza, line 101: A retried click is not idempotent. `click()` can raise after the browser dispatched the click, for example when the page navigates while the command runs. The retry then clicks a second time.

`test_on_load_navigate_non_dynamic` asserts an exact event count (`["/static/x-no page id"] * 5`). So a spurious retry replaces a stale-element flake with a count-mismatch flake, which is harder to diagnose.

Restricting the caught exceptions to stale and not-found removes most of this, because both raise before the click is dispatched.

tests/integration/utils.py
```diff
@@ -64,6 +64,43 @@ def poll_for_navigation(
...
+        except Exception as exc:
+            last_exc = exc
+            raise
```
FarhanAliRaza: Replace this `raise` with `return False`.

`AppHarness._poll_for` runs the target inside `contextlib.suppress(Exception)` (`reflex/testing.py:543`), so the re-raise is never observed. The `nonlocal last_exc` plus `raise` pair only exists to carry the exception out of a callable whose exceptions are already discarded. Returning `False` does the same with one branch less.

tests/integration/utils.py
```diff
@@ -64,6 +64,43 @@ def poll_for_navigation(
...
+        return True
+
+    if not AppHarness._poll_for(_click, timeout=timeout):
```
FarhanAliRaza: Use the public API instead of `AppHarness._poll_for`. `AppHarness.poll_for_or_raise_timeout` (`reflex/testing.py:765`) already polls and raises `TimeoutError`, and `poll_for_navigation` ten lines above in this same file uses public `AppHarness.expect`.

If the locator must stay in the error message, keep the explicit raise but poll through the public helper.

tests/integration/test_dynamic_routes.py
```diff
@@ -330,29 +329,24 @@ def test_on_load_navigate_non_dynamic(
...
     with poll_for_navigation(driver):
-        link.click()
+        click_element(driver, By.ID, "link_page_x")
     assert urlsplit(driver.current_url).path.removesuffix("/") == "/static/x"
     poll_assert_event_order(driver, ["/static/x-no page id", "/static/x-no page id"])
 
     for _ in range(3):
-        link = driver.find_element(By.ID, "link_page_x")
-        link.click()
+        click_element(driver, By.ID, "link_page_x")
```
FarhanAliRaza, line 342: Apply `click_element` to the remaining click sites in this file: lines 238/246, 276/278, 396/399 and 405/408.

Each one locates an element and then clicks it across a client-side navigation. That is the same shape as the line this PR fixes, so the same stale-element failure can occur there. Converting only some of them leaves the flake reachable in the tests this PR is fixing.

FarhanAliRaza (changes requested):
I ran `tests/integration/test_dynamic_routes.py` in a worktree at this branch, in two configurations.

First, a baseline. I reverted both changed files to their pre-PR content and ran the module repeatedly. The flake reproduced twice in 16 runs. Both failures were `test_on_load_navigate_non_dynamic[prod]`, and both raised `StaleElementReferenceException` from the `link.click()` that this PR replaces. This confirms the diagnosis in the description.

Second, the branch as submitted. The module passes.

I also checked the rest of the file for the same locate-then-click shape, and ran `ruff check`, `pyright`, and a McCabe complexity check at `max-complexity=4` on both changed files. All three are clean, so complexity is not a concern here.

The change fixes a real fault. My concerns are with the shape of the new helper and with how completely it is applied. See the inline comments.

tests/integration/utils.py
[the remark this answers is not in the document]
masenf: Partly addressed in 46d814f9b, partly pushing back.

The helper no longer retries lookups that can block on the implicit wait: it now catches only `StaleElementReferenceException`, so a missing element raises `NoSuchElementException` out of the first lookup instead of being re-polled. In the stale case the element does exist, so the re-lookup returns immediately and the deadline is honored.

Not toggling `driver.implicitly_wait(0)` inside the helper: the remaining worst case is a single `find_element` blocking for the implicit wait, which is what every other `find_element` in this module already does — the 30s wait at `tests/integration/test_dynamic_routes.py:194` is a deliberate (if temporary) fixture choice. Having one helper silently reset a session-wide setting would be a surprising side effect for a test util, and if it raised mid-loop the wait would be left at 0 for the rest of the session.

---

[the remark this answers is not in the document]
masenf: Same answer as the sibling thread on this line: the helper now retries only `StaleElementReferenceException`, so `find_element` is no longer re-polled for an absent element and can't stack implicit waits. Declining to toggle `implicitly_wait(0)` inside the helper — that mutates session-wide driver state for every caller, and a raise mid-loop would leave it at 0. The one remaining lookup that can block is no different from any other `find_element` in this module.

---

tests/integration/utils.py
```diff
@@ -64,6 +64,43 @@ def poll_for_navigation(
...
+        timeout: Time to wait for the click to succeed.
+
+    Raises:
+        TimeoutError: if the element could not be clicked within the timeout.
+    """
+    last_exc: Exception | None = None
+
+    def _click() -> bool:
+        nonlocal last_exc
+        try:
```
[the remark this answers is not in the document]
masenf: Agreed and fixed in 46d814f9b: the retry is now scoped to `StaleElementReferenceException` only, and everything else propagates immediately.

Not adding `NoSuchElementException` or `ElementClickInterceptedException` to that tuple: this module sets a 30s implicit wait (`test_dynamic_routes.py:194`), so a missing element already blocks that long per lookup and retrying it would multiply the cost of a genuinely bad locator; interception isn't a failure mode these link clicks have.

---

tests/integration/utils.py
masenf: Fixed in 46d814f9b — narrowed to `StaleElementReferenceException`, so a wrong locator, a wrong `By`, or a dead driver now fails fast with its original exception type.

I left `NoSuchElementException` out of the retry set on purpose: this module sets a 30s implicit wait (`test_dynamic_routes.py:194`), which already covers "not rendered yet", so retrying it would only multiply the cost of a genuinely bad locator.

On `WebDriverWait(..., ignored_exceptions=...)`: it would work, but it swallows the ignored exception and raises `TimeoutException` with no `__cause__`, and its message quotes the lambda rather than the locator. The explicit loop keeps the locator in the message and chains the last stale error via `raise ... from exc`.

---

tests/integration/utils.py
masenf: This was the strongest objection to the original version — a count-mismatch flake would indeed be worse than the stale-element one it replaced. Fixed in 46d814f9b by restricting the retry to `StaleElementReferenceException`, which chromedriver raises while validating the element reference, before the click is dispatched. Every other error, including anything raised after a click reached the browser, now propagates on the first attempt.

---

tests/integration/utils.py
masenf: Right that the re-raise was dead — and the same observation is why 46d814f9b drops `AppHarness._poll_for` here entirely rather than returning `False` into it. Blanket `contextlib.suppress(Exception)` is exactly the semantics this helper must not have: it needs one exception type retried and the rest raised. The helper is now a plain deadline loop, so `nonlocal last_exc` is gone and the last stale error is chained with `raise TimeoutError(...) from exc`.

---

tests/integration/utils.py
masenf: Addressed in 46d814f9b, though not by switching helpers: the private `_poll_for` call is gone, and so is the need for a polling helper at all.

`poll_for_or_raise_timeout` wouldn't work here — it delegates to `_poll_for`, so it inherits the `contextlib.suppress(Exception)` that makes a non-retryable error indistinguishable from a stale one and turns it into a bare `TimeoutError`. That's the behavior the other threads on this file asked to remove. `AppHarness.expect` in `poll_for_navigation` is fine by comparison because its target is a pure URL comparison that doesn't raise.

---

tests/integration/test_dynamic_routes.py
masenf: Done in 46d814f9b — all remaining find-then-click-across-navigation sites in this file now use `click_element`:

- `test_on_load_navigate`: the `link_page_next` click loop (the pre-loop lookup and the in-loop re-lookup are both gone, since the helper re-locates on each click), and the post-redirect click.
- `test_render_dynamic_arg`: both `next-page` clicks.

Kept out of scope, as follow-up work rather than in this PR:

1. The same pattern exists in other integration modules (e.g. `test_deploy_url.py`, `test_client_storage.py`); sweeping them is a bigger, unrelated diff and none of them is currently flaking on this.
2. Two band-aids in this module carry a `TODO: drop after flakiness is resolved` — `driver.implicitly_wait(30)` (line 194) and `await asyncio.sleep(3)` in `test_render_dynamic_arg`. Worth re-testing whether they can go now that the clicks re-locate, but that needs several clean CI runs to justify, so not here.

---

masenf: Correction to the file list in follow-up (1) above — I checked rather than guessed this time. The modules that actually have a locate-then-click-across-navigation window are `test_login_flow.py` (the `doit` button located right after a `poll_for_navigation`, lines 132-134) and `test_navigation.py` (`external`/`external2` located before the click that follows a navigation, lines 79-92). `test_client_storage.py` and `test_deploy_url.py` locate and click without a navigation in between, so they're not affected.

---

tests/integration/utils.py
masenf: Updating this: in a4be554d4 the helper now retries `NoSuchElementException` as well, reversing what I said above. The scope grew to the other Selenium modules with this same stale-click shape (`test_navigation.py`, `test_login_flow.py`, `test_event_chain.py`), and unlike `test_dynamic_routes.py` none of them sets an implicit wait — they got "not rendered yet" coverage from an `AppHarness.poll_for_or_raise_timeout(lambda: driver.find_element(...))` before the click, which the helper has to subsume to replace those call sites.

Your fail-fast concern still holds where it bites hardest, though: `InvalidSelectorException` derives from `WebDriverException`, not from `NoSuchElementException`, so a wrong `By` constant or a malformed selector still raises on the first attempt. A dead driver likewise. What now waits out the timeout is a locator whose element genuinely never appears — which is exactly what the presence polls it replaced already did.

Retry is still confined to errors raised while locating or validating the reference, never after the click is dispatched, so the double-click hazard stays closed.

---

tests/integration/utils.py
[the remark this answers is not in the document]
masenf: Amending this: a4be554d4 adds `NoSuchElementException` back to the retry set, so your original suggestion is now closer to what landed. Reason: the change was extended to the other Selenium modules with the same stale-click shape, and those set no implicit wait — they previously got presence coverage from a poll before the click, which the helper now subsumes.

`ElementClickInterceptedException` still propagates: it is raised after the element resolved, and none of these link/button clicks has an overlay that could intercept.

---

tests/integration/test_dynamic_routes.py
masenf: Follow-up (1) is no longer a follow-up — it's in this PR as of a4be554d4, at masenf's request to fix the class across the Selenium integration tests rather than defer it: `test_navigation.py` (`external`/`external2` after `driver.back()`), `test_login_flow.py` (`doit` after the /login navigation, plus `login`/`logout`), and `test_event_chain.py` (`unmount` after `driver.get()` + hydration wait).

Deliberately still not converted, since they aren't this class: single-expression `find_element(...).click()` calls with no window between locating and clicking (`test_auto_memo.py`, `test_deploy_url.py`), and clicks that never cross a navigation (`test_client_storage.py`, `test_connection_banner.py`, `test_upload.py` — those re-locate after each refresh already).

Follow-up (2), the `implicitly_wait(30)` and `asyncio.sleep(3)` band-aids, still stands as follow-up work. Dropping the implicit wait would mean converting every remaining bare `find_element` in `test_dynamic_routes.py` to a polling lookup, in the same file this PR is trying to stabilize, with no way to demonstrate the result short of repeated CI runs.

---

FarhanAliRaza (approved):

PR #6944, opened 2026-08-25
resolve type aliases

benedikt-bartscher:
<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/6944?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #6986, opened 2026-08-28
Resolve PEP 695 type aliases in _isinstance and typehint_issubclass

masenf:
### All Submissions:

### Type of change

### Changes To Core Features:

## Defect

Found in 0.9.9a1 pre-release testing as **FINDING-002 (HIGH)** and **FINDING-006 (MEDIUM)**. PR #6944 added `resolve_type_alias()` to `reflex_base/utils/types.py` but wired it only into `Var.guess_type`, so state vars annotated with a PEP 695 alias (`type Key = Literal["a", "b"]`, or the `typing_extensions.TypeAliasType` backport that 3.10/3.11 users hit) compiled while every runtime path that re-checks the raw annotation still choked on the `TypeAliasType`:

- **FINDING-002:** `State.__setattr__` validates assignments through `_isinstance()`, which fell through to the bare `isinstance()` call and raised `TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union` on every assignment to an alias-annotated var. The guard was designed to only `logger.error` a mismatch, but the TypeError escaped and aborted event processing — alias vars compiled but every mutating event handler died.
- **FINDING-006:** an event handler whose argument is alias-annotated, passed **uncalled** to an event trigger (`rx.input(on_change=S.choose)`), crashed page compile: `typehint_issubclass()` reached the bare `issubclass()` for a plain alias (`TypeError: Could not compare types <class 'str'> and Key`) and failed the origin comparison for a subscripted alias (`Pair[str, str]`), escalating what should at most be a warning into a fatal `EventHandlerArgTypeMismatchError`.

## Fix

Resolve `TypeAliasType` at the entry of `_isinstance` (the annotation operand) and `typehint_issubclass` (both operands), reusing the existing `resolve_type_alias()` helper, which already handles parameterized aliases (`Items[str]`) and aliases nested in unions (`Key | None`). Nested annotations (e.g. an alias inside `list[...]`) are covered by both functions' per-argument recursion resolving at each entry. Alias-annotated vars and handler arguments now behave exactly like their resolved annotations at runtime; the resolution is a no-op fast path for non-alias types. News fragment added in `packages/reflex-base/news/`.

## Test plan

Regression tests written first and shown to fail on unfixed `main` with exactly the reported errors:

- `tests/units/reflex_base/utils/test_types.py::test_isinstance_resolves_type_alias` and `::test_typehint_issubclass_resolves_type_alias` — plain alias, `Literal` alias, parameterized generic alias, and `alias | None`, parameterized over the `typing_extensions` backport and (on 3.12+) the native `typing.TypeAliasType` that the `type` statement produces.
- `tests/units/test_state.py::test_setattr_alias_annotated_var` — a state with all four alias-annotated var shapes assigns them via an event handler; also pins that a mismatched value is logged, not raised.

Verified locally: `uv run pytest tests/units/utils/test_types.py tests/units/reflex_base/utils/test_types.py tests/units/reflex_base/vars/test_base.py tests/units/test_state.py tests/units/test_event.py tests/units/components tests/units/test_app.py` (4600+ passed), `uv run ruff check .`, `uv run ruff format .`, `uv run pyright reflex tests` (0 errors), `uv run pyright packages/reflex-base` (0 errors, 1 pre-existing warning in an untouched file). Both pre-release repro scripts (`repro_alias_setattr.py`, `repro_alias_event_arg.py`, Python 3.14 native `type` statements) now pass, and a parameterized-alias `on_submit` handler behaves identically to a plain `dict[str, str]` annotation.

https://claude.ai/code/session_01EMjBXPozsNeQNSBZecNH8x

---
_Generated by [Claude Code](https://claude.ai/code/session_01EMjBXPozsNeQNSBZecNH8x)_

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/6986?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


benedikt-bartscher:
@masenf with https://github.com/reflex-dev/reflex/pull/6862 merged in this (https://github.com/reflex-dev/reflex/pull/6988), codspeed is green: [codspeed](https://app.codspeed.io/reflex-dev/reflex/branches/benedikt-bartscher%3Arel-fix-pep695-alias-runtime-perf-isinstance?utm_source=github&utm_medium=comment-v2&utm_content=button)

packages/reflex-base/src/reflex_base/utils/types.py
```diff
@@ -817,6 +817,9 @@ def _isinstance(
     Returns:
         Whether the object is an instance of the class.
     """
+    # A PEP 695 type alias is opaque to every check below, so unwrap it first.
+    cls = resolve_type_alias(cls)
```
FarhanAliRaza: `_isinstance` is a hot path (it dominates the wire/serialization benchmarks), and this call now runs on every entry, including each per-argument recursion. Measured on the branch (min of 5x100k, `nested=1`):

| annotation | `resolve_type_alias` | `_isinstance` total |
|---|---|---|
| `int` | 278ns | 1422ns |
| `list[str]` | 270ns | 1744ns |
| `str \| None` | 1747ns | 5860ns |

So ~15-20% on plain types and ~30% on `X | None` (the most common state annotation). The union cost comes from `resolve_type_alias` eagerly resolving every union member at the top, after which `_isinstance` recurses into each member and resolves it again.

Since both `_isinstance` and `typehint_issubclass` already recurse per argument, the entry only needs the top-level unwrap, not the union walk. Suggest a cheap helper for these two call sites, e.g.

```python
def _unwrap_alias(cls):
    if isinstance(cls, TypeAliasTypes) or isinstance(get_origin(cls), TypeAliasTypes):
        return resolve_type_alias(cls)
    return cls
```

and keep the full `resolve_type_alias` for `Var.guess_type`. Worth checking the CodSpeed report on this PR either way.

FarhanAliRaza (commented):
Validated end to end in a real app (Python 3.14 native `type` statements: plain, `Literal`, `Items[T]`, `Pair[K, V]`, `Key | None`): alias-annotated var assignment works, uncalled alias-annotated handlers on `on_change`/`on_submit` compile and fire, mismatches log instead of raise, plain vars unaffected. Regression tests fail without the source change. One perf concern inline.

masenf:
@benedikt-bartscher awesome thanks for that. will merge yours in and rebase this one. 0.9.9 is getting close 🤞

benedikt-bartscher:
I guess we can improve performance for the new benchmarks a bit

masenf:
> I guess we can improve performance for the new benchmarks a bit

yeah, that unconditional type alias unpacking is more expensive than i realized. probably need to cache it and see if that makes a difference

masenf:
Pushed e688d7d taking a slightly different route than caching: the unconditional `resolve_type_alias()` at the entry of `_isinstance`/`typehint_issubclass` is gone entirely — resolution now happens only at the exact points where an alias would be opaque (a zero-cost `try/except TypeError` around the final `isinstance`/`issubclass` mirroring `safe_issubclass`, a guard before union decomposition so an alias-of-a-union keeps union semantics, and the union branch of `resolve_type_alias` no longer allocates unless a member actually resolved). Local numbers vs main: `_isinstance('x', str|None)` 1.61µs → 4.78µs on the previous head → 1.60µs now; `typehint_issubclass(str, str|None)` 1.81 → 5.98 → 1.89µs; `test_process_event` mean back inside main's noise band. All alias-semantics tests still pass (plain/Literal/parameterized, alias|None, native + typing_extensions), plus new alias-of-union coverage and a trigger-path regression test for the uncalled-handler crash. CodSpeed should confirm on this head; caching remains an option if it still flags anything.

---

masenf:
i wouldn't be surprised if that last remaining codspeed issue is really just a fluke. 4% regression on _different_ hardware than the baseline, and we have our threshold set at 3%

masenf:
CodSpeed status on this PR, for whoever merges: the genuine regression is fixed and the remaining flag looks like environment noise.

- Head `7e845d2` (unconditional resolution): 8 regressed benchmarks, -33.7% — real, root-caused, fixed in `e688d7d`.
- Head `e688d7d`: only `test_isinstance_container[list_dict]` at -7.58% — also real (the per-call `isinstance(origin, TypeAliasTypes)` probe), fixed in `758bf18` by moving the check into the final fallback's `TypeError` handler, so alias-free checks pay nothing. Interleaved local A/B on that benchmark: parity with main (8.4 vs 8.45ms; previous head 9.07ms).
- Head `758bf18`: `list_dict` is green, but CodSpeed now flags a *different* benchmark, `test_get_all_imports[_complicated_page]` at -4.47%, with its own "different runtime environments detected" warning. Interleaved local timeit A/B of `page._get_all_imports()` on the same `_complicated_page` fixture shows identity between this head and main after warm-up (0.4617 vs 0.4647 ms; 0.4567 vs 0.4579 ms) — and this diff only touches `_isinstance`/`typehint_issubclass`/`resolve_type_alias`, none of which are on the import-collection path.

Each successive flag has been a different benchmark under a cross-environment comparison, so I'm not pushing further speculative changes. If you agree it's noise, acknowledging on CodSpeed (or a re-run against a same-environment baseline) should clear the check.

---

benedikt-bartscher:
> i wouldn't be surprised if that last remaining codspeed issue is really just a fluke. 4% regression on _different_ hardware than the baseline, and we have our threshold set at 3%

maybe https://github.com/reflex-dev/reflex/pull/6993 helps a tiny bit

FarhanAliRaza (approved):

PR #6994, opened 2026-08-28
Clear fatal-error interrupt window unconditionally on body exit

masenf:
### Type of change

### Changes To Core Features:

### Description

Follow-up to #6990 (merged while this review round was in flight), addressing the two review threads that were still open there:

1. **cubic's finding is valid**: `run_concurrently_context` cleared `in_body` only on the normal post-body path and in the `KeyboardInterrupt` handler. If the with-body raises any other exception, `in_body` stays `True` — and since the executor shuts down with `wait=False`, a task failing after the context has unwound could still deliver a stray SIGINT into unrelated caller code. The fix wraps the `yield` in a `try/finally` that clears `in_body` however the body exits, replacing the two scattered clears — a net simplification of the mechanism, not an addition (per the review discussion, no extra guarded setup path).
2. **greptile's nit**: the hang regression test's bare `5`-second assertion threshold now carries a comment explaining its relationship to the body's 10s wait.

Adds `test_run_concurrently_context_no_interrupt_after_body_exception`, which pins the escaped-interrupt case (late-failing task + body exception → the body's own `ValueError` propagates, no stray `KeyboardInterrupt`).

### Test Plan

- `uv run pytest tests/units/utils/test_processes.py -k run_concurrently`: 4 passed (the new test plus the three from #6990).
- `uv run ruff check` / `ruff format --check` clean on changed files; `uv run pyright` 0 errors on changed files.
- `reflex-release changelog-check --base-ref origin/main` passes.
- The two `test_is_process_on_port_*` failures in this sandbox are pre-existing on clean main here (container network stack), unrelated to the diff.

https://claude.ai/code/session_01EMjBXPozsNeQNSBZecNH8x

---
_Generated by [Claude Code](https://claude.ai/code/session_01EMjBXPozsNeQNSBZecNH8x)_

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/6994?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/news/+interrupt-window-followup.bugfix.md
+++ b/news/+interrupt-window-followup.bugfix.md
@@ -1,0 +1,1 @@
+Fixed an edge in the new fatal-error interrupt handling where a task failing after the with-body raised its own exception could deliver a stray SIGINT to unrelated caller code.
--- a/reflex/utils/processes.py
+++ b/reflex/utils/processes.py
@@ -354,11 +354,14 @@ def run_concurrently_context(
             # Don't enter (and block in) the body if a task has already failed.
             raise_first_failure(tasks)
 
-            # Yield control back to the main thread while tasks are running.
-            yield tasks
-
-            with interrupt_lock:
-                in_body = False
+            try:
+                # Yield control back to the main thread while tasks are running.
+                yield tasks
+            finally:
+                # However the body exits, stop interrupting the main thread:
+                # tasks outlive this context (shutdown below does not wait).
+                with interrupt_lock:
+                    in_body = False
 
             # Get the results in the order completed to check any exceptions.
             for task in futures.as_completed(tasks):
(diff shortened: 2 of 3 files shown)


masenf:
Note: this was opened moments before your replies landed on the #6990 threads ("if we're going down, we're going down" / "test code"), so calling out the delta explicitly: `executor.shutdown(wait=False)` means a worker can still be running after the context unwinds, so when the body exits via a non-KeyboardInterrupt exception (where `in_body` stayed `True`), a late task failure could deliver SIGINT into caller code beyond the context — the new test pins that concrete case, and the diff is a net simplification (one `finally` replaces the two scattered clears). If you still consider the edge not worth carrying, feel free to close this — no strong attachment.

---

PR #7004, opened 2026-08-29
fix: avoid duplicating inner tag in MemoComponent wrapper tag

### LINKED ISSUE #6955: Auto-memo wrapper tags for @rx.memo components repeat the inner tag
Follow-up from #6949 (auto-memoize `@rx.memo` call sites with stateful props). Cosmetic only.

### Problem

The generated wrapper around a user `@rx.memo` component gets a tag like:

```
Memocomponent_card_98ffd1e1_card_98ffd1e1_e52460d9ed338b99ee51cd718b2b8fa7_98ffd1e1
```

The inner tag appears twice. The name is noisy in compiled `app_components/*.jsx` output, React DevTools, and stack traces.

### Cause

`Component._compute_memo_tag()` (`packages/reflex-base/src/reflex_base/components/component.py`) builds the prefix as `f"{type(self).__qualname__}_{self.tag or 'Comp'}_{comp_hash}"`. For a `MemoComponent` instance, `type(self)` is the dynamic subclass from `_get_memo_component_class`, whose `__qualname__` equals its `tag` (`Card_98ffd1e1`), so the qualname and tag segments are identical. The qualname is there to keep distinct classes with identical render output from colliding; for memo subclasses the tag already carries that identity. The trailing `_98ffd1e1` is the module suffix added a second time by `_create_component_definition`.

### Suggested fix

In `_compute_memo_tag`, skip the `self.tag` segment when it equals `type(self).__qualname__` (or override `_compute_memo_tag` on `MemoComponent` to use `Memocomponent_{tag}_{hash}`). Keep the class identity in the name so the collision guarantee from the docstring holds. Add a unit test in `tests/units/compiler/test_memoize_plugin.py` asserting the wrapper tag for a user memo contains the inner tag once.

### Repro

Any `@rx.memo` component called with a state Var prop; inspect `.web/app_components/<app>/<app>.jsx` after `reflex run`.

LeonxLJX:
Override _compute_memo_tag on MemoComponent to skip the self.tag segment, which is already embedded in the dynamic subclass __qualname__ (MemoComponent_<tag>). Previously the wrapper tag contained the inner tag twice (e.g. Memocomponent_card_98ffd1e1_card_98ffd1e1_<hash>).

Adds a regression test asserting the inner tag appears exactly once.

Closes #6955

### All Submissions:

- [ ] Have you checked to ensure there aren't any other open [Pull Requests](https://github.com/reflex-dev/reflex/pulls) for the desired changed?

### Type of change

### New Feature Submission:

### Changes To Core Features:

    b. Describe your changes.

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/7004?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/reflex-base/src/reflex_base/components/memo.py
+++ b/packages/reflex-base/src/reflex_base/components/memo.py
@@ -398,6 +398,34 @@ class MemoComponent(Component):
             children: The children of the component (ignored).
         """
 
+    def _compute_memo_tag(self) -> str:
+        """Compute a stable tag name for this memo component.
+
+        Overrides ``Component._compute_memo_tag`` to avoid duplicating the
+        wrapped component's tag. For a ``MemoComponent`` the dynamic subclass
+        ``__qualname__`` is ``MemoComponent_<tag>`` (see
+        :func:`_get_memo_component_class`), which already encodes the inner
+        tag. Appending ``self.tag`` again would produce the inner tag twice
+        (e.g. ``Memocomponent_card_98ffd1e1_card_98ffd1e1_<hash>``).
+
+        The class identity is still preserved via the qualname prefix, so the
+        collision guarantee from the base implementation holds.
+
+        Returns:
+            The stable tag name.
+        """
+        from reflex_base.components.memoize_helpers import (
+            MemoizationStrategy,
+            get_memoization_strategy,
+        )
+
+        comp_hash = self._get_component_hash(
+            shallow=get_memoization_strategy(self) == MemoizationStrategy.PASSTHROUGH
+        )
+        return format.format_state_name(
+            f"{type(self).__qualname__}_{comp_hash}"
+        ).capitalize()
+
     def _post_init(self, **kwargs):
         """Initialize the memo component.
 
--- a/tests/units/compiler/test_memoize_plugin.py
+++ b/tests/units/compiler/test_memoize_plugin.py
@@ -2593,3 +2593,35 @@ def test_each_memo_wrapper_emits_one_component_module_file() -> None:
         "for Plain, one for WithProp, and one snapshot wrapper for the "
         f"LeafComponent boundary. Got: {sorted(ctx.memoize_wrappers)}"
     )
+
+
+def test_memo_wrapper_tag_contains_inner_tag_once() -> None:
+    """Auto-memo wrapper tags must not duplicate the wrapped component's tag.
+
+    Regression test for #6955: ``MemoComponent._compute_memo_tag`` previously
+    concatenated ``type(self).__qualname__`` (which already embeds the inner
+    tag as ``MemoComponent_<inner_tag>``) with ``self.tag`` again, producing
+    names like ``Memocomponent_plain_abc123_plain_abc123_<hash>``. The inner
+    tag must appear exactly once in the wrapper tag.
+    """
+    ctx, _page_ctx = _compile_single_page(lambda: Plain.create(STATE_VAR))
+
+    assert len(ctx.memoize_wrappers) == 1, (
+        f"Expected exactly one auto-memo wrapper, got: {sorted(ctx.memoize_wrappers)}"
+    )
+    wrapper_tag = next(iter(ctx.memoize_wrappers))
+    # The inner component's tag is "Plain"; after format_state_name it becomes
+    # lowercase "plain" in the wrapper tag. Count occurrences of the inner tag
+    # segment (case-insensitive, word-bounded to avoid matching hash substrings).
+    inner_tag_lower = Plain.tag.lower()
+    # Split on underscores and count exact matches of the inner tag
+    segments = wrapper_tag.lower().split("_")
+    inner_tag_occurrences = segments.count(inner_tag_lower)
+    assert inner_tag_occurrences == 1, (
+        f"Inner tag '{Plain.tag}' should appear exactly once in wrapper tag, "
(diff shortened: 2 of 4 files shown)


LeonxLJX:
Added the missing `reflex-base` news fragment (`news/7004.bugfix.md`) so the release workflow is satisfied.

**Root cause.** `MemoComponent._compute_memo_tag` rebuilt the tag by re-prepending the wrapped component's class-qualified prefix on top of the already-prefixed base, so the generated memo-wrapper name carried the inner tag twice. That made memoized component names unstable / collision-prone across recompiles.

**Fix.** Override `_compute_memo_tag` to preserve the class-qualified prefix + component hash exactly once, and added a compiler regression test that asserts the produced tag is no longer duplicated. Tag shape is unchanged for the common case, so existing apps keep the same memo identity.

If you'd prefer the fragment wording tuned (e.g. mention the failing case explicitly), happy to adjust.

LeonxLJX:
Bumping for visibility. Greptile flagged a P1 on the `reflex-base` component cache behavior — basically that under `MemoComponent`'s wrapper-tag dedup, a stale memoized tag can survive across user-state changes if the inner component's tag depends on state that changed. Before I push a fix I want to confirm the desired behavior with maintainers: should the dedup invalidate when the inner component's state changes (current PR does this implicitly via the wrapper-tag recompute), or should we expose an explicit `invalidate()` API on `MemoComponent`? I lean toward the former to keep the API surface minimal — but want to check before iterating.

tests/units/compiler/test_memoize_plugin.py
```diff
@@ -2593,3 +2593,35 @@ def test_each_memo_wrapper_emits_one_component_module_file() -> None:
...
+def test_memo_wrapper_tag_contains_inner_tag_once() -> None:
+    """Auto-memo wrapper tags must not duplicate the wrapped component's tag.
+
+    Regression test for #6955: ``MemoComponent._compute_memo_tag`` previously
+    concatenated ``type(self).__qualname__`` (which already embeds the inner
+    tag as ``MemoComponent_<inner_tag>``) with ``self.tag`` again, producing
+    names like ``Memocomponent_plain_abc123_plain_abc123_<hash>``. The inner
+    tag must appear exactly once in the wrapper tag.
+    """
+    ctx, _page_ctx = _compile_single_page(lambda: Plain.create(STATE_VAR))
```
FarhanAliRaza, line 2607: This test fails on this branch.

`STATE_VAR` is passed as a child, not a prop. `Component.create` wraps it in a `Bare`, so the tree is `Plain(Bare(STATE_VAR))` and the state hooks ride on the `Bare`. The auto-memoize pass wraps the `Bare`. That wrapper is an ordinary `Component`, not a `MemoComponent`, so `MemoComponent._compute_memo_tag` never runs and the inner tag `Plain` never appears in the wrapper tag.

```
AssertionError: Inner tag 'Plain' should appear exactly once in wrapper tag,
found 0 times. wrapper_tag=Bare_comp_bba9dad596f8328a01eabe6aaa98e665
assert 0 == 1
```

The doubling this PR fixes only occurs when a `@rx.memo` component receives a state Var as a prop. In that case the export name on `main` is `Memocomponent_statefulcard_14db30e6_statefulcard_14db30e6_dbfbf34c...`, and with this fix it is `Memocomponent_statefulcard_14db30e6_dbfbf34c...`.

tests/units/compiler/test_memoize_plugin.py
```diff
@@ -2593,3 +2593,35 @@ def test_each_memo_wrapper_emits_one_component_module_file() -> None:
...
+
+    assert len(ctx.memoize_wrappers) == 1, (
+        f"Expected exactly one auto-memo wrapper, got: {sorted(ctx.memoize_wrappers)}"
+    )
+    wrapper_tag = next(iter(ctx.memoize_wrappers))
+    # The inner component's tag is "Plain"; after format_state_name it becomes
+    # lowercase "plain" in the wrapper tag. Count occurrences of the inner tag
+    # segment (case-insensitive, word-bounded to avoid matching hash substrings).
+    inner_tag_lower = Plain.tag.lower()
```
FarhanAliRaza, line 2616: `Component.tag` is `str | None`, so the type check gate fails here.

```
uv run pyright tests/units/compiler/test_memoize_plugin.py
test_memoize_plugin.py:2616:33 - error: "lower" is not a known attribute of "None" (reportOptionalMemberAccess)
```

packages/reflex-base/src/reflex_base/components/memo.py
```diff
@@ -398,6 +398,34 @@ def _validate_component_children(self, children: list[Component]) -> None:
             children: The children of the component (ignored).
         """
 
+    def _compute_memo_tag(self) -> str:
```
FarhanAliRaza, line 401: This override repeats the whole body of `Component._compute_memo_tag`. Only one segment of the f-string differs. The imports, the strategy lookup, the hash call and the `format_state_name(...).capitalize()` now exist in two places, so the hashing policy can drift between them.

FarhanAliRaza (changes requested):
See the inline comments.

Issue #7003, opened 2026-08-28
Migrate `rx.moment` to react-moment 2.x (prop restructure + new peer dep)


masenf:
react-moment 2.0.0 is a breaking major that `reflex-components-moment` cannot take as a straight version bump. The pin is currently held at `react-moment@1.2.2`, and `react-moment` is filtered out of the release branch's outdated-dependency check with a pointer to this issue's problem.

## Why it's blocked

The 2.0 release reworks the component's public surface:

- global configuration moved into a `settings` prop on the `Moment` component rather than module-level config
- the component's props were restructured
- the singular `filter` prop was removed
- a new peer dependency, `moment-duration-format ^2.2.2`, joins `moment` and `react`

`Moment` in `packages/reflex-components-moment/src/reflex_components_moment/moment.py` declares roughly two dozen props one-to-one against the 1.x API (`interval`, `format`, `trim`, `parse`, `add`/`subtract` via `MomentDelta`, `from_now`, `from_now_short`, `from_now_during`, `to_now`, `with_title`, `title_format`, `diff`, `decimal`, `unit`, `duration`, `duration_from_now`, `unix`, `local`, `tz`, `locale`, `on_change`). Each needs checking against the 2.x prop layout, and any that moved or disappeared needs a deprecation path since `rx.moment` is public API.

## What needs doing

- [ ] Diff the 1.x → 2.x prop surface against the props declared on `Moment` and write up the mapping (upstream's 2.0.0 migration guide is the reference)
- [ ] Deprecate or rename any prop that moved, per the project's deprecation policy (`console.deprecate()` with a `deprecation_version` of the next dot version, `removal_version` of the next major)
- [ ] Add `moment-duration-format` to `lib_dependencies` so the new peer dep resolves
- [ ] Bump the `react-moment` pin to 2.x and regenerate `.pyi` stubs
- [ ] Add an integration test covering the props that changed, and a news fragment under `packages/reflex-components-moment/news/`
- [ ] Drop `react-moment` from the frontend ignore list in `.github/workflows/check_outdated_dependencies.yml`

## Notes

`moment-timezone` was already bumped to 0.6.3 separately; it is unaffected by this migration. react-moment 2.0.2 was published 2026-05-26, so there is no urgency from a freshness standpoint — this is purely about doing the prop migration properly.

Upstream changelog: https://github.com/headzoo/react-moment/blob/master/CHANGELOG.md

PR #7006, opened 2026-08-30
Migrate react-moment to 2.0.2

harsh21234i:
### Summary

  Migrates rx.moment from react-moment@1.2.2 to react-moment@2.0.2.

  ### Changes

  - Added moment-duration-format@2.2.2.
  - Updated trim to support boolean and string values.
  - Updated parse to support string and list values.
  - Regenerated Moment type metadata and lockfiles.
  - Added unit and integration regression tests.
  - Confirmed compatibility in both development and production modes.

  ### Testing

  - 121 passed unit tests
  - 2 passed integration tests
  - Ruff formatting and lint checks passed
  - Lockfile and generated metadata validation passed

  Related issue: #7003

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/7006?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/reflex-components-moment/src/reflex_components_moment/moment.py
```diff
@@ -31,8 +32,8 @@ class Moment(NoSSRComponent, MemoizationLeaf):
 
     tag: str | None = "Moment"
     is_default = True
-    library: str | None = "react-moment@1.2.2"
-    lib_dependencies: list[str] = ["moment@2.30.1"]
+    library: str | None = "react-moment@2.0.2"
+    lib_dependencies: list[str] = ["moment@2.30.1", "moment-duration-format@2.2.2"]
```
FarhanAliRaza: `lib_dependencies` adds the package to `package.json`. It does not emit an import.

react-moment 2.0.2 requires the plugin to be **loaded**, not only installed. Without an import, `rx.moment(duration=...)` throws and the whole page fails.

Add the import in `add_imports()`:

```python
if self.duration is not None or self.duration_from_now is not None:
    imports["moment-duration-format@2.2.2"] = ImportVar(tag=None)
```

`ImportVar(tag=None)` is required here. An empty tag `""` collides with the `moment-timezone` side-effect import, and compilation fails.

tests/integration/test_moment.py
```diff
@@ -0,0 +1,77 @@
...
+            rx.el.input(
+                id="token",
+                value=State.router.session.client_token,
+                is_read_only=True,
+            ),
+            rx.moment(
+                "2026-08-30",
+                format="YYYY-MM-DD",
+                parse=["YYYY-MM-DD"],
+                trim="h [hours]",
```
FarhanAliRaza: Two problems on this line.

**1. `trim` has no effect in this test.** react-moment 2.x reads `trim` only inside `formatDurationContent`. That function runs only when you set `duration` or `durationFromNow`. This app sets neither prop. The test therefore never reaches the code path that this PR breaks. CI stays green while `rx.moment(duration=...)` crashes the page.

Please add a `duration` case:

```python
rx.moment(
    duration="2026-08-30T00:00:00",
    date="2026-08-30T05:30:00",
    format="h [hrs] m [min]",
    id="moment-duration",
),
```

Then assert the formatted result, for example `5 hrs 30 mins`. This test fails on the current branch and passes after you add the import.

**2. `h [hours]` is not a valid `trim` value.** `trim` accepts `large`, `small`, `both`, `all`, `final`, `left`, `right`, or a boolean. `h [hours]` is a format template. It looks like a mix-up with the `format` prop.

tests/units/compiler/test_memoize_plugin.py
```diff
@@ -1841,6 +1841,24 @@ def test_moment_with_stateful_var_child_does_not_wrap_bare_independently() -> No
     )
 
 
+def test_moment_uses_react_moment_2_props_and_dependencies() -> None:
+    """The wrapper should support the changed react-moment 2.x prop types."""
```
FarhanAliRaza: The docstring is not accurate. react-moment 1.2.2 already declares:

```js
parse: PropTypes.oneOfType([PropTypes.string, PropTypes.array])
trim:  PropTypes.oneOfType([PropTypes.string, PropTypes.bool])
```

The 2.x upgrade did not change these prop types. The Python types were too narrow before this PR. The widening is correct and I confirmed both forms render. Please describe the reason correctly.

`trim="h [hours]"` on line 1854 is also an invalid `trim` value. See my comment on the integration test.

FarhanAliRaza (changes requested):
## Request changes

This PR breaks the `duration` and `duration_from_now` props of `rx.moment`.

### Cause

react-moment 1.2.2 bundles `moment-duration-format` inside its own dist file. react-moment 2.0.2 removes it. Version 2.0.2 loads the plugin with a `require()` call. A bundler replaces that call with a stub in an ESM build. The load fails without a message. The plugin never patches `moment`. The component then throws:

```
Error: [react-moment] Duration formatting requires moment-duration-format.
Install the `moment-duration-format` package and ensure it is loaded before
using the `duration` or `durationFromNow` props.
```

React unwinds to the error boundary. The complete page fails. The failure is not limited to the moment element.

### Evidence

I built an app with nine `rx.moment` variants. I ran the same app on `main` and on this branch. I changed only the library version.

| Props | `main` (1.2.2) | This PR (2.0.2) |
| --- | --- | --- |
| `format="YYYY-MM-DD"` | `2026-08-30` | page crash |
| `parse="DD-MM-YYYY"` | `2026-08-30` | page crash |
| `parse=["YYYY-MM-DD", "DD-MM-YYYY"]` | `2026-08-30` | page crash |
| `duration=...` | `5 hrs 30 mins` | page crash |
| `trim=True` | `5 hrs 0 mins` | page crash |
| `trim="large"` | `30 mins` | page crash |

`main` renders all cases. This branch renders none.

### Fix

`lib_dependencies` installs the package. It does not import it. Add the import in `add_imports()`. Use the same pattern as `tz`.

```python
from reflex_base.utils.imports import ImportDict, ImportVar


    def add_imports(self) -> ImportDict:
        ...
        if self.tz is not None:
            imports["moment-timezone@0.6.2"] = ""
        if self.duration is not None or self.duration_from_now is not None:
            imports["moment-duration-format@2.2.2"] = ImportVar(tag=None)

        return imports
```

Use `ImportVar(tag=None)`. Do not use `""`. An empty tag collides with the `moment-timezone` import. The compiler then fails:

```
ValueError: Can not compile, the tag  is used multiple time from
moment-timezone@0.6.2 and moment-duration-format@2.2.2
```

I applied this patch on the PR head. All nine cases render. `reflex export --no-zip` succeeds. The production bundle contains the plugin. The 121 unit tests pass. The browser console is clean.

### Also required

Add a `duration` case to the integration test. The current test cannot detect this failure. See the inline comments.

FarhanAliRaza (dismissed):

packages/reflex-components-moment/src/reflex_components_moment/moment.py
```diff
@@ -31,8 +31,8 @@ class Moment(NoSSRComponent, MemoizationLeaf):
...
```
masenf:
```suggestion
    lib_dependencies: list[str] = ["moment@2.30.1"]
```

does it still work when you make the moment-duration-format a conditional dep?

harsh21234i: Yes, you’re right. lib_dependencies is unconditional, so it installs moment-duration-format for date-only rx.moment usage. I’ll remove it from lib_dependencies, keep the conditional ImportVar(tag=None) in add_imports(), regenerate the lockfiles, and verify both date-only and duration cases.

FarhanAliRaza (dismissed):

FarhanAliRaza (approved):

FarhanAliRaza (approved):

PR #7010, opened 2026-08-31
docs: add a complete auth example app and sharpen authentication-vs-authorization framing

FarhanAliRaza:
### All Submissions:

### Type of change

### Changes To Core Features:

Docs-only change (ENG-10954, follow-up to the initial auth docs). The builder agent retrieves these pages during memory retrieval, so this is the knowledge layer behind generating correct `rxe.AuthPlugin` code.

- **New page `docs/enterprise/auth/example-app.md`** — one small, complete app (public landing page, protected dashboard, admin-only action) showing the whole pattern in ~100 lines: plugin config with placeholder `OIDC_*` vars, checks in one `authz.py` module, state with `auth=False` opt-outs and an `auth=is_admin` action, `rxe.App()`, and a run walkthrough. Ends with the best-practice rules and the failure modes (`rx.App()` vs `rxe.App()`, blank-when-logged-out, `redirect_uri_mismatch`, the deprecated `register_auth_endpoints`).
- **Authentication vs authorization framing** — the page opens with the two questions ("who is this user?" is the plugin's job; "what may this user do?" is `auth=`) and a decision table for picking the `auth=` value; the overview gains a matching paragraph.
- **Sidebar** — the page is added to the enterprise auth section after Secure by Default; the overview and secure-by-default pages link to it.

Docs app tests pass locally (`tests/test_sidebar.py`, `test_doc_links.py`, `test_routes.py`, `test_urls.py`, `test_frontmatter_meta.py`, `test_doc_description.py`, `test_breadcrumbs.py`).

Linear: [ENG-10954](https://linear.app/reflex-dev/issue/ENG-10954)

---
_Generated by [Claude Code](https://claude.ai/code/session_01LoNAFswJdYdYfFGa63qFPg)_

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/7010?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/enterprise/auth/example-app.md
+++ b/docs/enterprise/auth/example-app.md
@@ -64,9 +64,8 @@ team_notes/
     └── team_notes.py       # state, pages, app
 ```
 
-The notes live in an `rx.SharedState` (available from reflex 0.8.23) linked
-to one team token, so every signed-in user reads and writes the same board
-and edits propagate live. A real app would also persist the notes to a
+The notes live in an `rx.SharedState` linked to one team token, so every
+signed-in user reads and writes the same board and edits propagate live. A real app would also persist the notes to a
 database — shared state is memory-resident — but nothing about the auth
 pattern changes.
 
@@ -192,9 +191,8 @@ Four things to notice:
 
 - **The notes are genuinely shared.** `NotesState` is an
   [`rx.SharedState`](/docs/state-structure/shared-state/): every signed-in
-  user links to the same `"team-notes"` token (linked tokens may not contain
-  underscores), so an added or cleared note propagates to everyone on the
-  dashboard. The protected `join_team_board` event does the linking, and the
+  user links to the same `"team-notes"` token, so an added or cleared note
+  propagates to everyone on the dashboard. The protected `join_team_board` event does the linking, and the
   page guard guarantees a resolved user before it runs.
 - **`UserExtras` exists for the UI, not for security.** The common claims
   (`name`, `email`, `sub`, `picture`) are already frontend Vars on `User`; any
--- a/docs/enterprise/auth/example-app.md
+++ b/docs/enterprise/auth/example-app.md
@@ -126,6 +126,11 @@ single string, and Python's `in` on a string is a substring test —
 `"admins" in "not-admins"` is true. Checking the type first fails closed on
 any unexpected claim shape.
 
+```md alert warning
+# The claim has to come back from the userinfo endpoint
+`extra_scopes` asks the provider for a scope, but the claims your check reads are whatever the provider's **userinfo endpoint** returns — that response is what populates `ctx.auth_user_state.userinfo`. A `groups` claim mapped only into the ID token never reaches the check, which then returns `False` for everyone with nothing in the logs to explain it. Map the claim onto the userinfo response, and confirm it arrived by printing `ctx.auth_user_state.userinfo` once rather than assuming. Claims are cached for about 30 minutes, so a group added at the identity provider is not immediate.
+```
+
 Annotating `ctx` with the `AuthContext` union makes `is_admin` usable on any
 surface — an event, a field, a var, or the plugin's global default. A check
 never runs for an anonymous caller, and a check that raises fails closed. See
@@ -328,6 +333,55 @@ linked to the shared `"team-notes"` state. The clear button appears only if
 your IdP puts you in the `admins` group — and refuses everyone else even if
 they dispatch the event by hand.
 
+## When the roles are your app's, not your provider's
+
+`is_admin` reads a group the identity provider already manages, which is the
+cheapest check there is: `ctx.auth_user_state.userinfo` is a dict resolved once
+at sign-in and held in memory, so it costs nothing per event.
+
+That only works when the organization models the role in the IdP. Many apps
+have roles the provider has never heard of — `author`, `editor`, `reviewer` —
+and need an admin to promote someone *from inside the app*, which no OIDC flow
+can do: there is no write-back to the identity provider. Those roles live in
+your own table, and the check reads them through a state:
+
+```python
+# team_notes/roles.py
+class RoleState(rx.State):
+    """The signed-in member's role, loaded once per session."""
+
+    role: str = ""
+
+    @rxe.event
+    async def load_role(self):
+        claims = await User.current() or {}
+        if subject := str(claims.get("sub") or ""):
+            self.role = await member_role(subject)  # your table
+
+
+# team_notes/authz.py
+async def is_editor(ctx: AuthContext) -> bool:
+    """Editors and admins. Reads the row RoleState already loaded."""
+    roles = await ctx.auth_user_state.get_state(RoleState)
+    return roles.role in ("editor", "admin")
+```
+
+Run the loader on the pages that need it (`on_load=RoleState.load_role`), and
+again after any write that changes a role so the cached value stays honest.
+
+Two rules keep this both safe and fast:
+
+- **Load once, read many.** A check runs on *every* gated event, so opening a
+  database session inside one puts a query behind every click. Load the row
+  into a state at sign-in and let the check read memory.
+- **Key the row on `sub`, not `email`.** `sub` is the stable identifier for an
+  identity; `email` is optional in OIDC and can change. In a multi-provider app
+  key on `(provider_name, sub)`, since `sub` is unique only per issuer.
+
+An async check is also where you call an external authorization service, or
+re-read a permission that must not be up to 30 minutes stale. See
+[authorization checks](/docs/enterprise/auth/secure-by-default/#authorization-checks).
+
 ## The rules this app follows
 
 1. **Never hand-roll authentication.** No login form, no password column, no
@@ -359,6 +413,9 @@ they dispatch the event by hand.
   and protected vars their `initial_value` until login. Mark truly public
   surfaces `auth=False`, and give the rest placeholders that read as
   intentional.
+- **Querying the database inside a check.** A check runs on every gated event,
+  so a session opened there is a query per click. Load the role into a state at
+  sign-in and have the check read that state.
 - **`redirect_uri_mismatch` at the IdP.** The exact callback URL (scheme,
   host, and path) is not registered with the identity provider. Register the
   full URL for each environment. See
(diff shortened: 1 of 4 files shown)


docs/enterprise/auth/example-app.md
```diff
@@ -0,0 +1,359 @@
...
+    def is_admin_user(self) -> bool:
+        return "admins" in (self.userinfo.get("groups") or [])
+
+
+class NotesState(rx.State):
+    # Public: the landing page uses this before anyone logs in.
+    dark_mode: rx.Field[bool] = rxe.field(False, auth=False)
+
+    # Protected by default: withheld from the client until login.
+    notes: rx.Field[list[str]] = rx.field([])
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid observation, addressed in 9cc0c8f by making the doc honest rather than by adding persistence: the intro no longer calls it a "shared notepad", and the project-layout note now states that Reflex state is per-client — each user sees only their own notes — with a real team app storing them in a database shared by every user. Backing the example with a database would double its size while changing nothing about the auth pattern, which is the page's subject; the auth docs' other examples make the same trade.

---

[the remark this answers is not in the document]
FarhanAliRaza: Valid observation, addressed in 9cc0c8f by making the doc honest rather than by adding persistence: the intro no longer says "shared notepad", and the layout note now states that state is per-client (each user sees only their own notes) and that a real team app stores notes in a database shared by every user. Adding a database to the example would double its size while changing nothing about the auth pattern this page teaches, so the wording fix is the right scope here.

---

docs/enterprise/auth/example-app.md
```diff
@@ -0,0 +1,359 @@
...
+    """Project the ``groups`` claim onto the frontend, for rendering only."""
+
+    @rx.var
+    def is_admin_user(self) -> bool:
+        return "admins" in (self.userinfo.get("groups") or [])
+
+
+class NotesState(rx.State):
+    # Public: the landing page uses this before anyone logs in.
+    dark_mode: rx.Field[bool] = rxe.field(False, auth=False)
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in 9cc0c8f. The landing page now reads `dark_mode`: the button label flips between "Switch to dark"/"Switch to light" and the page's `background_color`/`color` bind to the field through `rx.cond`, so the public field and its `auth=False` handler have a visible effect for anonymous visitors, which is exactly the point the example makes with it.

---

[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in 9cc0c8f. The landing page now consumes the field: the toggle's label reflects the current mode and the page's `background_color`/`color` bind to `dark_mode` via `rx.cond`, so the control has a real, visible effect. Kept as page-level styling rather than the framework color-mode toggle because the field's job in this example is to be a *state* surface an anonymous visitor can read and write (`auth=False` on both), which the color-mode API would not demonstrate.

---

docs/enterprise/auth/example-app.md
```diff
@@ -0,0 +1,359 @@
...
+
+- **`UserExtras` exists for the UI, not for security.** The common claims
+  (`name`, `email`, `sub`, `picture`) are already frontend Vars on `User`; any
+  other claim you want to *render* — here, whether to show the admin button —
+  needs a computed var on an `AuthUserState` substate. The enforcement is the
+  `auth=is_admin` on the handler, never the rendering.
+- **`await User.current()`** returns the claims dict inside any event handler,
+  or `None` when anonymous. Because `add_note` requires login, the user is
+  always resolved there.
+- **Protected computed vars get an `initial_value`.** The placeholder is baked
```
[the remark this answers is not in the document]
FarhanAliRaza: Declining this one. The Google-docstring convention in CLAUDE.md governs framework source under `reflex/` and `packages/`, not the code blocks inside markdown docs — the established auth pages already follow the lighter style this example uses: `profile()` in `docs/enterprise/auth/overview.md` and the check functions in `docs/enterprise/auth/secure-by-default.md` carry either a one-line docstring or none, and no docs example in `docs/enterprise/` writes Args/Returns sections. This page keeps that convention: one-line docstrings where they teach something (the two pages, `UserExtras`) and none where they'd just restate the name (`navbar`, two-line handlers). Full Google sections would roughly double the listing for a page whose value is being small.

---

docs/enterprise/auth/example-app.md
```diff
@@ -0,0 +1,367 @@
...
+```text
+team_notes/
+├── rxconfig.py             # plugin + provider configuration
+└── team_notes/
+    ├── authz.py            # the authorization checks
+    └── team_notes.py       # state, pages, app
+```
+
+Notes live in Reflex state to keep the example focused on auth. State is
+per-client, so each user sees only their own notes; a real team app stores
```
masenf: i understand that this vastly simplifies the code, but i think it makes the example less usable. you might consider implementing the notes store as a SharedState so it is at least truly multi-user.

docs/enterprise/auth/example-app.md
```diff
@@ -0,0 +1,367 @@
...
+    """Project the ``groups`` claim onto the frontend, for rendering only."""
+
+    @rx.var
+    def is_admin_user(self) -> bool:
+        return "admins" in (self.userinfo.get("groups") or [])
+
+
+class NotesState(rx.State):
+    # Public: the landing page uses this before anyone logs in.
+    dark_mode: rx.Field[bool] = rxe.field(False, auth=False)
```
masenf: this seems like a weird thing to have in an official example because it's not the reflex-supported way of handling dark mode. maybe instead this could be a computed var that indicates whether the app is in maintenance mode?

masenf (commented):
left a few comments. overall i think the example is good. but we want to make sure we're not demonstrating less-than-desirable style

docs/enterprise/auth/example-app.md
FarhanAliRaza: Done in 810adf5 — `NotesState` is now an `rx.SharedState`. A protected `join_team_board` event links to a single `"team-notes"` token and runs from the dashboard's `on_load` (the guard is prepended, so only signed-in users ever link), which makes the board genuinely multi-user. The per-client `draft` field became an `rx.form` submission so nothing per-user leaks into the shared instance, and the prose now walks through the two-browser behavior and links to the Shared State docs.

---

docs/enterprise/auth/example-app.md
FarhanAliRaza: Agreed — the ad-hoc dark mode is gone in 810adf5. I went with a close cousin of your suggestion: the landing page now carries a dismissible promo banner (`SiteState.show_promo`, `auth=False` field + `auth=False` dismiss event) instead of a maintenance-mode computed var, because the example needs a public *event* an anonymous visitor can trigger as well as a public value they can read — the banner demonstrates both without shadowing the framework's color-mode support. Happy to switch it to the maintenance-mode shape if you'd rather.

---

docs/enterprise/auth/example-app.md
```diff
@@ -151,42 +152,50 @@ class UserExtras(AuthUserState):
...
         self.notes = []
 ```
 
-Three things to notice:
+Four things to notice:
 
+- **The notes are genuinely shared.** `NotesState` is an
+  [`rx.SharedState`](/docs/state-structure/shared-state/): every signed-in
+  user links to the same `"team-notes"` token (linked tokens may not contain
+  underscores), so an added or cleared note propagates to everyone on the
```
masenf: <del>(linked tokens may not contain underscores)</del>

the linked doc mentions this, and it's not relevant content for this page.

docs/enterprise/auth/example-app.md
```diff
@@ -64,10 +64,11 @@ team_notes/
     └── team_notes.py       # state, pages, app
 ```
 
-Notes live in Reflex state to keep the example focused on auth. State is
-per-client, so each user sees only their own notes; a real team app stores
-them in a database shared by every user. Nothing about the auth pattern
-changes either way.
+The notes live in an `rx.SharedState` (available from reflex 0.8.23) linked
```
masenf: <del>(available from reflex 0.8.23)</del>

again, this doesn't seem helpful here since the whole auth plugin depends on reflex > 0.9 i thought

masenf (commented):
few more things, but i like this example modifications a lot more!

docs/enterprise/auth/example-app.md
FarhanAliRaza: Removed in c3aefea — the bullet now just links to the Shared State doc.

---

docs/enterprise/auth/example-app.md
FarhanAliRaza: Right — the plugin already requires a newer reflex, so the note added nothing. Removed in c3aefea.

---

masenf (dismissed):

docs/enterprise/auth/example-app.md
```diff
@@ -0,0 +1,443 @@
...
+    # Protected by default; the placeholder shows until the user is resolved.
+    @rxe.var(initial_value=0)
+    def note_count(self) -> int:
+        return len(self.notes)
+
+    @rxe.event  # default auth=True: only signed-in users join the board
+    async def join_team_board(self):
+        if not self._linked_to:
+            await self._link_to("team-notes")
+
```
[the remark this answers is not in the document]
FarhanAliRaza: Declined — this is the example's documented design, not a defect.

"Team Notes" is introduced as *"a shared team notepad"* (line 9), the `What we will build` table lists one dashboard reachable by all logged-in users, and the prose directly under this code block states it explicitly: *"every signed-in user links to the same `\"team-notes\"` token, so an added or cleared note propagates to everyone on the dashboard."* A single global board is the behaviour the page is teaching — `rx.SharedState` linking plus `auth=` enforcement — so there is no team or tenant claim in the example for the token to violate.

There is also nothing to reproduce: this PR is docs-only and adds no shipped code path.

An app that *does* have tenants would derive the link token from a tenant claim, but adding that dimension here would obscure the one mechanism the page exists to demonstrate.

masenf (approved):

PR #7014, opened 2026-09-01
fix(proxy): carry dataclass class-level metadata onto synthesized proxy classes

masenf:
### All Submissions:

### Type of change

### Changes To Core Features:

### Description

`MutableProxy.__new__` builds one wrapper class per wrapped dataclass type and copied `__dataclass_fields__` onto it, which is what `dataclasses.is_dataclass` tests for. The rest of the metadata `@dataclass` writes was left behind, so the wrapper answered yes to `is_dataclass(type(value))` and then raised `AttributeError` for the next thing a caller reads:

```python
@dataclasses.dataclass(frozen=True)
class Frozen:
    x: int = 1

class S(rx.State):
    f: Frozen = Frozen()

value = state.f
dataclasses.is_dataclass(type(value))       # True
type(value).__dataclass_params__            # AttributeError, before this change
type(value).__match_args__                  # AttributeError, before this change
value.__dataclass_params__                  # worked: wrapt forwards it to the wrapped object
```

The instance lookups happened to work because wrapt forwards them to the wrapped object, which is why this only bit code going through the class — the standard idiom right after an `is_dataclass` check.

### What is copied, and what is not

Every class attribute `@dataclass` writes (CPython 3.10–3.14 `_process_class`) was audited against what the synthesized proxy class exposes. They fall into three groups:

| | Attributes | Handling |
| --- | --- | --- |
| Plain data, read off the class | `__dataclass_fields__`, `__dataclass_params__`, `__match_args__` | **Copied** — instance forwarding cannot answer for a class-level read |
| Generated methods | `__init__`, `__repr__`, `__eq__`, `__lt__`/`__le__`/`__gt__`/`__ge__`, `__hash__`, frozen `__setattr__`/`__delattr__`, `__replace__`, `__getstate__`/`__setstate__` | **Not copied** — already resolve through the wrapped object; rebinding them to the proxy would sidestep its dirty tracking |
| Class layout and values | `__slots__`, field defaults, ClassVars | **Not copied** — a class attribute is found before `__getattr__` runs, so copying these would shadow the wrapped instance's own values |

Metadata the wrapped class was declared without is not invented either: `@dataclass(match_args=False)` leaves no `__match_args__` on the wrapper.

### Testing

- New tests in `tests/units/istate/test_proxy.py`: metadata is carried (plain, frozen and `slots=True` models), absent metadata is not invented (`match_args=False`), no behavior or layout is copied, and a frozen dataclass stays frozen through its proxy. The metadata assertions fail with `AttributeError` before the change.
- Full unit suite: 8013 passed. The two `tests/units/utils/test_processes.py` failures are pre-existing and environmental (no IPv6 in the test container) — they fail identically on a clean checkout of `main`, and are deselected in that count.
- `ruff check`, `ruff format`, and `pyright` clean.

https://claude.ai/code/session_01VFzJWvNNwomK7nhsEN2ELR


FarhanAliRaza (changes requested):
Tested in a Reflex app built on a worktree of this branch, and in the same app against `main` as a baseline. The app holds a plain dataclass state var, a nested dataclass, and a `frozen=True, eq=True` dataclass.

On `main`, an event handler that reads `type(state.frozen).__dataclass_params__` after an `is_dataclass` test fails with `AttributeError`. On this branch it returns the correct flags, for nested proxies as well.

Mutation tracking, `dataclasses.replace`, `asdict`, and `FrozenInstanceError` on assignment to a frozen value all still work. `tests/units/istate` and `tests/units/test_state.py` pass. The new test fails when I revert the `__new__` hunk.

Two requested changes are inline.

reflex/istate/proxy.py
```diff
@@ -453,16 +464,15 @@ def __new__(
...
                     wrapper_cls_name,
                     (cls,),
                     {
-                        dataclasses._FIELDS: getattr(  # pyright: ignore [reportAttributeAccessIssue]
-                            wrapped_cls,
-                            dataclasses._FIELDS,  # pyright: ignore [reportAttributeAccessIssue]
-                        ),
+                        attr: getattr(wrapped_cls, attr)
+                        for attr in _DATACLASS_CLASS_ATTRS
+                        if hasattr(wrapped_cls, attr)
```
FarhanAliRaza: `is_dataclass(wrapped)` at line 464 already guarantees `__dataclass_fields__`, so this guard can only ever filter `__dataclass_params__`. Drop it and copy both attributes directly, or state in the comment that it covers the params attribute only.

reflex/istate/proxy.py
```diff
@@ -40,6 +40,17 @@
 # from `dataclasses.asdict`/`astuple` internals on the proxy read hot-path.
 _DATACLASSES_FILE = dataclasses.__file__
 
+# Class-level metadata that `@dataclass` writes, copied onto the proxy class
+# synthesized per wrapped dataclass type. `__dataclass_fields__` is what
+# `dataclasses.is_dataclass` tests for, and `__dataclass_params__` carries the
+# decorator's flags (`frozen`, `eq`, `slots`, ...) that callers read right after
+# that test, so copying only the first yields a class that answers yes to
+# `is_dataclass` and then raises on how it was declared.
```
FarhanAliRaza: `@dataclass` also writes `__match_args__` at class level, which this tuple does not carry. Reword to say these are the two attributes callers read after an `is_dataclass` test.

reflex/istate/proxy.py
masenf: The guard covers more than the params attribute now, so I've kept it and explained it in the code instead of dropping it.

Two cases reach it:

- `@dataclass(match_args=False)` writes no `__match_args__` at all, so there is nothing to copy (`test_dataclass_proxy_class_omits_absent_metadata` locks that in — the wrapper must not invent metadata the wrapped class was declared without).
- A class can carry `__dataclass_fields__` without `__dataclass_params__`. That is not hypothetical: it is exactly what these wrapper classes themselves did before this PR, and `MutableProxy(existing_proxy, ...)` re-wraps one. Without the guard that path raises `AttributeError` at proxy construction.

The guard also grew a third job in 23c3fc1c: it now skips any name the wrapped dataclass declares as a *field*, since copying such a name shadows the instance's value. The whole namespace construction moved into `_dataclass_proxy_namespace` so both skips are stated where they happen rather than inline in `__new__`.

---

reflex/istate/proxy.py
masenf: Agreed, and fixed in 9354043c — `__match_args__` is copied now rather than the comment being reworded around its absence. It was the same bug as `__dataclass_params__`: `type(value).__match_args__` raised while `value.__match_args__` worked.

I audited the rest of `_process_class` (CPython 3.10–3.14) while I was there, and those three are the complete set of plain-data attributes `@dataclass` writes. The comment above the tuple now also records what deliberately stays off the wrapper — generated methods (they resolve through the wrapped object, and rebinding them would sidestep dirty tracking) and `__slots__`/field defaults/ClassVars (they would shadow the wrapped instance).

---

reflex/istate/proxy.py
```diff
@@ -40,6 +40,28 @@
...
+# Its generated methods (`__init__`, `__repr__`, `__eq__`, the ordering set, the
+# frozen `__setattr__`/`__delattr__` pair, `__replace__`, `__getstate__`) are
+# already reached through the wrapped object, and rebinding them here would
+# sidestep the proxy's dirty tracking. Field defaults, ClassVars and `__slots__`
+# would be worse: a class attribute is found before `__getattr__` runs, so
+# copying those would shadow the wrapped instance's own values.
+_DATACLASS_CLASS_ATTRS = (
+    dataclasses._FIELDS,  # pyright: ignore [reportAttributeAccessIssue]
+    dataclasses._PARAMS,  # pyright: ignore [reportAttributeAccessIssue]
+    "__match_args__",  # absent when declared `@dataclass(match_args=False)`
```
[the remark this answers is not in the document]
masenf: Valid — reproduced and fixed in 23c3fc1c.

`__match_args__` has trailing underscores, so it is not name-mangled and is a legal field name. With a field of that name, `@dataclass` leaves the field's default on the class (`_set_new_attribute` does not overwrite an existing class attribute), so the copy took the default and shadowed the instance:

```
asdict(wrapped) -> {'__match_args__': ('live',), 'other': 1}
asdict(proxy)   -> {'__match_args__': (),        'other': 1}   # before the fix
```

`__dataclass_params__` has the same exposure, and there the class attribute is the params object rather than the field default, so the proxy read returned a `_DataclassParams` where the instance held an `int`.

Rather than special-casing one name, the copy now applies to itself the rule the surrounding comment already states for field defaults and ClassVars: any name the wrapped dataclass declares as a field is skipped, because a class attribute is found before `__getattr__` forwards. `test_dataclass_proxy_class_never_shadows_a_field` covers both names.

---

reflex/istate/proxy.py
```diff
@@ -40,6 +40,50 @@
...
+    """
+    fields = getattr(wrapped_cls, _DATACLASS_FIELDS)
+    return {
+        attr: getattr(wrapped_cls, attr)
+        # A dataclass may declare a field named like one of these attributes,
+        # and that value belongs to the instance: copying it would shadow the
+        # wrapped object, since a class attribute is found before `__getattr__`
+        # forwards. `@dataclass(match_args=False)` leaves nothing to copy.
+        for attr in _DATACLASS_CLASS_ATTRS
+        if attr not in fields and hasattr(wrapped_cls, attr)
```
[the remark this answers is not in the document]
masenf: Valid — reproduced and fixed in 2021705c.

`__dataclass_fields__` carries ClassVar and InitVar entries alongside the real fields, so keying the skip on it was too broad. Neither pseudo-field can be shadowed by the copy: both keep their value on the class, and the instance has no attribute of its own for a class attribute to hide.

The `__dataclass_fields__` case was the sharp one — the wrapper stopped impersonating a dataclass entirely:

```python
@dataclasses.dataclass
class ClassVarFields:
    __dataclass_fields__: ClassVar[dict] = {}
    other: int = 0

dataclasses.is_dataclass(type(proxy))   # False, between 23c3fc1c and 2021705c
```

The skip now keys on `dataclasses.fields()`, which lists exactly the entries that *can* be shadowed — the real instance fields — and excludes ClassVar and InitVar by definition, so no private `_field_type` check is needed. `test_dataclass_proxy_class_copies_class_level_pseudo_fields` covers both declarations, and `test_dataclass_proxy_class_never_shadows_a_field` still holds for real fields of the same names.

---

masenf:
@FarhanAliRaza ready to go now 🚀

FarhanAliRaza (approved):

Issue #6975, opened 2026-08-28
compile_app() discards user bundle_library() registrations, and the dynamic serializer never rewrites subpath imports of bundled libs


masenf:
**Describe the bug**

Two compounding defects in the bundled-libraries plumbing for dynamic components (relocated but unchanged by the RegistrationContext work):

1. `compile_app()` calls `reset_bundled_libraries()` at the start of every frontend compile and re-adds **only** `plugin.get_frontend_dependencies()`. Any user/app-level `bundle_library()` call made at module import is silently dropped from `window.__reflex`. Meanwhile the backend worker short-circuits `compile_app` before the reset and KEEPS the registration — frontend bundle and backend registry end up permanently out of sync.
2. The dynamic serializer's window-rewrite loop matches only the exact substring `from "<lib>"`. Subpath imports of a lib that IS in the registry (e.g. `rx.icon`'s package_path form `lucide-react/dist/esm/icons/*.mjs`) are left as bare specifiers inside the eval'd data-URI module, which browsers can never resolve (no import map applies).

Net effect: the documented-for-plugins `bundle_library()` pattern is defeated for any library imported via subpaths.

**To Reproduce**

App that calls `bundle_library("lucide-react")` at module import time plus a computed var returning an `rx.Component` that contains `rx.icon(...)`, e.g.:

```python
import reflex as rx
from reflex.components.dynamic import bundle_library

bundle_library("lucide-react")

class State(rx.State):
    @rx.var
    def dynamic_icon(self) -> rx.Component:
        return rx.vstack(rx.icon("apple"))

def index():
    return rx.box(State.dynamic_icon)

app = rx.App()
app.add_page(index)
```

Run `reflex run` and load the page with the browser console open.

**Observed behavior**

- `window.__reflex` never contains `lucide-react` (registration wiped by the compile-time reset).
- Serialization passes whose context DOES hold the lib emit un-rewritten bare specifiers like `import LucideApple from "lucide-react/dist/esm/icons/apple.mjs"` inside the eval'd data-URI module, and the browser throws `TypeError: Failed to resolve module specifier "lucide-react/dist/esm/icons/apple.mjs"`.
- Serializer-level check with zero network: with `"lucide-react"` registered, serializing `rx.vstack(rx.icon("apple"))` emits the bare subpath import — the rewrite loop in `reflex_base/components/dynamic.py` only matches the exact `from "<lib>"` substring, and `format_library_name` strips only the `@version` suffix, not subpaths.

[...]

PR #7016, opened 2026-09-01
Fix bundled library subpath serialization

harsh21234i:
## Summary

  Fixes issue #6975.

  `compile_app()` was clearing explicit `bundle_library()` registrations during frontend compilation. Dynamic components
  also failed to rewrite bundled library subpath imports, such as Lucide deep imports.

  ## Changes

  - Preserve user-registered bundled libraries during compilation.
  - Rewrite bundled default, named, and namespace subpath imports through `window.__reflex`.
  - Add unit tests for bundled-library preservation and subpath serialization.
  - Add browser integration coverage using a dynamically rendered Lucide icon.
  - Add required bugfix news fragments.

  ## Tests

  - `uv run pytest tests/units/compiler/test_dynamic_components_codegen.py tests/units/compiler/test_compiler.py -q`
  - `uv run pytest tests/integration/test_dynamic_components.py -q`
  - Ruff checks and formatting pass.
  - Pyright passes for the changed source modules.

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/7016?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


masenf:
can you explain what scenarios result in triggering this bug? i understand what's going on, but it seems like most situations that matter, like hot-reload would end up re-bundling the needed libraries anyway.

harsh21234i:
I agree that normal hot-reload may re-run the app module and re-register the library, which can mask the problem. The bug is not limited to source-file hot-reload, though.

 The failure occurs when bundle_library() is called during app/component registration and compile_app() later resets the bundled-library list before compilation. 
This can happen even on the first compile, or on a later compile where the existing app and RegistrationContext are reused without re-executing the registration code.

 Examples include repeated compilation in the same process, cached app modules, compiler/AppHarness tests, export/build paths, and dynamic component serialization. In those cases, the explicit library registration is lost even though the app/context is still valid. Hot- reload may re-register it, but the compiler should not depend on hot- reload re-executing user code to preserve an explicit registration.

masenf:
Consolidated into #7109, which now includes the bundled-subpath serialization, unique aliases, registration preservation, Radix compatibility, and integration coverage from this PR. Harsh Thakare is credited as co-author of the combined implementation. Further review and merging will continue on #7109.

PR #7033, opened 2026-09-02
feat: allow explicit only event handlers

### LINKED ISSUE #7032: disable implicit @rx.event handlers
currently all non-underscored state methods are implicitly event handlers. Add an option to only allow explicit annotated functions with `@rx.event`

benedikt-bartscher:
Closes #7032

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/7033?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/reflex-base/src/reflex_base/config.py
```diff
@@ -403,10 +407,10 @@ def _post_init(self, **kwargs):
         self._non_default_attributes = set(kwargs.keys())
         self._replace_defaults(**kwargs)
 
-        # Publish for State-class creation so it never re-enters get_config()
-        # (which AttributeErrors if a State is defined while rxconfig.py is mid-import).
-        global _state_auto_setters
-        _state_auto_setters = self.state_auto_setters
+        # Publish to the in-progress rxconfig.py import, if any, so States defined
+        # after this Config in rxconfig.py resolve their flags from it.
+        if (load := _config_load.get()) is not None:
```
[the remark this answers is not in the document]
benedikt-bartscher: Not applicable — a `Config` only takes effect as the `config` object in `rxconfig.py`, loaded via `get_config()`/`reload_config()`. That's true for every option (`state_auto_setters`, `redis_url`, `state_manager_mode`, …), and `_set_config` has exactly two callers, both fed by `_get_config()`. An ad-hoc `rx.Config(...)` has never governed anything, so the new flag is consistent with the rest.

Auto-attaching direct constructions to a context without a config would be a regression: it gives `rx.Config(...)` a hidden side effect (test_config.py builds dozens of throwaway instances, the hosting CLI builds them too), and since `get_config()` only loads when `ctx._config is None`, a stray instance would preempt the rxconfig load entirely. That is the "last constructed Config wins" leak removed earlier in this PR. The `_config_load` slot intentionally publishes only while `_get_config()` is importing `rxconfig.py` — the one moment a fresh `Config` is *the* app config but not yet on the context.

news/7033.feature.md
```diff
@@ -0,0 +1 @@
+Add the `state_explicit_event_handlers` config option. When enabled, only methods decorated with `@rx.event` become event handlers; other public state methods stay plain Python methods.
```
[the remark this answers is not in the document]
benedikt-bartscher: Done — both fragments are now `news/7033.feature.md` and `packages/reflex-base/news/7033.feature.md`.

masenf (commented):
great, this gives us a lever and a chicken switch for eventually defaulting to requiring the decorator for event handlers.

need to ensure that reflex-enterprise's auth plugin and other integration points can still work with this new structure, but it seems like it would at first glance.

docs/events/events_overview.md
```diff
@@ -46,6 +46,17 @@ Whenever the user hovers over the heading, the `next_word` **event handler** wil
 
 Adding the `@rx.event` decorator above the event handler is strongly recommended. This decorator enables proper static type checking, which ensures event handlers receive the correct number and types of arguments.
 
+By default, every public method of a state (one whose name does not start with `_`) is treated as an event handler, whether or not it is decorated. To make `@rx.event` mandatory and keep undecorated public methods as plain Python helpers, enable `state_explicit_event_handlers` in `rxconfig.py`:
```
abulvenz, line 49: Might also become default one day, following the Zen of Python
>    Explicit is better than implicit.

abulvenz (approved):

docs/events/events_overview.md
benedikt-bartscher: yeah, i wanted to go the same route as i did with state_auto_setters.

1. introduce a flag to enable explicit mode - let feature stabilize and some projects adopt
2. change the default to explicit
3. drop implicit mode

benedikt-bartscher:
@masenf i think this PR is important from a security perspective - can we get it in?

packages/reflex-base/src/reflex_base/config.py
```diff
@@ -902,29 +906,66 @@ def _record_imports() -> Iterator[_ImportRecorder]:
...
+        The resolved flag value.
     """
-    if _state_auto_setters is not None:
-        return _state_auto_setters
-    env_val = os.environ.get(Config._prefixes[0] + "STATE_AUTO_SETTERS")
+    load = _config_load.get()
+    config = (
+        load.config
+        if load is not None
+        else RegistrationContext.ensure_context()._config
```
FarhanAliRaza, line 951: This reads the cached config and never loads it. The backend worker does not have a config cached when the user states are created. `reflex run` calls `get_config()` in the main process before `get_app()`, so compile sees the flag. The uvicorn worker (`reload=True`) and the prod uvicorn/granian subprocesses import `<module>:app` in a fresh process. The user states are created before any `get_config()` call, so `_config` is `None` here and the flag falls back to env/default.

Repro with the app used for this review (`state_explicit_event_handlers=True`):

```python
from uvicorn.importer import import_from_string
import_from_string("pr7033_repro.pr7033_repro:app")
sorted(ExplicitState.event_handlers)
# [..., "mixin_helper", ..., "step", ...]  <- undecorated helpers are handlers
```

In the running app, a computed var that returns `",".join(sorted(type(self).event_handlers))` shows the same list in the browser. The same happens when a user imports their state module in pytest.

`_config_load.get()` being `None` means no rxconfig import is in progress, so it is safe to load the config there:

```python
load = _config_load.get()
if load is None:
    return getattr(get_config(), name)
config = load.config
```

With this change the worker-style import gives the correct handler set, and test_state, test_event and test_config pass. It also removes the import-order dependence for `state_auto_setters`, which uses the same helper.

tests/units/test_state.py
```diff
@@ -3998,6 +4000,209 @@ class TestState(State):
         assert "setvar" in TestState.event_handlers
 
 
+def test_explicit_event_handlers(tmp_path, forked_registration_context):
```
FarhanAliRaza, line 4003: All the new tests call `reload_config()` before they define a state, so only the "config already cached on the context" path is covered. The backend worker path (state defined before any config load) is not.

Please add a test that chdirs into a project with `state_explicit_event_handlers=True`, enters a fresh `RegistrationContext()`, and defines a state without a prior `reload_config()`/`get_config()` call. It must not register the undecorated method. This test fails on the current branch.

FarhanAliRaza (changes requested):
Tested with a small app that sets `state_explicit_event_handlers=True`. The state has decorated handlers (plain, `background=True`, `stop_propagation=True`, an `on_load` handler, an explicit setter), undecorated helper methods called from the handlers, a mixin with one decorated and one undecorated method, and two `rx.ComponentState` instances. A computed var lists the state's registered event handler names on the page.

Ran it with `reflex run` in dev mode and exercised every control in the browser. All decorated handlers, the background handler, stop propagation, the mixin handler, the setter and both component states work. No console errors.

The handler list rendered from the backend still contains the undecorated helpers. The compile process sees the flag, but the backend worker does not, because it imports the app module before any config is loaded. A worker-style import (`uvicorn.importer.import_from_string`) reproduces this without a browser. I verified a candidate fix in `_get_state_flag` locally: the worker-style import then registers only the decorated methods, and the state, event and config unit tests pass.

The PR unit tests pass as-is. See the inline comments for the requested changes.

packages/reflex-base/src/reflex_base/config.py
benedikt-bartscher: Good catch, thanks. Fixed in acc796dd6e: `_get_state_flag` now calls `get_config()` when no rxconfig import is in progress, so the worker-style import registers only the decorated methods. Verified with your repro before/after.

tests/units/test_state.py
benedikt-bartscher: Added `test_state_flags_load_config_on_demand` in acc796dd6e: chdir into an explicit-mode project, fresh `RegistrationContext()`, State defined without a prior config load. It fails without the fix and passes with it.

packages/reflex-base/src/reflex_base/config.py
benedikt-bartscher: Update: the on-demand `get_config()` call had to go. If rxconfig.py imports project modules (the docs app does this for its plugins), loading the config from inside State-class creation re-enters the module that is defining the State and raises a circular ImportError (see the docs-tests failure on acc796dd6e).

77fe965698 takes a different route: `run_backend` / `run_backend_prod` export the State-class flags via `Config._set_persistent` before spawning workers, and `_get_state_flag` already falls back to those `REFLEX_*` env vars when no config is cached. Verified with `reflex run --backend-only`: the granian worker registers only the decorated handlers. Entry points outside `reflex run` (e.g. a test suite importing state modules) need the env var or a prior config load; documented in events_overview.md.

reflex/utils/exec.py
```diff
@@ -496,6 +496,8 @@ def run_backend(
     if not frontend_present:
         notify_backend(host)
 
+    get_config()._persist_state_flags()
```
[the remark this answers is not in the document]
benedikt-bartscher: Not the case: `Config.__init__` calls `update_from_env`, which applies `REFLEX_<FIELD>` for every dataclass field, so an env-only `REFLEX_STATE_EXPLICIT_EVENT_HANDLERS=true` lands in `config.state_explicit_event_handlers` and `_persist_state_flags` writes back `True`. Verified locally. This is the same mechanism the existing `_set_persistent` port propagation relies on.

packages/reflex-base/src/reflex_base/config.py
```diff
@@ -833,6 +837,14 @@ def _set_persistent(self, **kwargs):
         self._non_default_attributes.update(kwargs)
         self._replace_defaults(**kwargs)
 
+    def _persist_state_flags(self):
+        """Export the State-class creation flags to the environment for subprocesses.
+
+        Backend workers import the app module before any config is loaded, so
+        their user States resolve these flags from the environment.
+        """
+        self._set_persistent(**{name: getattr(self, name) for name in _STATE_FLAGS})
```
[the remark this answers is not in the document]
benedikt-bartscher: `run_backend` / `run_backend_prod` block in the server for the rest of the process, so the parent never builds another `RegistrationContext` afterwards. The env fallback is process-global by design; that is what makes it reach the worker subprocesses, exactly like the existing `_set_persistent` calls for ports and URLs. AppHarness runs the backend in-process and never hits these launchers.

packages/reflex-base/src/reflex_base/config.py
```diff
@@ -422,10 +426,10 @@ def _post_init(self, **kwargs):
...
         self._replace_defaults(**kwargs)
 
-        # Publish for State-class creation so it never re-enters get_config()
-        # (which AttributeErrors if a State is defined while rxconfig.py is mid-import).
-        global _state_auto_setters
-        _state_auto_setters = self.state_auto_setters
+        # Publish to the in-progress rxconfig.py import, if any, so States defined
+        # after this Config in rxconfig.py resolve their flags from it.
+        if (load := _config_load.get()) is not None:
+            load.config = self
```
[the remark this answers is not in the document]
benedikt-bartscher: The window here is narrower than described: `_config_load` is set only for the duration of the `import_module("rxconfig")` call in `_get_config` and reset in the `finally`, so "after `rxconfig.config` is assigned" is already outside it — at that point `_get_state_flag` reads `RegistrationContext._config` again. The only way to hit this is a second `Config(...)` constructed inside rxconfig.py's own import graph, after the real one and before a State class in that same graph.

Neither of the two suggested triggers applies:

- Imported application code doesn't construct `Config`, it calls `get_config()`. A nested `get_config()` -> `_get_config()` during the outer load sets and resets its own `_ConfigLoad` token, so the outer one is restored intact.
- Plugins are constructed *inside* `Config.__init__` (`_normalize_plugins` / `_add_builtin_plugins`), i.e. before this line publishes. A plugin that built a `Config` would publish first and then be overwritten by the outer one — the opposite of the reported failure.

What remains is an rxconfig.py that builds a throwaway `Config` after the real one and then imports State-defining modules. Last-one-wins is a heuristic there, but so is first-one-wins: which object ends up bound to `rxconfig.config` isn't knowable until the import completes. Leaving as is.

reflex/state.py
```diff
@@ -742,10 +744,11 @@ def __init_subclass__(cls, mixin: bool = False, **kwargs):
             cls._init_var(name, prop)
 
         # Set up the event handlers.
+        explicit = cls.is_user_defined() and get_state_explicit_event_handlers()
```
FarhanAliRaza, line 747: With the flag on, wiring an undecorated method to a trigger falls through to the plain-callable path in `call_event_fn`, which calls the method at compile time with the component as `self`:

```python
class S(rx.State):
    n: int = 0
    def helper(self):
        self.n += 1

rx.button(on_click=S.helper)
```
```
TypeError: unsupported operand type(s) for +=: 'ObjectItemOperation' and 'int'
```

A helper that does not touch `self` gets this instead:

```
EventHandlerValueError: Invalid event chain for on_click: <function S.helper at 0x...> -> None: A lambda inside an EventChain list must return `EventSpec | EventHandler | EventChain | EventVar | FunctionVar` or a heterogeneous sequence of these types. Got: <class 'NoneType'>.
```

Neither mentions `@rx.event` or `state_explicit_event_handlers`, and every migration to this flag hits this exact mistake. It needs a targeted error that names the flag and the decorator, raised before the method is called.

PR #7035, opened 2026-09-02
Fix config reload state imports (#7028)

### LINKED ISSUE #7028: Reloading a config whose rxconfig.py imports a state-defining module raises StateValueError
**Describe the bug**

If `rxconfig.py` imports a project-local module that defines an `rx.State` subclass, reloading the config in the same `RegistrationContext` crashes:

```
StateValueError: The substate class 'appmod____my_state' has been defined multiple times.
Shadowing substate classes is not allowed.
```

`_get_config` (`packages/reflex-base/src/reflex_base/config.py`) drops every recorded project-local dependency from `sys.modules` before each load, so the next load re-imports them and re-executes any `rx.State` class body they contain. The second class object collides with the first, which is still registered as a substate of its parent in the current context, and the shadowing check in `State.__init_subclass__` (`reflex/state.py:583`) fires.

This is deterministic, not a race — it happens on every reload for such a project.

**To Reproduce**

Two files in a project directory:

```python
# appmod.py
import reflex as rx


class MyState(rx.State):
    value: str = ""
```

```python
# rxconfig.py
import appmod  # noqa: F401
import reflex as rx

config = rx.Config(app_name="statereload")
```

Then:

```python
import reflex_base.config as c
from reflex_base.registry import RegistrationContext

with RegistrationContext():
    c.get_config()
    c.reload_config()  # StateValueError
```

`get_config(reload=True)` fails the same way, since it delegates to `reload_config()`.

Scope of the failure:

| scenario | result |
| --- | --- |
| two separate `RegistrationContext`s, one load each | works — each context has its own substate registry |
| `reload_config()` / `get_config(reload=True)` in one context | **StateValueError** |

**Expected behavior**

Reloading the config re-reads `rxconfig.py` from disk and returns a fresh `Config`, without redefining states that the previous load already registered.

- Python Version: 3.14.7
- Reflex Version: main @ 812bb475 (reflex-base 0.9.10)
- OS: Linux

**Additional context**

The eviction exists so that two different project directories don't reuse each other's modules — that part works and should be preserved; interleaved loads across two projects with same-named helper modules each correctly resolve their own copy.

Two directions for a fix:

1. Scope the eviction to actual project changes — only drop the recorded deps when the project root differs from the one they were recorded under. `rxconfig.py` itself is still evicted every load, so a reload still re-reads it. Small and contained, but a same-root reload then stops picking up edits to helper modules imported by `rxconfig.py` (edits to `rxconfig.py` itself are still picked up).
2. Unregister the states before re-importing. This is what `reload_state_module` (`reflex/state.py`) already does for hot reload, but it lives in the `reflex` package and `reflex_base` imports nothing from `reflex`, so it isn't reachable from `config.py` today. Doing it properly likely means giving `RegistrationContext` an unregister-by-module capability.

Noticed while reviewing https://github.com/reflex-dev/reflex/pull/6933; it is unrelated to that PR's `sys.path` changes and reproduces identically on `main`.

harsh21234i:
• ## Summary
Fixes: #7028 
  Fix reload_config() raising StateValueError when rxconfig.py imports a
  module that defines a rx.State class.

  ## Changes

  - Preserve project-local dependencies during reloads within the same
    RegistrationContext.

  - Continue evicting dependencies for new contexts or different project
    roots.

  - Added regression tests for state reloads, fresh contexts, and cross-
    project isolation.

  - Added a reflex-base bugfix changelog fragment.

  ## Testing

  - Focused regression tests: 3 passed
  - Ruff check and formatting: passed

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/7035?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/reflex-base/src/reflex_base/config.py
```diff
@@ -919,16 +928,17 @@ def _get_config(project_root: Path | None = None) -> Config:
         cwd = str(project_root)
         sys.path.insert(0, cwd)
         try:
-            # Never cache rxconfig or its project-local dependencies — each load
-            # goes to disk so different RegistrationContexts hold independent
-            # Config instances resolved against the current project. Evict
-            # before probing: find_spec answers from sys.modules, so modules
-            # left behind by another project directory would fake the existence
-            # check below.
+            # Always reload rxconfig, but retain its dependencies when reloading
```
FarhanAliRaza: Too long comments, explaining what is already can be seen in code. happens in many places. 
we might want to clean up these.

packages/reflex-base/src/reflex_base/config.py
```diff
@@ -1021,6 +1031,6 @@ def reload_config() -> Config:
         The freshly loaded app config.
     """
     ctx = RegistrationContext.ensure_context()
-    config = _get_config()
+    config = _get_config(reload_dependencies=ctx._config is None)
```
FarhanAliRaza: `ctx._config is None` is the wrong signal. `RegistrationContext.fork()` copies `base_states` but resets `_config` to None. So a forked context evicts and re-imports the state module into a context that already holds the class, and the shadow check fires again.

Repro with the `appmod.py` / `rxconfig.py` pair from #7028:

```python
with RegistrationContext() as ctx:
    c.get_config()
    forked = ctx.fork()
    tok = RegistrationContext.set(forked)
    c.reload_config()
    # StateValueError: The substate class 'appmod____my_state' has been defined multiple times.
```

This is the path `AppHarness` takes (`reflex/testing.py:286`: fork, then `reload_config()`), so the harness still crashes on such a project. Please key the decision on whether the current context already holds the states those modules registered, not on whether it has a cached config. Add the fork case to the tests.

packages/reflex-base/src/reflex_base/config.py
```diff
@@ -919,16 +928,17 @@ def _get_config(project_root: Path | None = None) -> Config:
...
+            # Always reload rxconfig, but retain its dependencies when reloading
+            # the same project. Re-importing a helper that defines State would
+            # redefine the class in the active RegistrationContext. Before
+            # switching projects, evict dependencies so find_spec and imports
+            # cannot reuse modules from the prior root.
             sys.modules.pop(constants.Config.MODULE, None)
-            for dep in _config_module_deps:
-                sys.modules.pop(dep, None)
-            _config_module_deps.clear()
+            if reload_dependencies or _config_module_deps_root != project_root:
```
FarhanAliRaza: The `_config_module_deps_root != project_root` branch is not exercised. With the comparison removed, all 118 tests in `tests/units/test_config.py` still pass. `test_get_config_evicts_dependencies_from_another_project` calls `_get_config()` with the default `reload_dependencies=True`, so it never reaches this check.

Either drop `_config_module_deps_root` or add a test that reloads in one context after the cwd moved to a second project. Note that in that scenario a same-named state module would still hit the shadow error, so the guard may not buy anything.

tests/units/test_config.py
```diff
@@ -1029,20 +1029,118 @@ def test_get_config_accepts_explicit_project_root(
...
+        (project / "config_reload_dependency.py").write_text(
+            f"APP_NAME = {app_name!r}\n"
+        )
+        (project / "rxconfig.py").write_text(
+            "import config_reload_dependency\nimport reflex as rx\n\n"
+            "config = rx.Config(app_name=config_reload_dependency.APP_NAME)\n"
+        )
+
+    assert reflex_base.config._get_config(first_project).app_name == "first"
+    assert reflex_base.config._get_config(second_project).app_name == "second"
```
FarhanAliRaza: Both calls use `reload_dependencies=True`, so this passes on `main` too (after the fixture is adjusted) and does not cover the new root check. To cover it, load `first_project` into a context, then call `reload_config()` from `second_project` in the same context.

FarhanAliRaza (changes requested):
Tested with the `appmod.py` / `rxconfig.py` pair from #7028.

On `main`, `get_config()` followed by `reload_config()` in one `RegistrationContext` raises `StateValueError`. The new regression test fails there with that error and passes on this branch. `ruff`, `pyright`, and `tests/units/test_config.py` pass on the branch.

I also ran the same reload on a forked context, the shape `AppHarness` uses. That still raises the same `StateValueError`. I removed the `_config_module_deps_root` comparison and re-ran the config tests to check the root guard. They all pass without it.

harsh21234i:
Addressed the remaining review feedback in commit 4d8f57fb.

- Config reloads now restore only project-local modules that define states already registered in the active RegistrationContext. Ordinary imported config dependencies are evicted and re-imported, so edited values are not stale.
- Added a same-project edited-dependency regression assertion.
- Extracted the test module identifiers into named constants.
- The forked-context, older-context, fresh-context, project-switch, and edited-dependency cases are covered.

Verification:
- Focused reload tests: 4 passed.
- Full tests/units/test_config.py: 117 passed; 2 unrelated deprecation tests could not run in my borrowed environment because granian is unavailable.
- Ruff check, formatting, and git diff --check pass.

Please re-review the latest commit.

packages/reflex-base/src/reflex_base/config.py
```diff
@@ -1021,6 +1040,7 @@ def reload_config() -> Config:
         The freshly loaded app config.
     """
     ctx = RegistrationContext.ensure_context()
-    config = _get_config()
+    preserve_dependencies = bool(ctx._config_module_deps or ctx.base_states)
+    config = _get_config(reload_dependencies=not preserve_dependencies)
```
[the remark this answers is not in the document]
harsh21234i: Fixed in 4d8f57fb. Reload now restores only modules that define states already registered in the active context; ordinary project-local dependencies are evicted and re-imported. The regression test edits the dependency in place and verifies the updated config value is loaded.

tests/units/test_config.py
```diff
@@ -1029,14 +1029,169 @@ def test_get_config_accepts_explicit_project_root(
...
+            assert reflex_base.config.reload_config().app_name == "state_reload"
+
+
+def test_get_config_reloads_project_state_for_each_context(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch, clean_config_modules: None
+):
+    """Initial config loads register project state in each context.
+
+    Args:
+        tmp_path: The pytest tmp_path fixture.
```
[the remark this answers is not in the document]
harsh21234i: Fixed in 4d8f57fb. The test module identifiers are now named constants and reused by the generated fixtures and assertions.

harsh21234i:
hey @FarhanAliRaza  please  take a look whenever you get time :)

PR #7036, opened 2026-09-02
perf(state): skip unchanged scalar writes and resolve stale computed vars lazily

FarhanAliRaza:
## Problem

Every `__setattr__` on a state var runs the full computed-var invalidation cascade immediately, restarting from the whole `dirty_vars` set. N assignments in one handler cost N graph walks. There is no equality check, so `self.x = self.x` dirties `x`, invalidates every dependent computed var, and ships everything. A computed var that recomputes to an equal value still cascades to its dependents and is re-sent.

## Change

- **Scalar cutoff in `__setattr__`.** Assigning an `int`, `float`, `str`, `bool`, or `None` equal to the stored value returns before anything is marked dirty. Objects are never compared, so reassigning a mutated object still marks it dirty.
- **Stale marking instead of cache deletion.** Dependents go into a per-instance `_stale_computed_vars` set. Marking stops at a var that is already stale, so repeated writes in one handler do not re-walk the graph.
- **Lazy resolution.** A stale var is settled on read or in `get_delta`: it resolves stale computed dependencies first, recomputes only when a dependency is in `dirty_vars`, and keeps the cache when the new value is equal. That stops the cascade at that point. Unresolved stale vars drop their cache in `_clean`, so the next event recomputes them as before.
- **Class-level sets** for frontend computed vars and interval computed vars replace the per-call scans in `get_delta`.
- **`is_hydrated` is always sent.** The client resets it locally on navigation and flushes queued events only when a delta carries it back, so those writes bypass the cutoff.
- Async computed vars keep the delete-and-resend path since they cannot be resolved on a sync read.

Names already in `dirty_vars` are still sent unconditionally, so callers that add a computed var name to force delivery keep working.

## Benchmarks

New file `tests/benchmarks/test_state_dirty.py`. Medians on one machine:

| Benchmark | Before | After |
|---|---|---|
| 200 unchanged writes, 50 dependents | 35.8 ms | 0.33 ms |
| 200 changed writes, 50 dependents | 35.9 ms | 3.0 ms |
| 200 writes, no dependents | 2.56 ms | 2.08 ms |
| chain write + get_delta | 1.05 ms | 0.66 ms |
| fan-out write + get_delta | 0.80 ms | 0.85 ms |
| get_delta, nothing dirty | 10 µs | 8 µs |

## Behavior changes a reviewer should check

1. `reset()` dirties only vars whose value actually changed. Substates already at defaults no longer appear in `dirty_substates`.
2. A cached computed var that reads a `cache=False` var is still recomputed every delta but only sent when its value changes.
3. Five existing tests asserted the old eager marking or re-sending of unchanged values; they now assert the stale set and the smaller deltas.

## Tests

- Three new regression tests were committed failing first: unchanged write is not dirty, unchanged write keeps the dependent cache, equal computed value stops the cascade.
- Two guard tests pin the semantics that must survive: a read between writes sees each fresh value, and several writes then one read run the getter once.

https://claude.ai/code/session_01U3KUBonGWkaQm2D7sWKQ5T

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/7036?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/news/+dirty-tracking-cutoff.performance.md
+++ b/news/+dirty-tracking-cutoff.performance.md
@@ -1,1 +1,1 @@
-Skip dirty tracking when a scalar state var is assigned the value it already holds, and stop re-sending a cached computed var whose recomputed value is unchanged. Repeated writes in one event handler no longer re-walk the computed var dependency graph.
+Stop re-sending a cached computed var whose recomputed value is unchanged, and stop re-walking the computed var dependency graph on every write within one event handler.
--- a/news/+dirty-tracking-cutoff.performance.md
+++ b/news/+dirty-tracking-cutoff.performance.md
@@ -1,0 +1,1 @@
+Skip dirty tracking when a scalar state var is assigned the value it already holds, and stop re-sending a cached computed var whose recomputed value is unchanged. Repeated writes in one event handler no longer re-walk the computed var dependency graph.
--- a/packages/reflex-base/news/+dirty-tracking-cutoff.performance.md
+++ b/packages/reflex-base/news/+dirty-tracking-cutoff.performance.md
@@ -1,0 +1,1 @@
+Cached computed vars are now resolved lazily: a stale value is recomputed only when one of its dependencies actually changed, and an unchanged result keeps the cache and does not cascade to dependents.
--- a/packages/reflex-base/src/reflex_base/utils/types.py
+++ b/packages/reflex-base/src/reflex_base/utils/types.py
@@ -185,7 +185,13 @@ PrimitiveToAnnotation = {
     dict: Dict,  # noqa: UP006
 }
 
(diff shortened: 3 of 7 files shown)


masenf:
this is definitely a behavior change that will affect some apps i've seen in the wild that use self-assignment to manually trigger a cascade of dependent computed vars  (which might have dependencies that cannot be tracked by the state system). Obviously the solution for those cases is to just explicitly mark the affected vars dirty, but the API for doing that is not currently very nice.

FarhanAliRaza:
@masenf droped that. Other changes are worth a look i think.

PR #7040, opened 2026-09-03
[HTTPXodus] migrate httpx to httpx2 (hard switch)

### LINKED ISSUE #7034: [HTTPXodus] Consider migrating from `httpx` to `httpx2` (the actively maintained fork)
**Title:** [HTTPXodus] Consider migrating from `httpx` to `httpx2` (the actively maintained fork)

**Body:**

HTTPXodus — The Great HTTP Ecosystem Migration

> 🏷️ *This issue is part of **HTTPXodus**, a community effort to help major projects plan their path off the stalled `httpx` stable line onto [`httpx2`](https://github.com/pydantic/httpx2), the actively maintained fork by Pydantic Services (with original httpx author Tom Christie involved). One coordinated issue per project — no spam, no drive-by PRs.*

## Background — what happened to httpx

`httpx` is one of the most important HTTP clients in the Python ecosystem (370k+ dependent repositories). However:

- The last **stable** release is **0.28.1 (December 2024)** — over 20 months ago.
- For most of 2025 the repository was dormant; maintenance trackers (e.g. Snyk) now flag it as **inactive**.
- Encouragingly, the project has recently resumed activity: `1.0.dev4–dev6` shipped between **2026-08-19 and 2026-08-31**. A 1.0 is clearly in the works, but there is **no committed date for a stable 1.0**, and the dev line carries breaking changes.

Meanwhile, the ecosystem has already begun moving. Projects that have adopted `httpx2`:

| Project | Stars | Status |
|---------|-------|--------|
| [Starlette](https://github.com/encode/starlette) | 12.5k+ | ✅ Migrated |
| [FastAPI](https://github.com/fastapi/fastapi) | 102k+ | ✅ Dual support (httpx2 test dependency) |
| [OpenAI Python SDK](https://github.com/openai/openai-python) | 31.5k+ | ✅ Migrated |
| [Anthropic Python SDK](https://github.com/anthropics/anthropic-sdk-python) | 3.9k+ | ✅ Migrated in v1.0 |
| [MCP Python SDK](https://github.com/modelcontextprotocol/python-sdk) | 24k+ | ✅ Migrated in v2.0 |

## What is httpx2

[`httpx2`](https://github.com/pydantic/httpx2) is a fork of httpx 0.28.1 maintained by **Pydantic Services Inc.**, with Tom Christie involved. It is actively released (v2.0.0 → v2.12.0 since May 2026) and keeps a **compatible public API**:

```python
import httpx2
r = httpx2.get("https://example.org")
```

## Why migrate — the benefits

1. **Active maintenance** — regular releases, reviewed PRs, and a funded maintainer team, versus an uncertain stable-release timeline.
2. **Modern TLS by default** — certificates are verified against the **OS trust store** instead of the bundled `certifi`, removing a common source of stale-CA issues.
3. **New capabilities** — built-in Server-Sent Events (`client.sse()`) and WebSocket support (`httpx2[ws]`), without extra dependencies.
4. **Ecosystem alignment** — as frameworks and SDKs migrate, staying on httpx increasingly means duplicate HTTP stacks in one dependency tree.

## Risks of staying on httpx

- **Security exposure**: no stable-line releases means no stable-line security fixes. If a CVE lands in 0.28.x today, there is no maintained branch to patch it.
- **Dependency conflicts**: packages that pin `httpx<1.0` already conflict with migrated peers; the longer the wait, the worse the resolver pain for downstream users.
- **Compounding migration cost**: the gap between 0.28.x and whatever 1.0 becomes keeps growing; migrating to httpx2 now is a small, well-documented step (official migration guide: https://pydantic.dev/docs/httpx2/get-started/migration/).

## What migration could look like here

The good news first: in Reflex, **httpx is used only by internal framework / CLI tooling — never by end-user app code** (user apps are served over ASGI via Starlette/Granian; httpx never enters the request path). All call sites are **synchronous** — there is no `AsyncClient` anywhere in the package — and most are lazily imported inside functions. Verified against the 0.9.10 sdist:

- **`reflex/utils/net.py`** — the central helper. `_httpx_client()` builds one shared `httpx.Client` with a custom `httpx.HTTPTransport(local_address=..., verify=...)` and per-scheme `mounts` that set `httpx.HTTPTransport(proxy=httpx.Proxy(url=url), verify=...)`. It also powers plain `httpx.head()` connectivity probes (e.g. `http://1.1.1.1`, IPv6) and wraps `get`/`post`/`head` for the rest of the CLI.
- **`reflex/utils/telemetry.py`** — `_send_event()` fires anonymous usage telemetry to PostHog via `httpx.post(POSTHOG_API_URL, ...)` (all exceptions swallowed).
- **`reflex/utils/js_runtimes.py`**, **`reflex/utils/templates.py`**, **`reflex/utils/frontend_skeleton.py`** — download Node/Bun runtimes, app templates, and the frontend skeleton; each catches `httpx.HTTPError`.
- **`reflex/utils/registry.py`** and **`reflex/custom_components/custom_components.py`** — query / publish to the custom-component registry via `httpx.post`, checking `response.status_code == httpx.codes.FORBIDDEN`.

Because the blast radius is a handful of internal, sync-only call sites, this is one of the lower-risk migrations in the ecosystem. One concrete thing to verify up front (it matters for **either** option below):

> ⚠️ **`net.py` reaches into a private module**: `from httpx._utils import get_environment_proxies`. That underscore module is not part of httpx's public API, so before switching you'll want to confirm `httpx2` still ships an equivalent helper (and that the `mounts=` / `httpx.Proxy(...)` construction behaves the same). This is the single non-mechanical change in an otherwise drop-in migration.

Two viable paths, depending on your compatibility goals:

- **Option A — dual support (recommended for libraries)**: prefer httpx2 when available, fall back to httpx:

  ```python
  try:
      import httpx2 as httpx
  except ModuleNotFoundError:
      import httpx
  ```

  with `httpx2; python_version >= "3.10"` added as an extra/optional dependency. This keeps Reflex installable alongside apps that still pin `httpx`.

- **Option B — hard switch (fine here too)**: replace `httpx >=0.26,<1.0` with `httpx2` and switch imports to `import httpx2 as httpx`. Because every call site is internal and synchronous, this is genuinely low-risk for Reflex — arguably cleaner than dual support, since nothing in the public API re-exports httpx types. Requires Python ≥ 3.10, which Reflex already mandates (`>=3.10,<4.0`).

⚠️ **One caveat worth calling out**: httpx2 verifies TLS against the **OS trust store** instead of the bundled `certifi`. This is especially relevant to Reflex because `net.py` already has first-class proxy / custom-`verify` handling for exactly the corporate-proxy and locked-down-container users this change affects — those environments may need `SSL_CERT_FILE` or system CA configuration after the switch. Worth a line in the changelog either way.

## Notes

- **Current constraint**: `httpx >=0.26,<1.0` — a **runtime** dependency in `[project.dependencies]` (`pyproject.toml`, line 25), not a dev/optional extra. **Python requirement**: `>=3.10,<4.0`, so httpx2's Python ≥ 3.10 floor is already satisfied. (Verified against the published 0.9.10.post1 sdist.)
- **Ecosystem alignment, close to home**: Reflex already depends on `starlette >=1.3.1`, and Starlette has adopted httpx2 (in its TestClient). Migrating keeps Reflex on the same HTTP line as one of its own core dependencies.
- **Sibling distribution**: `reflex-hosting-cli` (a separate package Reflex depends on) independently pins `httpx >=0.25.1,<1.0`. A complete migration story would want to cover it too; `reflex-base` has **no** httpx dependency.
- **No duplication**: as of 2026-09-02 there is no existing httpx2-related issue or PR in this repository that we could find.

Happy to open a PR for either option if the maintainers are interested — and equally happy to close this if you'd rather wait for httpx 1.0 stable. No pressure; the goal is just to make sure the decision is an informed one. 🙏

ProgrammerPlus1998:
Closes #7034

> 🏷️ *Part of [HTTPXodus](https://github.com/ProgrammerPlus1998/httpxodus) — a community effort to help major Python projects plan their path off the stalled `httpx` stable line onto [`httpx2`](https://github.com/pydantic/httpx2), the actively maintained fork by Pydantic Services.*

## What this PR does

**Complete hard switch** of the main `reflex` package from `httpx` to `httpx2`:
- Direct `import httpx2` everywhere — no `try/except ImportError` fallback, no `import httpx2 as httpx` aliasing
- Removes the `httpx` runtime dependency; `httpx2>=2.12.0` unconditional (`requires-python` is already `>=3.10`, so no version-marker needed)
- `httpx.Client`, `httpx.AsyncClient`, `httpx.Limits`, `httpx.HTTPTransport`, `httpx.Proxy`, `httpx.HTTPError` → `httpx2.X`
- `from httpx._utils import get_environment_proxies` → `from httpx2._utils import get_environment_proxies` (httpx2 ships the same helper)

## Diff summary

**12 files** (commit `e71e2e5`):

| Area | Change |
|---|---|
| `reflex/utils/` (net, telemetry, registry, js_runtimes, templates, frontend_skeleton) | `import httpx2 as httpx` → `import httpx2`; all usages renamed |
| `reflex/custom_components/custom_components.py` | Same |
| `tests/units/test_telemetry.py`, `tests/units/utils/test_utils.py`, `tests/test_node_version.py` | Remove the last two `ModuleNotFoundError` fallbacks; patch targets and mock dependency data updated |
| `pyproject.toml` | Drop `httpx >=0.26,<1.0`; `httpx2 >=2.12.0` without version marker |
| `uv.lock` | Record the httpx2 swap; also repairs a corrupt duplicate `ty` entry (0.0.69 pointing at truststore artifacts) that current uv releases refuse to parse |

## Scope

`packages/reflex-hosting-cli` (a separate distribution that pins `httpx <1.0`) and `packages/reflex-site-shared` are unchanged — they keep httpx for now, so httpx remains installed transitively via `reflex-hosting-cli`. Migrating the hosting CLI (50+ lazy imports in `hosting.py` alone) is a natural follow-up PR.

## Test results

- Full `pytest tests/units`: **7997 passed**, 18 skipped, in a fresh venv installed from this branch's `uv.lock` (`uv sync --frozen`) with httpx2 2.12.0
- The 163 failures are all in `tests/units/reflex_cli/` (the hosting-CLI sub-package this PR deliberately does not touch): they fail with "Reflex version 0.0.0.post2.dev0 is not compatible with reflex-hosting-cli", an artifact of the local editable install having no git tags (uv-dynamic-versioning fallback), not of this change

## Notes for reviewer

- ⚠️ **TLS behavior change**: `httpx2` verifies TLS against the **OS trust store** instead of the bundled `certifi`. Self-hosted Reflex deployments behind corporate proxies or in minimal containers that relied on certifi's CA bundle may need `SSL_CERT_FILE` / `SSL_CERT_DIR` after the switch. Worth a line in the changelog.
- The `uv.lock` repair (corrupt `ty` entry) is included here because the dependency swap requires re-locking anyway; happy to split it out if you prefer.

Happy to revise per review — and equally happy to close this PR if the maintainers would rather wait for `httpx` 1.0 stable. 🙏

### DIFF
--- a/reflex/custom_components/custom_components.py
+++ b/reflex/custom_components/custom_components.py
@@ -634,7 +634,10 @@ def _collect_details_for_gallery():
     Raises:
         SystemExit: If pyproject.toml file is ill-formed or the request to the backend services fails.
     """
-    import httpx
+    try:
+        import httpx2 as httpx
+    except ModuleNotFoundError:
+        import httpx
     from reflex_cli.utils import hosting
 
     console.rule("[bold]Authentication with Reflex Services")
--- a/reflex/custom_components/custom_components.py
+++ b/reflex/custom_components/custom_components.py
@@ -634,10 +634,7 @@ def _collect_details_for_gallery():
     Raises:
         SystemExit: If pyproject.toml file is ill-formed or the request to the backend services fails.
     """
-    try:
-        import httpx2 as httpx
-    except ModuleNotFoundError:
-        import httpx
+    import httpx2
     from reflex_cli.utils import hosting
 
     console.rule("[bold]Authentication with Reflex Services")
@@ -669,18 +666,18 @@ def _collect_details_for_gallery():
         # Send a POST request to achieve two things at once:
         # 1. Check if the package is already shared by the user. If not, the backend will return 403.
         # 2. If this package is not shared before, this request records the package name in the backend.
-        response = httpx.post(
+        response = httpx2.post(
             post_custom_components_gallery_endpoint,
             headers={"Authorization": f"Bearer {access_token}"},
             data=params,
         )
-        if response.status_code == httpx.codes.FORBIDDEN:
+        if response.status_code == httpx2.codes.FORBIDDEN:
             logger.error(
                 f"{package_name} is owned by another user. Unable to update the information for it."
             )
             raise SystemExit(1)
         response.raise_for_status()
-    except httpx.HTTPError as he:
+    except httpx2.HTTPError as he:
         logger.error(f"Unable to complete request due to {he}.")
         raise SystemExit(1) from None
 
@@ -707,7 +704,7 @@ def _collect_details_for_gallery():
     # Now send the post request to Reflex backend services.
     try:
         logger.debug(f"Sending custom component data: {params}")
-        response = httpx.post(
+        response = httpx2.post(
             post_custom_components_gallery_endpoint,
             headers={"Authorization": f"Bearer {access_token}"},
             data=params,
@@ -716,7 +713,7 @@ def _collect_details_for_gallery():
         )
         response.raise_for_status()
 
-    except httpx.HTTPError as he:
+    except httpx2.HTTPError as he:
         logger.error(f"Unable to complete request due to {he}.")
         raise SystemExit(1) from None
 
--- a/news/7040.breaking.md
+++ b/news/7040.breaking.md
@@ -1,0 +1,1 @@
+Replace the runtime HTTP dependency `httpx` with `httpx2` (the actively maintained fork). TLS verification now uses the OS trust store instead of the bundled `certifi` CA bundle; deployments behind corporate proxies or in minimal containers may need `SSL_CERT_FILE` / `SSL_CERT_DIR` set.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,7 @@ dependencies = [
   "click >=8.2",
   "granian[reload] >=2.7.4",
   "httpx >=0.26,<1.0",
+  "httpx2 >=2.0; python_version >= '3.10'",
   "packaging >=24.2,<27",
   "psutil >=7.0.0,<8.0; sys_platform == 'win32'",
   "python-multipart >=0.0.32,<1.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,8 +22,7 @@ requires-python = ">=3.10,<4.0"
 dependencies = [
   "click >=8.2",
   "granian[reload] >=2.7.4",
-  "httpx >=0.26,<1.0",
-  "httpx2 >=2.0; python_version >= '3.10'",
+  "httpx2 >=2.12.0",
   "packaging >=24.2,<27",
   "psutil >=7.0.0,<8.0; sys_platform == 'win32'",
   "python-multipart >=0.0.32,<1.0",
--- a/reflex/utils/frontend_skeleton.py
+++ b/reflex/utils/frontend_skeleton.py
@@ -141,7 +141,10 @@ def initialize_agents_md(
     """
     plan = _plan_agents_md(agents_file, claude_file)
 
-    import httpx
+    try:
+        import httpx2 as httpx
+    except ModuleNotFoundError:
+        import httpx
 
     logger.debug(f"Fetching {url}")
     try:
--- a/reflex/utils/frontend_skeleton.py
+++ b/reflex/utils/frontend_skeleton.py
@@ -141,16 +141,13 @@ def initialize_agents_md(
     """
     plan = _plan_agents_md(agents_file, claude_file)
 
-    try:
-        import httpx2 as httpx
-    except ModuleNotFoundError:
-        import httpx
+    import httpx2
 
     logger.debug(f"Fetching {url}")
     try:
         response = net.get(url, timeout=5)
         response.raise_for_status()
-    except httpx.HTTPError as e:
+    except httpx2.HTTPError as e:
         logger.warning(f"Failed to fetch AGENTS.md from {url} due to {e}. Skipping.")
         return
 
--- a/reflex/utils/js_runtimes.py
+++ b/reflex/utils/js_runtimes.py
@@ -234,7 +234,10 @@ def download_and_run(url: str, *args, show_status: bool = False, **env):
     Raises:
         SystemExit: If the script fails to download.
     """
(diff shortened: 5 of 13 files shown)


reflex/custom_components/custom_components.py
```diff
@@ -634,7 +634,10 @@ def _collect_details_for_gallery():
     Raises:
         SystemExit: If pyproject.toml file is ill-formed or the request to the backend services fails.
     """
-    import httpx
+    try:
+        import httpx2 as httpx
+    except ModuleNotFoundError:
```
FarhanAliRaza: we can just outright replace httpx with httpx2 no need for this fallback.

masenf (commented):
i dont understand how the "dual import" thing really helps us here.

if we're adding httpx2 to the framework deps, it will always be installed and thus we should just import httpx2.

if downstream code was depending on httpx and not declaring it as their own dependency, then they will be broken, but i'm okay with that. downstream should not be assuming that the framework dependencies can just be used because they happen to be there.

ProgrammerPlus1998:
Both greptile findings addressed in bccf225:

- **P1 (lockfile)**: removed the stale `testing` extra block from the `reflex` entry in `uv.lock` — it was carried over from upstream's lockfile; the branch manifest only declares `db`/`pydantic` extras. Verified with `uv sync --frozen` + full unit suite (7997 passed).
- **P2 (news fragment)**: added `news/7040.breaking.md` describing the httpx2 switch and the TLS trust-store behavior change.

PR #7049, opened 2026-09-05
Reduce dev startup memory by deferring unused integrations

Alek99:
Reflex dev startup imports optional data, plotting, image, database, and admin integrations even when the app does not use them. Spawned Granian workers also cannot reuse app modules preloaded by the supervisor, leaving an unnecessary copy resident between reloads.

Resolve optional serializers from already-loaded class identities, defer unused database/admin imports, and preload the app only for forked Granian workers. Preserve custom serializer overrides, exact-type/subclass precedence, runtime annotations, and cold direct-SQLModel relationship data.

This PR now also includes the backend-only follow-up:
- State mutation tracking recognizes already-loaded model classes without eagerly importing SQLAlchemy. Model import order, cached classification, and explicit/wildcard access to `MUTABLE_TYPES` remain supported.
- Move the pure `write_file` helper into the lightweight path-operations module, preserving its old compiler reexport. This keeps the compiler/state dependency graph out of the backend supervisor.

### Latest backend-only measurements

Database-free blank app, Granian dev mode, macOS on Apple M5 Pro, Python 3.14.5, with optional dependencies installed. These are **additional improvements over this PR's initial commit**, measured again on both sides:

| Metric | Initial commit, remeasured | With follow-up | Additional improvement |
| --- | ---: | ---: | ---: |
| Startup process-tree RSS | 209.11 MiB | 163.95 MiB | 21.6% lower |
| Startup to ready | 456.1 ms | 336.5 ms | 26.2% faster |
| Python reload to ready | 524.8 ms | 398.4 ms | 24.1% faster |
| Post-reload process-tree RSS | 209.13 MiB | 164.02 MiB | 21.6% lower |

Each side uses 5 fresh starts and 15 reloads; values are medians. The final source was verified unchanged after benchmarking. Startup ranged from 332.6–351.9 ms, reloads from 380.4–421.2 ms, and post-reload RSS from 163.91–164.28 MiB.

After the regression fixes in `b2bf7009`, a fresh 5-start/15-reload check measured **164.00 MiB startup RSS, 338.6 ms startup, 400.5 ms reload, and 164.03 MiB post-reload RSS**. This retains the earlier savings; it is a separate-session sanity check, not an interleaved A/B benchmark.

After the second-audit fixes, an alternating same-session comparison against `b2bf7009` (5 fresh starts and 15 reloads per revision, separate excluded warm-ups) measured **377.35 → 372.56 ms startup, 428.70 → 429.55 ms reload, 164.66 → 164.34 MiB startup RSS, and 164.59 → 164.41 MiB post-reload RSS**. This small sample shows no noticeable regression; it does not establish a new speedup. Parent and spawned-worker import paths were verified for each revision. A separate final standalone run measured 163.89 MiB startup RSS.

For context, the earlier measurement of the first commit showed:

| Metric | Original framework | Initial PR commit |
| --- | ---: | ---: |
| Process-tree RSS | 375.27 MiB | 209.13 MiB |
| Startup to ready | 0.906 s | 0.446 s |
| Python reload to ready | 1.070 s | 0.533 s |

The original baseline was `c57b32cc8b0cbcb61ffe303b18a94f5cb3ca1a10`; the follow-up baseline was `cc9107032f71aa83b31427764d3cd2a0694899ec`. The initial-commit values above come from separate benchmark passes, so small timing differences are expected.

Startup waits for `/ping`; reload timing starts at a Python source-file touch and waits for a replacement worker and a successful `/ping`. The compile-skip marker is removed before each fresh start. Telemetry and version checks are disabled on both sides.

RSS sums the supervisor, backend worker, and multiprocessing helper; it excludes Vite/browser memory and can double-count shared pages. Savings depend on installed and used integrations; database-using apps still pay their database import cost. These are fresh-process launches with warm filesystem caches. Linux/fork, other Python versions, large-app compile time, production performance, and long-duration leak freedom have not been established.

### Validation

The latest fixes address all eight findings from the second adversarial audit:
- Preserve SQLModel relationship data in cold state/event payloads, and preserve the historical serializer import, keyword argument, annotations, and pickle path.
- A later compatibility import of `reflex.model` no longer silently replaces a custom SQLModel serializer or broad custom fallback. This deliberately removes the old import-order-dependent override; both cases have regression coverage.
- Recognize actual optional-library classes by identity, so unrelated user packages named `pandas` do not trigger dependency imports.
- Resolve defaults without importing libraries or rebuilding the registry. This removes the global serializer lock, destructive reordering, custom-hash lock inversion, and reentrant registration loss.
- Finish Plotly's deferred JSON/engine initialization before fork only when Plotly Figure/Template classes are already loaded. Regression tests overlap fork with Template, JSON-helper, and optional `orjson` imports.
- Count directly imported SQLModel database usage without loading the unused database integration.
- Bound all serializer subprocess tests with timeouts.

Validation of the current fixes:
- Final full unit suite: **8,360 passed, 19 skipped; 75.96% coverage** on Python 3.14.5.
- Focused serializer modules: **88 passed, 1 skipped** (the explicit `orjson` case skips because that package is not installed in the Python 3.14 environment).
- Two additional test-only broad-fallback cases subsequently passed; the expanded cold-process module has **15 passing cases and 1 optional-orjson skip** on Python 3.14.
- Final cold-process regression module: **16 passed** on Python 3.12.13 with `orjson` installed; the JSON-helper and `orjson` fork cases also passed **8 fresh-process repeats each**.
- Plotly auto/json/orjson configurations and simulated unavailable-orjson fallback passed. No installed dependencies were changed for the fallback checks.
- Full Playwright suite: **253 passed**.
- Docs suite on Python 3.12: **419 passed, 1 skipped, 1 expected failure**.
- Real docs backend: **8 WebSocket assertions across initial startup and 3 worker reloads**.
- Public counter, SQLite CRUD, and Excel/pandas/SQLModel data-visualization apps: **all 8 dev/production runs passed**, including browser interactions, reload/state survival, and a separate fresh Excel ingestion fixture.
- **100 seeded base/current comparisons (6,700 mixed operations per revision)** matched exactly. Missing optional dependencies, annotations, pickle, and repeated module reload checks passed; a real user app named `pandas` now compiles on both revisions.
- Independent review also matched **624 optional/default/override inheritance combinations** against the original base.
- All configured pre-commit hooks passed: Ruff, codespell, stub generation, Pyright, ty, and Biome.
- Regression failures were captured before implementing their fixes, including removing only the final warm-up call to reproduce the `orjson` hang.
- News fragments included for both touched packages.

The full Playwright, docs, and public-app runs preceded the final narrowly scoped Plotly JSON-engine warm-up adjustment. The full unit suite, focused cross-environment fork tests, and startup benchmark were rerun after it.

Earlier broad validation on `46f9c7bf` included large-state apps through 10,000 fields and normal/strict Granian plus Uvicorn reloads. Its selected state/event matrix reached **98 distinct passing cases and 10 skips across initial and isolated runs**; failures in the initial combined batch also reproduced on base (including macOS select-all and fixture-isolation issues).

CI on preceding head `b2bf7009` finished with **109 successes and 2 skips**, including docs production checks. New-commit CI must run separately. Local full docs frontend/browser validation still requires Enterprise login; backend and docs-unit checks are not substitutes for it.

The fork tests cover the identified deferred-import races, not a blanket guarantee that arbitrary third-party code is safe to fork from a multithreaded process. Direct mutation of internal serializer registries is not a supported compatibility contract.

### Changes to core features

- [x] Explained the change and its motivation.
- [x] Added regression tests and ran the unit suite locally.
- [x] Reviewed the patch adversarially and addressed compatibility findings.
- [x] Checked related open PRs; #6930 is related import work and also touches `types.py`.

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/7049?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/news/+dev-mode-imports.performance.md
+++ b/news/+dev-mode-imports.performance.md
@@ -1,0 +1,1 @@
+Reduce development startup and reload memory by deferring unused database and admin integrations and avoiding redundant app preloads in spawned Granian supervisors.
--- a/news/+dev-mode-imports.performance.md
+++ b/news/+dev-mode-imports.performance.md
@@ -1,1 +1,1 @@
-Reduce development startup and reload memory by deferring unused database and admin integrations and avoiding redundant app preloads in spawned Granian supervisors.
+Reduce development startup and reload time and memory by deferring unused database, admin, and compiler imports in the backend launcher and state mutation tracking, and by avoiding redundant app preloads in spawned Granian supervisors.
--- a/packages/reflex-base/src/reflex_base/utils/serializers.py
+++ b/packages/reflex-base/src/reflex_base/utils/serializers.py
@@ -8,6 +8,7 @@ import functools
 import inspect
 import json
 import logging
+import os
 import uuid
 import warnings
 from collections.abc import Callable, Mapping, Sequence
@@ -37,6 +38,15 @@ SERIALIZER_TYPES: dict[type, type] = {}
 _SERIALIZER_LOCK = RLock()
 _OPTIONAL_SERIALIZER_LOADERS: dict[str, Callable[[], None]] = {}
 
+if hasattr(os, "register_at_fork"):
+    # Wait for other threads to finish registry operations before forking.
+    # The surviving thread owns the lock, so both processes can release it.
+    os.register_at_fork(
+        before=_SERIALIZER_LOCK.acquire,
+        after_in_parent=_SERIALIZER_LOCK.release,
+        after_in_child=_SERIALIZER_LOCK.release,
+    )
+
 SERIALIZED_FUNCTION = TypeVar("SERIALIZED_FUNCTION", bound=Serializer)
 _REGISTRY_VALUE = TypeVar("_REGISTRY_VALUE")
 
@@ -58,10 +68,12 @@ def _load_optional_serializer(type_: type) -> None:
         type_: The value type that may belong to an optional dependency.
     """
     for value_type in getattr(type_, "__mro__", ()):
-        module = value_type.__module__.partition(".")[0]
+        module_name = value_type.__module__
+        if not isinstance(module_name, str):
+            continue
+        module = module_name.partition(".")[0]
         if loader := _OPTIONAL_SERIALIZER_LOADERS.get(module):
             loader()
-            return
 
 
 @overload
--- a/packages/reflex-base/src/reflex_base/utils/serializers.py
+++ b/packages/reflex-base/src/reflex_base/utils/serializers.py
@@ -2,21 +2,24 @@
 
 from __future__ import annotations
 
+import base64
 import dataclasses
 import decimal
 import functools
 import inspect
+import io
 import json
 import logging
 import os
+import sys
 import uuid
 import warnings
 from collections.abc import Callable, Mapping, Sequence
+from contextlib import suppress
 from datetime import date, datetime, time, timedelta
 from enum import Enum
 from importlib.util import find_spec
 from pathlib import Path
-from threading import RLock
 from typing import Any, Literal, TypeVar, get_type_hints, overload
 from uuid import UUID
 
@@ -35,17 +38,9 @@ Serializer = Callable[[Any], SerializedType]
 
 SERIALIZERS: dict[type, Serializer] = {}
 SERIALIZER_TYPES: dict[type, type] = {}
-_SERIALIZER_LOCK = RLock()
-_OPTIONAL_SERIALIZER_LOADERS: dict[str, Callable[[], None]] = {}
-
-if hasattr(os, "register_at_fork"):
-    # Wait for other threads to finish registry operations before forking.
-    # The surviving thread owns the lock, so both processes can release it.
-    os.register_at_fork(
-        before=_SERIALIZER_LOCK.acquire,
-        after_in_parent=_SERIALIZER_LOCK.release,
-        after_in_child=_SERIALIZER_LOCK.release,
-    )
+_OPTIONAL_SERIALIZERS: dict[str, Serializer] = {}
+_OPTIONAL_SERIALIZER_TYPES: dict[str, type] = {}
+_OPTIONAL_SERIALIZER_MODULES: dict[str, tuple[str, ...]] = {}
 
 SERIALIZED_FUNCTION = TypeVar("SERIALIZED_FUNCTION", bound=Serializer)
 _REGISTRY_VALUE = TypeVar("_REGISTRY_VALUE")
@@ -61,19 +56,23 @@ deserializers = {
 }
 
 
-def _load_optional_serializer(type_: type) -> None:
-    """Load the optional serializer associated with a value type.
+def _get_optional_type_name(type_: type) -> str | None:
+    """Identify an optional type by identity in an already-loaded module.
 
     Args:
         type_: The value type that may belong to an optional dependency.
+
+    Returns:
+        The optional type name, or None for an unrelated type.
     """
-    for value_type in getattr(type_, "__mro__", ()):
-        module_name = value_type.__module__
-        if not isinstance(module_name, str):
-            continue
-        module = module_name.partition(".")[0]
-        if loader := _OPTIONAL_SERIALIZER_LOADERS.get(module):
-            loader()
+    name = getattr(type_, "__name__", None)
+    if not isinstance(name, str):
+        return None
+    for module_name in _OPTIONAL_SERIALIZER_MODULES.get(name, ()):
+        module = sys.modules.get(module_name)
+        if module is not None and vars(module).get(name) is type_:
+            return name
+    return None
 
 
 @overload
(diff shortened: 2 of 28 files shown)


packages/reflex-base/src/reflex_base/utils/serializers.py
```diff
@@ -181,6 +210,51 @@ def serialize(
     return serialized
 
 
+def _find_serializer(
```
FarhanAliRaza: `_find_serializer` has cyclomatic complexity 10 (`ruff --select C901` with `max-complexity=8` flags it). On `main`, `get_serializer` and `get_serializer_type` are 4 each. Please split the optional-MRO pass and the registry scan into two helpers so each stays under the threshold.

packages/reflex-base/src/reflex_base/utils/serializers.py
```diff
@@ -215,33 +278,26 @@ def get_serializer_type(type_: type) -> type | None:
...
-    serializer_for_type = get_serializer(type_)
-    return serializer_for_type is not None and (
-        into_type is None or get_serializer_type(type_) == into_type
+    if get_serializer(type_) is None:
+        return False
+    if into_type is None:
+        return True
+    serializer_type = get_serializer_type(type_)
+    return (
+        serializer_type == into_type or types.get_origin(serializer_type) == into_type
```
FarhanAliRaza: Matching `get_origin(serializer_type)` against `into_type` is not related to import deferral. It changes `can_use_in_object_var` from False to True for every SQLModel class and for any custom serializer annotated with a generic dict. ObjectVar attribute access on Model vars already works on `main` (checked with `State.author.name` on both branches). Please move this to its own PR, or state in the PR body why the lazy path needs it.

packages/reflex-base/src/reflex_base/utils/serializers.py
```diff
@@ -408,105 +464,164 @@ def serialize_color(color: Color) -> str:
...
+    "Image": ("PIL.Image",),
+    "SQLModel": ("sqlmodel.main",),
+}
+_OPTIONAL_SERIALIZER_ORDER = {
+    name: len(_INITIAL_SERIALIZER_TYPES) + index
+    for index, name in enumerate(_OPTIONAL_SERIALIZERS)
+}
+
+if hasattr(os, "register_at_fork"):
+    os.register_at_fork(before=_prepare_serializers_for_fork)
```
FarhanAliRaza: This installs a process-wide fork hook on import of `reflex_base`. It runs a Plotly `to_json` before every fork in any process that has `Figure` loaded, including user multiprocessing pools. Please call `_prepare_serializers_for_fork` explicitly from the Granian fork path in `reflex/utils/exec.py` instead of registering it globally.

reflex/istate/proxy.py
```diff
@@ -415,21 +416,49 @@ def mark_dirty(self):
...
 
-if find_spec("sqlalchemy"):
-    from sqlalchemy.orm import DeclarativeBase
+_MUTABLE_MODEL_BASES = (
+    ("sqlalchemy.orm.decl_api", "DeclarativeBase"),
+    ("pydantic.main", "BaseModel"),
+)
+
+
+def __dir__() -> list[str]:
```
FarhanAliRaza: Nothing in the repo or docs references `MUTABLE_TYPES` besides the new tests. If it stays for compatibility, `__getattr__` alone is enough. Please drop `__dir__` and the synthesized `__all__` branch.

news/+dev-mode-imports.performance.md
```diff
@@ -0,0 +1 @@
+Reduce development startup and reload memory by deferring unused database and admin integrations and avoiding redundant app preloads in spawned Granian supervisors.
```
FarhanAliRaza: This overlaps with `news/+backend-startup-followup.performance.md`. Please merge the two into one performance fragment.

FarhanAliRaza (dismissed):
Tested this PR against `main` with two real apps on Python 3.14 (Linux, forkserver start method), both in Granian dev mode.

The first app uses a pandas DataFrame, a DataFrame subclass with a custom `@rx.serializer(to=str)`, a Plotly figure, a PIL image, an `rx.Model` with a relationship list, a pydantic model, in-place mutation handlers, and a configured `admin_dash`. All values rendered, the custom serializer took precedence over the optional default, and the mutation handlers re-rendered without reassignment. A source edit triggered a Granian reload and served the new state.

The second app imports `sqlmodel` directly, never `rx.Model`, and reports the backend worker's loaded modules from an `on_load` handler. The worker had only `sqlalchemy` and `sqlmodel` loaded; pandas, plotly, PIL, starlette_admin and `reflex.model` were absent. On `main`, `import reflex.app` loads all of them. The relationship list rendered in the cold payload. `reflex export --no-zip` succeeded.

Unit tests for the touched modules pass. The order-dependent `test_upload_*` failures in `tests/units/test_app.py` reproduce on `main`. Ruff and pyright are clean.

FarhanAliRaza (approved):

PR #7054, opened 2026-09-06
Speed up deploy preparation and add opt-in frontend build reuse

Alek99:
Repeated deploys redo frontend dependency and archive work, and unchanged production frontends always rerun Vite. This PR reuses matching installed framework pins, avoids unnecessary manifest rewrites and sidecar recompression, and adds opt-in reuse of deterministic production frontend builds.

### Changes

- Retain equivalent package manifests without rewriting them, preserving main's deterministic dependency-only install fingerprint.
- Reuse matching framework pins after a successful persisted-lock install. Verify npm caret declarations against both lock metadata and installed package versions, use `--save-dev`, and include development dependencies under `NODE_ENV=production`.
- Cache archive exclusion identities and avoid recompressing frontend gzip/Brotli/Zstandard sidecars. Keep normal frontend/backend files deflated.
- Preserve main's per-package 24-hour version-check cache and one-hour failure cooldown; handle legacy/local and UTC timestamps, reject failed HTTP responses, and skip provider lookup when no provider prompt is needed. Required hosting compatibility checks remain active.
- Add `REFLEX_FRONTEND_BUILD_CACHE=true`, disabled by default. Cache pristine Vite output and replay post-build plugins, fallback handling, compression, and frontend-path processing. Validate inputs/output, invalidate changed inputs, fall back on unsupported/corrupt snapshots, and publish only after a successful complete build.

### Current rebase validation (`87ea648d`)

Rebased onto `main` at `77cfe61b` on September 8, 2026. The rebase preserves main's lazy CLI startup, dependency-only install fingerprint, per-package version checks, and failure cooldown.

- Production workspace locking now also works on Windows, even with build reuse disabled. It waits through contention, propagates other acquisition failures, supports nested same-thread calls, and releases after failed builds. Opt-in build reuse itself remains macOS/Linux only.
- Build fingerprints ignore the new per-package version-check and attempt timestamps, avoiding needless cache misses after routine metadata updates.
- Regression failures were captured before fixing Windows lock bypass and timestamp-driven cache invalidation. Separate spawned-process locking tests run on all CI platforms with caching both enabled and disabled.
- The final `87ea648d` follow-up only consolidates documentation and adds duration comments. Its Python AST is identical to the fully tested `b9230726` revision.
- Full final local unit suite: **8,479 passed, 18 skipped; 76.45% coverage** on Python 3.14.5. An earlier run alongside other checks hit a Redis lock-expiry timing failure; the isolated final suite passed without changes to the Redis implementation or that test.
- Focused rebase/cache/deploy suite: **341 passed**; final cache/prerequisites check after the platform-guard adjustment: **204 passed**.
- Real production build with caching enabled: **9 passed**, covering gzip/Brotli/Zstandard sidecars and HTTP content negotiation; 9 development-mode cases intentionally skipped.
- All repository pre-commit hooks passed: Ruff, codespell, stub generation, Pyright, ty, and Biome.

### Historical local deploy measurements (`93c8683`, before rebase)

These measurements predate this rebase. Main now includes independent startup and dependency-cache improvements, so these numbers are not a performance comparison against current main.

Same nine-page dashboard, two warmups and three fresh-process samples per variant, rotating run order. Times include imports, preparation, compilation/dependency handling, build or restore, postprocessing/compression, and both ZIPs. They exclude cloud authentication/provider APIs, upload, remote builds/rollout, optional version requests, telemetry delivery, interpreter teardown, and artifact verification.

| Unchanged frontend | Original | New default | Build cache enabled |
|---|---:|---:|---:|
| npm | 5.792 s | 2.533 s (56.3% less time) | 1.248 s (78.4% less time) |
| Bun | 2.551 s | 2.464 s (3.4% less time) | 1.244 s (51.2% less time) |

All unchanged candidate runs skipped dependency-manager commands; enabled-cache runs also skipped Vite. Python hash seeds vary between interpreters while the child build environment remains constant.

| Changed API/deploy URLs | Original | New default | Build cache enabled, miss |
|---|---:|---:|---:|
| npm | 5.787 s | 3.390 s | 4.188 s |
| Bun | 2.541 s | 2.510 s | 3.299 s |

Cache misses add about 0.8 seconds over the new default. The cache is for repeat deterministic builds on local macOS/Linux filesystems with reliable file timestamps; Windows bypasses it. External/clock/remote build inputs and excluded generated caches require a fresh build. Set `REFLEX_FRONTEND_BUILD_CACHE=false` to discard the snapshot and rebuild. Source/snapshot contents are hashed; installed dependencies use identity, size, mtime, and ctime. Documentation covers ignored private telemetry timestamps and symlink handling.

Measured on macOS arm64, Python 3.14.5, Node 26.0.0, npm 11.12.1, Bun 1.4.0 against base `c49a85df9c41ddffbe9fc2ac37d495d363f4e313`. No competing builds/tests ran during timing. These are warm local results, not total cloud deploy latency. Small differences can be noise. The original npm changed-URL samples were 8.489, 5.787, and 5.776 seconds; all are retained in the median. Avoiding sidecar recompression increased the earlier dashboard ZIP by 0.063%; upload impact was not measured.

### Earlier validation before rebase

- Full unit suite: **8,441 passed, 18 skipped, 76.38% coverage** (72% required).
- Production integration: **14 passed each with Bun and npm**, with caching enabled; real JS/CSS compression and HTTP negotiation plus Chromium chart rendering/state updates. Each manager skipped 9 development cases and deselected 5 development browser cases.
- Fixed-clock real exports cover unchanged repeats, URL/Python/CSS/SVG mutations, and dependency addition/removal. **96 cross-variant archive comparisons plus 48 reference checks** verify raw extracted payloads, ordered names, counts, modes, current assets/content, and actual cache hit/miss commands. npm comparisons permit dependency-section moves and `dev`/`devOptional` flags while requiring other manifest/lock metadata and resolved versions to agree.
- The archive matrix preceded the last symlink-redirect invalidation fix; its new failing-before/passing-after regression and the final full unit/production suites cover that correction. Earlier validation also covered clean `.web` restoration and npm `NODE_ENV=production`.
- Ruff, formatting, `pyright reflex tests`, whitespace checks, and patch application against the base pass. All applicable pre-commit hooks pass, including codespell, stub generation, Pyright, and ty. Independent review identified and resolved serializer and symlink edge cases. Regression tests, documentation, and news fragments for each touched package are included.

Full cloud deployment and arbitrary concurrent workspace writers (initialization/dev hot reload/external tools) were not exercised. The new concurrency regressions use separate processes with controlled compiler/build/ZIP test doubles; production behavior is also checked by the Bun/npm integration tests.

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/7054?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### Earlier review follow-up (`4da5154`, before rebase)

- Serialize initialized production exports from environment/URL configuration through compilation, setup, build/cache operations, and ZIP creation. Production/preview startup uses the same reentrant POSIX lock through frontend build and releases it before serving; direct builds and cache-disabled calls participate too. Forked children drop inherited lock ownership safely. Lock acquisition failures stop the operation. Initialization, dev hot reload, and unrelated workspace writers remain outside this scope.
- Bump the cache schema to discard snapshots created before locking.
- Expire successful version checks after 24 hours; expired, malformed, and future timestamps refresh. Preserve legacy local timestamps, write UTC for new checks, and keep network/HTTP/parse failures retryable.
- Added regressions for process overlap, distinct export URLs, ZIP interleaving, mixed cache settings, reentrancy/fork cleanup, unsafe/failed locks, expiry boundaries, timezone handling, and retries. **8,441 unit tests, 28 production integration checks, and all applicable pre-commit hooks pass.**

Candidate-only follow-up on the same warmed unchanged-frontend fixture (two warmups, three fresh-process runs per variant, alternating order):

| Manager | Default | Build reuse enabled |
|---|---:|---:|
| npm | 2.601 s | 1.253 s |
| bun | 2.527 s | 1.268 s |

All measured repeats skipped dependency-manager commands; enabled-cache repeats skipped Vite and reran compression. The original baseline and changed-URL cases were not rerun in this follow-up; the earlier table remains the initial revision's comparison. Timing still covers local preparation through ZIP creation, excluding cloud and interpreter teardown.


reflex/utils/build.py
```diff
@@ -235,7 +263,31 @@ def build():
...
+        "export",
+    ]
+    with build_cache.frontend_build_cache(wdir, command) as cache:
+        if cache is None or not cache.restore():
+            _build_frontend(wdir, command)
+            if cache is not None:
+                cache.capture()
+        _postprocess_frontend(wdir)
+        if cache is not None:
+            cache.commit()
```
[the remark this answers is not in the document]
Alek99: Fixed in [4da5154](https://github.com/reflex-dev/reflex/commit/4da5154bca6aef42ad5148eca06c41589c855179). A persistent, reentrant POSIX workspace lock now covers export configuration, compilation, setup, Vite/cache capture, postprocessing, publication, and ZIP creation. Direct builds and cache-disabled invocations participate; production/preview startup releases the lock before serving. Forked children discard inherited ownership safely, and schema 2 invalidates older unguarded snapshots. Added process-based regressions for overlapping builds, distinct export URLs/ZIP interleaving, mixed cache settings, reentrancy, and fork cleanup. Full validation: 8,441 unit tests (76.38% coverage), 28 production integration checks, and pre-commit hooks pass. Initialization/dev hot reload and unrelated workspace writers remain outside this documented lock.

reflex/utils/prerequisites.py
```diff
@@ -88,6 +88,8 @@ def check_latest_package_version(package_name: str):
     if environment.REFLEX_CHECK_LATEST_VERSION.get() is False:
         return
     try:
+        if get_or_set_last_reflex_version_check_datetime(set_if_missing=False):
```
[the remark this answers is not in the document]
Alek99: Fixed in [4da5154](https://github.com/reflex-dev/reflex/commit/4da5154bca6aef42ad5148eca06c41589c855179). Successful version checks now expire after 24 hours. Expired, malformed, or future timestamps trigger a new request; legacy local and timezone-aware timestamps are supported, with new records written in UTC. Only successful HTTP responses with valid versions refresh the timestamp, so network/HTTP/parse failures remain retryable. Added failing-before/passing-after regressions for the expiry boundary, timestamp variants, and retries; the full 8,441-unit suite passes.

reflex/utils/build.py
```diff
@@ -235,7 +263,31 @@ def build():
         SystemExit: If the build process fails.
     """
     wdir = prerequisites.get_web_dir()
+    command = [
+        *js_runtimes.get_js_package_executor(raise_on_none=True)[0],
+        "run",
+        "export",
+    ]
+    with build_cache.frontend_build_cache(wdir, command) as cache:
```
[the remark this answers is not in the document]
Alek99: Fixed in [4da5154](https://github.com/reflex-dev/reflex/commit/4da5154bca6aef42ad5148eca06c41589c855179). A persistent, reentrant POSIX workspace lock now covers export configuration, compilation, setup, Vite/cache capture, postprocessing, publication, and ZIP creation. Direct builds and cache-disabled invocations participate; production/preview startup releases the lock before serving. Forked children discard inherited ownership safely, and schema 2 invalidates older unguarded snapshots. Added process-based regressions for overlapping builds, distinct export URLs/ZIP interleaving, mixed cache settings, reentrancy, and fork cleanup. Full validation: 8,441 unit tests (76.38% coverage), 28 production integration checks, and pre-commit hooks pass. Initialization/dev hot reload and unrelated workspace writers remain outside this documented lock.

reflex/utils/prerequisites.py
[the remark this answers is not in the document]
Alek99: Fixed in [4da5154](https://github.com/reflex-dev/reflex/commit/4da5154bca6aef42ad5148eca06c41589c855179). Successful version checks now expire after 24 hours. Expired, malformed, or future timestamps trigger a new request; legacy local and timezone-aware timestamps are supported, with new records written in UTC. Only successful HTTP responses with valid versions refresh the timestamp, so network/HTTP/parse failures remain retryable. Added failing-before/passing-after regressions for the expiry boundary, timestamp variants, and retries; the full 8,441-unit suite passes.

reflex/utils/build_cache.py
```diff
@@ -0,0 +1,439 @@
...
+            info.st_mtime_ns,
+            info.st_ctime_ns,
+        ))
+    payload = [
+        2,  # Discard snapshots created before production workspace locking.
+        str(web_dir),
+        constants.Reflex.VERSION,
+        [str(arg) for arg in command],
+        runtimes,
+        sorted(os.environ.items()),
```
FarhanAliRaza, line 317: Hash only the environment that reaches Vite, not all of `os.environ`.

With the full environment in the key, ordinary variation between two runs of the same command defeats the cache. Reproduced on an unchanged app, each case rebuilt and replaced the snapshot, so the next plain run missed again:
- `reflex export --loglevel debug` (sets `REFLEX_LOGLEVEL`)
- `OLDPWD=/tmp reflex export`
- the same command in the foreground and in a background subshell (`SHLVL` differs)

In CI, a per-run variable such as a job id makes the cache never hit. Hash an allowlist (for example `PATH`, `NODE_*`, `NPM_CONFIG_*`, `VITE_*`, `REFLEX_*` minus `REFLEX_LOGLEVEL`), or the explicit env dict that `_build_frontend` passes to the subprocess.

reflex/utils/build_cache.py
```diff
@@ -0,0 +1,439 @@
...
+    Raises:
+        OSError: A file cannot be inspected or links outside the tracked tree.
+        ValueError: The tree contains unsupported file types or invalid metadata.
+    """
+    if root.is_symlink():
+        msg = "Build cache cannot inspect a symlink as its root"
+        raise ValueError(msg)
+    digest = hashlib.sha256()
+
+    def visit(directory: Path) -> None:
```
FarhanAliRaza, line 197: Split `visit` by entry type.

`_tree_digest` measures cyclomatic complexity 15 and `visit` 13 (`ruff --select C901` at max 8). One helper per entry kind (symlink, `node_modules` metadata, `reflex.json`, regular file) keeps each path readable and testable on its own.

reflex/utils/js_runtimes.py
```diff
@@ -468,6 +469,81 @@ def _is_bun_package_manager(package_manager: str) -> bool:
     return Path(package_manager).stem.lower() == "bun"
 
 
+def _npm_installed_package_sections(
```
FarhanAliRaza, line 472: Extract the per-package verification into a helper.

This function measures cyclomatic complexity 13. The loop body is a chain of guard clauses that reads better as `_verified_installed_version(name, declaration, entry) -> str | None`, with this function reduced to the loop and the assignment.

reflex/utils/build.py
```diff
@@ -118,7 +137,16 @@ def _zip(
         for file in files_to_zip:
             logger.debug(f"{target}: {file}", extra={"progress": progress})
             progress.advance(task)
-            zipf.write(file, Path(file).relative_to(root_directory))
+            # Sidecars are already compressed for serving the frontend.
+            compress_type = (
```
FarhanAliRaza, line 141: Move the compress-type choice into a small helper.

`_zip` now measures cyclomatic complexity 9. It was under 8 on `main`. A helper such as `_zip_compress_type(component_name, file)` keeps the write loop flat.

reflex/utils/export.py
```diff
@@ -80,26 +63,41 @@ def _time_phase(name: str) -> Iterator[None]:
...
-        if zipping:
-            with _time_phase("zip_duration"):
-                build.zip_app(
-                    frontend=frontend,
-                    backend=backend,
-                    zip_dest_dir=zip_dest_dir,
-                    include_db_file=upload_db_file,
-                    backend_excluded_dirs=backend_excluded_dirs,
-                )
+        with build_cache.frontend_build_lock(prerequisites.get_web_dir()):
```
FarhanAliRaza, line 66: Take the lock only when `frontend` is true.

A backend-only export in a project with no `.web` now creates a `.web` directory that contains only `.reflex-build.lock`. Reproduced with `reflex export --backend-only` after removing `.web`.

reflex/utils/prerequisites.py
```diff
@@ -100,13 +102,18 @@ def check_latest_package_version(package_name: str):
...
         logger.debug(f"Latest version of {package_name}: {latest_version}")
         current_version_parsed = version.parse(current_version)
         latest_version_parsed = version.parse(latest_version)
         path_ops.update_json_file(
             get_web_dir() / constants.Reflex.JSON,
-            {_version_check_timestamp_key(package_name): str(datetime.now())},
+            {
+                _version_check_timestamp_key(package_name): datetime.now(
+                    timezone.utc
+                ).isoformat()
```
FarhanAliRaza, line 115: Use `isoformat()` for the attempt timestamp as well.

The attempt key at line 193 still writes `str(now)`, so `reflex.json` mixes `2026-09-10 18:19:33+00:00` and `2026-09-10T18:19:34+00:00`. Both parse, but one format is simpler to reason about.

FarhanAliRaza (changes requested):
Tested with a two-page counter app that has one static asset, built against this branch.

Ran `reflex export` with and without zip, with the cache enabled and disabled, with bun and with npm. Repeat exports with the cache enabled skipped Vite and restored the snapshot. Setting the variable to `false` rebuilt and removed the cache directory. Two concurrent exports serialized on the lock, and the second one reused the first one's snapshot when the environment was identical. The frontend zip stored the `.gz` sidecars and deflated the other entries.

Ran `reflex run --env prod` with the cache enabled. The restart restored the cached build. In the browser the counter updated over the websocket, the asset loaded, page navigation worked, and gzip content negotiation returned compressed assets. No console errors.

The unit tests for the touched files pass.

The inline comments hold the requested changes. The first one, about the cache key, is the one that matters most: the same command with a different log level, a different `OLDPWD`, or a different `SHLVL` rebuilt every time.

reflex/utils/build_cache.py
FarhanAliRaza: Valid — fixed in 7ebe20fd8. Cache fingerprints now use PATH and frontend build environment prefixes, excluding log-level and unrelated shell state. Regression test: test_non_build_environment_change_reuses_cache.

reflex/utils/build_cache.py
FarhanAliRaza: Valid — fixed in 7ebe20fd8. Split tree entry handling into symlink, regular-file, and dispatch helpers. The max-complexity check now passes.

reflex/utils/js_runtimes.py
FarhanAliRaza: Valid — fixed in 7ebe20fd8. Extracted _verified_installed_version and lockfile parsing so the section loop only verifies and assigns.

reflex/utils/build.py
FarhanAliRaza: Valid — fixed in 7ebe20fd8. Extracted _zip_compress_type and archive exclusion identity checking. The max-complexity check now passes.

reflex/utils/export.py
FarhanAliRaza: Valid — fixed in 7ebe20fd8. Backend-only exports now avoid creating or acquiring the frontend lock. Regression test: test_backend_only_export_does_not_lock_frontend.

reflex/utils/prerequisites.py
FarhanAliRaza: Valid — fixed in 7ebe20fd8. Failed-check attempt timestamps now use isoformat(), matching successful timestamps. Regression coverage: test_check_latest_package_version_refreshes_stale_or_invalid_timestamp.

reflex/utils/build_cache.py
```diff
@@ -0,0 +1,523 @@
...
+
+    Returns:
+        Sorted build-relevant environment key-value pairs.
+    """
+    return sorted(
+        (key, value)
+        for key, value in os.environ.items()
+        if (key == "PATH" or key.startswith(_BUILD_ENVIRONMENT_PREFIXES))
+        and key not in _IGNORED_BUILD_ENVIRONMENT_KEYS
+    )
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in 8ad34c900b279b3fd39f11f68af842c608fc3280. The cache fingerprints the effective frontend subprocess environment, including custom variables, and invalidates snapshots created with the old filter. Only REFLEX_LOGLEVEL and SHLVL are excluded, as documented. Regression tests cover adding, changing, and removing APP_RELEASE, CI, and lowercase npm variables, plus effective allocator/color overrides. Real npm exports confirmed that unchanged inputs reuse the snapshot while changed or removed APP_RELEASE reruns postexport and updates its artifact.

reflex/utils/export.py
```diff
@@ -80,26 +63,45 @@ def _time_phase(name: str) -> Iterator[None]:
...
-                    backend=backend,
-                    zip_dest_dir=zip_dest_dir,
-                    include_db_file=upload_db_file,
-                    backend_excluded_dirs=backend_excluded_dirs,
-                )
+        with (
+            build_cache.frontend_build_lock(prerequisites.get_web_dir())
+            if frontend
+            else nullcontext()
+        ):
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in 8ad34c900b279b3fd39f11f68af842c608fc3280. Backend-only exports acquire the workspace lock while packaging an existing frontend workspace; exports without that workspace still avoid creating it. Regression test test_backend_archive_waits_for_frontend_compilation uses a separate spawned process and a real backend ZIP with caching enabled and disabled. It failed before the fix because packaging completed while the compiler lock was held, and now waits for the final marker before archiving.

reflex/utils/js_runtimes.py
```diff
@@ -468,6 +469,108 @@ def _is_bun_package_manager(package_manager: str) -> bool:
...
+    if (
+        not isinstance(entry, dict)
+        or entry.get("link")
+        or entry.get("name", name) != name
+    ):
+        return None
+    installed_version = entry.get("version")
+    if not isinstance(installed_version, str) or declaration != f"^{installed_version}":
+        return None
+    resolved = entry.get("resolved")
```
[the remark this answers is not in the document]
FarhanAliRaza: Declined: the cited rule is from reflex-dev/flexgen and is not in this repository's AGENTS.md. Existing code uses literal schema keys consistently: frontend_skeleton._compile_package_json reads dependencies and devDependencies, and prerequisites.check_latest_package_version reads info/version directly. The npm keys identify fields in an external JSON schema. Adding a parallel set of constants would add indirection without addressing a correctness defect.

reflex/utils/js_runtimes.py
```diff
@@ -468,6 +469,108 @@ def _is_bun_package_manager(package_manager: str) -> bool:
...
+    if isinstance(resolved, str) and resolved.startswith((
+        "file:",
+        "link:",
+        "workspace:",
+    )):
+        return None
+    package_dir = get_web_dir() / "node_modules" / name
```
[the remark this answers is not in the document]
FarhanAliRaza: Updated in 8ad34c900b279b3fd39f11f68af842c608fc3280. Installed-package paths, npm lockfile paths, and build-cache dependency-directory checks now use Javascript.NODE_MODULES from reflex_base.constants.base, where the existing constant is defined. The package-install and build-cache regression suites pass.

FarhanAliRaza:
For unchanged exports, npm went from 7.06 s to 2.50 s by default and 1.01 s with caching. Bun went from 2.53 s to 2.49 s by default and 1.06 s with caching.
Cache misses added about 0.7 s over the new default.

FarhanAliRaza (commented):
Withdrawn as out of scope for #7054. The two findings previously listed here concern `reflex/state.py`, which this PR does not change. Both were already present in its merge base, originating in #6946 and #7068. No change is requested in #7054 for either finding.

FarhanAliRaza (approved):

masenf (changes requested):
@FarhanAliRaza this needs to be split up into separate patches, because i want some of them. but other ones make me nervous and add extra complexity, so we need to make sure that the speedup is actually worth the cost in complexity.

this seriously feels like 5 or 6 separate things

PR #7071, opened 2026-09-09
Keep React context objects stable across hot updates

FarhanAliRaza:
## Problem

The generated `utils/context.js` created every React context with `createContext()` at module scope, and the static runtime (`state.js`, `react-theme.js`) imported app data from it. Vite re-executes every module between a changed file and a react-refresh boundary. So any state default change rewrote `context.js` and re-ran the whole runtime. A component refreshed after the update then held different objects than the still-mounted tree:

- a different context object, so `useContext` returned `null` and the page crashed with `Cannot read properties of null (reading '<var>_rx_state_')`
- a reset `addEvents` slot, so clicks and frontend error reports during the overlap were dropped
- a fresh, empty `refs` object, so ref-based events like `rx.set_focus` were no-ops during the overlap

On top of that, the generated module was never registered with fast refresh, because Vite's React plugin skips a `.js` file without JSX. Every re-execution produced new provider component types, and React remounted the whole provider subtree on each compile: websocket torn down and reopened, client state reset, and a second update in the same window closed a socket that was still connecting (`WebSocket is closed before the connection is established`).

## Fix

Three commits.

1. **Contexts move to a static registry.** `utils/context-registry.js` imports only React and owns every context object. The generated module re-exports the fixed contexts and builds `StateContexts` through `getStateContext(name)`. All existing importers and generated call sites are unchanged.
2. **The runtime stops importing the generated module.** The generated module calls `registerApp({...})` with its data and the runtime reads it from the registry at call time. The event loop dispatchers live in the registry as well. Nothing the compiler writes sits below the runtime any more, so a compile re-executes only generated files and components, and the circular import between `context.js` and `state.js` is gone.
3. **The generated module is emitted as `context.jsx`.** The refresh runtime now maps new provider functions onto the mounted ones, so a compile re-renders the providers in place instead of remounting them. A stale `context.js` is removed so extensionless resolution of `$/utils/context` cannot pick it over the `.jsx` file. Import specifiers are unchanged.

## Verification

Playwright scripts against a counter app under `reflex run` with HMR on, editing Python source and holding module requests to force a consumer to meet a newer context instance than the mounted provider. The event loop effect and `connect()` were instrumented to count provider remounts.

| Check under the forced ordering | main | this branch |
|---|---|---|
| page crash | `TypeError: Cannot read properties of null (reading 'count_rx_state_')` | none |
| browser refetches `state.js` during the update | yes | no |
| `refs` object seen by refreshed modules | new, empty | same as mounted tree |
| `+` click while the provider refresh is held | dropped | counted |
| `rx.set_focus` while the provider refresh is held | no-op | input focused |
| provider remounts per compile | 1 | 0 |
| `WebSocket is closed before the connection is established` | n/a (one reconnect per compile) | none |
| Vite frames | `update` only | `update` only, no `full-reload` |

Integration tests in `tests/integration/tests_playwright/test_hmr.py` pass. The overlap test loads a route by navigation while the root refresh is held, since with a self-accepting provider module a later refresh cannot apply until the earlier one completes. A mutation that hands out a fresh context per lookup makes that test fail with the original `null` error.

## Notes

- `ColorModeContext` and `ThemeContext` now have fixed fallback values. They are only observable without their providers, which the app root always mounts.
- An existing `.web` from a dev checkout needs re-initialization to receive the new files. Released versions re-copy templates on version change.

https://claude.ai/code/session_01T7KbvgyqeruWJetigu4e6Q


packages/reflex-base/src/reflex_base/.templates/web/utils/state.js
```diff
@@ -1199,11 +1193,11 @@ export const useEventLoop = (
     }
 
     // Equivalent to routeChangeComplete - runs after navigation completes
-    addEvents(onLoadInternalEvent());
+    addEvents(app.onLoadInternalEvent());
 
     // Update the ref
     prevLocationRef.current = location;
-  }, [location, dispatch, onLoadInternalEvent, addEvents]);
+  }, [location, dispatch, addEvents]);
```
masenf, line 1200: i think we'd need to add `app` to this dependency array now, right?

FarhanAliRaza: react only rerenders on assignments, changing properties do not rerender.

FarhanAliRaza: `app.onLoadInternalEvent` maybe ?

reflex/compiler/compiler.py
```diff
@@ -725,6 +725,9 @@ def compile_contexts(
     """
     # Get the path for the output file.
     output_path = utils.get_context_path()
+    # A leftover ``.js`` module would win extensionless resolution of
+    # ``$/utils/context`` over the ``.jsx`` file written here.
+    Path(output_path).with_suffix(constants.Ext.JS).unlink(missing_ok=True)
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in 7cf018e69. The unlink now runs after the dry-run return, next to the memo pruning, so `--dry` leaves `utils/context.js` alone. Regression test: `test_compile_dry_run_does_not_prune_or_write_manifest` now seeds a stale `context.js` and asserts it survives a dry run.

packages/reflex-base/src/reflex_base/.templates/web/utils/context-registry.js
```diff
@@ -0,0 +1,43 @@
+// React context objects shared by the generated context module and by every
+// component that reads state. A provider and its consumers only connect when
+// they hold the same context object, so the objects live here rather than in
+// a module the compiler rewrites: Vite re-executes a rewritten module on hot
+// update, and every module it imports along with it, but this module imports
+// nothing that can change and is never re-executed.
+import { createContext } from "react";
+
+export const ColorModeContext = createContext({
+  colorMode: "system",
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in 7cf018e69. The fallback key is now `rawColorMode`, matching what the provider publishes and what the `color_mode` var reads.

tests/units/compiler/test_compiler.py
```diff
@@ -476,12 +476,15 @@ def test_compile_app_root_includes_radix_window_library_when_bundled():
         reset_bundled_libraries()
```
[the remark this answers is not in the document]
FarhanAliRaza: Declined. Those tests asserted substrings of the rendered template, so they restated the template text rather than checking behavior and would break on any harmless rewording. The guarantee they described (contexts, dispatcher, and refs stay stable while a consumer refreshes before its provider) is covered end to end by `test_route_loaded_during_held_provider_refresh_uses_mounted_provider` in `tests/integration/tests_playwright/test_hmr.py`, which fails with the original `Cannot read properties of null` error if the registry hands out fresh contexts.

packages/reflex-base/src/reflex_base/.templates/web/utils/state.js
FarhanAliRaza: `app` is no op. `app.onLoadInternalEvent` double fires.

masenf: gotcha. probably fine to leave it off, since each time the function runs, it will capture the latest value hanging off of `app` and since `app` is const and never reassigned, this works fine.

masenf (approved):

PR #7073, opened 2026-09-09
Stop the rehydrate fanning out on events that carry no route

masenf:
Rebased onto `main` now that [#7072](https://github.com/reflex-dev/reflex/pull/7072) has merged. Single commit; no longer stacked.

### Type of change

### Changes To Core Features

An event that reaches expired state without `router_data` — as one enqueued from the backend does, e.g. `app.event_processor.enqueue(token, ...)` — rehydrated against the empty route. `on_load_internal` resolved that to the 404 loaders and chained them, and those chained events carried no `router_data` either, so each one rehydrated in turn: **1136 `on_load` runs** in the regression test before the queue shut down. Nothing on this path ever sets `router_data`, so the rehydrate also repeated for every such event — a full state dict emit and a `_reset_client_storage()` each time.

`_rehydrate` now hydrates and marks the state hydrated without running page loaders when there is no route, and skips a state that is already hydrated and still routeless. `_execute_event` and `needs_to_rehydrate` are untouched, so the routed path is unchanged; in particular a routeless hydrate leaves `router_data` unset, so the first event that *does* carry a route still rehydrates and gets its page's `on_load`.

### Tests

- `test_event_without_router_data_hydrates_once_without_loaders` — app with `/` and `/404` loaders, two backend-initiated events; no loader runs, `is_hydrated` ends `True`, and exactly one hydrate delta is emitted. Fails in 2.4s without the fix, with 1136 `on_load` runs. It drives the processor with `join` rather than the event's future, because the pre-fix fan-out outruns `wait_all` and wedges instead of failing.
- `test_routed_event_after_a_routeless_rehydrate_loads_its_page` — a backend event followed by a `/page-b` event; the page's loader still runs. This one passes without the fix too; it is there to stop a future "mark hydrated, never rehydrate again" shortcut from eating the page's loaders.

Both read their final state back through `_read_back` (added in #7072), so they hold under opportunistic locking.

Verified against a real redis in all three configurations the unit-test workflow runs: plain (8367), `REFLEX_REDIS_URL` (8381), and `REFLEX_OPLOCK_ENABLED` on top (8381). `ruff` and `pyright` clean.

https://claude.ai/code/session_0171CdXddfL4FTfSAg6WUmj6


FarhanAliRaza (dismissed):
code-review, minor: the news fragment packages/reflex-base/news/+rehydrate-without-router-data.bugfix.md says the empty route's loaders are "the 404 ones". Both the live base run and app.router("") resolve the empty path to the index route, so it is the / page's loaders, with 404 only as the fallback when / has no on_load. The fragment is also narrative-heavy for CLAUDE.md's guidance. A one-sentence version would do: backend-initiated events reaching expired state now hydrate it once without running page loaders, and the next routed event still runs its page's on_load.

PR #7079, opened 2026-09-10
perf: preload the app and fork production granian workers

FarhanAliRaza:
## Summary

- On Linux, `run_granian_backend_prod` now forces the `fork` start method, imports the app in the supervisor, and runs `gc.collect(); gc.freeze()` before granian starts workers. Workers share the framework and app pages copy-on-write instead of each re-importing ~1,800 modules.
- `REFLEX_BACKEND_START_METHOD` (`fork` / `spawn` / `forkserver`) overrides the choice for apps that are not fork-safe. Unset keeps the interpreter default on macOS and Windows (`spawn`), so nothing changes there.
- A forked child inherits the telemetry `ThreadPoolExecutor` but not its worker thread, so any event submitted from a worker would queue forever. An `os.register_at_fork` hook drops the inherited pool in the child. This also affected dev mode on Python <= 3.13, where `fork` was already the Linux default.

## Measurement

Blank `reflex init` app, `GRANIAN_WORKERS=4`, `reflex run --env prod --backend-only`, Python 3.14, granian 2.8.1. PSS summed over the whole process tree from `/proc/<pid>/smaps_rollup`, after 32 `/ping` requests, median of 3 runs.

| | PSS | private dirty | port open after |
|---|---|---|---|
| main | 721 MB | 692 MB | 1.82 s |
| this PR | 223 MB | 70 MB | 1.07 s |

Per worker: 137 MB PSS → 41 MB PSS. The supervisor goes from a 137 MB private copy to 47 MB PSS with 172 MB shared.

Functional check: 8 socket.io clients against the forked workers, each sending two state events and receiving the expected deltas (`count` 1 then 2), no backend errors.

## Notes for review

- `telemetry.send("run-prod")` leaves the telemetry thread alive at fork time, so Python emits its hidden-by-default `DeprecationWarning` about forking a multi-threaded process. This is the same situation dev mode has had on Python <= 3.13. Draining the executor before forking would block startup on the telemetry HTTP request, so I did not add that.
- Granian respawns (crash or `workers_max_rss`) fork from the supervisor again, which still holds the pristine preloaded app.
- Worker count default (`cpu × 2 + 1`) and the lazy-import work in #7049 are separate; this PR gains more once #7049 shrinks the shared image, but has no code dependency on it.

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/7079?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/news/+prod-fork-preload.performance.md
+++ b/news/+prod-fork-preload.performance.md
@@ -1,0 +1,1 @@
+Production backend workers on Linux are now forked from a supervisor that has already imported the app, so framework and app code are shared copy-on-write instead of re-imported per worker. A 4-worker blank app drops from about 720 MB to about 220 MB of proportional set size and starts faster. Set `REFLEX_BACKEND_START_METHOD=spawn` for apps that are not fork-safe.
--- a/packages/reflex-base/news/+prod-fork-preload.performance.md
+++ b/packages/reflex-base/news/+prod-fork-preload.performance.md
@@ -1,0 +1,1 @@
+Add the `REFLEX_BACKEND_START_METHOD` environment variable to choose how production backend workers are started (`fork`, `spawn`, or `forkserver`).
--- a/packages/reflex-base/src/reflex_base/environment.py
+++ b/packages/reflex-base/src/reflex_base/environment.py
@@ -708,6 +708,13 @@ class EnvironmentVariables:
     # Whether to run Granian in a spawn process. This enables Reflex to pick up on environment variable changes between hot reloads.
     REFLEX_STRICT_HOT_RELOAD: EnvVar[bool] = env_var(False)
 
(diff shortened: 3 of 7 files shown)


masenf (commented):
nice improvement.

the telemetry stuff does slightly make me nervous though. i think it's probably worth it to defer sending the "run" telemetry until after the fork and attempt to drain the executor (which theoretically should be empty on the run path if we defer the cli telemetry call). forking with threads is more than just a warning, it can lead to weird and hard-to-debug issues; if we can avoid it, then we should

PR #7127, opened 2026-09-13
fix(plotly): forward the id prop as react-plotly.js's divId

### LINKED ISSUE #6977: rx.plotly silently drops the id prop (react-plotly.js expects divId)
**Describe the bug**

`rx.plotly(..., id="the-plot")` renders the chart but the `id` never lands in the DOM — `document.getElementById("the-plot")` returns null. react-plotly.js's `Plot` component only forwards `divId` (plus `style`/`className`/`ref`) to the container div it renders; the framework-universal `id` prop that Reflex passes falls into Plot's `...eventProps` rest and is silently ignored.

**To Reproduce**

```python
import plotly.graph_objects as go
import reflex as rx

fig = go.Figure(go.Scatter(x=[1, 2, 3], y=[1, 3, 2]))

def index():
    return rx.vstack(
        rx.plotly(data=fig, id="the-plot"),
        rx.box(id="control-box"),  # control: gets its id fine
    )

app = rx.App()
app.add_page(index)
```

Run, load the page, and in the browser console check `document.getElementById("the-plot")` (null) vs `document.getElementById("control-box")` (found). Same result with a state-var figure.

**Observed behavior**

Chart renders; the id is absent from the DOM. The compiled JSX proves Reflex emits the prop (`jsx(Plot,{id:"the-plot",ref:ref_the_plot,...})`); react-plotly.js@4.0.0's `Plot` destructures `divId` and renders `React.createElement("div",{id:divId,style,ref,className})`, so `id` is discarded.

**Expected behavior**

Either the wrapper maps `id=` to react-plotly.js's `divId`, or the mismatch is documented (with `divId` exposed as a prop).

- Python Version: 3.11
- Reflex Version: 0.9.9a1 (reflex-components-plotly 0.9.5a1)
- OS: linux
- Browser (Optional): Chromium

**Additional context**

- Severity: low. Longstanding third-party-API mismatch, identical on 0.9.8 (same pinned react-plotly.js@4.0.0) — not a regression.
- Found during 0.9.9a1 pre-release testing; probe script and screenshots on branch `claude/reflex-prerelease-testing-t0sd90` under `prerelease-testing/2026-08-27-v0.9.9a1/devtools_perf/`.

Chirag6722:
### All Submissions:

### Type of change

### Summary

Fixes #6977.

react-plotly.js's `Plot` only forwards `divId`, `style`, `className` and `ref` to the div it renders. Reflex's framework-universal `id` prop fell into `Plot`'s `...eventProps` rest and was discarded, so `rx.plotly(data=fig, id="the-plot")` rendered the chart but `document.getElementById("the-plot")` returned `null`. The compiled JSX proved the prop was emitted; it just never reached the DOM.

The wrapper now renames `id` to `divId` at render time, in `Plotly._render`, next to the existing `locale` → `config` rewrite. The `ref` is attached separately by the base component and already pointed at the same container div, so `rx.get_ref` and scrolling were never affected and stay that way.

Of the two options in the issue, mapping `id` rather than exposing `divId` keeps the universal prop working the way it does on every other component, so users do not need to know which underlying library a chart happens to be backed by.

### Verification

- `uv run pytest tests/units/components/graphing/test_plotly.py` — 6 passed. Two tests added: one asserts `id` is gone from the rendered props and `divId` carries `"the-plot"` with `ref` still present; the other asserts no `divId` is emitted when no `id` was given.
- Reverting only the wrapper change and keeping the tests fails the first one with `'id'` still in the rendered props.
- `uv run pytest tests/units/components tests/units/compiler` — 4413 passed, 1 skipped, so nothing else was relying on `id` reaching `Plot`.
- `pre-commit` on the changed files: ruff-format, ruff-check, codespell, update-pyi-files, pyright and ty all pass. The `.pyi` needed no regeneration since the public signature is unchanged.
- News fragment added at `packages/reflex-components-plotly/news/6977.bugfix.md`.

Not verified in a browser: I did not run the Playwright integration test locally. The unit assertion is on the rendered tag props, which is exactly where the issue showed the prop being emitted, and `divId` is the documented prop `Plot` forwards.

---

AI help was taken for this change. Reviewed and run locally by me.

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/7127?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Chirag6722:
Closing: this duplicates #7082, which landed three days earlier with the same change and the same fragment, and is already under review. I ran the open-PR search and opened this in the same step instead of reading the result first, so the template checkbox above was ticked in error. Apologies for the noise; nothing here that #7082 does not already cover.

Issue #6981, opened 2026-08-28
Spurious ERROR log lines on normal SIGTERM shutdown of reflex run (granian "Unexpected exit from worker-1"; "Starting frontend failed with exit code 143")


masenf:
**Describe the bug**

A perfectly normal SIGTERM shutdown of a dev `reflex run` produces ERROR-level log records, which is noise that pollutes captures, alerting, and log-based CI checks:

1. granian logs `[ERROR] Unexpected exit from worker-1` — timing-dependent: some shutdowns produce the clean `Shutting down granian` / `Stopping worker-1` / `Stopped worker-1` sequence, others the spurious ERROR.
2. reflex itself logs `Starting frontend failed with exit code 143` followed by a frontend log dump: 143 (SIGTERM) is not in the accepted return codes for the frontend process — only SIGINT/130 is — so a SIGTERM teardown of the frontend child is reported as a startup failure.

**To Reproduce**

```sh
reflex init --template blank
reflex run --frontend-port 3000 --backend-port 8000 > run.log 2>&1 &
# wait for the app to come up, then:
kill -TERM %1
grep -n "ERROR\|failed" run.log
```

Repeat a few times; the granian ERROR appears on a large fraction of shutdowns (timing-dependent), the exit-code-143 ERROR reliably when the frontend child is torn down by the signal.

**Observed behavior**

ERROR-level records during an orderly, user-requested shutdown.

**Expected behavior**

A normal SIGTERM shutdown logs nothing above INFO: worker teardown should be sequenced so granian does not classify it as unexpected (or the message downgraded), and 143/SIGTERM should join 130/SIGINT in the accepted frontend exit codes.

**Specifics (please complete the following information):**

- Python Version: 3.11
- Reflex Version: 0.9.9a1
- OS: linux

**Additional context**

- Severity: low (cosmetic noise). Not a regression — both behaviors reproduce identically on 0.9.8.
- Found during 0.9.9a1 pre-release testing; server logs showing both shapes (and the clean-shutdown variant) are on branch `claude/reflex-prerelease-testing-t0sd90` under `prerelease-testing/2026-08-27-v0.9.9a1/state_concurrency/` and `prerelease-testing/2026-08-27-v0.9.9a1/logging_cli/`.

PR #7128, opened 2026-09-13
fix(processes): accept a SIGTERM teardown of the frontend as a clean exit

Chirag6722:
### All Submissions:

### Type of change

### Summary

Fixes the second half of #6981: the `Starting frontend failed with exit code 143` report on an ordinary SIGTERM shutdown.

`stream_logs` decides whether a finished child was a failure from an accepted-exit-code list:

```python
accepted_return_codes = [0, -2, 15, 130] if constants.IS_WINDOWS else [0, -2, 130]
```

That list recognises SIGINT in both of the forms it arrives in — `-2` when Popen saw the signal directly, `130` when a shell wrapper such as react router reported it as 128+2 — and SIGTERM in neither. So `kill -TERM` on `reflex run` tears the frontend child down with 143, the check calls it a failure, the frontend logs are dumped at ERROR, and `SystemExit(1)` is raised for a stop the user asked for. On the issue's own terms it is "cosmetic noise"; in practice it is also a non-zero exit from an orderly shutdown.

On POSIX the set is now built from the two interrupt signals in both forms, so the two signals are treated the same way instead of one being spelled out twice and the other not at all: `{0, -2, -15, 130, 143}`.

Windows is deliberately untouched, at `{0, -2, 15, 130}` as before. It has no POSIX signal exit codes: `os.kill(pid, SIGTERM)` calls `TerminateProcess` with the signal number, so SIGTERM surfaces as a bare `15` there, and `-15`/`143` are ordinary application exit codes that must keep counting as failures. cubic caught that the first push widened Windows for no reason.

### Not included

The other half of the issue, granian's timing-dependent `Unexpected exit from worker-1`, is inside granian's worker supervision rather than Reflex's code, and needs the teardown sequenced so granian does not classify it as unexpected. That is a separate change and I have not attempted it here.

### Verification

- `uv run pytest tests/units/utils/test_processes.py` — 29 passed. The new tests drive **both** platform branches by patching `IS_WINDOWS`: on POSIX the four signal-derived codes (`-2`, `130`, `-15`, `143`) are clean exits; on Windows `-2`, `130` and `15` are clean while `-15` and `143` must still log `failed with exit code` and raise `SystemExit`; and a real exit code `1` fails on both.
- Mutation check in both directions. Against upstream `main`, exactly the two POSIX SIGTERM cases fail, with the logged `Starting frontend failed with exit code 143`. Against the first push of this PR, exactly the two Windows `-15`/`143`-stay-failures cases fail. Each half of the change has a test that only it turns green.
- `pre-commit` on the changed files: ruff-format, ruff-check, codespell, update-pyi-files, pyright, ty all pass.
- News fragment at `news/6981.bugfix.md`.

Not reproduced end to end: I did not run the issue's `kill -TERM` script, since this machine is Windows and SIGTERM delivery to a backgrounded `reflex run` does not behave the same way there. The tests exercise `stream_logs` directly with the exit codes the issue reports.

---

AI help was taken for this change. Reviewed and run locally by me.


reflex/utils/processes.py
```diff
@@ -440,12 +440,21 @@ def stream_logs(
...
+    interrupt_signals = (int(signal.SIGINT), int(signal.SIGTERM))
+    accepted_return_codes = {
+        0,
+        *(-sig for sig in interrupt_signals),
+        *(128 + sig for sig in interrupt_signals),
+    }
+    if constants.IS_WINDOWS:
+        # Windows uvicorn bug: SIGTERM surfaces as a bare 15.
+        # https://github.com/reflex-dev/reflex/issues/2335
+        accepted_return_codes.add(int(signal.SIGTERM))
```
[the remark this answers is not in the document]
Chirag6722: Agreed, and fixed in 32d99fb.

Checked the mechanism before changing it: on Windows `os.kill(pid, SIGTERM)` calls `TerminateProcess` with the signal number, so a SIGTERM teardown surfaces as a bare `15`, which the Windows set already accepted. `-15` and `143` never arise from a signal there, so widening the Windows set to include them bought nothing and would have let a real failure with either code skip the error log and the `SystemExit`.

The Windows set is now the explicit `{0, -2, 15, 130}` it had before this PR, unchanged. Only POSIX gets the derived set with both SIGTERM forms, which is where the reported bug lives (the issue was filed from Linux).

Tests exercise both branches by patching `IS_WINDOWS`, and two new cases pin exactly this: `-15` and `143` on Windows must still log `failed with exit code` and raise `SystemExit`. Against the previous commit those two fail; against upstream `main` the two POSIX SIGTERM cases fail instead. So each half of the change has a test that only it turns green.

news/6981.bugfix.md
```diff
@@ -0,0 +1 @@
+`reflex run` no longer reports "Starting frontend failed with exit code 143" and exits non-zero when the frontend child is stopped by SIGTERM. A SIGTERM teardown is now accepted alongside SIGINT, in both the direct (`-15`) and shell-reported (`143`) forms.
```
masenf:
```suggestion
Stopping `reflex run` with SIGTERM no longer reports "Starting frontend failed with exit code 143" and now exits cleanly.
```

Changelog entries should always be user-facing and observable without mentioning irrelevant internal implementation details.

reflex/utils/processes.py
```diff
@@ -440,12 +440,26 @@ def stream_logs(
...
+        # bare 15, and Node reports Ctrl+C as 130 by convention. -15 and 143
+        # are ordinary application exit codes here and must stay failures.
+        # https://github.com/reflex-dev/reflex/issues/2335
+        accepted_return_codes = {0, -2, 15, 130}
+    else:
+        # On POSIX each signal shows up two ways: negative when Popen saw it
+        # directly, 128+N when a shell wrapper such as react router reported
+        # it (130 for SIGINT, 143 for SIGTERM). Recognising SIGINT in both
+        # forms but SIGTERM in neither is what turned an orderly `kill -TERM`
+        # into "Starting frontend failed" (#6981).
```
masenf:
```suggestion
        # it (130 for SIGINT, 143 for SIGTERM).
```

Narratives should be in PR or commit message, not code comments

masenf (approved):

PR #7155, opened 2026-09-15
Fix OpenTelemetry spans lost during dev compilation

### LINKED ISSUE #7095: Initial dev compilation loses OpenTelemetry spans when its worker exits
- Cluster: `otel` | Regression: n/a | Changelog: reflex-base 0.9.11a1 (#6900), README "one
  `reflex.compile` span per app compile ... with the stages as child spans"
- Under dev `reflex run` the initial compile happens in a `ProcessPoolExecutor` worker that exits via
  `os._exit`, so neither `atexit` nor provider shutdown runs: with OTLP the tree never arrives, and
  with a fast batch processor only two stage children escape, pointing at a parent span id that was
  never exported. Prod, `reflex export` and hot reload all export the full tree.
- Evidence: `otel/evidence/compile-span-loss.txt`, `run1-spantree.txt`.

Source: prerelease finding 028, recorded against Reflex 0.9.11a1 and compared with 0.9.10.post2. See the [full campaign report](https://github.com/reflex-dev/reflex/blob/c8801b060039ca7e1c9e74996595a6971772eba5/prerelease-testing/2026-09-10-v0.9.11a1/FINDINGS.md) and [cluster notes and reproduction artifacts](https://github.com/reflex-dev/reflex/tree/c8801b060039ca7e1c9e74996595a6971772eba5/prerelease-testing/2026-09-10-v0.9.11a1/otel). Findings and verification status above are from that campaign; this filing does not claim a new browser rerun.

harsh21234i:
## Summary

Fixes #7095.

The initial development compilation runs in a worker process that exits with `os._exit`, which bypasses OpenTelemetry provider shutdown and can lose the root compile span and child stage spans.

This change:

- Flushes the OpenTelemetry provider when the compile worker finishes.
- Keeps the flush best-effort so telemetry failures do not break compilation.
- Preserves existing behavior for production builds, exports, and hot reloads.
- Adds regression coverage for the worker flush path and provider behavior.
- Adds a changelog entry.

## Validation

- 51 affected unit tests passed.
- Targeted Ruff and formatting checks passed.
- Changed-file Pyright passed.
- Codespell passed.


packages/reflex-base/src/reflex_base/otel.py
```diff
@@ -196,6 +207,9 @@ def enable(
     )
 
     _remote_propagator = TraceContextTextMapPropagator()
+    _tracer_provider = (
+        tracer_provider if tracer_provider is not None else trace.get_tracer_provider()
+    )
```
FarhanAliRaza: This captures the provider at `enable()` time. The docstring of `enable()` supports a different order: call `enable()` first, then configure the SDK. In that order, `trace.get_tracer_provider()` returns the API's `ProxyTracerProvider` here. The proxy has no `force_flush`, so `flush()` returns `True` without exporting anything — and the compile spans are still lost, because the proxy tracer does resolve them against the late provider.

Repro on this branch: `otel.enable()` → `trace.set_tracer_provider(TracerProvider(...))` with a `BatchSpanProcessor` → end a span → `otel.flush()` returns `True`, exporter receives 0 spans.

Suggested change of the same size: store only the explicit provider here (it may be `None`), and resolve `_explicit_provider or trace.get_tracer_provider()` inside `flush()`. Please also add a regression test for the late-configuration order; the current test fixture always passes an explicit provider, so it cannot catch this.

harsh21234i: Fixed in `8afa643a`. `enable()` now stores only an explicitly supplied tracer provider; when none is supplied, `flush()` resolves `trace.get_tracer_provider()` at flush time. Added `test_flush_resolves_provider_configured_after_enable`, which enables OTEL first, configures a late `TracerProvider` with a queued `BatchSpanProcessor`, ends a span, and verifies that `flush()` exports it. Explicit-provider behavior remains unchanged. Targeted OTEL tests, pre-commit checks, type checks, and diff validation all pass.

packages/reflex-base/src/reflex_base/otel.py
```diff
@@ -227,6 +239,31 @@ def disable() -> None:
     _message_size = _ws_connections = _NOOP_INSTRUMENT
 
 
+def flush(timeout_millis: int = 5000) -> bool:
+    """Flush spans already ended by the current process.
+
+    This is needed before a compile worker exits with ``os._exit``, which
+    bypasses the SDK's normal interpreter shutdown hooks.
```
FarhanAliRaza: `os._exit` appears nowhere in Reflex itself — it's multiprocessing's fork/forkserver child exit path, and only on Python <= 3.12: since 3.13 children run `atexit` hooks, so the SDK's registered shutdown flush does run there (verified empirically on 3.12 vs 3.14 — the flush is still needed for the supported Python 3.10–3.12 range). Naming the real condition keeps the docstring accurate for readers on newer interpreters:

```suggestion
    This is needed before a compile worker exits: on Python <= 3.12,
    multiprocessing fork and forkserver children exit with ``os._exit``,
    skipping the ``atexit`` hook where the SDK registers its shutdown flush.
```

---

FarhanAliRaza (commented):
Reviewed by exercising the change in a real app (dev run with `OTEL_TRACES_EXPORTER=console` and a 60s batch delay): the compile worker exports the full `reflex.compile` trace at exit, runtime event spans are unaffected, and an isolated repro confirms the loss is real on Python <= 3.12 (fork/forkserver) and fixed by this flush. One docstring nit inline; the change itself looks good.

---

FarhanAliRaza (approved):

packages/reflex-base/src/reflex_base/otel.py
```diff
@@ -227,6 +239,32 @@ def disable() -> None:
...
+    skipping the ``atexit`` hook where the SDK registers its shutdown flush.
+
+    Args:
+        timeout_millis: Maximum time to wait for exporters to flush.
+
+    Returns:
+        Whether the provider accepted and completed the flush.
+    """
+    if not enabled:
+        return True
```
FarhanAliRaza: why return bool whne not needed anywhere?

FarhanAliRaza (changes requested):
why return bool whne not needed anywhere?

packages/reflex-base/src/reflex_base/otel.py
FarhanAliRaza: why return a bool when not needed anywhere?

packages/reflex-base/src/reflex_base/otel.py
FarhanAliRaza: Verified on the branch: `otel.enable()` → `trace.set_tracer_provider(TracerProvider())` with a queued `BatchSpanProcessor` → end a span → `otel.flush()` exports it (exporter receives 1 span). Also confirmed end to end in real `ProcessPoolExecutor` workers: on Python 3.12 (fork and forkserver) the spans are lost without the flush and exported with it. Resolving.

packages/reflex-base/src/reflex_base/otel.py
FarhanAliRaza: Nothing reads it, so dropped in 4517ed41b. 3d7dfe314 then removed the `try/except` around `force_flush` (the SDK never raises from it) and logs at debug level when it returns False instead.

packages/reflex-base/src/reflex_base/otel.py
```diff
@@ -227,6 +239,25 @@ def disable() -> None:
...
+
+    Args:
+        timeout_millis: Maximum time to wait for exporters to flush.
+    """
+    if not enabled:
+        return
+    provider = _tracer_provider or trace.get_tracer_provider()
+    force_flush = getattr(provider, "force_flush", None)
+    if force_flush and not force_flush(timeout_millis=timeout_millis):
+        logger.debug("OpenTelemetry span flush timed out after %d ms", timeout_millis)
```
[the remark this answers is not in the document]
FarhanAliRaza: Declined. The SDK's `TracerProvider.force_flush` and `SynchronousMultiSpanProcessor.force_flush` contain no raise path; exporter exceptions are caught inside `BatchProcessor._export` and logged by the SDK. Repro on this branch with an exporter whose `export()` raises `RuntimeError`: the SDK logs "Exception while exporting Span" and `otel.flush()` returns normally. The guard you propose was removed on purpose in 3d7dfe314 because it was dead code.

packages/reflex-base/src/reflex_base/otel.py
```diff
@@ -227,6 +239,25 @@ def disable() -> None:
...
+    created before the worker started.
+
+    Args:
+        timeout_millis: Maximum time to wait for exporters to flush.
+    """
+    if not enabled:
+        return
+    provider = _tracer_provider or trace.get_tracer_provider()
+    force_flush = getattr(provider, "force_flush", None)
+    if force_flush and not force_flush(timeout_millis=timeout_millis):
```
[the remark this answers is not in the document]
FarhanAliRaza: Declined. `force_flush` on the SDK provider does not raise: exporter errors are caught in `BatchProcessor._export` and `SynchronousMultiSpanProcessor.force_flush` only returns False. Reproduced with a raising exporter: the SDK logs the error and `otel.flush()` returns normally. A user-written provider that raises from `force_flush` would break every SDK shutdown path too, not just this one; that is not a case this PR guards against.

packages/reflex-base/src/reflex_base/otel.py
```diff
@@ -227,6 +239,25 @@ def disable() -> None:
     _message_size = _ws_connections = _NOOP_INSTRUMENT
 
 
+def flush(timeout_millis: int = 5000) -> None:
+    """Flush spans already ended by the current process.
+
+    Multiprocessing children exit with ``os._exit``. Before Python 3.13 that
+    skips ``atexit`` entirely, and since 3.13 it only runs hooks registered
```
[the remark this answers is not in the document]
FarhanAliRaza: Declined. Since 3.13, `multiprocessing/popen_fork.py` (and the forkserver child) calls `atexit._clear()` right after the fork and only later runs `atexit._run_exitfuncs()`, so inherited hooks are dropped and only hooks registered in the child run. Verified on 3.14.6: a provider created in the parent before the fork loses its spans without `flush()` (0 exported), while a provider created inside the worker flushes via its own atexit hook. The docstring describes exactly that.

FarhanAliRaza (approved):

PR #7159, opened 2026-09-15
fix(state): persist explicitly supplied states in StateManagerDisk

### LINKED ISSUE #6806: StateManagerDisk drops debounced set_state updates for non-BaseState tokens

FarhanAliRaza:
## Summary

`StateManagerDisk.set_state` silently lost data when called with a state instance that was not the one returned by `get_state`:

- The debounced write queue kept the first queued object and ignored later calls, so the new instance never reached disk.
- The in-memory cache (`self.states`) was never updated, so subsequent `get_state` calls returned the stale instance.
- A freshly constructed instance has no touched flags set, so even the immediate write path skipped it entirely.

## Approach

Alternative implementation of #6807 (thanks @harsh21234i for the analysis and regression tests). Instead of threading a `force_write` parameter through `QueueItem`, `set_state_for_substate`, and both flush paths, this encodes the persistence decision where it already lives: the state's own `_was_touched` flag.

`set_state` does one O(1) identity check against the cache. If the supplied instance is not the cached one, it marks the whole tree touched — the existing touched-driven machinery then persists it with **zero signature changes** to `set_state_for_substate` or the flush paths. The flag is sticky until write time, which also covers accumulation across debounced updates for free. Setting `_was_touched` externally is an established pattern (`_screenshot.py`, `vars/base.py`), the flag is stripped on pickling, and its docstring states it exists exactly for this purpose ("determines whether this state instance should be persisted").

Also:
- Queued `QueueItem`s are updated in place (`queued_item.state = state`) instead of ignored, per masenf's preference in #6807.
- The cache is updated on both paths (after the write on the immediate path, so a failed write is not cached).
- `set_state` records `_token_last_touched`, so a token that is only ever written still expires instead of leaking cache entries.

Non-`BaseState` tokens need no marking: their `get_and_reset_touched_state` always returns `True`.

## Testing

Regression tests (ported from #6807, written to fail before the fix): latest-queued-value flush, cache replacement for arbitrary instances (both debounce modes), and persistence of an untouched `BaseState` (both debounce modes).

Fixes #6806

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/7159?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tests/units/test_state.py
```diff
@@ -4486,6 +4486,81 @@ async def test_state_manager_disk_close_resets_write_queue_task():
     assert state_manager._write_queue_task is None
 
 
+@pytest.mark.asyncio
+async def test_state_manager_disk_debounced_set_state_flushes_latest_value(
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in ad5de3efc. Moved the three new StateManagerDisk tests to tests/units/istate/manager/test_disk.py and gave the untouched-persistence test a local DiskPersistState class so the module stays self-contained.

reflex/istate/manager/disk.py
```diff
@@ -338,17 +363,24 @@ async def set_state(
             context: The state modification context.
         """
         token = self._coerce_token(token)
+        _mark_replacement_state_touched(self.states.get(token.cache_key), state)
+        self._token_last_touched[token.cache_key] = time.time()
         if self._write_debounce_seconds > 0:
             # Deferred write to reduce disk IO overhead.
-            if token not in self._write_queue:
+            self.states[token.cache_key] = state
```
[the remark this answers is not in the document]
FarhanAliRaza: This failure mode is pre-existing and unchanged by this PR. On main, the flush loop already pops the item before writing (`self._write_queue.pop(token).state`) and the except handler logs and continues, dropping the item; the touched flag is likewise reset before the write attempt in BaseStateToken.get_and_reset_touched_state (token.py), which this PR does not touch. The cache also already holds the unpersisted instance on main in this scenario — get_state caches it — so this assignment introduces no new inconsistency; it writes the same instance the flush would have persisted.

Updating the cache only after the write succeeds is not possible for a debounced write: the write happens up to debounce_seconds later, and the cache must serve get_state the latest value in the meantime — that staleness is the very bug this PR fixes. A retry/requeue mechanism for failed flush writes would be a behavior change to pre-existing error handling, out of scope here.

reflex/istate/manager/disk.py
```diff
@@ -36,6 +36,31 @@ class QueueItem(Generic[TOKEN_TYPE]):
...
+    """Mark a state tree as touched when it replaces the cached instance.
+
+    A state instance not obtained from get_state carries no touched tracking,
+    so mark the whole tree touched to ensure it gets persisted.
+
+    Args:
+        cached_state: The instance currently cached for the token, if any.
+        state: The instance supplied to set_state.
+    """
+    if isinstance(state, BaseState) and state is not cached_state:
```
masenf: is this extra `isinstance` on a relatively hot path noticable in benchmarks? i don't think we have specific benchmarks for get_state / set_state in the different state managers.

masenf (commented):
i'm trying to see how possible it is to add benchmarks for the state managers so we actually can easily answer the question i've asked here.

reflex/istate/manager/disk.py
masenf:
```suggestion
    if state is not cached_state and isinstance(state, BaseState):
```

if we invert the conditional, then we only pay for the extra isinstance when the identity check fails rather than every time

masenf: the new benchmark tests did find this added function, and it adds 0.3% in overall cost. with the inverted conditional, this should be happening much more rarely that it wont matter

masenf (approved):

PR #7162, opened 2026-09-16
Fix superseding shared child events

### LINKED ISSUE #7041: Supersedes does not cancel events from different chains.
**Describe the bug**

If you have multiple events which return/yield the same event supersession does not cancel the first run of the returned event.

e.g.

You have event_a and event_b both return the background task do_stuff.
event_b is called while event_a is still running which causes do_stuff to run two times.
do_stuff is not canceled.

I wrote some tests regarding that behavior here
https://github.com/tim-haselhoff/reflex/pull/6

Is this intended behavior or a development oversight?

**To Reproduce**

- Code/Link to Repo:
https://github.com/tim-haselhoff/reflex/pull/6

**Expected behavior**

returned event is canceled.

**Screenshots**

- Python Version: 3.14
- Reflex Version: 0.9.10
- OS: Linux
- Browser (Optional): Chromium

**Additional context**

harsh21234i:
Fixes #7041.

When two different superseding root events enqueue the same superseding child event, the second child event did not cancel the first because supersession was only applied to root-level enqueues. This could allow stale background work to continue running alongside the newer event.

This change applies the existing supersession-key logic to chained events while preserving existing behavior for non-superseding handlers.

**Included:**

- Regression test covering distinct root chains sharing a superseding child event.

- Documentation update for chained-event supersession behavior.

- `reflex-base` changelog fragment.

**Validation:**

- 45 event processor tests passed.

- Changed-file pre-commit checks passed, including Ruff, codespell, pyright, ty, and the pyi hook.

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/7162?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/reflex-base/src/reflex_base/event/processor/event_processor.py
```diff
@@ -573,7 +590,8 @@ def _supersede_previous(
             return
         key = (event.name, token)
         previous = self._superseded.get(key)
-        if previous is not None and not previous.all_done():
+        is_ancestor = previous is not None and _is_future_ancestor(previous, tracked)
+        if previous is not None and not is_ancestor and not previous.all_done():
```
FarhanAliRaza: Untested behavior change worth pinning with a unit test: because the supersession key is `(event.name, token)` and ignores args, a handler that yields the same superseding handler twice in one batch — e.g. `return [State.do_other("x"), State.do_other("y")]` — now has `y` cancel `x` before it even starts (verified in a live app on this branch; previously both ran, since chained enqueues never superseded). The siblings share a parent, so the ancestor check doesn't protect them.

This is consistent with latest-wins and with how roots already behave, so it looks intended — but it's downstream-visible beyond the #7041 scenario and no test covers it. Suggest adding a test asserting only the last sibling runs, so the semantics are locked in deliberately.

---

packages/reflex-base/src/reflex_base/event/__init__.py
```diff
@@ -584,13 +584,14 @@ def is_background(self) -> bool:
...
-        When True, enqueuing this handler as a chain root cancels the previous
-        unfinished event chain rooted at the same handler for the same client
-        token. Cancellation is cooperative: a handler that never yields to the
-        event loop runs to completion, and only its not-yet-started chained
-        events are skipped.
+        """Whether a newer invocation supersedes an older one.
+
+        When True, enqueuing this handler cancels the previous unfinished
+        invocation chain for the same handler and client token. This also
+        applies when the handler is yielded by multiple distinct roots.
```
FarhanAliRaza, line 591: Edge case that might deserve a sentence here: cancellation is latest-*enqueue*-wins, not latest-user-action-wins. Roots with different names don't supersede each other, so if the older root chain is still running and enqueues the shared superseding child *after* the newer chain already did, the stale chain's invocation cancels the newer one. Unlikely in practice (children are usually enqueued promptly), and inherent to the `(event.name, token)` key design — just worth documenting so the semantics aren't surprising.

---

packages/reflex-base/news/7162.bugfix.md
FarhanAliRaza, line 1: Nit: the `+slug` form is for before the PR number is known — this can be renamed to `7162.bugfix.md` now.

---

FarhanAliRaza (commented):
Reviewed by exercising this branch in a real app (two roots returning a shared `@rx.event(background=True, supersedes=True)` child, a root-supersession pair, a self-chaining superseding poll loop, and a same-batch sibling yield), driven headless in Chromium, plus a static pass.

**Verdict: LGTM** — the fix does what it claims, with only minor non-blocking notes (inline).

Verified live on this branch:
- The #7041 repro is fixed: clicking Event A then Event B while the shared child sleeps yields `do_stuff(a) started → do_stuff(b) started → do_stuff(b) finished`, with the stale `a` invocation cancelled mid-sleep. No console errors.
- Root-level supersession is unchanged (two rapid clicks of the same superseding root → only the second finishes).
- The `_is_future_ancestor` guard works: a self-chaining superseding handler runs its full chain to completion.
- Plain (non-superseding) events unaffected.

Test/static evidence:
- The new regression test fails on `main`'s `event_processor.py` and passes with the fix — and the `main` failure mode also shows the old code leaked the `_superseded` tracking entry, which this change cleans up.
- Full `tests/units/reflex_base` suite passes (443 passed, 1 skipped); pyright and ruff clean on the changed files; the touched functions stay under the C901 complexity threshold.
- No component changes, so no `pyi_hashes.json` update needed; backend-only, no prod-build impact.

Inline comments: one suggested test to pin the (apparently intended) new sibling-yield cancellation semantics, one docstring suggestion for a latest-enqueue-wins edge case, and a news-fragment naming nit.

---

harsh21234i:
Hi @FarhanAliRaza , I’ve addressed the review feedback on PR #7162 and pushed the latest changes. The final commit is `f7e3d2b3`. I also added regression coverage, updated the changelog, and verified that 51 targeted tests and pre-commit checks pass. The branch is ready for another review. Thanks!

FarhanAliRaza:
superceded by this https://github.com/reflex-dev/reflex/pull/7168

PR #7164, opened 2026-09-16
fix: persist latest debounced disk state

### LINKED ISSUE #6839: StateManagerDisk debounced writes can persist stale state after repeated set_state calls

harsh21234i:
Fixes #6839.

When multiple `StateManagerDisk.set_state` calls occur during the debounce window, the queued write now uses the latest state value instead of persisting the first queued value.

### Changes

- Replace the existing queued item with the latest state while preserving debounce timing.

- Update the state cache with the supplied state instance.

- Track touched tokens for correct expiry behavior.

- Persist replacement `BaseState` trees even when they are not marked as touched.

- Add regression coverage in `tests/units/istate/manager/test_disk.py`.

- Add the required bugfix changelog fragment.

### Validation

- Focused StateManagerDisk regression tests passed.

- Existing StateManagerDisk tests passed.

- Ruff, Pyright, and pre-commit checks passed.

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/7164?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


FarhanAliRaza:
duplicated https://github.com/reflex-dev/reflex/pull/7159

PR #7171, opened 2026-09-16
docs: pass a backend function in the DnD provider example

Darksinian:
The manual provider example passes `backend="HTML5"`, but react-dnd expects a backend function. Replace it with `backend=rxe.dnd.HTML5Backend` and correct the mobile alternative to `rxe.dnd.TouchBackend`.

One line changed. All memoization and provider-discovery changes have been removed; framework fixes remain separate.

Validation: verified both backend values render as unquoted function references; Ruff formatting and `git diff --check` pass.


docs/enterprise/drag-and-drop.md
```diff
@@ -12,7 +12,9 @@ import reflex_enterprise as rxe
 Reflex Enterprise provides comprehensive drag and drop functionality for creating interactive UI elements using the `rxe.dnd` module. Built on top of react-dnd, it offers both high-level components for common use cases and low-level hooks for advanced scenarios.
 
 ```md alert warning
-# Important: Always decorate functions defining `rxe.dnd.draggable` components with `@rx.memo` to avoid compilation errors.
+# Important: Decorate functions defining `rxe.dnd.draggable` or `rxe.dnd.drop_target` with `@rx.memo`.
+
+Each drop target needs its own hook and ref. Pass dynamic destinations as `rx.Var` props using keyword arguments, especially inside `rx.foreach`. Place these memoized components below a shared [provider](#provider).
```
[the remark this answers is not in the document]
Darksinian: Fixed in 60187fa8. Narrowed the top guidance and Best Practice #1 to helpers rendered multiple times (including rx.foreach), and explained that a single instance can use the page hook scope. Also memoized the repeated drop_zone example with typed keyword props and a shared parent provider. Both changed demos compile independently with their own destination/position bindings; all six demos compile together.

docs/enterprise/drag-and-drop.md
```diff
@@ -293,7 +295,10 @@ def draggable_list_item(item: rx.Var[ListItem]) -> rx.Component:
     )
 
 
-def droppable_list(title: str, items: list[ListItem], list_id: str):
+@rx.memo
+def droppable_list(
+    title: rx.Var[str], items: rx.Var[list[ListItem]], list_id: rx.Var[str]
+):
```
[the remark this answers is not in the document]
Darksinian: Fixed in 60187fa8 by adding -> rx.Component to droppable_list. A targeted check against _warn_missing_annotations reproduced the deprecated fallback before this change and passes afterward. The compiled example still owns its useDrop hook, binds listIdRxMemo in the drop event, and uses the HTML5 backend function.

docs/enterprise/drag-and-drop.md
```diff
@@ -12,7 +12,9 @@ import reflex_enterprise as rxe
 Reflex Enterprise provides comprehensive drag and drop functionality for creating interactive UI elements using the `rxe.dnd` module. Built on top of react-dnd, it offers both high-level components for common use cases and low-level hooks for advanced scenarios.
 
 ```md alert warning
-# Important: Always decorate functions defining `rxe.dnd.draggable` components with `@rx.memo` to avoid compilation errors.
+# Important: Decorate functions defining `rxe.dnd.draggable` or `rxe.dnd.drop_target` with `@rx.memo`.
```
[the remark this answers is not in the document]
Darksinian: Fixed in 60187fa8. The guidance now applies to repeatedly rendered helpers and explains why single-instance page examples can use the page hook scope. The four-instance drop_zone helper is now memoized, takes rx.Var[int] by keyword, declares -> rx.Component, and renders under a shared provider. An isolated compilation check failed before the change because DropZone lacked a React component boundary, and now verifies its own useDrop and position prop. All six demos compile together; the unrelated single-instance demos are unchanged.

Darksinian:
The CodSpeed report flags `test_get_state[disk]` at 937.6 µs vs 908.6 µs (-3.09%). I checked the compared revisions (`1bc8936` → `60187fa`): the only changed file is `docs/enterprise/drag-and-drop.md`; framework/package source, benchmark fixtures, dependency lockfile, and performance workflow are identical. This state-manager benchmark does not load that Markdown page, so there is no changed benchmark input or runtime code path explaining the result. I am rerunning the benchmark job once to check reproducibility, without changing code or acknowledging away the required check.

docs/enterprise/drag-and-drop.md
```diff
@@ -596,7 +606,7 @@ def custom_collect_example():
 
 ## Provider
 
-Drag and drop functionality requires the `rxe.dnd.provider` component to wrap your app. The provider is automatically added when using `draggable` or `drop_target` components.
+Drag and drop requires `rxe.dnd.provider` above the components that own its hooks. High-level components register it automatically, but discovery can miss components inside `@rx.memo` boundaries (observed with Reflex `0.9.11.post1` and Reflex Enterprise `0.9.5`). Use a shared parent provider for memoized boards, as in the dynamic-list example, to avoid `Expected drag drop context` in the browser.
```
masenf: i am working on a framework fix for this. i'd rather have it working as intended instead of biting users and making them manually figure out where to stash the provider

docs/enterprise/drag-and-drop.md
```diff
@@ -605,19 +615,20 @@ def app():
...
 
 ```md alert warning
-# Do not add a second provider
+# Keep hook-owning components below the provider
 
-Because the provider is added automatically when `draggable` or `drop_target` components are used, wrapping the app in an additional `rxe.dnd.provider` results in duplicate providers and breaks drag and drop. Only use manual control when the automatic provider does not fit (e.g. to select the touch backend), and make sure it is the only provider in the tree.
+A provider returned inside the same memoized function cannot supply context to hooks called by that function. Put the provider in a parent of the memoized draggable and drop-target components. Share it across the board instead of adding one per card or column.
 ```
 
+The default backend is `rxe.dnd.HTML5Backend`. To select a backend explicitly, pass `backend=rxe.dnd.HTML5Backend` or `backend=rxe.dnd.TouchBackend`, not a string such as `"HTML5"` or `"Touch"`. Strings can cause `backendFactory is not a function`.
```
masenf: this is another place where i'd rather ensure the rxe implementation has proper typing and rejects known-invalid prop values at compile time instead of letting it bubble as a nasty frontend error

docs/enterprise/drag-and-drop.md
Darksinian: Removed the explicit provider wrappers and all provider-discovery workaround guidance. The Provider section is restored unchanged. This PR now only corrects the repeated target helpers (memo boundaries, typed props/returns, keyword calls); their hook/destination binding checks pass independently. End-to-end provider behavior is deferred to your framework fix.

docs/enterprise/drag-and-drop.md
Darksinian: Removed the backend-value/error guidance along with the provider workaround; the entire Provider section is unchanged from the base. Keeping this PR scoped to the independent memoization example fixes.

masenf (approved):

PR #7177, opened 2026-09-16
Split Reflex and Docs navbar logo links

Alek99:
Clicking either wordmark previously returned to the docs overview. Split the navbar logo so **Reflex** links to https://reflex.dev/ and **Docs** links to the docs overview, preserving the layout and giving each link an accessible name and keyboard focus indicator.

The Reflex wordmark uses `rx.el.elements.a` (a raw HTML anchor) so the off-site URL bypasses React Router, and its href comes from the shared `REFLEX_URL` constant. The Docs wordmark stays on `rx.el.a` so it resolves against the `/docs` basename.

Validation: all 13 navbar tests and pre-commit checks pass. Verified both rendered destinations and clicked Docs from an API reference page to confirm navigation to `/docs/` in the development preview.


docs/app/reflex_docs/views/docs_navbar.py
```diff
@@ -36,8 +36,10 @@ def github_button() -> rx.Component:
 
 
 def logo() -> rx.Component:
-    return rx.el.a(
-        rx.el.div(
+    """Link each wordmark to its respective site overview."""
+    focus_class = "rounded-sm focus-visible:outline-2 focus-visible:outline-offset-4 focus-visible:outline-ring"
+    return rx.el.div(
+        rx.el.elements.a(
```
masenf:
```suggestion
        rx.el.a(
```

like on line 57

masenf (dismissed):

FarhanAliRaza (approved):

PR #7193, opened 2026-09-18
Keep Granian logs out of JSON stdout

### LINKED ISSUE #6972: `reflex run --json --loglevel debug`: granian's own startup/shutdown lines are plain text on stdout, breaking JSON-lines parsing
**Describe the bug**

In dev mode with `--json`, granian's own logger writes plain-text lines directly to stdout, interleaved with the JSON records. At `--loglevel debug` these include `[INFO] Starting granian (main PID: ...)`, `[INFO] Listening at: ...`, `[INFO] Spawning worker-1 ...`, `[INFO] Started worker-1`, and the shutdown lines (`Shutting down granian`, `Stopping/Stopped worker-1`, `Granian shutdown completed, see ya!`). A strict JSON-lines consumer of stdout fails on them.

Not present at `--loglevel info`/default (granian's level is derived via `subprocess_level()`, so its INFO lines are filtered) or in `--env prod`.

**To Reproduce**

```
reflex init --template blank
reflex run --json --loglevel debug --frontend-port 3260 --backend-port 8260 > out 2> err
# wait for startup, then stop the server, then try to parse every stdout line as JSON
```

In a verified run, stdout contained 106 lines: 98 valid JSON records and 8 plain-text granian lines (4 startup + 4 shutdown). Depending on shutdown timing, `[ERROR] Unexpected exit from worker-1` can appear as a 9th plain-text line. stderr stays empty.

**Observed behavior**

Plain-text granian lines mixed into `--json` stdout.

**Expected behavior**

If `--json` guarantees parseable stdout, every line on stdout should be a JSON record (or granian's own logging should be routed/suppressed accordingly).

**Root cause (from source)**

`reflex/utils/exec.py::run_granian_backend` (and `run_granian_backend_prod`) construct `granian.server.Server` with `log_level=LogLevels(loglevel.value)` but never pass `log_dictconfig`/`log_enabled`, so granian's `_granian` logger keeps its default plain-text stdout handler (`"[%(levelname)s] %(message)s"`, `propagate=False`) even when JSON mode is on. The reflex logging pipeline (`reflex_base/utils/log.py`) never configures `_granian`. Fix direction: pass a JSON-formatting `log_dictconfig` (granian supports it), or `log_enabled=False` + re-emit, when JSON mode is active — in both the dev and prod granian paths.

- Python Version: 3.11
- Reflex Version: 0.9.9a1
- OS: linux

**Additional context**

- Severity: low. Not a regression (the `--json` pipeline is new in the 0.9.9 series).
- Found during 0.9.9a1 pre-release testing; test app, logs, and a `check_jsonlines.py` validator are on branch `claude/reflex-prerelease-testing-t0sd90` under `prerelease-testing/2026-08-27-v0.9.9a1/logging_cli/`.

flcrom:
Granian writes its own lifecycle messages straight to stdout. With `reflex run --json --loglevel debug`, those plain-text lines break JSONL consumers.

This disables Granian's native logger when Reflex JSON logging is active, for both dev and prod. Normal logging is unchanged.

I reproduced this on current main (8 non-JSON lines), then reran the same command after the change (0 non-JSON lines).

Tests: `uv run pytest -q tests/units/utils/test_exec.py` (17 passed), ruff, format check, pyright.

Closes #6972

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/7193?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


FarhanAliRaza (changes requested):
I dont think supression is the only fix we can pass custom handler to granian and get back json logs. 

```
log_dictconfig={
    "handlers": {
        n: {"()": "reflex_base.utils.log.JsonHandler"}
        for n in ("console", "access")
    }
} if log.is_json_mode() else None,

```

Granian's Server accepts a `log_dictconfig` argument.

FarhanAliRaza (approved):

PR #7216, opened 2026-09-19
fix: only record an uncached var as sent once its delta is delivered

### LINKED ISSUE #7212: 0.9.12a1: @rx.var(cache=False) withheld from a delta by a downstream get_delta filter is never re-sent (#6946 records "last sent" while building the delta, not on delivery)
**Describe the bug**

#6946 dedupes unchanged uncached computed vars by remembering the last value sent per client. The memo is written while the delta is *built*, not when it is delivered: `BaseState.get_delta` (`reflex/state.py` ~2380) calls `cvar._record_delta_value(...)`, and `ComputedVar._record_delta_value` (`packages/reflex-base/src/reflex_base/vars/base.py:2689`) immediately stores `instance.__last_delta_<js_expr> = (token, key)`. Nothing rolls it back when a downstream `get_delta` wrapper removes the entry, or when delivery fails. So a `@rx.var(cache=False)` that a filter withholds is marked "sent"; once the filter stops withholding it, the same value is deduped away and the client keeps the stale value until the var's value changes again.

This is not a synthetic pattern: the `get_delta` docstring documents that downstream code overrides it with a self-only signature, and reflex-enterprise 0.9.5 does exactly that for post-event auth filtering (`reflex_enterprise/auth/oidc/state.py:2529`); its `redeliver_protected()` (`auth/enforcement.py:960-995`) re-marks withheld vars dirty on the assumption that the end-of-event delta re-delivers them — the assumption this breaks for `cache=False` vars. Regression vs 0.9.11.post1, which delivers the value. The existing primitive for this, `_suppress_delta_recording()` (`reflex/state.py:317`), is private and only used by `reflex/istate/shared.py:122`.

**To Reproduce**

Browser-free (`uv pip install --prerelease=allow 'reflex==0.9.12a1'`):

```python
import reflex as rx

_SHARED = {"v": "default"}
AUTHENTICATED = False


class ProbeState(rx.State):
    hits: int = 0

    @rx.var(cache=False)
    def shared_label(self) -> str:
        return f"shared={_SHARED['v']}"


_orig = rx.State.get_delta


def filtered_get_delta(self):
    delta = _orig(self)
    if AUTHENTICATED:
        return delta
    return {sn: {k: v for k, v in sd.items() if "shared_label" not in k} for sn, sd in delta.items()}


rx.State.get_delta = filtered_get_delta


def sent(label, st):
    d = st.get_delta()
    print(label, any("shared_label" in k for sd in d.values() for k in sd), d)
    st._clean()


st = ProbeState(_reflex_internal_init=True)
st._mark_dirty(); sent("1 anon initial       ", st)
_SHARED["v"] = "SERVER-ONLY-SECRET"; st.hits += 1; sent("2 anon event         ", st)
AUTHENTICATED = True; st.hits += 1; sent("3 after login        ", st)   # expected: True
st.hits += 1; sent("4 after another event", st)                           # expected: True
```

- 0.9.12a1: steps 3 and 4 print `False` — `shared_label` is never delivered again.
- 0.9.11.post1: `True` for both.

Same result in a browser with a State subclass overriding `get_delta` (dev/in-memory, dev + redis single worker, prod + redis 9 workers): the websocket delta after the filter is lifted carries only the other vars and the page shows the stale value until the var's value changes again. Under redis the stale `__last_delta_…` memo is visible inside the pickled state, so this is neither a prod-build nor a multi-worker artifact.

**Expected behavior**

An uncached var that was withheld from the delivered delta is re-sent as soon as it is no longer withheld (0.9.11.post1 behaviour).

**Suggested fix shape**

Record the "last sent" key after the delta has been filtered/emitted, or expose `_suppress_delta_recording()` / a post-filter hook to `get_delta` overrides — instead of recording while building the delta.

- Python Version: 3.11.15
- Reflex Version: 0.9.12a1 / reflex-base 0.9.12a1 (regression vs 0.9.11.post1)
- OS: Linux (Ubuntu 24.04 container)

**Additional context**

0.9.12a1 pre-release QA campaign FINDING-003, reproduced through reflex-enterprise auth, in pure reflex (dev, dev + redis, prod + redis) and by two independent adversarial verifiers. Evidence and scripts: [`prerelease-testing/2026-09-18-v0.9.12a1/`](https://github.com/reflex-dev/reflex/tree/claude/reflex-prerelease-testing-t0sd90/prerelease-testing/2026-09-18-v0.9.12a1) — `event_loop/` (page `/filtered` of `elapp`, `scripts/s_filtered.py`, `verification/`), `ent_mcp_oidc/verification/2026-09-19-adversarial/pure_delta_memo.py`, and `FINDINGS.md`.

masenf:
### All Submissions:

### Type of change

### Changes To Core Features:

## Summary

closes #7212

#6946 dedupes unchanged `@rx.var(cache=False)` values by remembering the last value sent per client. The memo was written while the delta was *built* (`BaseState.get_delta` → `ComputedVar._record_delta_value`), so a downstream `get_delta` override that withholds an entry — the pattern the `get_delta` docstring documents, and what reflex-enterprise 0.9.5's auth filtering does — left the memo claiming the client had a value it never received. Once the filter stopped withholding, the identical value was deduped away and the client stayed stale until the value changed again. Regression vs 0.9.11.post1.

## What changed

- `get_delta` gathers candidate `_DeltaRecord`s (state, key, value, instance, memo attribute, stored key) into a `ContextVar` while building the delta; `ComputedVar._record_delta_value` became `_pending_delta_record` and no longer writes the memo.
- `_get_resolved_delta` — the single chokepoint every delivery path uses (`app.py`, `istate/proxy.py`, reflex-base `base_state_processor.py`, `istate/shared.py`), and the one place that sees the delta after the override chain with coroutines resolved — commits a record only where `delta[state][key] is value`, i.e. the very object that was computed survived. Dropped or replaced entries are owed to the client again.
- `_suppress_delta_recording()` is still honoured, and a suppressed traversal nested inside a recording one records nothing.
- Behaviour change to be aware of: a bare `get_delta()` call no longer dedupes (only delivered deltas record). Ten existing #6946 unit tests moved onto `await state._get_resolved_delta()` with their assertions unchanged.

## Tests

- New: `test_uncached_var_withheld_by_delta_override_is_resent` (dropped / replaced-with-placeholder × sync / async) and `test_uncached_computed_var_recorded_only_once_delivered`. The sync cases and the latter fail on `main` and pass here; the async cases document that a filter closes the coroutine before it can reach the memo.
- Mutation checks: making `_delta_value_key` return a bare value, or a constant instead of `_UNKEYABLE_VALUE`, fails the corresponding #6946 tests again (a first version of this change had left two of them on the non-recording path; fixed during review).
- `uv run pyright reflex tests`: 0 errors. ruff / format clean. `uv run pytest tests/units --ignore=tests/units/reflex_cli`: 9137 passed together with the other three campaign fixes.

## Verification against the real breakage

- Browser-free probe from the pre-release campaign: on the published 0.9.12a1 the withheld var is never re-sent (steps 3/4 `False`); on this branch it is delivered (`True`), matching 0.9.11.post1.
- Playwright repro (`elapp` page `/filtered`), dev in-memory and dev + redis: the published alpha shows the stale `secret-0` after the filter is lifted and `secret-2` after the next round; this branch shows `secret-1` and `secret-3`, with zero console/page/network errors. #6946's savings are intact: the campaign's `/uncached` probe sends the same 24 delta frames with the same keys (15.3 KB inbound vs 26.1 KB on 0.9.11.post1).

## Notes for review

- The commit still happens before `emit_update`, so a delta the socket never delivers is recorded as sent (pre-existing; a reconnect re-hydrates via `dict()`). Moving it past the emit is a separate decision.
- A downstream package that filters *after* `_get_resolved_delta` would still get values recorded; reflex-enterprise 0.9.5 wraps only `get_delta` and `dict` (verified from the wheel), passing surviving values through by reference.
- Survival is tested by identity; a filter that returns a copy of a value loses the dedupe for that var (chattier, never stale).
- Withholding an *async* uncached var leaves the inner coroutine unawaited (`RuntimeWarning`; pre-existing, reproduced on `main`, filtered in the new test).
- Independent adversarial review blocked the first version (two pyright errors, two vacuous tests); both fixed in the amended commit. Report with the review and follow-up: [fixes/f003/REPORT.md](https://github.com/reflex-dev/reflex/blob/claude/reflex-prerelease-testing-t0sd90/prerelease-testing/2026-09-18-v0.9.12a1/fixes/f003/REPORT.md).

https://claude.ai/code/session_01NeyxWsuC9hqSKq8YcEwZyY

---
_Generated by [Claude Code](https://claude.ai/code/session_01NeyxWsuC9hqSKq8YcEwZyY)_

<a href="https://cubic.dev/pr/reflex-dev/reflex/pull/7216?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


reflex/state.py
```diff
@@ -330,6 +352,36 @@ def _suppress_delta_recording() -> Iterator[None]:
...
+    replaced it with a placeholder, and what the client never received has to
+    be sent again later.
+
+    Args:
+        pending: The records gathered while the delta was built.
+        delta: The delta as it comes back out of ``get_delta``, resolved.
+    """
+    for state_name, key, value, instance, attr, stored in pending:
+        subdelta = delta.get(state_name)
+        if subdelta is None or key not in subdelta or subdelta[key] is not value:
```
FarhanAliRaza: P2: Invalidate the previous memo when the filter sends a replacement.

In a real app, I sent `cycle-1`, changed the value to `cycle-2` while the filter replaced it with `anon`, then released the filter and restored `cycle-1`. This branch skips the replacement but retains the memo for `cycle-1`. The next delta omits the value, so the page stays on `anon`, even after another event. The same app restores `cycle-1` on the base commit. A focused regression test passes on the base and fails here.

Clear the stored memo when the key remains in the outgoing delta with a different value, mark the state touched so this persists, and add this cycle to the regression tests.

reflex/state.py
```diff
@@ -2371,23 +2433,36 @@ def get_delta(self) -> Delta:
...
                 if inspect.iscoroutine(value):
-                    value = _drop_unchanged_delta_value(cvar, self, value, token)
-                elif not cvar._record_delta_value(self, value, token):
-                    continue
-            subdelta[prop + FIELD_MARKER] = value
+                    value = _drop_unchanged_delta_value(
+                        cvar, self, value, token, full_name, key, pending
+                    )
+                else:
+                    record = cvar._pending_delta_record(self, value, token)
```
FarhanAliRaza: P3: Share the record-or-drop logic with `_drop_unchanged_delta_value`.

This block repeats the `_pending_delta_record` call, the unchanged-value branch, and the conditional `_DeltaRecord` append from the async helper. Extract a helper that processes a resolved value and returns that value or `_DROP_FROM_DELTA`. Call it from both paths so changes to memo handling remain consistent.

FarhanAliRaza (changes requested):
Tested this commit in a real Reflex app in Chrome and compared it with the main-branch base (`4cba00435`). The app uses downstream `get_delta` filters that drop values or replace them with placeholders. I also checked async uncached values, mutable lists, cached dependencies, and unchanged-value deduplication through DOM assertions and socket frames.

The base restores the earlier value in the placeholder cycle; this commit leaves the placeholder displayed. A focused regression test passes on the base and fails on this commit. The existing state, computed-variable, shared-state, proxy, and event-processor unit tests pass.

reflex/state.py
masenf: Confirmed and fixed in 3c8cb5b60 — thanks, this was a real hole. Your cycle reproduces exactly as described: `secret-1` delivered, `secret-2` replaced with `anon`, then back to `secret-1` and the value is deduped away while the page still shows `anon`.

`_commit_delta_records` now splits the two withholding cases instead of treating them alike:

- key **dropped** → the client is still on the value the record names, so the record stands;
- key **replaced** → the client holds something this side never computed, so the record is discarded (`delattr` + `_was_touched = True`) and the next value is sent whatever it is.

The async case needed one more thing: a filter closes the wrapper coroutine before the value it carries can ever be compared, so there was no record to invalidate. `get_delta` now stakes the key on the wrapper at build time with an invalidating record; the real record is appended as the delta resolves, comes later in the list, and wins for the values that survive. Same cycle, same outcome, sync and async.

The cycle is in the regression tests: `test_uncached_var_withheld_by_delta_override_is_resent` now continues past the release with a withheld `secret-2` and a restored `secret-1`, asserting the value is re-sent for `replaced` and correctly still omitted for `dropped` (there the client never left `secret-1`). Both parametrizations fail without their half of the fix — reverting the replacement branch fails `[False-replaced]`, dropping the wrapper record fails `[True-replaced]`.

---

reflex/state.py
masenf: Done in 3c8cb5b60. Extracted `_record_or_drop_delta_value(cvar, instance, value, token, state_name, key, pending)`, which does the `_pending_delta_record` call, the unchanged-value branch and the `_DeltaRecord` append, and returns the value or `_DROP_FROM_DELTA`. `_drop_unchanged_delta_value` is now just `return _record_or_drop_delta_value(cvar, instance, await value, ...)`, and the sync path in `get_delta` calls it directly and `continue`s on the sentinel.

`pending` also lost its `| None` on both: after the sibling fix below, neither path is reached without a collector, so the conditional append is gone from the shared helper too.

---

reflex/state.py
```diff
@@ -2371,23 +2433,36 @@ def get_delta(self) -> Delta:
         always_dirty_computed_vars = self._always_dirty_computed_vars
         # Token of the client this delta is for, used to know which values it has.
         token = self.router.session.client_token if always_dirty_computed_vars else ""
+        # Where to leave the values this traversal sends, for whoever delivers
+        # the delta to record them.
+        pending = _pending_delta_records.get() if always_dirty_computed_vars else None
```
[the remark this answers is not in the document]
masenf: Valid — fixed in 3c8cb5b60, though at the source rather than here.

`_suppress_delta_recording()` now clears the collector as well as the flag:

```python
token = _record_delta_values.set(False)
records_token = _pending_delta_records.set(None)
```

That keeps the two ContextVars consistent by construction, so a block entered anywhere — including a `get_delta` override partway through a delivered traversal — records nothing, and `get_delta` does not pay a second ContextVar lookup per state in the tree to re-derive it.

Covered by `test_suppressed_delta_inside_a_delivered_one_records_nothing`: an override that wraps `super().get_delta()` in the suppressor still yields the value on every delta. It fails without the two lines above.

---

reflex/state.py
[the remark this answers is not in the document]
masenf: Agreed — fixed in 3c8cb5b60. The dedupe branch is now gated on the collector:

```python
if pending is not None and prop in always_dirty_computed_vars:
```

Leaving a value out of a delta is only safe where delivering that delta is what records it, so a delta nobody collects now neither records nor dedupes. `token` moved behind the same condition, which also skips the `router.session` lookup when nothing is being recorded.

`test_uncached_computed_var_recorded_only_once_delivered` now asserts a bare `get_delta()` after a delivered one still carries the value; it fails with the old unconditional branch.

---

FarhanAliRaza (approved):

PR #7269, opened 2026-09-22
fix(min-deps): resolve workspace siblings from the workspace, and reuse the build jobs' wheels

masenf:
Rebased onto main now that #7268 has merged, so this is script and CI only — no version bumps.

### Type of change

### Changes To Core Features

**The blind spot.** `scripts/check_min_deps.py` fails a package only on pyright errors that are *new* at its declared floors, measured against a "latest" resolution. Both ends resolved from PyPI, so an error present in both cancelled out — which is exactly what happens to a cross-package API for as long as it is unpublished.

#7121 added `reflex_base.components.tags.CommonTag` and five component packages imported it without raising their floor on `reflex-base`. No published `reflex-base` had the symbol, both resolutions reported the same `"CommonTag" is unknown import symbol`, and the check passed on every run — until `reflex-base` 0.9.12 reached PyPI, at which point it went red on already-merged code (#7268 fixed those floors). The window in which it is blind is precisely the development cycle of a release train, which is the window in which a floor should have been raised.

**The fix.** Build the workspace siblings into the wheelhouse that until now held only `*.dev` pins, and have the baseline resolution install those builds, so `latest` means *this workspace* for a sibling and PyPI for everything else. A symbol that exists at HEAD but not at the floor is then new at the minimum, and the check reports it on the PR that introduces the coupling — asking for the `*.dev` pin the release pipeline already knows how to rewrite. Third-party dependencies are untouched: `--no-sources` still forces them from PyPI at both ends, and the cancelling-out that suppresses `sqlalchemy`/`pandas`/circular-import noise is unchanged.

**And the CI side.** Each check job was building those siblings itself, so the same wheels were produced once per job. The build workflow already builds every package, so it now uploads each wheel as an artifact and the min-deps check moved in beside it, where `needs:` can reach those jobs. `--wheelhouse DIR` seeds the index from them rather than replacing the build step, so anything the directory does not cover — a release train whose `*.dev` floor has outrun the tags, which the build jobs know nothing about — is still built at the floor. Local runs without the flag are unchanged.

### Commits

Reviewable in order; the last four are each a defect the previous one exposed.

1. **`resolve workspace siblings from the workspace`** — the fix above.
2. **`build each sibling once per package, and serialize the builds`** — sibling builds race under `-j`: the pyi build hook rewrites the stubs in a sibling's own source tree, and nine packages declare `reflex-components-core`.
3. **`build the whole workspace closure, and check only real source`** — the closure, [reported by cubic](https://github.com/reflex-dev/reflex/pull/7269#discussion_r4067985121): a sibling's build environment resolves *its own* runtime requirements when its backend sets `require-runtime-dependencies`, as the root package's does, so `reflex-docgen` built `reflex` without the `reflex-base` its build then asked for. Plus: naming the source *directory* to pyright let it check the generated `.pyi` under `src/` as if it were source.
4. **`build the sibling wheels once per run, before any check starts`** — serializing builds against each other was not enough, because the other side of the race is a pyright run. Building everything up front means no build overlaps a check, and each sibling is built once per run rather than once per package that declares it.
5. **`take the min-deps sibling wheels from the build jobs`** — the CI change above.

### Verification

End to end, by adding a symbol to `reflex-base` HEAD and importing it from `reflex-components-core` — the situation #7121 created. Same working tree, both scripts:

```
$ python scripts/check_min_deps.py reflex-components-core        # before this PR
  PASS               reflex-components-core

$ python scripts/check_min_deps.py reflex-components-core        # after
  FAIL (min-version) reflex-components-core
  .../base/bare.py:10:41 - error: "UNPUBLISHED_SENTINEL" is unknown import symbol
```

Repeated against wheels built the way the build jobs build them, so the `--wheelhouse` path is covered too: detection unchanged, the unpublished-floor case still resolves, and no sibling's source tree is touched at all. Full sweeps pass repeatedly either way — 85 s with a seeded wheelhouse against 111 s without:

```
$ uv run python scripts/check_min_deps.py -j 4
All 20 package(s) passed.
```

`tests/units/test_check_min_deps.py` is at 61 tests. Two are regressions for the original bug (`_local_sources` selecting plain sibling floors, and `reflex-components-core` recording `reflex-base`) and fail against the old selection logic; the rest cover the transitive closure and its ordering, the baseline pin, the pin being withheld from the minimum resolution, a build skipped when it falls below its floor, a build redone at a declared `*.dev` floor, a prebuilt wheel being kept, and a prebuilt wheel below a floor still being redone. `ruff check`, `ruff format --check` and `pyright` clean.

### One thing for a maintainer

`min_deps.yml` is deleted, so its check names change from `check (<package>, <version>)` to `check-min-deps (<package>, <version>)`, now under `Build all packages`. **Required status checks need updating to match**, or PRs will wait on contexts that no longer exist.

The build jobs' triggers are deliberately left as they were: `min_deps.yml` had `paths-ignore: ["**/*.md"]`, and a `paths-ignore` on a required check never reports on a docs-only PR. The cost is that min-deps now also runs on docs-only PRs.

No news fragment: `scripts/` and `tests/` are not package source prefixes, so `changelog-check` reports no packaged source changes.

https://claude.ai/code/session_01RYkJBgnaybdNTyWLZVWmBm


scripts/check_min_deps.py
```diff
@@ -682,6 +889,13 @@ def main() -> int:
         default=1,
         help="Number of packages to check in parallel (default: 1).",
     )
+    parser.add_argument(
+        "--wheelhouse",
+        type=Path,
+        help="Directory of prebuilt workspace wheels to use instead of building them. CI "
+        "points this at the artifacts the build workflow already produced; anything the "
+        "selection needs that is missing from it is still built.",
```
masenf: > anything the selection needs that is missing from it is still built.

i would suggest making wheel building an opt-in flag (for local runs, like `--build-wheels`), otherwise we should proceed with only what is provided by the earlier CI jobs.

otherwise, we could get into a drift situation where the `build` workflow creates wheels that this script deems as unusable. for one, it would be a waste of build minutes; but more crucially it would be a incongruence between how the system performs the wheel build and how it checks that that deps are properly aligned.

masenf: Agreed, and done in `6d24b94`. Building is now `--build-wheels`, which local runs pass and CI does not; a sibling the wheelhouse doesn't hold is an error naming it rather than something built over.

One case I split out rather than failing on, because no build can fix it: a wheel that is *present* but can't satisfy what's declared against it. A package is numbered from the newest tag its checkout reaches, so a release tagged off a branch leaves the workspace permanently below a floor raised to that release. That's reported and left to PyPI.

Which is how your instinct turned up a live one — the repo is in exactly that state today:

```
warning: reflex-components-moment builds as 0.9.3.post249.dev0 here, which does not
satisfy reflex-components-moment>=0.9.4. Resolving it from PyPI instead of the
workspace, so a change to it is not covered by this run.
```

`reflex-components-moment-v0.9.4` isn't an ancestor of `main`, so the root package's floor on it has never actually been checked against the workspace. That was silent before and is printed on every run now. Happy to hard-fail on it instead if you'd rather force the tag onto `main`, but that would take CI red today, so I left it as a warning.

Full sweep passes both ways: 93 s from a wheelhouse, 119 s building. `AGENTS.md` now documents `--build-wheels` as the local invocation.

---

masenf:
On Greptile's P1 (*Fallback Restores Blind Spot*) — the finding is correct, and it's live rather than hypothetical. `reflex-components-moment` is in exactly that state for the root package right now, so `reflex`'s check resolves moment from PyPI at both ends and a new moment API used by `reflex` would cancel out, just as `CommonTag` did.

I did not close it by fabricating a version that satisfies the floor, which is the obvious fix and the wrong one. `reflex-components-moment-v0.9.4` was tagged off a branch and never landed on `main`, so there's no reason to believe main's moment source is what 0.9.4 shipped. Pinning the workspace build *as* 0.9.4 would type-check the root against source that may predate the release it claims to be — the check would stop reporting the gap while quietly being wrong. Falling back to PyPI is the honest answer; getting the tag onto `main` is what actually restores the coverage.

What was fair in the finding is "silently". It said so once, early, in a log that then runs for minutes. As of `420c728` the uncovered siblings print next to the PASS/FAIL list:

```
  PASS               reflex

Not covered by the workspace (resolved from PyPI at both ends):
  reflex-components-moment: builds as 0.9.3.post249.dev0 here, which does not
  satisfy reflex-components-moment>=0.9.4

All 1 package(s) passed.
```

@masenf — this is the same decision I put to you on [the `--build-wheels` thread](https://github.com/reflex-dev/reflex/pull/7269#discussion_r4068353264), so it's one call, not two:

1. **Leave as is** — degraded coverage for moment only, printed on every run, until the tag reaches `main`.
2. **Hard-fail instead** — forces the tag onto `main`, but takes the `reflex` check job red today and blocks every PR until someone does it.

I've left it at (1) because (2) is a call about your release process rather than about this script, and it would go red on `main` the moment this merges. Happy to flip it in a line if you'd rather have the forcing function.

---

scripts/check_min_deps.py
```diff
@@ -402,50 +477,202 @@ def _dev_build_version(project: dict, dependency_name: str) -> str | None:
...
+                return {}, detail
+            continue
+        # The workspace cannot number this sibling high enough for what is declared against
+        # it: its release was tagged off a branch, so the newest tag this checkout reaches is
+        # older than the floor. No build gets out of that, so say so and leave it to PyPI,
+        # which is what the pin would otherwise have overridden.
+        print(
+            f"warning: {name} builds as {built} here, which does not satisfy "
+            f"{', '.join(str(r) for r in declared)}. Resolving it from PyPI instead of the "
+            "workspace, so a change to it is not covered by this run."
```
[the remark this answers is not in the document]
masenf: Confirmed, and it's live rather than hypothetical — `reflex-components-moment` is in this state for the root package today. Full reasoning in [this comment](https://github.com/reflex-dev/reflex/pull/7269#issuecomment-5771155547); the short version:

"Use the local sibling at a usable version" is the fix I deliberately didn't take. `reflex-components-moment-v0.9.4` was tagged off a branch and never landed on `main`, so pinning the workspace build *as* 0.9.4 would type-check the root against source that may predate the release it claims to be — the check would stop reporting the gap while quietly being wrong, which is worse than the gap.

"Fail closed" is a real option and I've put it to @masenf, but it takes the `reflex` check job red on `main` the moment this merges and blocks every PR until the tag is moved — his call, not this script's.

What I did take from the finding is "silently": as of `420c728` the uncovered siblings print next to the PASS/FAIL list rather than once, early, in a log that runs for minutes.

---

masenf:
**CI is red on `f007bf38`, intentionally, and needs a tag pushed before it can go green.**

6 of 64 jobs fail — `check-min-deps` for `reflex`, `reflex-components-internal` and `reflex-docgen` on both Python versions. Those are exactly the three packages whose workspace closure reaches `reflex-components-moment`; the other 17 pass.

```
the wheelhouse cannot serve every workspace sibling:
  reflex-components-moment: builds as 0.9.3.post253.dev0+5656328 here, which does not
  satisfy reflex-components-moment>=0.9.4
```

This is the hard-fail @masenf asked for, working as intended: `main` carries a `reflex-components-moment >= 0.9.4` pin that no revision of `main` can satisfy, because `reflex-components-moment-v0.9.4` points at `9edd00ad4` — a release-from-prerelease commit that is not an ancestor of `main`. The newest moment tag `main` reaches is `v0.9.3`.

**The remedy, which only someone with push access to the repo can apply:**

```bash
git tag reflex-components-moment-v0.9.4.post1 8b393b587
git push origin reflex-components-moment-v0.9.4.post1
```

`8b393b587` is *"fix(moment): isolate each component from imported global locales"* (#7110). Its `packages/reflex-components-moment/src` tree hashes to `30d403a9fded`, byte-identical to the `v0.9.4` tag's, and the whole package directory diffs empty against that tag outside `CHANGELOG.md` and `news/`. Nothing has touched `moment/src` on `main` since, so it is both the defining commit and a stable choice.

With that tag in place the checkout derives `0.9.4.post1.dev101`, which satisfies the floor. Verified locally: 20/20 packages pass in both `--wheelhouse` and `--build-wheels` modes.

I have not pushed the tag myself — it is a release-shaped ref on this repository, and @masenf asked for the command to run rather than for me to run it. I have not re-run the jobs either: the failure is deterministic, not flaky, and a re-run without the tag would fail identically. Once the tag is on the remote, this PR needs no new commit — just a re-run.

---

scripts/check_min_deps.py
```diff
@@ -402,50 +477,204 @@ def _dev_build_version(project: dict, dependency_name: str) -> str | None:
...
+        floor = _dev_build_version(declared)
+        if build and floor is not None:
+            detail = _build_sibling(source, wheelhouse, floor)
+            if detail is not None:
+                return {}, detail
+            continue
+        unusable.append(
+            f"  {name}: builds as {built} here, which does not satisfy "
+            f"{', '.join(str(requirement) for requirement in declared)}"
+        )
```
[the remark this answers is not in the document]
masenf: Correct, and intended — @masenf asked for fail-closed here precisely so a pin `main` cannot satisfy stops being mergeable. "Tagging the matching source on main" is the resolution, and the commit is identified: `8b393b587`, whose `moment/src` tree is byte-identical to the `v0.9.4` tag's.

```bash
git tag reflex-components-moment-v0.9.4.post1 8b393b587
git push origin reflex-components-moment-v0.9.4.post1
```

Details and verification in [this comment](https://github.com/reflex-dev/reflex/pull/7269#issuecomment-5771427167). Leaving this thread open — it is genuinely blocking until that tag is on the remote, at which point these six jobs pass on a re-run with no new commit.

---

scripts/check_min_deps.py
masenf: Closing this out: you took the hard-fail option, `reflex-components-moment-v0.9.4.post1` is now on `main`, and `f007bf38` drops the PyPI fallback entirely. A sibling the wheelhouse can't serve — absent, or present at a version below its declared floor — is an error naming it and the `git tag` command that fixes it. `--build-wheels` stays opt-in for local runs, so CI proceeds with exactly what the `build` jobs produced and nothing else.

`Build all packages` is green on `0cf472a8`, all 64 jobs including the six `check-min-deps` jobs that this was failing closed on.

---

scripts/check_min_deps.py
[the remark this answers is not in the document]
masenf: Resolved as predicted. `reflex-components-moment-v0.9.4.post1` was pushed to `8b393b587`, so `git describe` from `main` now reaches it and the workspace wheel numbers as `0.9.4.post1.dev…`, satisfying `reflex-components-moment>=0.9.4`.

The six `reflex` / `reflex-components-internal` / `reflex-docgen` jobs needed a fresh `build` matrix rather than a job re-run — the artifacts they consume were produced before the tag existed — so `0cf472a8` re-triggered the whole workflow. `Build all packages` is green there, 64/64.

---

scripts/check_min_deps.py
```diff
@@ -402,50 +477,204 @@ def _dev_build_version(project: dict, dependency_name: str) -> str | None:
...
+        built = _wheel_versions(wheelhouse).get(name)
+        if built is None and build:
+            detail = _build_sibling(source, wheelhouse, None)
+            if detail is not None:
+                return {}, detail
+            built = _wheel_versions(wheelhouse).get(name)
+        if built is None:
+            unusable.append(f"  {name}: absent from the wheelhouse")
+            continue
+        if _satisfies(declared, built):
```
FarhanAliRaza, line 631: This skips the build at a declared `*.dev` floor when the checkout version already satisfies it. That makes the `reflex` result depend on the uv version.

`reflex` declares `reflex-base >= 0.9.12.dev0`. The checkout builds `reflex-base` as `0.9.12.post12.dev0`. That version satisfies the floor, so no wheel is built at `0.9.12.dev0`. The minimum resolution then resolves from PyPI, where the specifier admits pre-releases:

- uv 0.11.23: `lowest-direct` installs `reflex-base==0.9.12a1`. That version has no `_validate_state_name` or `_reflex_state_root`, so `reflex` fails (`reflex/state.py:70`, `:1129`, `:1642`, `:1848`).
- uv 0.12.17 (CI): it installs `reflex-base==0.9.12`, so `reflex` passes.

`required-version = ">=0.9.17"` allows both. `uv pip compile --resolution lowest-direct` on `reflex-base >= 0.9.12.dev0` shows the difference directly.

On main, the wheel built at the `*.dev` floor sorted below every PyPI pre-release. So the minimum resolution always took the workspace source. Please keep that for a declared `*.dev` floor, or raise `required-version` to a uv that resolves this the same way as CI.

scripts/check_min_deps.py
```diff
@@ -710,13 +967,26 @@ def main() -> int:
...
-            )
-    else:
-        results = [check_package(p, args.python) for p in selected]
+    with tempfile.TemporaryDirectory(prefix="min-deps-wheelhouse-") as tmp:
+        wheelhouse = Path(tmp) / "wheelhouse"
+        if args.wheelhouse is not None:
+            # Copied rather than used in place, so a wheel built to cover a gap in the
+            # prebuilt set never lands in the caller's directory.
+            shutil.copytree(args.wheelhouse, wheelhouse)
+        versions, detail = build_wheelhouse(selected, wheelhouse, args.build_wheels)
```
FarhanAliRaza: A run without flags now fails. `uv run python scripts/check_min_deps.py reflex-components-core` exits 1:

```
the wheelhouse cannot serve every workspace sibling:
  reflex-base: absent from the wheelhouse
  ...
A sibling that is absent means this check and the build jobs that produce these wheels have drifted; pass --build-wheels to build it here instead.
```

Nothing drifted. The user only did not pass a flag. The PR description also says "Local runs without the flag are unchanged", which is not true after 6d24b94e4.

Please build by default when `--wheelhouse` is not given:

```python
build = args.wheelhouse is None or args.build_wheels
```

CI passes `--wheelhouse`, so it stays strict. `--build-wheels` can then go, and the AGENTS.md command stays as it was.

FarhanAliRaza (changes requested):
I tested the script on the PR head and compared it with the script from main.

First, I reproduced the original blind spot. I added an unpublished symbol to `reflex_base.components.tags` and imported it in `reflex-components-core`. The main script passes this. The PR script fails it with `--build-wheels`. It also fails it with `--wheelhouse`, where I built the sibling wheels the same way the build jobs do. The fix works on both paths.

Then I ran a full sweep on a clean tree with `--build-wheels -j 4`. I also ran the unit tests and ruff.

Two requested changes are inline.

scripts/check_min_deps.py
masenf: Confirmed, and fixed in `c2e53c7`. `reflex-base` has `0.9.12a1` and `0.9.12a2` on PyPI below `0.9.12`, and `>= 0.9.12.dev0` is itself the opt-in that puts them in reach — so `lowest-direct` had three candidates and picked by uv version. Locally on 0.12.15 I get your second result:

```
$ uv pip compile --no-sources --resolution lowest-direct --no-deps  # reflex-base >= 0.9.12.dev0
reflex-base==0.9.12
```

I took neither option you offered, because each leaves a hole. Rebuilding at the floor only works where building is allowed, so `--wheelhouse` runs — which is CI — would have kept resolving from PyPI and stayed uv-dependent. And raising `required-version` pins the symptom: the floor would still be satisfied by whichever pre-release a future uv decides to prefer.

Instead the minimum resolution now pins it. A `*.dev` floor names a version that was never published, so the workspace build is the only thing that can satisfy it on purpose rather than by accident; saying so outright is deterministic on both paths and needs no second build.

Scoped to what the package declares **itself**, since `lowest-direct` pins direct dependencies and a floor one sibling declares against another is that sibling's own to test:

```
reflex:                 minimum: ['reflex-base==0.9.12.post12.dev0+abc']
reflex-components-core: minimum: []
```

That second line is the part that matters — `reflex-components-core`'s `reflex-base >= 0.9.12` is a published floor, still resolved from PyPI, so the case this check exists to catch is untouched. `check_min_deps.py --python 3.12 reflex` passes on the new head.

---

scripts/check_min_deps.py
masenf: You're right on both counts — the failure and the stale claim in the description. Done as you suggested in `c2e53c7`:

```python
versions, detail = build_wheelhouse(selected, wheelhouse, build=args.wheelhouse is None)
```

`--build-wheels` is gone, and `AGENTS.md` is back to the plain invocation. A caller who supplied `--wheelhouse` means those wheels to be the whole story; one who did not has nowhere else to get them — which keeps masenf's requirement that CI proceed with exactly what the build jobs produced, without making a local run guess a flag. The "absent from the wheelhouse" message now points at `--wheelhouse` rather than a flag that no longer exists.

```
$ uv run python scripts/check_min_deps.py --python 3.12 reflex-components-core
  PASS               reflex-components-core
```

I've corrected the PR description too.

---

masenf:
**`CodSpeed Performance Analysis` is red on `c2e53c7`, and it is not this PR's.** Standing down on it, with reasons — and I can't re-run it myself.

It reports `-3.16%` on `tests/benchmarks/test_state_proxy.py::test_var_access[mutable_dict]` (45.6 ms → 47.1 ms), 66 untouched.

This PR cannot reach that benchmark. Against `main` it changes five files:

```
.github/workflows/build_packages.yml
.github/workflows/min_deps.yml
AGENTS.md
scripts/check_min_deps.py
tests/units/test_check_min_deps.py
```

Nothing under `reflex/`, `packages/`, or `tests/benchmarks/`. `scripts/check_min_deps.py` is a standalone CI script that nothing in the framework imports, and `test_var_access` exercises `reflex.state` proxy access.

The previous head is the control. On `0cf472a8` CodSpeed reported **"will not alter performance"** — 67 untouched, 0 regressed — against the identical `reflex/` tree. The only difference between `0cf472a8` and `c2e53c7` is that script, its unit tests, and `AGENTS.md`. The same runtime code went from 67-untouched to 1-regressed with no runtime change in between.

Two more things point the same way. CodSpeed's own report carries *"Different runtime environments detected — Some benchmarks with significant performance changes were compared across different runtime environments, which may affect the accuracy of the results."* And the comparison base moved between the two reports: `bbb5662` (a fallback, because no successful run existed on main `3f44673`) to main `84bb6ab`. Different baseline, different machine, 3% apart on one benchmark.

Worth separating: **the `performance-tests` workflow itself passed** ([run 35892028804](https://github.com/reflex-dev/reflex/actions/runs/35892028804)). The benchmarks ran clean; what is red is CodSpeed's threshold judgement on the measurement.

No fix to port — there's no broken base here, just a noisy comparison — and I'm not going to touch `tests/benchmarks/` to quiet it, since that is outside this PR and would be papering over a measurement rather than fixing code. Acknowledging the regression on the CodSpeed dashboard is a maintainer action. **I tried the one re-run the situation warrants and got `403 Resource not accessible by integration`**, so someone with Actions write access would need to re-run `performance-tests` to get a fresh measurement. I'm leaving the PR watched either way.

Meanwhile the part this commit *could* have broken is green: **all 40 `check-min-deps` jobs and all 23 `build` jobs pass on `c2e53c7`**, across both Python versions and all 20 packages. 107 checks green, 13 still running, this one failure.

---

FarhanAliRaza (approved):

